PR #14, opened 2022-01-28
Wiring

iajrz:
Introduces the first set of changes for the wiring: the build tool (mage), a build script, and injection of version information.
Also starts ignoreing the `build/` directory where build artifacts can live without inconveniences.


luyzdeleon:
Good effort as a first PR @iajrz ! However I wanna point a few things out:

1. We don't have any sort of QA/Testing document for this PR, if automation is not possible day 1, at least a guide on how to test this would be ideal.
2. There are versioning assumptions/functionality included here that is not documented.
3. There's the setting of the Golang version which is not documented. If we're going to use a specific version, this should be widely available in the repo.
4. The design decision of the application entrypoint is not documented other than a comment on the issue.

One other idea that comes to mind is that this PR adds a new dependency `mage`, I think we should make it standard practice to write a little document for every dependency added, what's the impact and scope with it, and how do we plan to maintain the codebase around these dependencies, but I'm open to debate that.

iajrz:
Excellent catches; as per the team call, this is the sort of thing I'm hoping the PR process will catch and nudge us to improving. Good show of working with a v1 mindset, and excellent feedback. I'll get to work on this feedback, which will likely include some conversations around the best way to do things. The nudge to document these conversations _here_ is implicit, but I'm explicitly stating that if the conversations happen offline I'll do an effort of documenting them here.

There are a few levels of documentation that we're looking at here:
- Docs to give certainty about quality
- Docs to communicate design
- Docs to explain the organization to other developers

I wonder where to place these; I mean, most likely within the `doc/` subtree for some of it, and some in the `README.md`; if you have any specifics about these, even source material specifying rules I need to conform to, I'd appreciate a link.

I think the explainer about the quality could live in the PR, but maybe it should live elsewhere to ensure things are tested properly when we introduce further changes here?

I posit that specifying the version needs no documentation; anyone who is a go developer knows where that's specified, and to look for the datum, which makes me not want to document it – it's developer documentation, and I expect developers to have minimal understanding of the tools, and the language version is part of that minimal understanding.

I agree about the dependency documentation; part of it is going to be, in this case, in the `README.md` as it will explain how to build the code. Do you have any place in mind for this to live besides that? And any specific questions – I guess we should have a set of default questions to think about answering when we introduce the dependencies.

cmd/pocket/main.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"flag"
+	"fmt"
+)
+
+var version string
+
+func main() {
+	v := flag.Bool("version", false, "")
```
Olshansk, line 11: I'm a bit confused by what this is trying to do. The passed in version is a bool and we're printing an unassigned version string?

Don't we want the flag to write the value in &version?

go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/pokt-network/pocket
+
+go 1.17
```
Olshansk, line 3: Any chance we can start of with 1.18. I was hoping to use generics!

mage.go
```diff
@@ -0,0 +1,49 @@
...
+package main
+
+import (
+	"fmt"
+	"github.com/magefile/mage/sh"
+)
+
+const versionStringEnvVarName = "POCKET_VERSION_STRING"
+const pocketPackage = "github.com/pokt-network/pocket/cmd/pocket"
+const ldflag = "-X main.version=$" + versionStringEnvVarName
```
Olshansk, line 12: s/ldflag/slFlag

mage.go
```diff
@@ -0,0 +1,49 @@
+//go:build mage
+
+package main
+
+import (
+	"fmt"
+	"github.com/magefile/mage/sh"
+)
+
+const versionStringEnvVarName = "POCKET_VERSION_STRING"
```
Olshansk, line 10: Why not use golang factoring here?

E.g. https://golangbyexample.com/multiple-constant-declarations-go/

mage.go
```diff
@@ -0,0 +1,49 @@
...
+const pocketPackage = "github.com/pokt-network/pocket/cmd/pocket"
+const ldflag = "-X main.version=$" + versionStringEnvVarName
+
+var env = map[string]string{}
+
+func init() {
+	env["GO111MODULE"] = "on"
+
+	fmt.Println("Getting branch and commit information")
```
Olshansk, line 19: NIT: I think we should use `log.Println` since eventually, we'll be passing a logger (with a bunch of bells and whistles on top) around.

mage.go
```diff
@@ -0,0 +1,49 @@
...
+	_, dirty := sh.Output("git", "diff", "--quiet")
+
+	if branch != "" {
+		branch = "-" + branch
+		hash = "/" + hash
+	}
+	if dirty != nil && dirty.Error() == "running \"git diff --quiet\" failed with exit code 1" {
+		hash += "+dirty"
+	}
+	env[versionStringEnvVarName] = "0.0.0" + branch + hash
```
Olshansk, line 31: NIT: I think the more idiomatic way to go about this is to use `fmt.Sprintf`. Example:

```
if branch != "" {
    fmt.Sprintf("0.0.0-%s/%s", branch, hash)
} else {
   ...
}
```

mage.go
```diff
@@ -0,0 +1,49 @@
...
+	fmt.Println("Getting branch and commit information")
+	branch, _ := sh.Output("git", "branch", "--show-current")
+	hash, _ := sh.Output("git", "rev-parse", "--short", "HEAD")
+	_, dirty := sh.Output("git", "diff", "--quiet")
+
+	if branch != "" {
+		branch = "-" + branch
+		hash = "/" + hash
+	}
+	if dirty != nil && dirty.Error() == "running \"git diff --quiet\" failed with exit code 1" {
```
Olshansk, line 28: Maybe moving this long string into an unexported constant?

mage.go
```diff
@@ -0,0 +1,49 @@
+//go:build mage
+
+package main
+
+import (
+	"fmt"
+	"github.com/magefile/mage/sh"
```
Olshansk, line 7: Why not use https://pkg.go.dev/os/exec#Cmd.Run which is a go native library instead of a 3rd party one?

Olshansk (changes requested):

go.mod
iajrz: 1.18 is still beta; I was hoping to migrate to it when it goes live.

cmd/pocket/main.go
iajrz: The variable is set via a linker flag (see `magefile.go`). This is something I need to make explicit as part of the documentation that Luis required of me 😄

mage.go
iajrz: explicitly ldflag, because it is the "value" part of the `-ldflags` argument to `go build`. Perhaps it should be `ldflags`, though.

mage.go
iajrz: The `mage.go` file is not part of the project per se; it's a separate source that is executed a la `rake` or `make` by `mage`. I don't think this will grow to have a logger soon.

mage.go
iajrz: it's used in a single place; I generally prefer to extract into declarations when the information is shared across places, otherwise it's just extra steps.

mage.go
iajrz: This contains a wrapper that builds the environment conveniently with just a hashmap of environment variables; we depend on `mage` to run the file anyhow, so we might as well leverage the niceties they include; the actual code that is executed when setting up the environment can be seen [here](https://github.com/magefile/mage/blob/526bf4775130d567de6b18fea8f3791d76c8be2a/sh/cmd.go#L126).

mage.go
iajrz: I think this case might not fit the idiom since there are fragments of the version string that are optional. I will consider it, though, while I go through the code and contemplate the suggestions and restructure things.

mage.go
iajrz: No particular reason. Taking this into account.

go.mod
luyzdeleon: This is the kind of reasoning that should live in documentation.

cmd/pocket/main.go
Olshansk: Made also add a comment in the code about the variable?

go.mod
Olshansk: I also think that adding a comment in the `go.mod` file above `go. 1.17` will make it very clear to the reader (without needing to look in a different place) why we're not on 1.18 yet.

mage.go
Olshansk: +1 to `ldflags`

mage.go
Olshansk: My preference is to change it.

iajrz:
I still need to document the choice for dependencies; I'm thinking about a directory for that sort of justification within the docs.

docs/contributing/CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
+## How to contribute
+
+Pocket is an open source project, and as such we welcome any contribution from anyone on the internet.
+
+Please fork, code and submit a Pull Request for the team to review and merge. We ask that you please follow the
+guidelines below in order to submit your contributions for review:
```
Olshansk, line 6: NIT: There's a line breakpoint between 5 & 6, even though line 5 wraps onto a new line.

Is this intentional or do you use a specific line length in your editor?

docs/contributing/CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+- Code must be documented adhering to the official Go commentary guidelines.
+- Pull requests need to be based on and opened against the `main` branch.
+
+### Criteria for PR Evaluation
+
+- Quality: the PR should include instructions on how to test the submitted code. Automated tests are not a must in all
+  situations, but always strongly encouraged
+- Criteria expressed in [this document](TASTE.md)
+- Context: make it easy for the reviewers to understand what is being addressed by the PR and why it's important to the
+  project
```
Olshansk, line 29: DITTO comment above around "breakpoints" being in different places.

docs/contributing/CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Code must adhere to the official Go formatting guidelines (i.e. uses [gofmt](https://golang.org/cmd/gofmt)).
+- (Optional) Use [Editor Config](https://editorconfig.org) to help your Text Editor keep the same formatting used
+  throughout the project.
+- Code must be documented adhering to the official Go commentary guidelines.
+- Pull requests need to be based on and opened against the `main` branch.
+
+### Criteria for PR Evaluation
+
+- Quality: the PR should include instructions on how to test the submitted code. Automated tests are not a must in all
```
Olshansk, line 25: Note for a future PR.

We can create a template for the PR with headings (e.g. SUmmary, Purpose, Test, etc...) and add details for the user to fill out via comments.

Anything inside of `<!-- -->` is not evaluated by markdown,

docs/testing/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+If built outside a repository, or without the `git` executable in the path, it should report version `UNKNOWN`.
+
+If built in a repository where uncommitted changes are present, the version should report a version with `+dirty`
+
+Version reported for a proper git repository with the `git` executable present in the path should be `0.0.0-branchname/commithash`, where the version number is fixed but the branchname reflects the current branch and the commit hash likewise the current commit's hash.
+
+## Listing Targets
+
+running `mage -l` should show the 4 targets with brief descriptions for each.
```
Olshansk, line 16: s/running/Running

go.mod
Olshansk, line 3: OPTIONAL: Add TODO to move to 1.18 once it's out of beta.

mage.go
```diff
@@ -0,0 +1,60 @@
...
+
+	fmt.Println("Getting branch and commit information")
+	branch, branchError := sh.Output("git", "branch", "--show-current")
+	hash, hashError := sh.Output("git", "rev-parse", "--short", "HEAD")
+	_, dirtyError := sh.Output("git", "diff", "--quiet")
+
+	versionNumber := "UNKNOWN"
+
+	// git repo invariant: we have a branch and a hash whenever version is known
+	if branchError == nil && hashError == nil {
```
Olshansk, line 29: Nice!

Olshansk (approved):
Very minor nits (feel free to ignore) but LGTM!

luyzdeleon (approved):
LGTM, all my comments from my initial review are addressed here and can be further developed with the project.

docs/contributing/CONTRIBUTING.md
iajrz: We could; the question of where to put the information, though, is partially answered by the very shape of this PR; testing information lives in the docs, for instance; summary and purpose should be tied to the origin document and the rest of the context that we'd provide in the goal. I think it's important that test information for each patch lives in the repo so that it can serve as a reference that lives with the repo, not necessarily get archived and forgotten with a closed PR.

docs/contributing/CONTRIBUTING.md
iajrz: Indeed; I will ignore for the moment since it's meaningless whitespace in markdown, but will include in the next PR.

PR #27, opened 2022-02-23
V1 Prototype Snapshot

Olshansk:
This PR is meant to address this issue: https://github.com/pokt-network/pocket/issues/22.

It is a functional snapshot of the V1 prototype that can be referenced and used as a baseline as V1 development continues.

# Testing

## 0. Prerequisite

The development environment and instructions here are not expected to be complete since the development environment is still a work in progress, but the two following assumptions are made:
1. Docker is installed and the daemon is running locally.
2. `protoc` is installed and running locally.

## 1. Pull changes:
```
$ git clone git@github.com:pokt-network/pocket.git pocket_v1_proto_snapshot
$ cd pocket_v1_proto_snapshot
$ git checkout -b integration/workshop2/attempt2 origin/integration/workshop2/attempt2
$ cd prototype
```

## 2. Clean environment:
```
$ cd prototype
$ make docker_wipe # Press `y` if prompted
$ make v1_generate_protos # Note that this is not using the docker protos at the moment
```

## 3. Run localnet

Example video can be found [here](https://drive.google.com/file/d/1Uc43PQ4t15GyT5t1bpsv_KpCwyu9HxBa).

### 3.1 Shell 1:
```
$ make compose_and_watch
```

### 3.2 Shell 2:
```
$ make client_start
$ make client_connect
```

### 3.3. Run commands

In Shell:
1. Select `TogglePaceMakerManualMode`
2. Select `TogglePaceMakerManualMode`

Issue #22, opened 2022-02-23
V1 Prototype Snapshot


Olshansk:
# Objective
Immutable code snapshot/reference of the 1st V1 prototype.

# Origin Document
One of the goals in mid-February was to build a foundation for the V1 module integration. Since it was just a prototype, the team is aware of the following:
* As V1 matures, the code will likely change quite a bit.
* To submit the current prototype to mainline, it'll be split into a handful (i.e. ~5) of smaller PRs.
* It is just a prototype, and there are a lot of missing gaps in the implementation.
* A lot of the existing code paths may be unused and there is no guarantee that all the tests are passing.

In order to have a reference to which we can revert, the snapshot of the code, as of Feb 18th 2022 will be submitted into a separate directory which can eventually be completely removed.

## Deliverables
- All the code used for the V1 end-to-end block demo
- Notes were taken during final integration meeting
- Notes/instructions on how to test

# Context

Note that the goal this issue is not a complete prototype, but rather a history of the first functional end-to-end block.

For example, the following will be done in subsequent iterations:
1. Removing unused code
2. Removing `pre_p2p` with `p2p`
3. Fixing the consensus testing framework.
4. Etc...

---

**Creator**: @olshansky        
**Co-Owners**: @andrewnguyen22 @derrandz @luyzdeleon

Olshansk:
Opened a PR here: https://github.com/pokt-network/pocket/pull/27

PR #29, opened 2022-02-24
Pocket node entrypoint, module interface and application specific bus 

Olshansk:
# Context

This is the first of several PRs to address https://github.com/pokt-network/pocket/issues/21.

A lot of the code here was copied over from the `prototype` in https://github.com/pokt-network/pocket/pull/27, with some fixes, cleanup, modularization. 

**IMPORTANT NOTE TO THE REVIEWER**: Note that some code was deleted so just a skeleton is submitted and should be added back when the module-specific code is added back in.

## Details

- Various additions to the .gitignore file
- Makefile addition: temporary / experimental targets to enable V1 prototype development
- Add Dockerfiles for validator, debug client and Protobufs generation
- Add a 1st iteration version of the configuration file
- Add a 1st iteration version of the genesis file
- Add a docker-compose file for 4 validators and a debug client
- Add a debug client with a placeholder trigger
- Skeleton implementations of all 4 modules
- Update all go dependencies (go.mod, go.sum, etc…)
- Migrated first iteration of crypto library from pre-prototype to the prototype

## Next Steps
* Submit this PR as the foundation for pocket V1
* Prepare a small "design document" (< 1 page) on the structure of the bus and modules
* Write up a small document on how to use these (for contributors and new employees)

## TODOs as a result of the discussion PR
** (will create an issue and add these before submission)
- [ ] **[Small]** Remove makefile altogether
- [ ] **[Small]** Remove protobuf generation from host machine to Dockerfiles
- [ ] **[Small]** Add documentation for Goland & DLV debugging
- [ ] **[Medium]** Migrate all useful targets from Makefile to mage
- [ ] **[Large]** Discuss & design how `LoadConfig` should be implemented
- [ ] **[Ongoing]** Iterate on and define the structure of the configuration file
- [ ] **[Ongoing]** Iterate on and define the structure of the genesis file

## 1. Pull changes (if not already local):
```
$ git clone git@github.com:pokt-network/pocket.git pocket
$ cd pocket
$ git checkout -b issue21/prototype/entrypoint origin/issue21/prototype/entrypoint
```

## [Optional] 2. View available commands:
```
$ make
```

## 3. Clean environment:
```
$ make docker_wipe # Press `y` if prompted
```

### 3.1 Shell 1:
```
$ make compose_and_watch
```

### 3.2 Shell 2:
```
$ make client_start
$ make client_connect
```

### 3.3. Run commands

Inspect that a placeholder events gets continually sent to the main bus

[...]

PR #31, opened 2022-02-24
V1 P2P Prototype

derrandz:
### Overview
* #### Issue: [Issue#25](https://github.com/pokt-network/pocket/issues/22)
* #### Depends_on: #27, 27 has to be merged before this comes in at it builds on top of it
* #### Impact:
    * Added a new module: P2P prototype
    * Updated the existing p2p mock to use P2P protobuf messages
    * Deprecated unused/unnecessary p2p types used by the p2p mock
* #### Description:

This PR adds in the p2p prototype on top of the v1 snapshot as in, and keeps in place the existing p2p mock (_at least for the time being_)

* #### Testing:

**_Prerequisite_**: The development environment and instructions [here](https://drive.google.com/file/d/1Uc43PQ4t15GyT5t1bpsv_KpCwyu9HxBa) are not expected to be complete since the development environment is still a work in progress, but the two following assumptions are made:

- Docker is installed and the daemon is running locally.
- protoc is installed and running locally.

* Test the p2p module
```
$ go test -v -race ./p2p
```

* Run the e2e script
```
$ E2E=true go test -v -run TestNetwork_E2E -v -race ./p2p
```

* Run the e2e stack
```
make p2p-run-e2e-stack
```

* Test the p2p integration with the rest of the modules:

Shell#1
```
$ make compose_and_watch
```

Shell#2
```zsh
$ USE_P2P_MOCK=false make client_start && make client_connect
```
After you are prompted with the choices, select whatever action you see fit.


Olshansk:
@derrandz Temporarily changing the base to `integration/workshop2/attempt2` so I can see the actual changes in the PR (not the code that is being submitted in the other branch).

Note:
1. We will change it back to `v1/milestone-protype` after my "snapshot" PR is submitted.
2. Please make sure not to merge this into the `integration/workshop2/attempt2` branch since branch protection isn't set up yet and I don't have permissions to enable it.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+	config *config.P2PConfig
+	c      *dcodec
+
+	id           uint64
+	protocol     string
+	address      string
+	externaladdr string
+
+	inbound  pipemap
+	outbound pipemap
```
Olshansk, line 44: Consider making the naming more descriptive. E.g. `outboundNetworkPipe`. Doesn't need to be super verbose, but maybe a comment could help too for the reader.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+
+	peerlist *types.Peerlist
+
+	listener struct {
+		sync.Mutex
+		*net.TCPListener
+	}
+	listening atomic.Bool
```
Olshansk, line 52: `s/listening/isListening`

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+
+	done    chan uint
```
Olshansk, line 54: 1. Consider create a type for `chan uint`.
2. It's unclear to me why all of these are unit channels. Consider adding a comment for what they're used for.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+	ready   chan uint
+	closed  chan uint
+	errored chan uint
+
+	sink     chan types.Work
+	handlers map[types.PeerEvent][]func(...interface{})
```
Olshansk, line 60: So every handler can point to an array of variadic functions?

Is this level of "generalizability" needed?

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+
+	logger struct {
+		sync.RWMutex
+		print func(...interface{}) (int, error)
+	}
+
+	err struct {
```
Olshansk, line 67: Question: are these tiny embedded structs a common pattern? I would have personally externalized them but also don't have as much experience with Go.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+		sync.Mutex
+		error
+	}
+}
+
+var (
+	ErrNotCreated = errors.New("Module error: P2P Module not created. Trying to start the p2p module before calling create.")
```
Olshansk, line 74: 1. Why not group these with the other vars/constants above?
2. Why var vs const?

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+func NewP2PModule() *P2PModule {
```
Olshansk, line 77: I like the `Create` function below but don't think we should be exposing a factory function that returns a pointer to the underlying struct in order to stay consistent with the rest of the codebase. So:

1. `P2PModule` struct shouldn't be exposed.
2. `NewP2PModule` should be local or removed altogether.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+	return &P2PModule{
+		protocol: Protocol,
```
Olshansk, line 79: 1. I think these shold come from the config.
1.1 If not, why not?
1.2 If this is a future item, please add a TODO.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+
+		inbound:  *NewIoMap(MaxInbound),
+		outbound: *NewIoMap(MaxOutbound),
+
+		c: NewDomainCodec(),
+
+		peerlist: nil,
+		sink:     make(chan types.Work, 100), // TODO: rethink whether this should be buffered
+
+		done:   make(chan uint, 1),
```
Olshansk, line 90: 1. If `chan uint` was a specific type, the code would be more self-explanatory.
2. Add a comment on why the size of the channel is `1`.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+		protocol: Protocol,
+		address:  Address,
+
+		inbound:  *NewIoMap(MaxInbound),
+		outbound: *NewIoMap(MaxOutbound),
+
+		c: NewDomainCodec(),
+
+		peerlist: nil,
+		sink:     make(chan types.Work, 100), // TODO: rethink whether this should be buffered
```
Olshansk, line 88: When leaving a TODO, it's good to do something like `TODO(derrandz): ...` so people know:

1. Who the TODO belongs to.
2. Who added the TODO without needing to do a `git blame`.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+
+func Create(config *config.Config) (modules.NetworkModule, error) {
+	module := NewP2PModule()
+
+	module.setLogger(func(args ...interface{}) (int, error) {
+		log.Println(args...)
+		return 0, nil
+	})
+
+	module.config = config.P2P
```
Olshansk, line 107: I think we should instantiate the struct in this function directly.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+
+		handlers: make(map[types.PeerEvent][]func(...interface{}), 0),
+		errored:  make(chan uint, 1),
+	}
+}
+
+func Create(config *config.Config) (modules.NetworkModule, error) {
+	module := NewP2PModule()
+
+	module.setLogger(func(args ...interface{}) (int, error) {
```
Olshansk, line 102: My preference: remove custom logger logic. This is something that all interfaces will need and we'll probably add as a single "infra" PR. 

I think @iajrz has been thinking about this.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (m *P2PModule) Send(addr string, msg *anypb.Any, topic string) error {
+	netmsg := &types.NetworkMessage{Data: msg, Topic: types.Topic(topic)}
+	encodedBytes, err := m.c.encode(netmsg)
+	if err != nil {
+		return err
+	}
+
+	return m.send(addr, encodedBytes, true) // true: meaning that this message is already encoded
```
Olshansk, line 167: What does the `true` in the broadcast mean then? Consider adding a comment above.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (m *P2PModule) AckSend(addr string, msg *types.NetworkMessage) (bool, error) {
```
Olshansk, line 170: Does `AckSend` need to be exposed to other modules?

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+	encodedBytes, err := m.c.encode(msg)
+	if err != nil {
+		return false, err
+	}
+
+	response, err := m.request(addr, encodedBytes, true) // true: meaning that this message is already encoded
```
Olshansk, line 176: Would it not be simpler if the functions either ONLY either an encoded message or not?

p2p/module.go
Olshansk, line 170: What does the `bool` in the return value represent?

prototype/Makefile
```diff
@@ -33,9 +33,15 @@ build_and_watch:
 client_start:
 	docker-compose -f build/deployments/docker-compose.yaml up -d client
 
+client_alpha_start:
```
Olshansk, line 36: What is `client_alpha`? Why can't we just use one client?

prototype/Makefile
```diff
@@ -33,9 +33,15 @@ build_and_watch:
...
+	docker-compose -f deployments/docker-compose.yaml up -d client_alpha
+
 ## Connect to the client daemon running in a background docker container.
 client_connect:
-	docker exec -it client /bin/bash -c "go run cmd/client/*.go"
+	docker exec -it client /bin/bash -c "go run cmd/client-with-pre-p2p/*.go"
+
+client_beta_connect:
```
Olshansk, line 43: There's a client, a client alpha, a client beta, etc...

This is fine for a local working branch but I don't think it should be committed to master.

prototype/build/node.Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM golang:1.17.6-bullseye
```
Olshansk, line 1: Why do we need a different Dockerfile for this? Ideally a single Dockerfile should be sufficient for all purposes.

Again, seems like the PR is not ready for submission?

prototype/build/stack.yml
```diff
@@ -0,0 +1,394 @@
+version: "3.5"
```
Olshansk, line 1: Maybe rename this to "p2p_test_stack.yml"?

prototype/cmd/client-with-p2p/neo4j.go
```diff
@@ -0,0 +1,103 @@
+package main
```
Olshansk, line 1: Remove. Neo4j was just an experiment that we can bring back later.

prototype/cmd/client-with-pre-p2p/main.go
```diff
@@ -0,0 +1,176 @@
+package main
```
Olshansk, line 1: Rather than copy-pasting main.go, maybe we can just create config internally. The goal of it was for both of them to have the same interface.

prototype/cmd/v1/client/neo4j.go
```diff
@@ -0,0 +1,104 @@
+package main
```
Olshansk, line 1: Remove.

prototype/p2p/codec.go
```diff
@@ -140,6 +140,31 @@ func (c *wcodec) decode(wiredata []byte) (nonce uint32, enc Encoding, data []byt
 	return
 }
 
+func (c *wcodec) decodeHeader(header []byte) (flagswitch []bool, nonce uint32, bodylength uint32, err error) {
```
Olshansk, line 143: `s/bodylength/bodyLength`

I'm pretty sure golang best practices is camelCase.

Ditto for `s/wcodec/wCodec`

prototype/p2p/errors.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func isErrEOF(err error) bool {
+	if errors.Is(err, io.EOF) {
+		return true
+	}
+
+	var netErr *net.OpError
+
+	if errors.As(err, &netErr) && netErr.Err.Error() == "use of closed network connection" {
```
Olshansk, line 28: You can replace this:
```
	if errors.As(err, &netErr) && netErr.Err.Error() == "use of closed network connection" {
		return true
	}

	return false
```

With:
return `errors.As(err, &netErr) && netErr.Err.Error() == "use of closed network connection"`

prototype/p2p/errors.go
```diff
@@ -0,0 +1,43 @@
...
+		return true
+	}
+
+	return false
+}
+
+func isErrTimeout(err error) bool {
+	var netErr *net.OpError
+
+	if errors.As(err, &netErr) && netErr.Timeout() {
```
Olshansk, line 38: Ditto. All of this can be replaced with `errors.As(err, &netErr) && netErr.Timeout()`

prototype/p2p/errors.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"net"
+)
+
+var (
+	ErrPeerHangUp func(error) error = func(err error) error {
```
Olshansk, line 11: Why create vars rather than just named functions?

prototype/p2p/messenger.go
```diff
@@ -0,0 +1,35 @@
+package p2p
```
Olshansk, line 1: Rename file to `message.go` instead of `messenger.go`.

prototype/p2p/messenger_test.go
```diff
@@ -0,0 +1,78 @@
+package p2p
+
+import (
+	"pocket/p2p/types"
+	"testing"
+)
+
+func TestMessager_Protobuff(t *testing.T) {
```
Olshansk, line 8: Pro tip: If you want people to run your tests, add a Makefile target to run them.

`go test ./...` will include it, but it's nice to have a dedicated target for it too.

Olshansk, line 8: Messager -> Messenger?
Protobuff -> Protobuf?

Why the `_` in the test name?

prototype/p2p/messenger_test.go
```diff
@@ -0,0 +1,78 @@
...
+	nonce, src, dst, level := int32(1), "10.0.0.1:1234", "9.0.0.2:5432", int32(2)
+
+	// test message instantiation
```
Olshansk, line 11: I think we should have different unit tests rather than sub-scoping everything within a single test.

prototype/p2p/messenger_test.go
```diff
@@ -0,0 +1,78 @@
...
+	{
+		p2pmsg := Message(nonce, level, types.PocketTopic_P2P_PING, src, dst)
+
+		if p2pmsg.Topic != types.PocketTopic_P2P_PING {
```
Olshansk, line 15: Highly recommend importing these dependencies:

```
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
```

The test could then be replaced with one line:

`assert.Equal(p2pmsg.Topic, types.PocketTopic_P2P_PING types.PocketTopic_P2P_PING, fmt.Sprintf("Protobuff messenger error: Failed to instantiate a ping message, expected topic: %s, got: %s", types.PocketTopic_P2P_PING, p2pmsg.Topic))`

See example here: https://github.com/pokt-network/consensus-pre-prototype/blob/main/pkg/consensus/leader_election/election_test.go

prototype/p2p/messenger_test.go
```diff
@@ -0,0 +1,78 @@
...
+			t.Errorf("Protobuff messenger error: Failed to instantiate a ping message, expected topic: %s, got: %s", types.PocketTopic_P2P_PING, p2pmsg.Topic)
+		}
+
+		if p2pmsg.Source != src {
```
Olshansk, line 19: Ditto for all the other tests below. We should definitely use assertify.

prototype/shared/node.go
```diff
@@ -30,6 +30,11 @@ func Create(config *config.Config) (n *Node, err error) {
 		return nil, err
 	}
 
+	//networkMod, err := p2p.Create(config)
```
Olshansk, line 33: Delete if not used or add a TODO.

prototype/p2p/module.go
```diff
@@ -1,56 +1,192 @@
 package p2p
 
 import (
-	"google.golang.org/protobuf/types/known/anypb"
+	"errors"
```
Olshansk, line 4: We already have a copy of this file at the root directory. 

I personally think that everything should be submitted to the "non-prototype" directory. It was meant to be a snapshot of last friday.

prototype/p2p/network_lifecycle.go
```diff
@@ -0,0 +1,61 @@
+package p2p
+
+import (
+	"pocket/p2p/types"
+	"strings"
+)
+
+func (g *P2PModule) configure(protocol, address, external string, peers []string) {
```
Olshansk, line 8: The variable is name `g`, I'm guessing for `gater` but it is an element of the `P2PModule` struct.

1. Rename to either `m` or `module` or `p`. 
2. `P2PModule` struct should not be exposed outside the `p2p` module.

prototype/p2p/network_lifecycle.go
```diff
@@ -0,0 +1,61 @@
...
+	g.protocol = protocol
+	g.address = address
+	g.externaladdr = external
+	g.peerlist = types.NewPeerlist()
+
+	// this is a hack to get going no more no less
```
Olshansk, line 14: Added a `TODO(derrandz)` for hacks so they are easy to identify, follow and track back. 

Also, for hacks, explain what the hack is doing.

prototype/p2p/network_lifecycle.go
```diff
@@ -0,0 +1,61 @@
...
+	for i, p := range peers {
+		peer := types.NewPeer(uint64(i+1), p)
```
Olshansk, line 16: What is the peer ID used for?

prototype/p2p/network_lifecycle.go
```diff
@@ -0,0 +1,61 @@
...
+		port := strings.Split(peer.Addr(), ":")
```
Olshansk, line 17: Doesn't strings.split return an array rather than just a single port value? The var name isn't intuitive.

prototype/p2p/network_lifecycle.go
```diff
@@ -0,0 +1,61 @@
...
+		myport := strings.Split(g.address, ":")
+		if port[1] == myport[1] {
+			g.id = peer.Id()
+		}
+		g.peerlist.Add(*peer)
```
Olshansk, line 22: Why derefernce the peer instead of passing in a pointer?

prototype/p2p/network_lifecycle.go
```diff
@@ -0,0 +1,61 @@
...
+	g.peerlist = types.NewPeerlist()
+
+	// this is a hack to get going no more no less
+	for i, p := range peers {
+		peer := types.NewPeer(uint64(i+1), p)
+		port := strings.Split(peer.Addr(), ":")
+		myport := strings.Split(g.address, ":")
+		if port[1] == myport[1] {
+			g.id = peer.Id()
+		}
```
Olshansk, line 21: Add an `else` that panics so it's easier to catch an error.

prototype/p2p/network_lifecycle.go
```diff
@@ -0,0 +1,61 @@
...
+		g.peerlist.Add(*peer)
+	}
+}
+
+func (g *P2PModule) init() error {
+	msg := Message(int32(0), 1, types.PocketTopic_P2P_PING, "", "")
```
Olshansk, line 27: Why do we only need to register one type of message? Maybe add a comment what this is doing.

prototype/p2p/network_mocks.go
```diff
@@ -0,0 +1,20 @@
+package p2p
```
Olshansk, line 1: We can autogenerate mocks with `mockgen`. 

I already have a `make target` for it in https://github.com/pokt-network/pocket/pull/29.

This way the autogenerated code doesn't need to be submitted to master.

prototype/p2p/pre_p2p/module.go
```diff
@@ -126,8 +129,8 @@ func (m *networkModule) BroadcastMessage(msg *anypb.Any, topic string) error {
 }
 
 func (m *networkModule) Send(addr string, msg *anypb.Any, topic string) error {
-	c := &types.P2PMessage{
-		Topic: topic, // TODO topic is either P2P (from this module) or consensus
+	c := &p2ptypes.NetworkMessage{
+		Topic: p2ptypes.Topic(topic), // TODO topic is either P2P (from this module) or consensus
```
Olshansk, line 133: Add names for TODOs. Could be someone else or just "team" if it can be handled by anyone.

prototype/p2p/proto/p2p.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package p2p;
+
+option go_package = "/types/";
+
+enum PocketTopic {
```
Olshansk, line 9: Do you think this shold be scoped to P2P or available elsewhere too?

prototype/p2p/testing_utils.go
```diff
@@ -0,0 +1,43 @@
+package p2p
+
+import (
+	"crypto/rand"
+	"sync/atomic"
+)
+
+type fnCallStub struct {
+	called *int32 // act as bool
```
Olshansk, line 9: Why the `act as` if it could be an int or a bool?

prototype/p2p/testing_utils.go
```diff
@@ -0,0 +1,43 @@
...
+		timesc: new(int32),
+	}
+}
+
+func (f *fnCallStub) trackCall() {
+	atomic.AddInt32(f.called, 1)
+	atomic.AddInt32(f.timesc, 1)
+}
+
+func (f *fnCallStub) wasCalled() bool {
```
Olshansk, line 25: I really like the idea of build custom tooling for our needs, but this seems to be something generic.

For example, if we use the `mockgen` library, we can literally tell it how many times it should be called and that way we are not rolling our own: 

```
	p2pNetworkMock.EXPECT().
		GetAddrBook().
		DoAndReturn(func() []*p2p_types.NetworkPeer {
			log.Println("[MOCK] Network GetNetwork", addrBook)
			return addrBook
		}).
		AnyTimes()
```		

Example: https://github.com/pokt-network/consensus-pre-prototype/blob/c54a9a062a7507639d30b6f920698c3d08605ffa/pkg/pocket/consensus_tests/utils_test.go#L32

We can discuss this in more detail

Olshansk (changes requested):
Reviewed most of the code except for `raintree.go` and `raintree_test.go`.

1. Lots of small comments throughout the codebase but generally LGTM.
2. Really good explanation in the PR description.
3. Need to discuss if this should be submitted to the `prototype` directory since it was intended to just be a snapshot of last Friday. Even if not, we should only have the PR submit code to either the root dir or the prototype dir, not both.

TODOs (for me):
1. [ ] Re-review the PR once comments are addressed
2. [ ] Do a team or 1:1 sync on what should and sholdn't go into this repo versus being left in our pre-prototypes repos. Fwiw, not planning to migrate any of the neo4j stuff for example in consensus.
3. [ ] Look through `raintree` and `raintree_test` in detail
4. [ ] Follow the instructions in the PR description to test locally

andrewnguyen22:
Will review after Olshanskys comments are addressed. No need to go through twice

.gitignore
```diff
@@ -16,3 +16,12 @@
 
 .DS_Store
 .idea/
+consensus/neo4j/
+
+// Proto generate messages
```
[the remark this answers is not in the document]
derrandz: I think it's a rebase slip.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+	"sync"
+
+	"go.uber.org/atomic"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+var (
+	MaxInbound           uint = 128
+	MaxOutbound          uint = 128
+	WireByteHeaderLength int  = 9
```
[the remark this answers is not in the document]
derrandz: Good catch

prototype/p2p/network_mocks.go
derrandz: I will look into mockgen

prototype/p2p/proto/p2p.proto
derrandz: It was initially intended only for P2P

prototype/p2p/testing_utils.go
derrandz: Because of the atomic nature of it, as it is used in goroutines by multiple concurrent routines and I did not want to use `atomic.Bool` from uber's atomic lib.

prototype/p2p/testing_utils.go
derrandz: I am interested in mockgen at this point. I've resorted to making my own "function call stub/spy" due to my ignorance of existing tools (_primarily_).

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+	ReadBufferSize       int  = (1024 * 4)
+	WriteBufferSize      int  = (1024 * 4)
+	ReadDeadlineMs       int  = 400
+)
+
+var (
```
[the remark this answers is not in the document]
derrandz: 1. Just because of prototyping, they should be consts
2. They should be loaded from the config, at the time it was more convenient to use var/consts as presumed default config values.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+	Protocol = "tcp"
+	Address  = "localhost:30303"
+)
+
+type P2PModule struct {
```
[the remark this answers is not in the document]
derrandz: Good catch.

p2p/module.go
```diff
@@ -0,0 +1,193 @@
...
+	bus    modules.Bus
+	config *config.P2PConfig
+	c      *dcodec
+
+	id           uint64
+	protocol     string
+	address      string
+	externaladdr string
+
+	inbound  pipemap
```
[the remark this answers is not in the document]
derrandz: It should be in `prototype/p2p/types/pipe_map.go`...

(_After I checked, indeed the file was missing, to my confusion, I suppose the double slash in the .gitignore must've done something as they were present locally and upon adding them, git complained that they were ignored._)

p2p/module.go
derrandz: Thanks vimmerman 🤣

p2p/module.go
derrandz: To be honest, I want to rename `pipe` into `socket`, but I wonder if that's too confusing... In which case it will be somewhat clear that `outbound socketmap` means it's a map of currently opened outbound connections.

p2p/module.go
derrandz: They are for signaling primarily for execution and what not. When we'll start using `context.Context`, this could be replaced with a cancellable context.

p2p/module.go
derrandz: Not really, this is pre-code to the handlers functionality, this was added in a hurry to "spy" on some events such as 'BroadcastingDone'.

I wonder if any arguments are needed at all, so I've left it general until decided upon granularly.

p2p/module.go
derrandz: I suppose they are, seen them quite a bit for small structs that won't be reused much outside of a file. I always use them when declaring a type is too cumbersome. I also mostly declare structs as types when I intend to add behavior to them and when their size and responsibility is relatively significant.

p2p/module.go
derrandz: Absolutely, thank you.

p2p/module.go
derrandz: TODOing it it is

p2p/module.go
derrandz: 1- I am thinking `type signal chan int`
2- Since this is primarily used to stop the listen routine, and the listen routine basically selects off of a single channel that is `done`, it will block unless is buffered. Whereas the main listening behavior happens as part of the `default` block of the `select`. So to guarantee that the listen behavior is proper, a buffered channel of 1 sufficient to keep the `default` case going unless the max buffer that is 1 is reached, which happens immediately on `.done <- 1`

p2p/module.go
derrandz: Good tip.

p2p/module.go
derrandz: I am unclear as to what you mean, do you mean:
```
module.config = &P2PConfig{}...
```

p2p/module.go
derrandz: Agreed, this is a placeholder as some other parts of this code. I expect that PR then to remove this.

p2p/module.go
derrandz: There is a comment already. True means that the message to send will be encoded, the reason why is because there will a type of unencoded messages, that will only use wire encoding, such as ping and pong to lax the size of such re-occurrent messages.

p2p/module.go
derrandz: I think so, I primarily think about syncing in this scenario, where a node might be interested in a response to a height request. Unclear still, will require some conversation.

p2p/module.go
derrandz: I will give this some thought.

p2p/module.go
derrandz: It signifies whether the send received an ACK back or not, this is a bool until an ACK type is defined for when responses will piggyback on ACKs. (_Think height response, that comes piggybacking on the actual ack_)

prototype/Makefile
derrandz: The client that will use the p2p module. I'd like to keep them both to allow easy switching off and on.

prototype/Makefile
derrandz: There should be a client and a client_alpha. The client_beta is a slip and leftover from the git rebase I was doing. Thanks for the catch, will clean immediately.

prototype/build/node.Dockerfile
derrandz: 1- This is the node's p2p docker image that is used to the e2e test. Should we use the other dockerfile instead? Not sure, but historically I think this precedes the other files that's why it's stood there.

2- Do you think so? I think the review is doing its job so no worries as functionality wise it's what I want to submit at this point, so that future work is streamlined and this does not lag behind and cause me conflict headache.

prototype/cmd/client-with-pre-p2p/main.go
derrandz: This was a mindless copy-paste and move-around to simply add p2p and keep everything that was working, working.

I am unclear about what you said about creating the config internally. We can make a separate PR for the client optimizations and refactors.

prototype/p2p/errors.go
derrandz: lol, true.

prototype/p2p/errors.go
derrandz: To keep the error semantics, in relation to the way other errors are defined. These are error factories.

A convention to declare errors and error factories in the same way. Not a pattern.

prototype/p2p/messenger_test.go
derrandz: To allow for targeted testing using `go test -run TestMessenger` or `go test -run TestNetwork`

prototype/p2p/messenger_test.go
derrandz: What do you mean?

prototype/p2p/messenger_test.go
derrandz: Good suggestion

prototype/p2p/module.go
derrandz: Ok. I suppose that's where I lost you in our last conversation and caused me to make a conflict-free PR moving things into prototype.

In either case I am happy with this going to the prototype, as this is a prototype, I want the code that will go to the non-prototype to include a few refactors I have in mind.

prototype/p2p/network_lifecycle.go
derrandz: For raintree purposes. What the ID should look like is undefined, but I expect it to be some uint64 representation of some type of byte sequence.

prototype/p2p/network_lifecycle.go
derrandz: Not sure why this or that. Needs some thought.

prototype/p2p/network_lifecycle.go
derrandz: It's only type of message we got atm

prototype/p2p/pre_p2p/module.go
derrandz: Those were yours. Should I olshansky'em?

derrandz:
Closing to open a new PR that ships the P2P code to the TLD.

Issue #32, opened 2022-02-25
V1 Pre2P (i.e. PreP2P) Module


Olshansk:
# Objective
A stop-gap implementation of the P2P module to unblock the implementation of Consensus & Utility in a localnet environment.

# Origin Document
Add a Pre2P (i.e. PreP2P) implementation to mock allow for sending and broadcasting messages between Pocket network actors and validator nodes. This is not a mock per say, but uses Goland's native networking modules to allow the broadcasting and sending messages to other actors in the Pocket Network. Though it is a full implementation of P2P communication, it is not optimized for large networks, nor does it follow the P2P spec in any way.

In short, it'll be a port of the code that was used in [consensus-pre-prototype](https://github.com/pokt-network/consensus-pre-prototype) to unblock work in the interim and allow for a very simple substitute for P2P in a localnet.

**Creator**: @Olshansk 
**Co-Owners**: @andrewnguyen22 @derrandz 
**Deliverables**: 
- Pre2P module implementation
- Documentation on how to use, build and test

PR #33, opened 2022-02-25
V1 PreP2P Module Prototype

Olshansk:
# Description

This PR is meant to address https://github.com/pokt-network/pocket/issues/32.

It is a temporary P2P module used to enable the V1 prototype on mainline.

# Details

- Consolidated PocketEvents/NetworkEvents/PocketMessage/NetworkMessage into one proto type
- Implement `pre2p`
    - Add a Network type
    - Supports basic `Send` and `Broadcast` using the golang std library
- Use proper types for address and privateKey in `config.go`
    - Add Unmarshalling of the Private & Public keys from file
- Remove all reference to `node_id` or `NodeId` from the pre-prototype
- Change the address type from `string` to `pcrypto.Address`
- Change the topic type from `string` to `types.PocketTopic`
- Add `protogen_local` to Makefile
- Minor update to .gitignore

## TODOs as a result of the discussion PR
** (will create an issue and add these before submission)
- [ ] **[Small]** Remove `pre2p` altogether
- [ ] **[Medium]** Discuss how to handle (avoid or do) self broadcasts in `pre2p`
- [ ] **[Large]** Discuss how to handle unexpected behavior. Panic? Fail Log? Something else?
- [ ] **[Large]** Add support for dynamic validator sets
- [ ] **[Large]** Proper discussion and design around a genesis file
- [ ] **[Large]** Proper discussion and design around how to access global state (i.e. remove `test_state.go`)

## Testing

Please note that until consensus is integrated into mainline, there is no business logic,
but we can connect to the open port via `telnet` once localnet starts up.

## 1. Pull changes (if not already local):
```
$ git clone git@github.com:pokt-network/pocket.git pocket
$ cd pocket
$ gco -b issue32/prototype/pre2p origin/issue32/prototype/pre2p
```

## [Optional] 2. View available commands:
```
$ make
```

## 3. Clean environment:
```
$ make docker_wipe # Press `y` if prompted
```

### 3.1 Shell 1:
```
$ make protogen_clean
$ make protogen_local
$ make compose_and_watch
```

### 3.2 Shell 2:
```
$ telnet localhost 8081
```

## [Optional] 4.  Foundation for unit tests:
```
$ make mockgen
$ make test_all
```


pre2p/types/validator.go
```diff
@@ -0,0 +1,40 @@
+package types
+
+type NodeId uint32
+
+type ValMap map[NodeId]*Validator
+
+// TODO(olshansky): Review all the attributes of this struct.
+type Validator struct {
+	NodeId     NodeId   `json:"node_id"`
```
andrewnguyen22, line 9: So many todo's haha. 

Similar to above, I'd recommend deprecating nodeId for the public key if possible. I see a TODO to remove it above, so won't harp or block cause of this

pre2p/genesis.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"os"
+	"pocket/pre2p/types"
+	"time"
+)
+
+// TODO(olshansky): This is an interim genesis structure that will be replaced with a real one. It is the bare minimum for prototyping purposes.
+type P2PGenesis struct {
+	GenesisTime time.Time          `json:"genesis_time"`
+	AppHash     string             `json:"app_hash"`
```
andrewnguyen22, line 14: No idea why P2PGenesis needs an appHash... Can this be removed?

pre2p/genesis.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (genesis *P2PGenesis) Validate() error {
+	if genesis.GenesisTime.IsZero() {
+		return fmt.Errorf("GenesisTime cannot be zero")
+	}
+
+	if len(genesis.Validators) == 0 {
+		return fmt.Errorf("Genesis must contain at least one validator")
+	}
+
```
andrewnguyen22, line 62: If you end up keeping the appHash (not recommended) then we need validation here

pre2p/network.go
```diff
@@ -0,0 +1,89 @@
...
+		// rather than having it be "generalized" through the P2P layer.
+		if self == peer.NodeId {
+			continue
+		}
+		client, err := net.DialTCP("tcp", nil, peer.ConsensusAddr)
+		if err != nil {
+			log.Println("Error connecting to peer: ", err)
+			continue
+		}
+		client.Write(data)
```
andrewnguyen22, line 42: Need error handling. Don't forget to close the client

pre2p/network.go
```diff
@@ -0,0 +1,89 @@
...
+		// TODO(olshansky): Quick hack to avoid sending network messages to self.
+		if node != peer.NodeId {
+			continue
+		}
+		client, err := net.DialTCP("tcp", nil, peer.ConsensusAddr)
+		if err != nil {
+			log.Println("Error connecting to peer: ", err)
+			continue
+		}
+		client.Write(data)
```
andrewnguyen22, line 59: Need error handling. Don't forget to close the client

pre2p/network.go
```diff
@@ -0,0 +1,89 @@
...
+// TODO(hack): Publically exposed for testing purposes only.
+func (n *network) GetAddrBook() []*types.NetworkPeer {
+	return n.AddrBook
+}
+
+func (n *network) connectToValidator(nodeId types.NodeId, v *types.Validator) error {
+	var tcpAddr *net.TCPAddr
+	var err, errDebug error
+
+	tcpAddr, err = net.ResolveTCPAddr("tcp4", fmt.Sprintf("%s:%d", v.Host, v.Port))
```
andrewnguyen22, line 76: Move "tcp4" to constant or find standard constant

pre2p/network.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// TODO(hack): Publically exposed for testing purposes only.
+func (n *network) GetAddrBook() []*types.NetworkPeer {
+	return n.AddrBook
+}
+
+func (n *network) connectToValidator(nodeId types.NodeId, v *types.Validator) error {
+	var tcpAddr *net.TCPAddr
+	var err, errDebug error
```
andrewnguyen22, line 74: I don't understand errDebug. Doesn't look used

pre2p/handlers.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"google.golang.org/protobuf/proto"
+)
+
+func (m *p2pModule) handleNetworkMessage(conn net.Conn) {
+	defer conn.Close()
+
+	data, err := ioutil.ReadAll(conn)
+	if err != nil {
+		log.Println("Error reading from conn: ", err)
```
andrewnguyen22, line 19: Seeing some fmt.Print... and some log.Println

I know we don't yet have a standard, but seems odd to not standardize the usage in this PR

andrewnguyen22 (changes requested):

pre2p/network.go
Olshansk: tl;dr Removed.

This was from when I was exposing one port for debugging/telemetry and one for the actual communication, but removing the debug/telemetry port in this initial version.

pre2p/network.go
Olshansk: Using `const NetworkProtocol = "tcp4"` everywhere in the file now.

pre2p/network.go
Olshansk: Done.

pre2p/genesis.go
Olshansk: The hash of the last block for the V0 blockchain when we do the migration?

pre2p/network.go
Olshansk: Let me know regarding my reasoning above.

pre2p/types/message.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+import "pocket/shared/types"
+
+type NetworkMessage struct {
+	Topic types.EventTopic
+	Data  []byte
```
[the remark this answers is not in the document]
Olshansk: Resolving this comment in favor of continuing the conversation in the thread above.

pre2p/types/proto/p2p.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package p2p;
+
+option go_package = "/types/";
+
+enum PocketTopic {
```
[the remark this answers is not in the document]
Olshansk: I consolidated PocketEvent and NetworkMessage in shared/events.proto.

It's a fair number of changes but simplifies the # of messages overall. Lmk what you think.

pre2p/genesis.go
Olshansk: tl;dr Validation added.

Added validation of the appHash for the time being, but can be removed depending on the outcome of the thread above.

pre2p/types/validator.go
Olshansk: Removing NodeId altogether. It'll make the consensus PR a little more complicated but I think it's the right move this early on.

pre2p/types/network.go
```diff
@@ -0,0 +1,21 @@
+package types
+
+import (
+	"crypto"
+	"net"
+)
+
+// TODO(olshansky): Try to find a way to remove `NodeId` from the entire codebase altogether; this is a stop-gap prototype measure.
```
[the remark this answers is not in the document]
Olshansk: Removing NodeId from broadcast.

pre2p/types/validator.go
Olshansk: Note that in order to remove NodeId from everywhere, I had to change the Address type from a byte slice to a byte array (i.e. fixed length with size `AddressLen`). Lmk if that's an issue.

.gitignore
```diff
@@ -24,9 +24,11 @@ vendor/
 shared/modules/mocks/
 
 # Proto generated files
-types/
+*.pb.go
```
derrandz, line 27: Good catch

p2p/pre2p/handlers.go
```diff
@@ -0,0 +1,38 @@
...
+	if err := proto.Unmarshal(data, &networkMessage); err != nil {
+		panic(err)
+	}
+	if err != nil {
+		log.Println("Error decoding network message: ", err)
+		return
+	}
+
+	event := types.PocketEvent{
+		Module: types.SourceModule_P2P_MODULE,
```
derrandz, line 32: I wonder if a `Module` property is necessary on an event. A topic should imply enough context. However, I see how useful can this be for debugging purposes.

Noting for later, this can benefit from some thought.

p2p/pre2p/module.go
```diff
@@ -49,10 +61,57 @@ func (m *p2pModule) GetBus() modules.Bus {
...
+		return err
+	}
+
+	return m.network.NetworkSend(data, addr)
 }
 
-func (m *p2pModule) Send(addr string, msg *anypb.Any, topic string) error {
-	panic("Send not implemented")
+func (m *p2pModule) GetAddrBook() []*pre2ptypes.NetworkPeer {
+	return m.network.GetAddrBook()
```
derrandz, line 116: Good looking file.

p2p/pre2p/network.go
```diff
@@ -0,0 +1,98 @@
...
+			log.Println("Error writing to one of the peers during broadcast: ", err)
+			continue
+		}
+	}
+	return nil
+}
+
+func (n *network) NetworkSend(data []byte, node pcrypto.Address) error {
+	for _, peer := range n.AddrBook {
+		// TODO(team): If the address book is a map instead of a list, we wouldn't have to do this loop.
```
derrandz, line 56: This contains the answer to the question at [line 35](https://github.com/pokt-network/pocket/pull/33/files#diff-4c195a5d75f98cf6df9e78d6fc0766ada49a8a40af22a3c4385d3c4becc008dcR35) as a list containing only neighbors would be a self-filtering mechanism to prevent self-broadcast, as long as other peers do not re-broadcast to not fall into infinite broadcast.

shared/handler.go
```diff
@@ -6,14 +6,12 @@ import (
...
 )
 
-// TODO: Only supporting a subset of topics because not all are used.
-func (node *Node) handleEvent(event *types.Event) error {
-	switch event.PocketTopic {
-	case types.ConsensusMessage:
-		node.GetBus().GetConsensusModule().HandleMessage(event.MessageData)
+func (node *Node) handleEvent(event *types.PocketEvent) error {
+	switch event.Topic {
+	case types.PocketTopic_CONSENSUS_MESSAGE_TOPIC:
```
derrandz, line 11: As a comment, I think I prefer shorter names. `types.EVENT_TOPIC_CONSENSUS_MESSAGE`. There is quite the redundancy in `PocketTopic_...._TOPIC`.

shared/handler.go
```diff
@@ -6,14 +6,12 @@ import (
...
+		node.GetBus().GetConsensusModule().HandleMessage(event.Data)
 	default:
-		// TODO(discuss): Should we panic here?
-		log.Printf("[WARN] Unsupported event: %s \n", event.PocketTopic)
+		log.Printf("Unsupported event: %s \n", event.Topic)
```
derrandz, line 14: We should think about handling undefined behavior, including panic recovery and unexpected behaviors such as these. I am talking in general as P2P also lacks this aspect (_intentionally deferred to be thought about thoroughly, perhaps collectively?_)

shared/modules/p2p_module.go
```diff
@@ -1,18 +1,13 @@
...
-type NetworkMessage struct {
-	Topic types.EventTopic
-	Data  []byte
-}
-
 type P2PModule interface {
 	Module
-	BroadcastMessage(msg *anypb.Any, topic string) error  // TODO(derrandz): get rid of topic
-	Send(addr string, msg *anypb.Any, topic string) error // TODO(derrandz): get rid of topic
+	BroadcastMessage(msg *anypb.Any, topic types.PocketTopic) error           // TODO(derrandz): get rid of topic
```
derrandz, line 11: I think these signatures are better.

shared/modules/pocket_bus_module.go
```diff
@@ -4,14 +4,14 @@ import (
...
+type EventsChannel chan types.PocketEvent
 
 type Bus interface {
 	// TODO: Do we want to implement `Module` here as well?
 
 	// Bus Events
-	PublishEventToBus(e *types.Event)
-	GetBusEvent() *types.Event
+	PublishEventToBus(e *types.PocketEvent)
+	GetBusEvent() *types.PocketEvent
```
derrandz, line 14: (_Just a comment_): I remember a short conversation about idiomatic go. I noticed also that in the generated protobuff golang code, they do not seem to respect the idiomatic way of naming getters. Overall, it's simpler to read `GetBusEvent` than `BusEvent`. Still, I want to hear other opinions, particularly yours.

shared/types/proto/events.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package pocket;
+
+option go_package = "/types/";
+
+enum SourceModule {
```
derrandz, line 9: I think it's more clean to name this something like: `Concern' or 'Scope' than 'Module'. Since a naming like "module" is somewhat meta (_as in meta-programming_) and reverts the attention off of the actual business of the code into the code structure itself.

I think `SCOPE` is more fitting.

derrandz (approved):
Overall LGTM. Minor comments to strike conversations, with relatively strong preferential opinions about naming.

Thanks for the good work good sir 👍🏻

pre2p/genesis.go
andrewnguyen22: Yes but why for P2P? Consensus might need one.

pre2p/types/validator.go
andrewnguyen22: Actually this is a problem

Protobuf only allows []byte not fixed length

So it's either we convert all slices to byte arrays using `copy` or we change address encoding to strings

p2p/pre2p/handlers.go
Olshansk: Removing. We can always add it back later if need be.

shared/handler.go
Olshansk: This is a proto generated type from an enum so outside of our control.

shared/handler.go
Olshansk: Adding a top-level TODO at the PR level. Would you want to add it to a future protocol hour and lead the discussion?

shared/modules/pocket_bus_module.go
Olshansk: I have no strong opinions either way. I think either approach enables us to do what we need.

shared/types/proto/events.proto
Olshansk: Removing altogether until we need it again.

pre2p/types/validator.go
Olshansk: Discussed this offline with Andrew directly. The decision was:
- Leave `Address` as a slice rather than a string or array for proto compatibility and confirmation to standards
- In the map (where a key cannot be a slice), simply use a string (hex representation of the address) without needing to create an extra type.

p2p/pre2p/network.go
derrandz: I suppose what I implied since Pre2P is always going to be used in controlled environments that we establish the expectation that the neighbors list will never contain own address, making of it a natural self-filtering mechanism. That's why making it a list and not a map will be much more useful.

shared/handler.go
derrandz: Yeah correct...

shared/handler.go
derrandz: Good suggestion Shans, will do.

p2p/pre2p/network.go
Olshansk: Makes sense. We'll just need to validate that when the static list / addrbook is created /configured, self is not in there.

Added a comment in the code.

shared/handler.go
Olshansk: Thanks Zanzan

Olshansk:
@andrewnguyen22 I believe I've addressed most of the issues. Could you take another look at this to see if everything 👀 good? If so, we can submit it for @derrandz to work on top of in his branch.

andrewnguyen22 (approved):

PR #42, opened 2022-03-02
Pre-Persistence Prototype Module First Iteration

### LINKED ISSUE #23: V1 Pre-Persistence Module
# Objective
Have an in-memory database with the necessary persistence operations that happen in the application.

# Origin Document
Add a pre-persistence implementation to mock needed storage ops. This mock should both unblock module developers and be utilized to demonstrate the storage needs of each module. This is meant to inform the development of the v1 persistence module while enabling integration of core modules.

Creator: @andrewnguyen22
Co-Owners: @iajrz  
Deliverables: 
- Pre-Persistence Prototype
- How to build guide
- How to use guide
- How to test guide

andrewnguyen22:
### Description
---
- **Name**: Pre-Persistence Prototype Module
- **About**: This issue is meant to unblock other module developers by providing a simple - in-memory - persistence module, to provide a mock for others module developers to utilize before the actual persistence module is implemented.
- **Title**: 'Pre-Persistence Prototype Module First Iteration'
- **Labels**: integration
- **Assignees**: '@andrewnguyen22'

### Objective
---
Have an in-memory database with the necessary persistence operations that happen in the application.

### Origin Document
---
Add a pre-persistence implementation to mock needed storage ops. This mock should both unblock module developers and be utilized to demonstrate the storage needs of each module. This is meant to inform the development of the v1 persistence module while enabling integration of core modules.

**Creator**: @andrewnguyen22
**Co-Owners**: @iajrz

### Deliverables:
---

- Pre-Persistence Prototype
- How to build guide
- How to use guide
- How to test guide

### Automation
---
closes #23


persistence/pre_persistence/proto/account.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+package utility;
+option go_package = "/utility/test/";
+
+message Account {
+  bytes Address = 1;
```
Olshansk, line 6: I realize that in golang, we use CamelCase and capitalize exported fields, but protobuf style guide prefers snake case: https://developers.google.com/protocol-buffers/docs/style

In addition, I confirmed locally that if you name an attribute `address_foo`, it translates to `AddressFoo` when `protoc` generates the Go structs.

persistence/pre_persistence/proto/account.proto
```diff
@@ -0,0 +1,13 @@
...
+  string Amount = 2;
```
Olshansk, line 7: Please add this comment: `# Amount is stored as a string in order to support BigInts at the application layer`.

persistence/pre_persistence/proto/account.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+package utility;
```
Olshansk, line 2: This is in pre_persistence. Is the package what you intend?

persistence/pre_persistence/proto/account.proto
```diff
@@ -0,0 +1,13 @@
...
+option go_package = "/utility/test/";
+
+message Account {
+  bytes Address = 1;
+  string Amount = 2;
+}
+
+message Pool {
```
Olshansk, line 10: Can you add a one line comment (if my understanding is correct): `// The in memory database will store a pool for every Account in the network keyed by it's Name`.

What is the name here?

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+	s, err := StringToBigInt(p.Account.Amount)
+	if err != nil {
+		return err
+	}
+	s2, err := StringToBigInt(amount)
+	if err != nil {
+		return err
+	}
+	s.Sub(s, s2)
```
Olshansk, line 72: 1. Do we want to add a check to make sure that s is greater than s2 before subtracting?
1.1 If `yes`, this might warrant adding a simple unit test since it seems to have fallen through the cracks.
1.2 If `not`, ignore this comment altogether.

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+	s.Add(s, s2)
+	p.Account.Amount = BigIntToString(s)
+	bz, err := cdc.Marshal(&p)
+	if err != nil {
+		return err
+	}
+	return db.Put(key, bz)
+}
+
+func (m *PrePersistenceContext) SubtractPoolAmount(name string, amount string) error {
```
Olshansk, line 51: The only difference between this and the function above is one line so you could create a helper and remove copy-pasted code. Tested here: https://go.dev/play/p/7pB70u5dWXS
Example:

```
func (m *PrePersistenceContext) AddPoolAmount(name string, amount string) error {
	add := func(s *big.Int, s1 *big.Int) {
		s.Add(s, s1)
	}
	m.operationPoolAmount(name, amount, add)
}

func (m *PrePersistenceContext) operationPoolAmount(name string, amount string, op func(*big.Int, *big.Int)) error {
	cdc := Cdc()
	p := Pool{}
	db := m.Store()
	key := append(PoolPrefixKey, []byte(name)...)
	val, err := db.Get(key)
	if err != nil {
		return err
	}
	err = cdc.Unmarshal(val, &p)
	if err != nil {
		return err
	}
	s, err := StringToBigInt(p.Account.Amount)
	if err != nil {
		return err
	}
	s2, err := StringToBigInt(amount)
	if err != nil {
		return err
	}
	op(s, s2)
	p.Account.Amount = BigIntToString(s)
	bz, err := cdc.Marshal(&p)
	if err != nil {
		return err
	}
	return db.Put(key, bz)
}
```

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func KeyFromHeightKey(heightKey []byte) (key []byte) {
+	k := strings.SplitN(string(heightKey), "/", 2)[1]
+	return []byte(k)
+}
+
+func PrefixEndBytes(prefix []byte) []byte {
+	if len(prefix) == 0 {
+		return nil
```
Olshansk, line 289: Should we panic or log an error here? Is there ever a use case where you call this with an empty length prefix?

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+	keyString := fmt.Sprintf("%s/%s", elenEncoder.EncodeInt(int(height)), k)
+	return []byte(keyString)
+}
+
+func KeyFromHeightKey(heightKey []byte) (key []byte) {
+	k := strings.SplitN(string(heightKey), "/", 2)[1]
+	return []byte(k)
+}
+
+func PrefixEndBytes(prefix []byte) []byte {
```
Olshansk, line 287: Since this function has no tests, I think a simple comment showing an example input & output, or an explanation would make it easier to understand.

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+	key := append(PoolPrefixKey, []byte(name)...)
+	bz, err := cdc.Marshal(&p)
+	if err != nil {
+		return err
+	}
+	return db.Put(key, bz)
+}
+
+func (m *PrePersistenceContext) SetPoolAmount(name string, amount string) error {
+	cdc := Cdc()
```
Olshansk, line 100: To discuss: per my comment above, I think it is pretty easy to generalize out a lot of the boilerplate in every function into a helper.

Olshansk (changes requested):
I took an initial stab and reviewed a small portion (10% or less of the PR) and will likely continue doing it in small bits given its size.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+		return nil, false, err
+	}
+	if bz == nil {
+		return nil, false, nil
+	}
+	if bytes.Contains(bz, DeletedPrefixKey) {
+		return nil, false, nil
+	}
+	if err = proto.Unmarshal(bz, app); err != nil {
+		return nil, true, err
```
Olshansk, line 46: should you return `false` here?

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	return app, true, nil
+}
+
+func (m *PrePersistenceContext) GetAllApps(height int64) (apps []*App, err error) {
+	cdc := Cdc()
+	apps = make([]*App, 0)
+	var it iterator.Iterator
+	if height == m.Height {
```
Olshansk, line 55: TODO: making a request to discuss this offline. Don't fully understand the design/scope.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+			return nil, err
+		}
+		apps = append(apps, &a)
+	}
+	return
+}
+
+func (m *PrePersistenceContext) InsertApplication(address []byte, publicKey []byte, output []byte, paused bool, status int, maxRelays string, stakedTokens string, chains []string, pausedHeight int64, unstakingHeight int64) error {
+	if _, exists, _ := m.GetApp(address); exists {
+		return fmt.Errorf("already exists in world state")
```
Olshansk, line 86: Consider putting the address or publicKey in the error message.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+func (m *PrePersistenceContext) UpdateApplication(address []byte, maxRelaysToAdd string, amountToAdd string, chainsToUpdate []string) error {
+	app, exists, _ := m.GetApp(address)
+	if !exists {
+		return fmt.Errorf("does not exist in world state")
+	}
+	cdc := Cdc()
+	db := m.Store()
+	key := append(AppPrefixKey, address...)
+	// compute new values
+	stakedTokens, err := StringToBigInt(app.StakedTokens)
```
Olshansk, line 119: What you're doing here is adding "two strings" which represent big ints.

IMO this warrants a helper that can easily be reused to shorten the code and make it more readable.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	maxRelaysToAddI, err := StringToBigInt(maxRelaysToAdd)
+	if err != nil {
+		return err
+	}
+	maxRelays.Add(maxRelays, maxRelaysToAddI)
+	// update values
+	app.MaxRelays = BigIntToString(maxRelays)
+	app.StakedTokens = BigIntToString(stakedTokens)
+	app.Chains = chainsToUpdate
+	// marshal
```
Olshansk, line 141: NIT: I don't really think we need this comment

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	// compute new values
+	stakedTokens, err := StringToBigInt(app.StakedTokens)
+	if err != nil {
+		return err
+	}
+	stakedTokensToAddI, err := StringToBigInt(amountToAdd)
+	if err != nil {
+		return err
+	}
+	stakedTokens.Add(stakedTokens, stakedTokensToAddI)
```
Olshansk, line 127: Is there validation somewhere that `stakedTokensToAddI` is non-negative?

If not, this goes against this spec.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	bz, err := cdc.Marshal(app)
+	if err != nil {
+		return err
+	}
+	return db.Put(key, bz)
+}
+
+func (m *PrePersistenceContext) DeleteApplication(address []byte) error {
+	if exists, _ := m.GetAppExists(address); !exists {
+		return fmt.Errorf("does not exist in world state")
```
Olshansk, line 151: Add address to the error message.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	db := m.Store()
+	key := append(AppPrefixKey, address...)
+	return db.Put(key, DeletedPrefixKey)
+}
+
+func (m *PrePersistenceContext) GetAppsReadyToUnstake(height int64, status int) (apps []*types.UnstakingActor, err error) { // TODO delete unstaking
```
Olshansk, line 158: Is the TODO outdated? Not sure what its intent is here.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	if err := proto.Unmarshal(val, &unstakingApps); err != nil {
+		return nil, err
+	}
+	for _, app := range unstakingApps.UnstakingActors {
+		apps = append(apps, app)
+	}
+	return
+}
+
+func (m *PrePersistenceContext) GetAppStatus(address []byte) (status int, err error) {
```
Olshansk, line 181: I think status should be an enum, because as a reviewer it's hard for me to understand the range of values this can take.

Feel free to add a TODO for this.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	app, exists, err := m.GetApp(address)
+	if err != nil {
+		return ZeroInt, err
+	}
+	if !exists {
+		return ZeroInt, fmt.Errorf("does not exist in world state")
```
Olshansk, line 187: Add hex encoded address to error.

Will stop leaving similar comments below.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	app.Status = int32(status)
+	// marshal
+	bz, err := cdc.Marshal(app)
+	if err != nil {
+		return err
+	}
+	if err := db.Put(key, bz); err != nil {
+		return err
+	}
+	unstakingKey := append(UnstakingAppPrefixKey, []byte(fmt.Sprintf("%d", unstakingHeight))...)
```
Olshansk, line 214: Use `strconv.Itoa` instead of `fmt.Sprintf` here.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+	if found := db.Contains(unstakingKey); found {
```
Olshansk, line 215: I don't fully understand the business logic here...

We're trying to unstake an app above, but later find out that it is already being unstaked in the DB, so we are appending the unstaking actor? Shouldn't there only ever be exactly one unstaking actor corresponding to a specific address?

Lmk if we should take this offline.

persistence/pre_persistence/app.go
```diff
@@ -0,0 +1,311 @@
...
+		OutputAddress: app.Output,
+	})
+	unstakingBz, err := cdc.Marshal(&unstakingApps)
+	if err != nil {
+		return err
+	}
+	return db.Put(unstakingKey, unstakingBz)
+}
+
+func (m *PrePersistenceContext) GetAppPauseHeightIfExists(address []byte) (int64, error) {
```
Olshansk, line 236: NIT: Height should be `uint64`.

Might warrant a refactor TODO.

Olshansk (changes requested):

persistence/pre_persistence/module.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"github.com/pokt-network/pocket/shared/config"
+	"github.com/pokt-network/pocket/shared/modules"
+	"github.com/pokt-network/pocket/shared/types"
+	"github.com/syndtr/goleveldb/leveldb/comparer"
+	"github.com/syndtr/goleveldb/leveldb/memdb"
+)
+
+func Create(cfg *config.Config) (modules.PersistenceModule, error) {
+	db := memdb.New(comparer.DefaultComparer, 888888888)
```
derrandz, line 12: (_Silly question_) Can this come from the config? If it's important at all it should.

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+	Cfg *config.Config
+}
+
+func NewPrePersistenceModule(commitDB *memdb.DB, mempool types.Mempool, cfg *config.Config) *PrePersistenceModule {
+	return &PrePersistenceModule{CommitDB: commitDB, Mempool: mempool, Cfg: cfg}
+}
+
+func (m *PrePersistenceModule) NewContext(height int64) (modules.PersistenceContext, error) {
+	newDB := NewMemDB()
+	it := m.CommitDB.NewIterator(&util.Range{Start: HeightKey(height, nil), Limit: HeightKey(height+1, nil)})
```
derrandz, line 52: (_minor comment_) I think for long pieces of code like this, it's very readable to put it on multiple lines:
```
it := m.CommitDb.NewIterator(&util.Range{
    Start: HeightKey(height, nil),
    Limint: HeightKey(height_1, nil),
})
```
and only fully use the for semantics:
```
defer it.Release()
for it.First(); it.Valid(); it.Next() {
  // ...
}
```

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+type PrePersistenceContext struct {
+	Height     int64
+	Parent     modules.PersistenceModule
+	SavePoints map[string]int // TODO save points not entirely implemented. Happy path only for now, rollbacks for later
+	DBs        []*memdb.DB
+}
+
+func (m *PrePersistenceContext) GetLatestBlockHeight() (uint64, error) {
```
derrandz, line 81: _Educational question_: why does the height have to be a `uint64`

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+		delete(m.SavePoints, key)
+	}
+	// rollback
+	m.DBs = m.DBs[:rollbackIndex]
+	return nil
+}
+
+// AppHash creates a unique hash based on the global state object
+// NOTE: AppHash is an inefficient, arbitrary, mock implementation that enables the functionality
+// TODO written for replacement, taking any and all better implementation suggestions - even if a temporary measure
```
derrandz, line 161: Assign this to either `team` or `self`

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+func (m *PrePersistenceContext) AppHash() ([]byte, error) {
+	result := make([]byte, 0)
+	index := len(m.DBs) - 1
+	db := m.DBs[index]
+	it := db.NewIterator(&util.Range{})
+	it.First()
+	for ; it.Valid(); it.Next() {
```
derrandz, line 168: Same minor comment about using the `for` semantics to the fullest

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	for _, key := range toDelete {
+		delete(m.SavePoints, key)
+	}
+	// rollback
+	m.DBs = m.DBs[:rollbackIndex]
+	return nil
+}
+
+// AppHash creates a unique hash based on the global state object
```
derrandz, line 159: Is `based on the global state` a requirement or a measure to guarantee maximum entropy?

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+	return m.RollbackToSavePoint(FirstSavePointKey)
+}
+
+// Commit the KV pairs to the parent (commit) db
+func (m *PrePersistenceContext) Commit() error {
+	index := len(m.DBs) - 1
+	db := m.DBs[index]
+	it := db.NewIterator(&util.Range{})
+	it.First()
+	for ; it.Valid(); it.Next() {
```
derrandz, line 191: 👆🏻 same

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+	return nil
+}
+
+func (m *PrePersistenceContext) Release() {
+	m.SavePoints = nil
+	for _, db := range m.DBs {
+		db.Reset()
+	}
+	m.DBs = nil
+	return
```
derrandz, line 221: (_minor comment_) Would be great to see this also reflected on a flag to prevent double releasing/useless releasing and also to allow for more fine-tuned testing and usage.

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+		err := dest.Put(it.Key(), it.Value())
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+var (
+	elenEncoder = lexnum.NewEncoder('=', '-')
```
derrandz, line 274: Can go to the top for clarity's sake

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+)
+
+func HeightKey(height int64, k []byte) (key []byte) {
```
derrandz, line 277: Actually perhaps all of this can live in a separate `height_utils.go` file?

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+	keyString := fmt.Sprintf("%s/%s", elenEncoder.EncodeInt(int(height)), k)
+	return []byte(keyString)
+}
+
+func KeyFromHeightKey(heightKey []byte) (key []byte) {
+	k := strings.SplitN(string(heightKey), "/", 2)[1]
```
derrandz, line 283: (_Just noting_) Good example of code whose failure we should look into how to handle. Recovery from panics and error returning?

derrandz (commented):
First small review, will look into continuing this FTMOR (_first thing in the morning_) as this is EOD for me.

I'm hestitant to request changes as those are only minor comments, I'll submit this review to start the discussion until I get the change to look at the full module.

Olshansk:
@andrewnguyen22 I never finished going through all the files when first reviewing this.

Should we hold off on reviewing the rest until you tend to the first round of comments or continue in parallel?

persistence/pre_persistence/proto/account.proto
andrewnguyen22: I assume there's no actionable here in this PR. Something to investigate

persistence/pre_persistence/account.go
andrewnguyen22: This is checked at the Utility level (an unfortunate byproduct of piece mealing the PR), but if we want to reject it at the persistence layer we could as well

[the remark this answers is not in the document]
andrewnguyen22: This is checked at the Utility level (an unfortunate byproduct of piece mealing the PR), but if we want to reject it at the persistence layer we could as well

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+	s.Sub(s, s2)
+	p.Account.Amount = BigIntToString(s)
+	bz, err := cdc.Marshal(&p)
+	if err != nil {
+		return err
+	}
+	return db.Put(key, bz)
+}
+
+func (m *PrePersistenceContext) InsertPool(name string, address []byte, amount string) error {
```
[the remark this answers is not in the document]
andrewnguyen22: The assumption was to make little assumptions about the capabilities and efficiencies of the db engine

persistence/pre_persistence/account.go
andrewnguyen22: Maybe we can work together to get there. The comment is a bit too generic for me to take action

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+			Limit: PrefixEndBytes(PoolPrefixKey),
+		})
+	} else {
+		key := HeightKey(height, PoolPrefixKey)
+		it = m.Parent.GetCommitDB().NewIterator(&util.Range{
+			Start: key,
+			Limit: PrefixEndBytes(key),
+		})
+	}
+	it.First()
```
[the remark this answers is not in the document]
andrewnguyen22: Okay this makes sense

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+		}
+		pools = append(pools, &p)
+	}
+	return
+}
+
+func (m *PrePersistenceContext) AddAccountAmount(address []byte, amount string) error {
+	cdc := Cdc()
+	account := Account{
+		Amount: BigIntToString(big.NewInt(0)),
```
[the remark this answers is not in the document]
andrewnguyen22: But what happens if we change the BigIntToString protocol?

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+	if x == nil {
+		return types.ErrEmptyAccount()
+	}
+	if x.Address == nil {
+		return types.ErrEmptyAddress()
+	}
+	if len(x.Address) != crypto.AddressLen {
+		return types.ErrInvalidAddressLen(crypto.ErrInvalidAddressLen())
+	}
+	amount := big.NewInt(0)
```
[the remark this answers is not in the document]
andrewnguyen22: Good catch

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+		return types.ErrInvalidAmount()
+	}
+	return nil
+}
+
+func (x *Account) SetAddress(address crypto.Address) types.Error {
+	if x == nil {
+		return types.ErrEmptyAccount()
+	}
+	if len(x.Address) != crypto.AddressLen {
```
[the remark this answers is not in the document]
andrewnguyen22: Nice catch!

persistence/pre_persistence/account.pb.go
```diff
@@ -0,0 +1,225 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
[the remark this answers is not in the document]
andrewnguyen22: This was automatic through protoc. Do you have a recommendation on how to make this go away?

persistence/pre_persistence/account_test.go
```diff
@@ -0,0 +1,107 @@
...
+const (
+	testPoolName  = "TEST_POOL"
+	testPoolName2 = "TEST_POOL2"
+)
+
+// NOTE: Pools encapsulate `accounts` so the functionality is tested
+func TestAddPoolAmount(t *testing.T) {
+	ctx := NewTestingPrePersistenceContext(t)
+	initialBalanceBig := big.NewInt(0)
+	initialBalance := BigIntToString(initialBalanceBig)
```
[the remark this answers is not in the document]
andrewnguyen22: See my comment about changing string protocol

persistence/pre_persistence/account_test.go
```diff
@@ -0,0 +1,107 @@
...
+	addedBalanceBig := big.NewInt(1)
+	addedBalance := BigIntToString(addedBalanceBig)
+	expectedBalanceBig := initialBalanceBig.Add(initialBalanceBig, addedBalanceBig)
+	expectedBalance := BigIntToString(expectedBalanceBig)
+	if err := ctx.InsertPool(testPoolName, nil, initialBalance); err != nil {
```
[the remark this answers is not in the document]
andrewnguyen22: Cool will do

persistence/pre_persistence/app.go
andrewnguyen22: In the utility module, I'll add one here too

persistence/pre_persistence/app.go
andrewnguyen22: It was to delete the unstaking status parameter. Made it more explicit

persistence/pre_persistence/app.go
andrewnguyen22: That doesn't Jive well with the intermodule primitive design strategy

persistence/pre_persistence/app.go
andrewnguyen22: Why should that error checking be here?

persistence/pre_persistence/app.go
andrewnguyen22: Hmmm is that the proper method for int64?

persistence/pre_persistence/app.go
andrewnguyen22: Let's take it offline

persistence/pre_persistence/app.go
andrewnguyen22: This requires discussion. Tendermint uses int64

persistence/pre_persistence/fisherman.go
```diff
@@ -0,0 +1,302 @@
...
+	fish.Status = int32(status)
+	// marshal
+	bz, err := cdc.Marshal(fish)
+	if err != nil {
+		return err
+	}
+	if err := db.Put(key, bz); err != nil {
+		return err
+	}
+	unstakingKey := append(UnstakingFishermanPrefixKey, []byte(fmt.Sprintf("%d", unstakingHeight))...)
```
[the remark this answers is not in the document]
andrewnguyen22: That logic feels dangerously close to state level validation which feels like Utility module scope

persistence/pre_persistence/fisherman.go
```diff
@@ -0,0 +1,302 @@
...
+		OutputAddress: fish.Output,
+	})
+	unstakingBz, err := cdc.Marshal(&unstakingActors)
+	if err != nil {
+		return err
+	}
+	return db.Put(unstakingKey, unstakingBz)
+}
+
+func (m *PrePersistenceContext) GetFishermanPauseHeightIfExists(address []byte) (int64, error) {
```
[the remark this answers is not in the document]
andrewnguyen22: Should return err on 229! good catch

persistence/pre_persistence/fisherman.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return ZeroInt, nil
+	}
+	if !exists {
+		return ZeroInt, fmt.Errorf("does not exist in world state")
+	}
+	return int64(fish.PausedHeight), nil
+}
+
+func (m *PrePersistenceContext) SetFishermansStatusAndUnstakingHeightPausedBefore(pausedBeforeHeight, unstakingHeight int64, status int) error {
```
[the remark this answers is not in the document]
andrewnguyen22: Agreed for this function, the others are hopefully pretty self-explanatory

persistence/pre_persistence/module.go
andrewnguyen22: Yes, will add

persistence/pre_persistence/module.go
```diff
@@ -0,0 +1,53 @@
...
+	state := GetTestState()
+	state.LoadStateFromConfig(cfg)
+	return NewPrePersistenceModule(db, types.NewMempool(10000, 10000), cfg), nil
```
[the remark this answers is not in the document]
andrewnguyen22: Yep

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+
+	"github.com/jordanorelli/lexnum"
+	"github.com/syndtr/goleveldb/leveldb/comparer"
+	"github.com/syndtr/goleveldb/leveldb/memdb"
+	"github.com/syndtr/goleveldb/leveldb/util"
+	"google.golang.org/protobuf/proto"
+)
+
+var (
+	FirstSavePointKey                                        = []byte("first_savepoint_key")
```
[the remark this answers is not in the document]
andrewnguyen22: They could be consts, but they need to be converted to []byte array anyway. Will make them consts

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	context.DBs = append(context.DBs, newDB)
+	return context, nil
+}
+
+func (m *PrePersistenceModule) GetCommitDB() *memdb.DB {
+	return m.CommitDB
+}
+
+type PrePersistenceContext struct {
```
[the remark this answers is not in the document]
andrewnguyen22: added TODO, will take another look once Utility is integrated

persistence/pre_persistence/persistence.go
andrewnguyen22: As opposed to what? A string? Height is a commonly used thing and we want to strike a balance between space saving (for blockchain storage) & future proofing

persistence/pre_persistence/persistence.go
andrewnguyen22: Yes

persistence/pre_persistence/persistence.go
andrewnguyen22: Not clear on the actionable here

persistence/pre_persistence/service_node.go
```diff
@@ -0,0 +1,347 @@
...
+	it.First()
+	defer it.Release()
+	for ; it.Valid(); it.Next() {
+		bz := it.Value()
+		if bytes.Contains(bz, DeletedPrefixKey) {
+			continue
+		}
+		sn := ServiceNode{}
+		if err := cdc.Unmarshal(bz, &sn); err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
andrewnguyen22: Is there a big disadvantage to using defer release?

persistence/pre_persistence/validator.go
```diff
@@ -0,0 +1,365 @@
...
+		OutputAddress: validator.Output,
+	})
+	unstakingBz, err := cdc.Marshal(&unstakingActors)
+	if err != nil {
+		return err
+	}
+	return db.Put(unstakingKey, unstakingBz)
+}
+
+func (m *PrePersistenceContext) GetValidatorPauseHeightIfExists(address []byte) (int64, error) {
```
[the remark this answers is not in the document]
andrewnguyen22: Likely, but in V0 we took the stance of keeping them separate so that we don't pigeonhole us into the actor specific functionality. To me, the current separation is the 'safer' approach right now, even though it's more code and not optimized

go.mod
```diff
@@ -3,23 +3,11 @@ module github.com/pokt-network/pocket
 go 1.17
 
 require (
-	github.com/golang/mock v1.6.0
```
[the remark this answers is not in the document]
andrewnguyen22: It shows up unused on my PR

persistence/pre_persistence/persistence.go
```diff
@@ -0,0 +1,306 @@
...
+// TODO written for replacement, taking any and all better implementation suggestions - even if a temporary measure
+func (m *PrePersistenceContext) AppHash() ([]byte, error) {
+	result := make([]byte, 0)
+	index := len(m.DBs) - 1
+	db := m.DBs[index]
+	it := db.NewIterator(&util.Range{})
+	it.First()
+	for ; it.Valid(); it.Next() {
+		result = append(result, it.Value()...)
+		// chunk into 100000 byte segments
```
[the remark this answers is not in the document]
andrewnguyen22: I need some help with this. Is this an immediate edit or just a comment? Can discuss offline

andrewnguyen22:
> @andrewnguyen22 I never finished going through all the files when first reviewing this.
> 
> Should we hold off on reviewing the rest until you tend to the first round of comments or continue in parallel?

I am off Utility now, and am ready to address stuff. Just did first round of comments

persistence/pre_persistence/account.pb.go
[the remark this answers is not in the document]
Olshansk: Yea, if you add we should add `*.pb.go` to `.gitignore`. 

If it's already there, but the protoc generated files are still included, you can remove them with `git rm --cached <file_name>`.

The makefile targets I added (which we will move to mage later) also help with cleaning and regenerating them. Lmk if you want to sync offline.

persistence/pre_persistence/proto/account.proto
Olshansk: The actionable would be to change `bytes Address = 1;` to `bytes address = 1;` in the *.proto files and the *.go files would not need to be changed.

persistence/pre_persistence/account.go
Olshansk: Got it, sgtm.

persistence/pre_persistence/account.go
Olshansk: I left a code block example in one of my previous comments: https://github.com/pokt-network/pocket/pull/42#discussion_r818120191

persistence/pre_persistence/fisherman.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return ZeroInt, err
+	}
+	if !exists {
+		return ZeroInt, fmt.Errorf("does not exist in world state")
+	}
+	return int(fish.Status), nil
+}
+
+func (m *PrePersistenceContext) SetFishermanUnstakingHeightAndStatus(address []byte, unstakingHeight int64, status int) error {
```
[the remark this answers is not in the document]
Olshansk: Figured it's worth pointing out that postgres does [support enums](https://www.postgresql.org/docs/9.1/datatype-enum.html), but It would require a lengthier discussion as to when/if/how the int <> enum conversion is done.

persistence/pre_persistence/util.go
```diff
@@ -0,0 +1,24 @@
+package pre_persistence
+
+import (
+	"github.com/pokt-network/pocket/shared/types"
+	"math/big"
+)
+
+const (
+	ZeroInt     = 0
```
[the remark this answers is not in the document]
Olshansk: Unnecessary abstraction? DId you do this to make the code easier to read?

persistence/pre_persistence/app.go
Olshansk: Makes sense to me.

persistence/pre_persistence/app.go
Olshansk: Not sure what we should use as the source of truth but [this](https://golangdocs.com/golang-int64-to-string-conversion) and [this](https://yourbasic.org/golang/convert-int-to-string/) suggest the use of `Itoa`. It seems weird for me to use string formatting when it's just a single string conversion.

persistence/pre_persistence/app.go
Olshansk: Added it to our 1:1 notion doc.

persistence/pre_persistence/app.go
Olshansk: Added to offline discussion list: https://www.notion.so/Andrew-Olshansky-1-1-d7d24cc5c6d54c3bad1f5c5af7b08b91#86be5aee6e0641029240125ceed7fbf3

persistence/pre_persistence/account.go
andrewnguyen22: Yes, and I used that code where commented for the add and subtract functions.

But specifically for the setter, what kind of 'boilerplate' are you thinking for 'every function'? It's a bit too generic for met to take any action

persistence/pre_persistence/account.pb.go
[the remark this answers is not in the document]
andrewnguyen22: Oh sorry, I thought you were saying that the comment shouldn't be generated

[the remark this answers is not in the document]
andrewnguyen22: It seems weird not to commit the generated files when they're needed for compilation... I can understand if we settle and agree upon a make file, but until we do I think we shouldn't change this

persistence/pre_persistence/util.go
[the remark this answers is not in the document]
andrewnguyen22: Yes and incase we change the param values to non integers could switch it to big.Int{}

persistence/pre_persistence/app.go
andrewnguyen22: I agree, but it requires a casting. You can't just use `strcov.Itoa()` on int64. 

So asking if casting is the best way to do it

persistence/pre_persistence/proto/account.proto
andrewnguyen22: Okay makes sense

persistence/pre_persistence/util.go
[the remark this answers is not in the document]
andrewnguyen22: Is there also an issue with `EmptyString = ""`

persistence/pre_persistence/account_test.go
```diff
@@ -0,0 +1,107 @@
...
+		t.Fatal(err)
+	}
+	if err := ctx.AddPoolAmount(testPoolName, addedBalance); err != nil {
+		t.Fatal(err)
+	}
+	actualBalance, err := ctx.GetPoolAmount(testPoolName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if actualBalance != expectedBalance {
```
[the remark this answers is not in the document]
andrewnguyen22: I don't think this is a good reason to block this PR since we don't have formal testing rules implemented

persistence/pre_persistence/app.go
andrewnguyen22: Let's do it

persistence/pre_persistence/persistence.go
andrewnguyen22: This seems like a preference rather than a needed change. I'd like to resolve this unless you feel strongly

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,377 @@
...
+const (
+	ServiceNodeStakePoolName = "SERVICE_NODE_STAKE_POOL"
+	AppStakePoolName         = "APP_STAKE_POOL"
+	ValidatorStakePoolName   = "VALIDATOR_STAKE_POOL"
+	FishermanStakePoolName   = "FISHERMAN_STAKE_POOL"
+	DAOPoolName              = "DAO_POOL"
+	FeePoolName              = "FEE_POOL"
+)
+
+func (m *PrePersistenceContext) AddPoolAmount(name string, amount string) error {
```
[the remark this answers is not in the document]
andrewnguyen22: Bump @iajrz

persistence/pre_persistence/account.go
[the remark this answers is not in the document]
andrewnguyen22: Bump @iajrz

persistence/pre_persistence/persistence.go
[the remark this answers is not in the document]
andrewnguyen22: Bump @iajrz

persistence/pre_persistence/service_node.go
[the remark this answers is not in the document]
andrewnguyen22: Bump @iajrz

persistence/pre_persistence/util.go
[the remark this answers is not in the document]
andrewnguyen22: Bump

persistence/pre_persistence/account_test.go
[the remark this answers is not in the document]
andrewnguyen22: Okay to resolve?

persistence/pre_persistence/fisherman.go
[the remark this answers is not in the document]
iajrz: Yup, this will take shape as the code evolves and adapts to datastores; at this point the observation and having it in our collective psyche is good enough.

persistence/pre_persistence/fisherman.go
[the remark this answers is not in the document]
iajrz: Indeed; the shape of validation will grow with the code, this is not a blocker, just a comment for reflection.

persistence/pre_persistence/persistence.go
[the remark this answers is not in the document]
iajrz: We agree that this is a placeholder for a placeholder.

iajrz (approved):

persistence/pre_persistence/account.pb.go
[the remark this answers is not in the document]
Olshansk: Going to resolve this one myself and push.

persistence/pre_persistence/proto/account.proto
Olshansk: Going to use my script to update these myself locally.

shared/types/mempool.go
```diff
@@ -0,0 +1,142 @@
...
+	hash := crypto.SHA3Hash(tx)
+	hashString := hex.EncodeToString(hash)
+	if _, ok := f.hashMap[hashString]; ok {
+		return ErrDuplicateTransaction()
+	}
+	f.pool.PushBack(tx)
+	f.hashMap[hashString] = struct{}{}
+	f.size++
+	f.transactionBytes += len(tx)
+	for f.size >= f.maxTransactions || f.transactionBytes >= f.maxTransactionsBytes {
```
Olshansk, line 58: Maybe we should check this first before pushing into the pool so we don't need to do an unnecessary pop?

shared/types/mempool.go
```diff
@@ -0,0 +1,142 @@
...
+func (f *FIFOMempool) PopTransaction() (tx []byte, size int, err Error) {
+	tx, err = popTransaction(f)
+	if err != nil {
+		return
+	}
+	size = len(tx)
+	return
+}
+
+func (f *FIFOMempool) Flush() {
```
Olshansk, line 104: Flush (in my mind) implies "write or dump" somewhere, this is more of a `Clear`. Rename?

persistence/pre_persistence/account.go
Olshansk: After reviewing utility, I think the verbosity is okay, especially since it's pre-persistence.

We can generalize things over time (in utility) as we see fit.

persistence/pre_persistence/account.go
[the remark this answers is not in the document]
Olshansk: Was this resolved offline?

persistence/pre_persistence/account_test.go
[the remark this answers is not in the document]
Olshansk: Yup, I'm going to have a script that auto refactors this.

Going to add an AI on myself.

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,357 @@
...
+	bz, err := codec.Marshal(&account)
+	if err != nil {
+		return err
+	}
+	return db.Put(key, bz)
+}
+
+func (m *PrePersistenceContext) GetAllAccounts(height int64) (accs []*Account, err error) {
+	codec := GetCodec()
+	accs = make([]*Account, 0)
```
Olshansk, line 244: Can we remove named params if we initialize these as well?

persistence/pre_persistence/account.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	amount := &big.Int{}
+	if _, ok := amount.SetString(x.Amount, 10); !ok {
+		return types.ErrInvalidAmount()
+	}
+	return nil
+}
+
+func (x *Account) SetAddress(address crypto.Address) types.Error {
+	if x == nil {
```
Olshansk: How can this ever be nill if we called it on an instance of the struct? Ditto below.

persistence/pre_persistence/block.go
```diff
@@ -0,0 +1,57 @@
+package pre_persistence
+
+import (
```
Olshansk, line 3: Should we add a TODO to move this into shared? Just add it in if so, otherwise I might need to understand more.

persistence/pre_persistence/codec.go
```diff
@@ -0,0 +1,54 @@
+package pre_persistence
+
+import (
```
Olshansk, line 3: Should we add a TODO to move this into shared? Just add it in if so, otherwise I might need to understand more.

persistence/pre_persistence/service_node.go
```diff
@@ -0,0 +1,346 @@
...
+	stakedTokensToAddI, err := StringToBigInt(amountToAdd)
+	if err != nil {
+		return err
+	}
+	stakedTokens.Add(stakedTokens, stakedTokensToAddI)
+	// update values
+	sn.ServiceUrl = serviceURL
+	sn.StakedTokens = BigIntToString(stakedTokens)
+	sn.Chains = chains
+	// marshal
```
Olshansk, line 113: NIT: I think these comments are unnecessary since the code is self exlpanatory.

persistence/pre_persistence/validator.go
```diff
@@ -0,0 +1,365 @@
...
+		key := HeightKey(height, ValidatorPrefixKey)
+		it = m.Parent.GetCommitDB().NewIterator(&util.Range{
+			Start: key,
+			Limit: PrefixEndBytes(key),
+		})
+	}
+	defer it.Release()
+	for valid := it.First(); valid; valid = it.Next() {
+		bz := it.Value()
+		//if bz == nil {
```
Olshansk, line 53: Remove?

Olshansk (changes requested):
Putting together a longer comment with followup action items, but PTAL at the changes and new comments.

Olshansk (approved):
I have a list of a few comments that I'll create an issue for but otherwise LGTM!

persistence/pre_persistence/persistence.go
derrandz: Never made it in time here. This will depend on how often do we reset, but if we reset once, we can do this:

```
if: already reset? return
else: iterate and release.
```

The persistence context can store that flag.

Issue #40, opened 2022-03-02
V1 Repurpose #31 to migrate the P2P Code to the TLD


derrandz:
### Description
---

Following a slight misunderstanding from my side, I've went a head and put all the p2p module code under the prototype directory in the repository, imitating #27 to avoid conflicts, which was intended only for snapshotting purposes.

After the review and a discussion with the team, we saw that it was more fitting and appropriate that the p2p module impacts the TLD directly.

### Acceptance Criteria
---

- [x] A module conforming to the module architecture established for V1
- [ ] Runnable passing tests for all the concerns of the module (unit and integration)
- [ ] Integration with the development client

### Deliverables
---

- [ ] a P2P module under `${TLD}/p2p` conforming to the v1 module architecture containing:
    - [x] Network IO and Socket capabilities: Read/Write/Poll/Answer
    - [ ] Network behavior and communication capabilities: Send/Broadcast/Ping/Pong/Handshake
    - [ ] Network structure capabilities: Raintree
- [ ] Unit tests for each networking concern/capabilitiy
   - [ ] `network_test.go` (_communication_)
   - [x] `socket_test.go` (_io and socket_)
   - [ ] `raintree_test.go` (_structure_)
- [ ] A development client version spinning 5 pocket nodes with this p2p module.

----

derrandz:
This has closed #31

derrandz:
As of our last team discussion around p2p, we decided to deprecate the P2P module in favor of a new version that will be spun out of a refactored Pre2P. Thus we will transfer all the learnings from the previous P2P module into the new one, which will be the mold of Pre2P and P2P.

PR #46, opened 2022-03-14
V1 P2P Module - First Iteration

derrandz:
### Description
----

The p2p socket is the representation of the connections (_inbound/outbound_) that will support the pocket network and help peers get and stay connected.

This component will be responsible for:

- [x] Establishing connections
- [x] Handling network IO, muxing, timeouts and errors
- [x] Handling network IO metadata (_ids, hashs, nonces_)
- [ ] Handling network security (_handshake and encryption_)

_The V1 P2P Module, this module implements the P2PModule interface defined in `shared/modules/p2p_module.go`._

### Acceptance Criteria

Please refer to #40 

### Details:
----

- [x] Add base p2p module and expose it to the other modules with unimplemented methods that panic
- [x] Add the p2p socket implementation in `socket.go`
- [x] Test the p2p socket implementation in `socket_test.go`
- [ ] Implement the socket handshake and test it

### Testing:
----
If switching from another branch, run these commands first:
```bash
$ go mod vendor && go mod tidy
$ make protogen_clean && make protogen_local
```

Then, to run all tests, run the following command:
```bash
$ make test_p2p_socket && make test_p2p_types && make test_p2p
```

-----
#### Owners: @derrandz


derrandz:
@Olshansk Removed you as a reviewer as I think that this is not ready for review. I keep on rewriting the history to fit my needs. Once I freeze the history, I will request your review immediately.

What's lacking to request a review is simply wrapping up the last test case refactor `TestSocket_Open`

Then I will open the PR for review knowing that I still have to add in the handshake logic to consider this PR complete.

p2p/types/request_map.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if exists {
+		rm.elements[index] = nil
+		fhalf := rm.elements[:index]
+		shalf := rm.elements[index+1:]
+		rm.elements = append(fhalf, shalf...)
+		return request.nonce, request.ch, exists
+	}
+
+	return nonce, nil, false
```
Olshansk, line 47: - The nonce will either match, or the value provided will be returning.
- Request will either have a value or be nil

I think you can do with one return statement, no?

p2p/types/request_map.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (rm *RequestMap) Delete(nonce uint32) bool {
```
Olshansk, line 50: Ditto my comments above for Golang style:

1. Use range for the loop
2. Max index an int pointer, and use its existence as the value for `exists`
3. Make the append of the first and second half a single statement

p2p/types/request_map.go
```diff
@@ -0,0 +1,78 @@
...
+		fhalf := rm.elements[:index]
+		shalf := rm.elements[index+1:]
+		rm.elements = append(fhalf, shalf...)
+	}
+
+	return exists
+}
+
+func NewRequestMap(cap uint) *RequestMap {
+	return &RequestMap{maxcap: uint32(cap), elements: make([]*Request, 0), nonces: uint32(0)}
```
Olshansk, line 77: Personal NIT (feel free to ignore): I think multi-lining this will make it more readable but its very much just a preference.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	buff = append(buff, (*readBuffer)[:s.headerLength+uint(n)]...)
+
+	return buff, n, err
+}
+
+// the read routine
+// this routine performs buffered reads (readBufferSize) on the established connection
+// and does two things as a consequence of a read operation:
+//    1- if it's a response to a request out of this socket, it will redirect the response to the request's response channel (see types/request.go)
+// .  2- if not, it will handover the read buffer as Work (check types/work.go) to the runner (check types/runner.go)
```
Olshansk, line 231: Remove `.` at the bigging of line

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+// this routine halfs if:
```
Olshansk, line 232: s/halfs/halts

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+//   - if the socket closes
+//   - if some party signals that this routine should stop (by closing the s.reading channel)
+//   - if there is an IO error: EOF, UnexpectedEOF, peer hang up, unexpected error
+func (s *socket) read(ctx context.Context) {
+	defer func() {
+		close(s.reading)
+		s.closed <- 1
+	}()
+
+	{
```
Olshansk, line 244: What's the purpose of this scope?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+// this routine halfs if:
+//   - the cancelable routine cancels
+//   - if the runner stops
+//   - if the socket closes
+//   - if some party signals that this routine should stop (by closing the s.reading channel)
+//   - if there is an IO error: EOF, UnexpectedEOF, peer hang up, unexpected error
+func (s *socket) read(ctx context.Context) {
+	defer func() {
+		close(s.reading)
+		s.closed <- 1
```
Olshansk, line 241: Maybe create a helper for `s.SignalClose()` that does this?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	}()
+
+	{
+		s.reading <- 1 // signal start
+	}
+
+reader:
+	for {
+		select {
+		// TODO(derrandz): replace with passed down context ?
```
Olshansk, line 251: Isn't that what it's already doing?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+		case <-s.done:
+			break reader
+
+		case _, open := <-s.reading:
+			if !open {
+				break reader
+			}
+
+		default:
+			{
```
Olshansk, line 267: What's the purpose of these extra `{}`?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+						break reader
+
+					default:
+						s.error(ErrUnexpected(err))
+						break reader
+					}
+				}
+
+				if n == 0 {
+					continue
```
Olshansk, line 287: SHould we log a warning when nothing is read?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+				}
+
+				nonce, _, data, wrapped, err := s.c.decode(buf)
+				if err != nil {
+					s.error(err)
+					break reader
+				}
+
+				if nonce != 0 {
```
Olshansk, line 296: when would a zero nonce happen? Does it warrant a waning log?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+
+	if _, err := s.writeChunk(b, false, requestNonce, wrapped); err != nil {
+		s.requests.Delete(requestNonce)
+		return types.NewWork(requestNonce, nil, "", false), err
+	}
+
+	var response types.Work
+
+	select {
+	case response = <-request.Response():
```
Olshansk, line 347: Why do we need the `=` here?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+//   - if the socket closes
+//   - if some party signals that this routine should stop (by closing the s.writing channel)
+//   - if there is a write error
+func (s *socket) write(ctx context.Context) {
+	defer func() {
+		close(s.writing)
+		s.closed <- 1
+	}()
+
+	{
```
Olshansk, line 369: Remove extra scope?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+		s.writing <- 1 // signal start
+	}
+
+writer:
+	for {
+		select {
+		case <-ctx.Done():
+			break writer
+		case <-s.runner.Done():
```
Olshansk, line 378: You have a newline in some cases but not others (like here).

MY personal preference is that we don't need them, but we should stay consistent.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+		case <-s.done:
+			break writer
+
+		case _, open := <-s.writing:
+			if !open {
+				break writer
+			}
+
+		case <-s.buffers.write.Signals(): // blocks
+			{
```
Olshansk, line 390: I believe these {} are unnecessary.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+// tracks and stores the encountered error
+func (s *socket) error(err error) {
+	defer s.err.Unlock()
+	s.err.Lock()
+
+	if s.err.error == nil {
+		s.err.error = err
+	}
+
+	s.errored <- 1
```
Olshansk, line 420: Thoughts on having a helper for this? e.g. `s.SignalError()`

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+}
+
+// A constructor to create a socket
+func NewSocket(readBufferSize uint, packetHeaderLength uint, readTimeoutInMs uint) *socket {
+	wc := newWireCodec()
```
Olshansk, line 425: why not just inline this on line 427 below?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	pipe := &socket{
+		c: wc,
+
+		kind:         UndefinedSocketKind,
+		headerLength: packetHeaderLength,
+		bufferSize:   readBufferSize,
+		readTimeout:  readTimeoutInMs,
+
+		buffers: struct {
```
Olshansk, line 434: I had a question earlier about why this is a custom inline struct that I want to reiterate here.

However, if it is really needed, we should define a type for it so we don't have to redefine it inline twice (here and in `type socket struct`).

p2p/testutils/fn.go
```diff
@@ -0,0 +1,36 @@
+package testutils
```
Olshansk, line 1: When you cross this comment, let's either jump into MEWS, or schedule a call in protocol hour. I feel strongly about deleting this file
Screen Shot 2022-03-16 at 5 31 01 PM
 because `mockgen` allows this to be configured at test configuration time.

p2p/testutils/chunk.go
```diff
@@ -0,0 +1,34 @@
...
+
+type DataChunk struct {
+	Bytes   []byte // actual data
+	Encoded []byte // data after encoding
+	Error   error  // error encountered while reading or writing
+	Length  uint   // the length written or read
+}
+
+func (d *DataChunk) Randomize(length int) {
+	d.Bytes = GenerateByteLen(length)
```
Olshansk, line 13: Do we not need to encode the new bytes?

p2p/testutils/chunk.go
```diff
@@ -0,0 +1,34 @@
...
+		Length:  0,
+		Error:   nil,
+	}
+
+	dchunk.Randomize(l)
+
+	return dchunk
+}
+
+func GenerateByteLen(size int) []byte {
```
Olshansk, line 30: 1. s/GenerateByteLen/GenerateRandBytes
2. Should this be local (i.e. start with a lowercase)

p2p/wire_codec.go
```diff
@@ -0,0 +1,225 @@
...
+ @          is request __| |
+ @    is erroror end ______|
+ @
+ @
+ @    *: does the body have to be decoded at the application level (i.e by the domain codec)
+ @
+ @    1234: request number as uint16, empty if not a request
+ @    5678: bodyLength
+ @
+ @ body.
```
Olshansk, line 47: ?

p2p/wire_codec.go
```diff
@@ -0,0 +1,225 @@
...
+	sync.RWMutex
+}
+
+func (c *wireCodec) encode(encoding Encoding, iserror bool, reqnum uint32, data []byte, wrapped bool) []byte {
+	c.Lock()
+	defer c.Unlock()
+
+	var flags byte = 0x00
+
+	bodyLength := make([]byte, 4)
```
Olshansk, line 60: Can you make 4 a local constant with a semantic name

p2p/wire_codec.go
```diff
@@ -0,0 +1,225 @@
...
+	requestnumber := make([]byte, 4)
+
+	binary.BigEndian.PutUint32(bodyLength, uint32(len(data)))
+	binary.BigEndian.PutUint32(requestnumber, uint32(reqnum))
+
+	if wrapped {
+		flags ^= 16 // set the fifth bit to 1
+	}
+
+	if reqnum != 0 {
```
Olshansk, line 70: reqNum

p2p/wire_codec.go
```diff
@@ -0,0 +1,225 @@
...
+		flags ^= 8 // setting the fourth bit to 1
+	}
+
+	if iserror {
```
Olshansk, line 74: isError

p2p/wire_codec.go
```diff
@@ -0,0 +1,225 @@
...
+		flags |= 4 // setting the third bit to 1
+	}
+
+	switch encoding {
+
+	case Binary:
+		// set the second and first bits to 0 (they are already at 0 from initialization)
```
Olshansk, line 81: Either add a TODO, fallthrough, or panic

p2p/types/request_map_test.go
```diff
@@ -0,0 +1,94 @@
+package types
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRequestMap_Get(t *testing.T) {
+	rmap := NewRequestMap(100)
```
Olshansk, line 10: rMap

p2p/types/request_map_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	request := rmap.Get()
+
+	assert.Equal(
+		t,
+		request.Nonce(),
+		uint32(1),
+		"Request map error: failed to retrieve new request with valid nonce",
```
Olshansk, line 18: I'm confused by how this check for request retrieval when we're comparing nonce values.

p2p/types/request_map_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	assert.NotNil(
+		t,
+		request.Response(),
+		"Request map error: failed to retrieve request with a respond channel",
+	)
+}
+
+func TestRequestMap_Find(t *testing.T) {
+	rmap := NewRequestMap(100)
```
Olshansk, line 29: rMap

Olshansk (changes requested):
# General

A few of the comments will require further discussion, but we can schedule that offline or during protocol hour.

My biggest concern for me personally is that even though I reviewed most of the code, I still don't have a full grasp of what's going on, and might need another high level overview (design doc, etc..) to understand what's going on.

I was thinking that @oten91 might be a good person to review the business logic in a small subset of the files?

# Review Related
Can you do a `git checkout` on the prototype repo? This PR shouldn't be making any modifications to it.

Note to self: I did not fully review the following files, but will do so after the initial set of comments are addressed:
```
p2p/socket_test.go
p2p/types/request_map_test.go
p2p/wire_codec.go
p2p/wire_codec_test.go
```

p2p/types/runner.go
```diff
@@ -0,0 +1,6 @@
+package types
+
+type Runner interface {
+	Sink() chan<- Work
+	Done() <-chan uint
```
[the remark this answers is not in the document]
derrandz: `chan bool` vs `chan uint` vs `chan struct{}`. I honestly don't know what's the differentiating factor when it comes to choosing one instead of the other. Should research this later.

All I can say about `uint` is that I expect values to be positive (_although up to this moment I haven't made use of the actual sent values but rather just the event of receipt of new values_)

p2p/types/work.go
```diff
@@ -0,0 +1,32 @@
+package types
+
+type Work struct {
```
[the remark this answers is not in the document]
derrandz: Common, I am not sure. but what I can say is (_as I thought about renaming this more than once_):

1- Packet is not semantic when it comes to this "work". Maybe `P2PPacket`? dunno
2- Job? Yeah could `Work`. (lol sorry)

I'd say that I'd be happiest if I say `Packet`. Only that I need an alibi

p2p/types/work.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (w *Work) From() string {
+	return w.addr
+}
+
+func NewWork(nonce uint32, data []byte, addr string, encoded bool) Work {
+	return Work{nonce, data, addr, encoded}
+}
+
+func (w *Work) Implode() (uint32, []byte, string, bool) {
```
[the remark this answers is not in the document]
derrandz: Hahah! you caught me here. Tbh I've thought about `Explore/Extract`. Will rename then.

It's used in the `network`, the implementer of `Runner`. But it's gonna come in a next PR since this is a series of PRs.

shared/config/config.go
```diff
@@ -30,10 +30,15 @@ type Pre2PConfig struct {
...
-	Address    pcrypto.Address `json:"address"`
-	ExternalIp string          `json:"external_ip"`
-	Peers      []string        `json:"peers"`
+	Protocol         string          `json:"protocol"`
+	Address          pcrypto.Address `json:"address"`
+	ExternalIp       string          `json:"external_ip"`
+	Peers            []string        `json:"peers"`
+	MaxInbound       uint32          `json:max_inbound`
+	MaxOutbound      uint32          `json:max_outbound`
+	BufferSize       uint            `json:connection_buffer_size`
```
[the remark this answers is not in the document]
derrandz: Inbound/Outbound maxes are usually set around 128, 256 and what not, and since we don't want to limit the possible maxes (_which will heavily depend on the machines capacity_) I've went with 32 (_instead of 8 which will result in 256_). To come to think of it, `Uint16` is enough.

The other values such as buffer size and timeouts are freely set values and are convention based, so the only part that's required to be specified about their type is that they are positive, and they are integers.

shared/types/proto/events.proto
```diff
@@ -1,15 +1,14 @@
 syntax = "proto3";
-
 import "google/protobuf/any.proto";
 
 package pocket;
 
-option go_package = "/types/";
+option go_package = "github.com/pokt-network/pocket/shared/types;types";
```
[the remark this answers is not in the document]
derrandz: This has allowed me to use the shared proto definitions in P2P's definitions (_I needed that_). But since this is not part of this PR, I think I will remove this change until It's shipped with what it makes most sense with.

p2p/wire_codec.go
```diff
@@ -0,0 +1,225 @@
...
+
+import (
+	"encoding/binary"
+	"errors"
+	"sync"
+)
+
+type Encoding string
+
+const (
```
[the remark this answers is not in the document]
derrandz: Not opposed, feels like a go design, will try it out and see it

p2p/wire_codec.go
```diff
@@ -0,0 +1,225 @@
...
+/*
+ @ Idea inspired by scuttlebutt's secure p2p wire protocol
+ @
+ @ Layout:
+ @
+ @ * = 1 byte
+ @ + = 1 bit
+ @
+ @    9 bytes header
+ @    -----header---][----body----]
```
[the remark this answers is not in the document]
derrandz: Yes

p2p/wire_codec.go
derrandz: Yes sir

p2p/wire_codec.go
derrandz: fallthrough it is

derrandz (commented):
Thanks @Olshansk for your review, will add the requested changes

shared/config/config.go
```diff
@@ -30,10 +30,15 @@ type Pre2PConfig struct {
 }
 
 type P2PConfig struct {
-	Protocol   string          `json:"protocol"`
-	Address    pcrypto.Address `json:"address"`
-	ExternalIp string          `json:"external_ip"`
-	Peers      []string        `json:"peers"`
+	Protocol         string          `json:"protocol"`
+	Address          pcrypto.Address `json:"address"`
```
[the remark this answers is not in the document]
derrandz: It should, but since this PR is taken from the original P2P Module, keeping it as pcrypto.Address silenced errors I was not willing to deal with at that particular moment. Let's keep as is I will fix regardless. No need for a TODO as it already exists in the parts of the code that I will come in the next PRs.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+// kicks off the onSocketOpened event handler, returns if there is an error
+// launches both the reading and writing routines (read, write methods): both are blocking
+// when the write routine exists, the onSocketClosed event handler kicks off (returns an error if there is one)
+// the write and read routines will both exist for the same reasons, so having just one of them block is sufficient.
+func (s *socket) startIO(ctx context.Context, kind SocketKind, addr string, conn net.Conn, onopened SocketEventMonitor, onclosed SocketEventMonitor) {
+	defer func() {
+		s.close()
+	}()
+
+	s.addr = addr
```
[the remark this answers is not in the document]
andrewnguyen22: Agreed

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	s.kind = kind
+
+	s.conn = conn
+	s.reader = bufio.NewReader(conn)
+	s.writer = bufio.NewWriter(conn)
+
+	if err := onopened(ctx, s); err != nil {
```
andrewnguyen22, line 175: Not blocking, just a comment:
I'm not a fan of placeholder code. Looking through the tests, you're just using these callbacks for telemetry tracking. I would prefer if the telemetry engine was built into the lifecycle with a more linear design

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+// This is used by send/request/broadcast operations
+// upon each send, the write routine will receive a signal so that it may proceed to send the write over the network
+func (s *socket) writeChunk(b []byte, iserroreof bool, reqnum uint32, wrapped bool) (uint, error) {
+	defer s.buffers.write.Unlock()
+	s.buffers.write.Lock()
+
+	writeBuffer := s.buffers.write.Ref()
+
+	buff := s.c.encode(Binary, iserroreof, reqnum, b, wrapped)
+	(*writeBuffer) = append((*writeBuffer), buff...)
```
andrewnguyen22, line 325: parenthesis not needed

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+		}
+	}
+}
+
+// TODO(derrandz): add buffered write by writing exactly the writeBufferSize. Will need to think about how to split big payloads and sequence them
+// writes a chunk=writeBufferSize to the writer
+// this operation is blocking, and blocks until a waiter is ready to receive the signal from the write buffer (signaling that is has written)
+// This is used by send/request/broadcast operations
+// upon each send, the write routine will receive a signal so that it may proceed to send the write over the network
+func (s *socket) writeChunk(b []byte, iserroreof bool, reqnum uint32, wrapped bool) (uint, error) {
```
andrewnguyen22, line 318: snake case pls

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+
+	s.buffers.write.Signal()
+	return uint(len(b)), nil // TODO(derrandz): should length be of b or of the encoded b
+}
+
+// writeChunkAckful is a writeChunk that expects to receive an ACK response for the chunk it has written
+// This method will create a request, which is basically a nonce to identify the chunk to write, and a channel on which to receive the response
+// the channel is blocking, thus allowing the 'wait to receive the response' behavior.
+// the `read` routine takes care of identifying incoming responses (_using the nonce_) and redirecting them to the waiting channels of the currently-open requests.
+func (s *socket) writeChunkAckful(b []byte, wrapped bool) (types.Packet, error) {
```
andrewnguyen22, line 335: TODO? It's unused

p2p/socket_err.go
```diff
@@ -0,0 +1,23 @@
+package p2p
+
+import (
+	"fmt"
+)
+
+var (
+	ErrSocketEmptyContextValue func(string) error = func(value string) error {
```
andrewnguyen22, line 8: Omit all types

```
ErrSocketEmptyContextValue func(string) error = func(value string) error {
		return fmt.Errorf("socket error: the provided context does not have the value: %s", value)
	}
```

To 

```
ErrSocketEmptyContextValue = func(value string) error {
		return fmt.Errorf("socket error: the provided context does not have the value: %s", value)
	}
```

p2p/socket_mocks.go
```diff
@@ -5,8 +5,10 @@ import (
 	"context"
 	stdio "io"
 	"net"
-	"reflect"
+	"sync"
```
andrewnguyen22, line 8: If this mock is for testing purposes, it needs to be better organized within the code so that it's not included in `go get`

Make a test file may include `*testing.T` param to further enforce testing only

p2p/wire_codec.go
```diff
@@ -0,0 +1,234 @@
+package p2p
```
andrewnguyen22, line 1: Having a custom wire codec makes me very uncomfortable.

Let's take this offline on Friday's protocol hour

p2p/socket_mocks.go
andrewnguyen22: testutil seems like an appropriate dir

andrewnguyen22 (changes requested):

p2p/types/buffer.go
```diff
@@ -0,0 +1,81 @@
+package types
+
+import (
+	sync "sync"
+
+	"go.uber.org/atomic"
+)
+
+type Buffer struct {
```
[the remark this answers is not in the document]
derrandz: This is a type used primarily for tests. The justification for why there is a concurrent and non concurrent one is the actual test cases that either are multithreaded or not. I wanted the fact that we are used a concurrent buffer be explicit in concurrent test cases.

p2p/types/request.go
```diff
@@ -0,0 +1,18 @@
+package types
+
+type Request struct {
+	nonce uint32
```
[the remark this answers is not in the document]
derrandz: Noted.

p2p/types/request_map.go
```diff
@@ -0,0 +1,78 @@
+package types
+
+import sync "sync"
+
+type RequestMap struct {
+	sync.Mutex
+	maxcap   uint32
+	elements []*Request
```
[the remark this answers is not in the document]
derrandz: Good one

p2p/types/request_map.go
```diff
@@ -0,0 +1,78 @@
...
+	nonces   uint32
+}
+
+func (rm *RequestMap) Get() *Request {
```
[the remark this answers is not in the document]
derrandz: Good good point

p2p/socket.go
derrandz: 🤔 deserves some thought, you are correct in assuming that these callbacks served primarily with tests, and their importance will be more apparent in the incoming PRs since they will be used by other parts of the p2p module.

I would say that I can benefit from a discussion about the linear design approach. Adding this to the backlog of topics.

p2p/socket.go
derrandz: Good catch for the test, missed it. For it's usage, it will be used by the network struct/interface when it comes in later parts.

p2p/socket_mocks.go
derrandz: awesome suggestions, thank you

p2p/README.md
```diff
@@ -0,0 +1,3 @@
+## The P2P Module
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/types/runner.go
```diff
@@ -0,0 +1,6 @@
+package types
+
+type Runner interface {
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/wire_codec.go
derrandz: TODO

p2p/wire_codec.go
derrandz: TODO

derrandz (commented):
Resolved a few comments, marked others as TODOs

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	"net"
+	"sync"
+	"time"
+
+	"github.com/pokt-network/pocket/p2p/types"
+
+	"go.uber.org/atomic"
+)
+
+type SocketKind string
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+
+var (
+	Outbound            SocketKind = "outbound"
```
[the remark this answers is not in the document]
derrandz: Great suggestions. TODO

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	Inbound             SocketKind = "inbound"
+	UndefinedSocketKind SocketKind = "unspecified"
+)
+
+type SocketEventMonitor func(context.Context, *socket) error
+
+// socket is an abstraction around net.Conn to represent a p2p connection with full read/write capabilities
```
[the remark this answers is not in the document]
derrandz: 🤔 , TODO.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	bufferSize   uint
+	readTimeout  uint
+	addr         string
+	kind         SocketKind // inbound or outbound
+
+	// the actual network socket
+	conn net.Conn
+
+	// the wire codec
+	c *wireCodec
```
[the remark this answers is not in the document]
derrandz: noted, TODO

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+
+	// io buffers
+	buffers struct {
```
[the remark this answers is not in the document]
derrandz: Will add comments as requested, but the second buffer (writeBuffer) is not buffered yet as of this implementation.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+		read  *types.Buffer
+		write *types.ConcurrentBuffer
+	}
+
+	// the io reader/writer
+	reader *bufio.Reader
```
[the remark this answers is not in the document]
derrandz: Will answer in comments. But in short, the buffers are separate from their readers and writers and I did not see a reason as to why group them under the same struct

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	isOpen atomic.Bool
+
+	ready   chan uint // when the socket is opened and IO starts, this channel gets closed to signal readiness
+	writing chan uint // when the writing starts, this channel receives a new input; closes when done writing (i.e: stopped the socket)
+	reading chan uint // when the reading starts, this channel receives a new input; closes when done reading (i.e: stopped the socket)
+	done    chan uint // if this channel is closed or receives and input, it stops the socket and IO operations
+	closed  chan uint // this channel signals that the socket has been closed by receiving a new input
+
+	errored chan uint // on error, this channel receives a new input to signal the happening of an error
+	err     struct {  // the reference to store the encountered error.
```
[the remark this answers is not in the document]
derrandz: Yeah... as soon as it errors it stops, for the time being all errors that this struct stores are halting. This can definitely use some more thinking as telemetry and error handling are topics I haven't dealt with properly yet.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+
+// retrieves the network connection in question through the connector argument
+// and starts the IO operations on that connection, while putting in place event handlers for onSocketOpened and onSocketClosed events
+// returns an error on failure
+func (s *socket) open(ctx context.Context, connector func() net.Conn, onopened SocketEventMonitor, onclosed SocketEventMonitor) error {
+	s.buffers.write.Open()
+
+	conn := connector()
+
+	addr := ctx.Value("address").(string)
```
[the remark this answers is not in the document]
derrandz: Great tip!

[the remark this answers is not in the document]
derrandz: TODO

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	kind := ctx.Value("kind").(SocketKind)
+
+	if addr == "" {
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+		s.close()
+		return ErrSocketUndefinedKind(string(kind))
+	}
+
+	go s.startIO(ctx, kind, addr, conn, onopened, onclosed)
+
+	select {
+	case <-s.errored:
+		return s.err.error
+	case <-s.reading:
```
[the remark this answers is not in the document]
derrandz: 🤔 absolutely, will use a filler like that.

_But you might have noticed that logging is entirely absent in this part of the p2p module. Well, that primarily stems from the fact that I haven't decided on what to do regarding this topic. I would benefit from a discussion to unify logging/telemetry style across code-base._

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	select {
+	case <-s.errored:
+		return s.err.error
+	case <-s.writing:
+	}
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+
+	s.isOpen.Store(true)
+	close(s.ready)
+
+	return nil
+}
+
+// Stops the ongoing IO operations gracefully
```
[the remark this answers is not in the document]
derrandz: Hahah! I love these comments about comments. Will do.

TODO

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	s.isOpen.Store(false)
+}
+
+// creates a reader and writer for the network connection
+// kicks off the onSocketOpened event handler, returns if there is an error
+// launches both the reading and writing routines (read, write methods): both are blocking
+// when the write routine exists, the onSocketClosed event handler kicks off (returns an error if there is one)
+// the write and read routines will both exist for the same reasons, so having just one of them block is sufficient.
+func (s *socket) startIO(ctx context.Context, kind SocketKind, addr string, conn net.Conn, onopened SocketEventMonitor, onclosed SocketEventMonitor) {
+	defer func() {
```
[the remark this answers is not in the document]
derrandz: absolutely

p2p/socket.go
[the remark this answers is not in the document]
derrandz: Not a bad idea at all to make Socket like a module! but my initial intention with this "struct" or "file" is that it remains entirely private to the p2p module, hence the non-reliance on the already existing module pattern. This pattern however will be used with the network module, that will build on top of this, curious what you guys think @andrewnguyen22 @Olshansk ?

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+
+	if err := onclosed(ctx, s); err != nil {
+		s.error(err)
+		return
+	}
+}
+
+// the TLS handshake algorithm to establish encrypted connections
+func (s *socket) handshake() {
+
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/types/work.go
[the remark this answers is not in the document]
derrandz: I ended up going with Packet anyway. Lmk if this warrants a discussion.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+}
+
+// reads a chunk=readbufferSize amount of bytes out of the reader
+// this is used by the read routine to perform buffer reads
+// this operation blocks until the full body length is read
```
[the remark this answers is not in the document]
derrandz: will clarify properly in the comment

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+func (s *socket) readChunk() ([]byte, int, error) {
+	var n int
+
+	readBuffer := s.buffers.read.Ref()
+	if _, err := io.ReadFull(s.reader, (*readBuffer)[:s.headerLength]); err != nil {
+		return nil, 0, err
+	}
+	_, _, bodylen, err := s.c.decodeHeader((*readBuffer)[:s.headerLength])
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	if bodylen > uint32(s.bufferSize-s.headerLength) {
+		return nil, 0, errors.New(fmt.Sprintf("io pipe error: cannot read a buffer of length %d, the accepted body length is %d.", bodylen, s.bufferSize-s.headerLength))
+	}
+
+	if n, err = io.ReadFull(s.reader, (*readBuffer)[s.headerLength:bodylen+uint32(s.headerLength)]); err != nil {
+		return nil, 0, err
+	}
+
+	buff := make([]byte, 0)
+	buff = append(buff, (*readBuffer)[:s.headerLength+uint(n)]...)
```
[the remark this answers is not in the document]
derrandz: will that "copy"? will re-visit

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+		return nil, 0, err
+	}
+	_, _, bodylen, err := s.c.decodeHeader((*readBuffer)[:s.headerLength])
+	if err != nil {
+		return nil, 0, err
+	}
+
+	// TODO(derrandz): replace with configurable max value or keep it as is (i.e: max=chunk size) ??
+	if bodylen > uint32(s.bufferSize-s.headerLength) {
+		return nil, 0, errors.New(fmt.Sprintf("io pipe error: cannot read a buffer of length %d, the accepted body length is %d.", bodylen, s.bufferSize-s.headerLength))
```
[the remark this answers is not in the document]
derrandz: good catch

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	if n, err = io.ReadFull(s.reader, (*readBuffer)[s.headerLength:bodylen+uint32(s.headerLength)]); err != nil {
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/socket.go
derrandz: visual distinction no more no less.

p2p/socket.go
derrandz: Forgot to remove the comment, good catch

p2p/socket.go
derrandz: Visual distinction no more no less. Code organization.

p2p/socket.go
derrandz: Good idea, these are exactly the edge cases I want to fully cover. For now filling them with warnings is a good idea.

p2p/socket.go
derrandz: No, a non-zero nonce happens on nonced-respones, i.e: responses to already sent requests. Using the non-zero nonce, we are able to fetch the existing (waiting) request from the request map and pull out the channel on which this request expects to receive a response.

Zero nonces on the other hand are regular payloads and not response payloads.

p2p/socket.go
derrandz: As answered above. I really think this is comfortable for my eyes. Wdyt?

p2p/socket.go
derrandz: Good catch, agreed.

p2p/socket.go
derrandz: I feel that they comfort my eyes when the block is big and I am able to properly focus on that singular block. Not a strong preference, can remove.

p2p/socket.go
derrandz: I hestiate between inline and what's here often times. I like the philosophy of doing one thing per line and never doing many things at once, i.e: instantiate X somewhere, then move and assign it to Y where approriate... but this seems fair here. Doing it inline.

p2p/socket.go
derrandz: Not much reuse anywhere else for it become a separate type, this is a nesting in struct to further organize the struct fields, thus why I kept it inline. If it gets defined one more time elsewhere I will type it away.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	if addr == "" {
+		return ErrSocketEmptyContextValue("address")
+	}
+
+	if kind == "" {
+		return ErrSocketEmptyContextValue("kind")
+	}
+
+	switch kind {
+	case Outbound:
```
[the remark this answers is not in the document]
derrandz: So about this, fallthrough is not what I am looking for. I suppose switch is not what should be used then. The desired behavior is:

```
if kind != Outbound && kind != Inbound {
  // default
}
```

I think the switch statement is misleading. Will refactor to use an if statement

p2p/socket.go
derrandz: Awesome suggestion

p2p/socket.go
derrandz: So that it's returned on receipt. Check the return statement at the end of the function.

p2p/wire_codec.go
derrandz: Actually no, read the comment.

derrandz:
@oten91 Since you are back, give us some of your precious eyes 👁️ 😝

p2p/types/request_map.go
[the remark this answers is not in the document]
derrandz: TODO

p2p/wire_codec.go
```diff
@@ -0,0 +1,234 @@
...
+//
+//
+// *: does the body have to be decoded at the application level (i.e by the domain codec, think proto)
+//
+//  bytes 1234: request number/nonce/id as uint16, empty if not a request
+//  bytes 5678: bodyLength
+//
+
+type wireCodec struct {
+	sync.RWMutex
```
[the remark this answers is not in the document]
derrandz: TODO

p2p/socket.go
derrandz: Added as a backlog item for protocol hours, resolving for now.

p2p/socket.go
derrandz: TODO

p2p/socket.go
derrandz: TODO

p2p/socket.go
derrandz: TODO: add test and a comment to explain where this will be used

p2p/socket.go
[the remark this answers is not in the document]
Olshansk: I use our "module" pattern for the pacemaker module and the leader election module, which are internal to consensus only.

It allows using the same style throughout the same codebase so I'd say I feel strongly about doing it here too.

p2p/socket.go
```diff
@@ -0,0 +1,461 @@
...
+	if err := onopened(ctx, s); err != nil {
+		s.error(err)
+
+		close(s.writing)
+		close(s.reading)
+		return
+	}
+
+	go s.read(ctx) // closes s.reading when done
+	s.write(ctx)   // closes s.writing when done
```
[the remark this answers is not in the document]
Olshansk: In that case, why not just make sure both of them are closed?

```
func WaitMany(a, b chan bool) {
        for a != nil || b != nil {
                select {
                case <-a:
                        a = nil 
                case <-b:
                        b = nil
                }
        }
}
```

Taken from here: https://dave.cheney.net/2013/04/30/curious-channels

p2p/types/request_map.go
```diff
@@ -0,0 +1,78 @@
...
+	defer rm.Unlock()
+
+	rm.nonces++
+	nonce := rm.nonces
+	newreq := &Request{nonce: nonce, ch: make(chan Work)}
+	rm.elements = append(rm.elements, newreq)
+	return newreq
+}
+
+func (rm *RequestMap) Find(nonce uint32) (uint32, chan Work, bool) {
```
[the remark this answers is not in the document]
Olshansk: I meant a comment in the code. Going to copy-paste what you wrote here.

p2p/socket.go
Olshansk: Updated comment in the code.

p2p/socket.go
Olshansk: The same one is used on line 432 and 45.

I think this should be externalized into a local struct.

p2p/wire_codec.go
Olshansk: Adding this comment inline: `// NOTE: Left empty w/o a a fallthrough intentionally`

p2p/types/request_map_test.go
Olshansk: So the nonce is monotonically increasing, correct?

Is this documented somewhere?

Olshansk (changes requested):
## General
Overall, looks like it's almost ready as a first iteration that we can submit to main.

Even with the new document, my biggest issue is that I still have trouble really understanding everything that's going on (in detail) rather than a high level point of view. Perhaps a demo of some sort could help?

@derrandz Can you take a look at the commits that I pushed to make sure that the changes I made look good to you? Feel free to revert if there's a problem. I made sure to test everything with: `make test_p2p_socket && make test_p2p_types && make test_p2p`

## Changes

Similar to the utility PR, I made minor changes myself, left actionable comments inline and left longer ones as AIs here. We can create a separate issue and track them there once this PR is submitted.

### Short-term refactor
- [ ] Consider creating a helper for `s.closed <- struct{}{}` called `SignalClose` so it's more semantically explicit as one reads the codebase

### Mid-term refactor
- [ ] Sync with @Olshansk on how we can use mockgen for socket_mock.go (see utility in consensus PR as a reference)

### Large refactor
- [ ] Design & consolidate `logger` with the rest of the repository
- [ ] Replace `WireCodec` with `protobufs`

## Discuss
- [ ] `The write and read routines will both exit for the same reasons, so having just one of them block is sufficient` . IMO we should await on both (see [this](https://dave.cheney.net/2013/04/30/curious-channels) as a reference). If we don't, and simply assume they both exit for the same reason, it looks like an opportunity for a bug if one breaks but the other doesn't.

p2p/socket_test.go
```diff
@@ -82,6 +82,120 @@ func TestSocket_WriteChunk(t *testing.T) {
...
+			assert.Equal(
+				t,
+				response.Data,
+				[]byte("Bye bye world"),
+				"pipe writeAckful error: response data mismatch",
+			)
+		}
+	}
+
+	{ // test a non successful write with a timeout
```
Olshansk, line 150: Why not split this into two different tests?

andrewnguyen22 (approved):

Olshansk (approved):

PR #73, opened 2022-04-18
First iteration of a PostgreSQL based Persistence Schema

### LINKED ISSUE #68: [Persistence] Module First Implementation
Owner @andrewnguyen22 
Co-Owners: @iajrz  @Olshansk 

End Date May 1, 2022

## Objective

Create the first iteration of the Pre-Persistence module and integrate it with the other modules within the /github.com/pokt-network/pocket codebase

## Deliverables

Minimally implement the first iteration of the persistence layer.

- Create clean interfaces and MVP schemas for the P2P dataset, Consensus dataset, Mempool dataset, and the State dataset
- Use PostgreSQL as the first 'Database Engine'
- Minimally implement the State Versions Deduplication Strategy
- Minimally implement the Immutable State Schema (Patricia Merkle Tree) and AppHash functionality
- Integrate the DB Engine into the development infrastructure stack
- Integrate the persistence module with the other modules, deprecating pre-persistence

andrewnguyen22:
## Objective

Foundational iteration of PostgreSQL based persistence module implementation.

## Origin Document

https://github.com/pokt-network/pocket/issues/68

## Type of change

New major module implementation.

### Persistence-related core changes:
- List of actors / interfaces with MVP implementation:
    - Applications
    - Fisherman
    - ServiceNode
    - Accounts
    - Pools
- List of actors / interfaces with partial MVP implementation:
    - Validator
    - Gov params
- List of actors / interfaces with minorimplementation:
    - Block
- SQL Schema definition of the actors above
- SQL Query implementation for common actor persistence functionality
- PostgresContext implementation of the actors actors above
- Base infrastructure for fuzz testing

Non-persistence “fly-by” changes
- Updates to the PrePersistence module and utility module with breaking changes
- A few minor improvements/additions to the Makefile
- TODOs & comment cleanups throughout the codebase

### Unit Tests

```
make test_persistence
make test_all
```

### LocalNet
Ran a basic LocalNet following the instructions in the [development README](docs/development/README.md).


persistence/module.go
```diff
@@ -15,7 +15,10 @@ type persistenceModule struct {
 	bus modules.Bus
 }
 
-func Create(_ *config.Config) (modules.PersistenceModule, error) {
+func Create(c *config.Config) (modules.PersistenceModule, error) {
+	if err := connectAndInitializeDatabase(c.Persistence.PostgresUrl); err != nil {
```
derrandz, line 19: To ensure the statelessness of our code, I think we should not include the database "lifecycle" in the module's lifecycle. I'd argue that this will makes things easier especially in development.

In my opinion, we should have all database "scripts" live as separate tools or script, not even necessarily in Go.

For this PR, I would be satisfied with moving this function into a calling script under `tools/database/create.go` and keep everything in one script, so that we separate database and schema creation and validation from the persistence module's lifecycle.

In replacement for the "createIfNoExists" logic wither for the database, schema or tables, I suggest to do basic selects and expect the three to exist, as a validation layer in the `Create` phase of the persistence module, and throw an error indicating that the database is not OK if the database/schema/tables are not created.

About the scripts/tools idea for the database, ideally they should do exactly what this function does, arguably with the possibility of isolating tasks.

- `tools/database/create.(go|sh)`
- `tools/database/migrate.(go|sh)`
- `tools/database/destroy.(go|sh)`

1. Create should do the CreateIfNotExists thing
2. The migration should take care of creating our desired tables, changing their schemas and destroying them.
4. Destroy, wipe out the postgres server clean (_or just the database_)

This is a rough idea, a better idea is to write plain SQL for the schema and migrations and have them be ran optionally at DB start up from docker-compose.

persistence/sql.go
```diff
@@ -0,0 +1,48 @@
...
+	rand.Seed(time.Now().UnixNano())
+}
+
+func connectAndInitializeDatabase(postgresUrl string) error {
+	ctx := context.TODO()
+	conn, err := pgx.Connect(context.Background(), postgresUrl)
+	if err != nil {
+		log.Fatalf("Unable to connect to database: %v\n", err)
+	}
+	schema := os.Getenv(PostgresSchemaEnvVar)
```
derrandz, line 33: When you turn this function to a script, this can be a flag.

derrandz (changes requested):

persistence/sql.go
```diff
@@ -0,0 +1,48 @@
...
+	CreateTableIfNotExists  = "CREATE TABLE IF NOT EXISTS"
+	TableName               = "users"
+	TableSchema             = "(id int)"
+)
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+func connectAndInitializeDatabase(postgresUrl string) error {
```
Olshansk, line 27: Part of this (e.g. the conenction) is "production code", but the user piece is more "placeholder".

I think we should split it and add a comment that the user table piece is just a placerholder.

persistence/sql.go
Olshansk, line 33: I know we originally decided to put this as an env variable, but maybe it should be in the config?

persistence/module.go
andrewnguyen22: I agree - but not for this PR. I'd like to open a follow up issue on this

Olshansk: Prior to the schemas being pushed to the same branch & the same PR (different topic), though this suggestion has merit and deserves discussion, was larger than the actual PR reminding me of this: https://twitter.com/iamdevloper/status/397664295875805184?lang=en

I actually asked Andrew to do a small PR so we can just merge in the wiring local for postgres so that we can build on top of it in parallel: e.g. schemas, tooling, migrations, aforementioned design can be done independently while we discuss/review each other's work.

@derrandz For this specific example, I think it's clear that the connection/initialization is not intended to be proper production code in any way, so I think it's safe to close this out because properly designing out the persistence module is a give without needing to be said.

persistence/sql.go
Olshansk: Could we tend to this minor infra/wiking piece, split it out into the original PR, and merge it independently of the SQL migrations?

persistence/schema/gov.go
```diff
@@ -0,0 +1,124 @@
+package schema
+
+const ( // TODO (Team) - I'm sure there's obvious optimizations here
```
Olshansk, line 3: Long message incoming...

So this is weird because:
1. Params is a map, and not a structure SQL schema
2. There is only ever 1 record: the current set of params
3. This essentially create a SQL table that will only ever have 1 row

Though we are getting out of scope, the answer as to how we want to store this is the exact same as to how we want to store feature flags.

Per [this StackOverflow answer](https://stackoverflow.com/a/43264444/768439), we can go with either either `hstor`, `json` or `jsonb`. However, after reading [this](https://www.citusdata.com/blog/2016/07/14/choosing-nosql-hstore-json-jsonb/), the tl;dr is that we should use `jsonb` for key-value maps stored within json.

I understand why you'd want to explicitly list out all the columns here, but ultimately the source of truth for the "schema" of the government params is the protobuf, so my suggestion would be to:
1. Create a table called `params` or `maps`
2. Params will have a column `featureFlags JSONB NOT NULL` and `govParame JSON NOT NULL`

This loses the schema validation benefit at the database engine layer, but I can argue (preferably not over text) as to why that's fine.

persistence/schema/fisherman.go
```diff
@@ -0,0 +1,29 @@
+package schema
```
Olshansk, line 1: For this file and all the others, I wanted to add a point about migrations & schemas that I feel pretty strongly about before moving forward.

I looked at different [SQL migration tools](https://povilasv.me/go-schema-migration-tools/) in Golang and though I haven't picked a specific one, [golang-migrate/migrate](https://github.com/golang-migrate/migrate) seems to be the most popular one. However, instead of defining our tables in strings withing `.go` file, we should follow the best practices for how to manage migrations (note that creation is a migration in itself).

Source: golang-migrate/migrate
tl;dr
1. We store these in `.sql` files
2. Each change will have a `<date>_<migration>.up.sql` and a `<date>_<migration>.down.sql` file
3. This will enable:
- Cleaner & easier to read code (separation of go and SQL)
- Going back & forth during development
- Easy to track changes / migrations when we actually do them on chain.

Olshansk (changes requested):
Left a few non-trivial comments, so let's discuss tomorrow.

persistence/schema/gov.go
Olshansk: By `So this is weird because`, I implied `I'm not sure what the best approach here so here are just some thoughts I have`.

Per a suggestion from @iajrz offline, this schema was suggested

```
param values table
|------------+------------|
| field name | field type |
|------------+------------|
| height     | bigint     |
| param_id   | int        |
| value      | string     |
|------------+------------|


params table
|------------+------------|
| field name | field type |
|------------+------------|
| id         | int        |
| value_type | int        |
| param_name | string     |
|------------+------------|
```

My modification to it would be:

```

CREATE TYPE val_type AS ENUM ('STRING', 'BIGINT', 'SMALLINT');

CREATE TABLE params (
  name    string     PRIMARY_KEY,
  height  bigint     PRIMARY_KEY,
  enabled bool       NOT NULL,
  type    val_type   NOT NULL,
  value   string     NOT NULL,
)

CREATE TABLE flags (
  name    string     PRIMARY_KEY,
  height  bigint     PRIMARY_KEY,
  enabled bool       NOT NULL,
  type    val_type   NOT NULL,
  value   string     NOT NULL,
)
```

We can have very similar tables for flags and params and not how the primary key is a composite key of the name of the flag/param and the height, along with an enabled/disabled bool, so we can automatically (for example) in one snapshot of the state, predefined the behaviour of a flag only being applied for some number of blocks.

Olshansk (commented):
My comments might require an offline discussion since it is a large change to what was proposed, so lmk.

Otherwise:

1. Make sure to update the PR name & description since we'll probably end up submitting this all at once.

2. Some of the comments regarding the schema may be redundant, so feel free to resolve them if they were addressed elsewhere.

3. After reading https://github.com/pokt-network/pocket-network-protocol, my understanding is that the postgres state only represents the state of the blockchain at the height it's currently synched to. We don't need to store history in the psotgres DB (which will really bloat it over time), because that's guaranteed by the stateHash. In conjunction with the `created_at` and `deleted_at` fields, along with the stateHash stored in each block, we can replay transactions and recreate a small & efficient postgres DB at every height.

THis way, we can also use the `address` as a primary key, avoid two tables when one is sufficient, and overall keep it much simpler.

persistence/schema/account.go
```diff
@@ -0,0 +1,28 @@
+package schema
+
+const (
+	AccountTableName   = "Account"
+	AccountTableSchema = `(
+			id         BIGINT PRIMARY KEY,
+			address    STRING NOT NULL UNIQUE,
+		)`
+	AccountMetaTableName   = "Account_meta"
```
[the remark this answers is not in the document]
andrewnguyen22: > Having a metadata table implies it is metadata - not used for business logic, which is not the case here. Why split it out?
To not to have to duplicate the address field everytime you edit the balance
> We just started storing two BIGINTs id and account_id, and added in a foreign constraint w/ 2 tables. This can be made much simpler.
How so? Please leave actionables
> This is missing created_at and updated_at per the spec
Added created_at - updated_at is not needed

andrewnguyen22:
https://github.com/pokt-network/pocket/issues/76

persistence/schema/gov.go
andrewnguyen22: #76

persistence/schema/application.go
```diff
@@ -0,0 +1,29 @@
+package schema
+
+const (
+	AppTableName   = "app"
```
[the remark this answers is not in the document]
andrewnguyen22: Kept them all singular

persistence/schema/application.go
```diff
@@ -0,0 +1,29 @@
...
+	AppTableSchema = `(
+			id         BIGINT PRIMARY KEY,
+			address    BYTEA NOT NULL UNIQUE, /*look into this being a "computed" field*/
+			public_key BYTEA NOT NULL UNIQUE
+		)`
+	AppMetaTableName   = "app_meta"
+	AppMetaTableSchema = `(
+			app_id           BIGINT,
+			height           BIGINT NOT NULL,
+			staked_tokens    STRING NOT NULL,
```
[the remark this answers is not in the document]
andrewnguyen22: Hmm - not totally agains - might be a bit clunky / pigeon hole us into uPOKT...

persistence/schema/application.go
```diff
@@ -0,0 +1,29 @@
...
+	AppTableName   = "app"
+	AppTableSchema = `(
+			id         BIGINT PRIMARY KEY,
+			address    BYTEA NOT NULL UNIQUE, /*look into this being a "computed" field*/
+			public_key BYTEA NOT NULL UNIQUE
+		)`
+	AppMetaTableName   = "app_meta"
+	AppMetaTableSchema = `(
+			app_id           BIGINT,
+			height           BIGINT NOT NULL,
```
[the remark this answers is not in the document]
andrewnguyen22: It is not - it's the created at height

persistence/sql.go
andrewnguyen22: Separate issue?

persistence/sql.go
andrewnguyen22: https://github.com/pokt-network/pocket/pull/75

persistence/schema/fisherman.go
andrewnguyen22: https://github.com/pokt-network/pocket/issues/77

persistence/schema/application.go
```diff
@@ -0,0 +1,29 @@
...
+			id         UUID PRIMARY KEY,
+			address    BYTEA NOT NULL UNIQUE, /*look into this being a "computed" field*/
+			public_key BYTEA NOT NULL UNIQUE
+		)`
+	AppMetaTableName   = "app_meta"
+	AppMetaTableSchema = `(
+			app_id           UUID NOT NULL,
+			height           BIGINT NOT NULL,
+			staked_tokens    TEXT NOT NULL,
+			max_relays		 TEXT NOT NULL,  /*look into this being a "computed" field*/
```
Olshansk: Not a BIGINT?

persistence/sql.go
```diff
@@ -40,12 +39,111 @@ func connectAndInitializeDatabase(postgresUrl string, schema string) error {
...
 	}
+	if err := InitializeTables(ctx, db); err != nil {
+		log.Fatal(err.Error())
+	}
+	return nil
+}
 
-	// TODO(olshansky;github.com/pokt-network/pocket/issues/77): Enable proper up and down migrations
-	// pgx.MigrateUp(options, "persistence/schema/migrations")
+func InitializeTables(ctx context.Context, db *pgx.Conn) error {
```
Olshansk, line 56: Just FYI: once I implement migrations, we'll probably be able to delete all of this.

Olshansk (changes requested):
Consolidating feedback into one message here so it’s easier to track and discuss.

## Non Actionables - Points of agreement

**Dependency** - I put https://github.com/pokt-network/pocket/pull/75 for review which is a foundation for this, so we need to review & submit that first.

**Minor edits** - Any small edits I deemed “non-controversial”, I simply made myself and pushed to the branch. Please review.

**Followup Issues** - [👍](https://emojipedia.org/thumbs-up/) to the followup issues created so we can move forward faster with this PR.

**Meta tables** - Given that a UUID is 16 bytes and our implementation of addresses results in a 42 byte string, I am okay with splitting it into two tables for memory optimization purposes.

## Actionables

The following points are what I consider blockers that either need to be discussed or addressed. Some of them are related to naming and may seem pedantic, but I do think it’s important because even I was confused many times along the way (consider the public, new hires, etc..):
1. Core Table naming 
    - **Action**: Use plural naming 
    - **Explanation**: We should go with plural instead of singular names. For example, `app` should be `apps` because `SELECT id FROM apps` will give the id of all apps in the database where every row is 1 app.   
2. Meta Table naming 
    - **Action**: Rename to <singular>_data 
    - **Explanation**: The table corresponding to `apps` should be `app_data`. Note that it is singular and uses `data` instead of `meta`. Metadata is something that can be used for analysis but isn’t necessary in day-to-day operations (e.g. if we wanted to store the IP address of the validator that proposed a block), but something like `balance` is core data required for core logic. Secondly, `apps_data` or `app_datas` just sounds bad and doesn’t make sense because this table has a 1:1 mapping with the `apps` table. So we have an `apps` table containing all the apps, and an `app_data` table containing all the data associated with each app (which we will be filtering nay when querying). 
3. Update / Create columns
    - **Action**: Include `created_at_height` and `deleted_at_height` on every `data` table  
   - **Explanation**: In the spec [1] (image below), we state that EACH collection must have a `created_at` and `deleted_at` field to support CRUD operations, which is a core concept in the spec. While we don’t have to follow the spec 1:1 during implementation, I think that omitting this is a pretty fundamental change from the base concept. Secondly, I was confused by what `height` meant in the PR and I was confused by what `created_at` meant in the in the spec, hence the name suggestion. 
4. Address type inconsistency
    - **Action**: Choose if we want to use `TEXT` or `BYTEA` for representation the address  
   - **Explanation**: I think `BYTEA` is better, but either is fine. I care more about consistency.



[1] https://[github.com/pokt-network/pocket-network-protocol/tree/main/persistence#2213-state-versions-deduplication-strategy](http://github.com/pokt-network/pocket-network-protocol/tree/main/persistence#2213-state-versions-deduplication-strategy)
Screen Shot 2022-04-24 at 1 25 23 PM

persistence/application.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/pokt-network/pocket/shared/types"
+)
+
+const (
+	StakedStatus    = 2 // TODO (Team) get rid of status and/or move to shared
+	UnstakingStatus = 1
+	UnstakedStatus  = 0
+)
+
+func (p PostgresContext) GetAppExists(address []byte) (exists bool, err error) {
```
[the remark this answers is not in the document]
andrewnguyen22: But the point is that this is the postgres implementation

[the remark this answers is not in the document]
andrewnguyen22: The interface may be implemented by any number of engines

persistence/sql.go
Olshansk: Resolving because I added a TODO inline.

persistence/schema/application.go
Olshansk: Resolving because outdated.

persistence/application.go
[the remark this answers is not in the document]
Olshansk: +1 One of the interesting things I realized from the spec (and we can make it clearer in a future iteration of the spec) is that some of the implementation decisions outlined (e.g. use of protobufs) impacts the protocol, while others (e.g. use of Postgres) can be modified on a per client basis.

I have some local notes on my computer that I have moved into the spec update yet, but we can probably discuss it in a future protocol hour.

Screen Shot 2022-05-10 at 3 31 04 PM

derrandz:
I have a suggestion for the migration functionality.

We can make use of the following tool: [dbmate](https://github.com/amacneil/dbmate#postgresql). This tool will basically run in a container alongside the database, and will receive the database URL as an environment variable.

This will allow us to create very simple migrations in SQL, by simply using text directives:

```SQL
-- migrate:up

-- migrate:down
```

So an example of a migration would be:
```SQL
-- migrate:up
create table users (
  id integer,
  name varchar(255),
  email varchar(255) not null
);

-- migrate:down
```

This migration file will be generated using a command (`dbmate new create_users_table`) and we will fill it however we see fit.

Then, running the migrations up or down becomes a matter of calling a `make` command that will send the instruction to the tool in the container, which will subsequently run it against the database, the command would be something like:

```
migration-up:
	docker exec -v $(PWD)/migrations:/db migrater dbmate up

migration-down:
	docker exec -v $(PWD)/migrations:/db migrater dbmate rollback
```

And to generate new timestamped migrations:
```
generate_migration:
   docker exec -v $(PWD)/migrations:/db migrater dbmate new action_table
```

Olshansk:
@andrewnguyen22 I did not finish updating the whole PR as promised, but did make a substantial number of changes and made a list of notes (unedited) for us to reference in a discussion:

**Things not reviewed / changed**
- fisherman*.go (very similar to the changes in application.go)
- service_node_*.go
- validator_*.go
- Blockers:
    - Code duplication
    - Lack of confidence in tests (e.g. see usage of `Service_node_chains` in fisherman_test.go)
    - Code simplicity (e.g. See `clearFunctions` in sql.go`)

**Out of scope changes added**
- Added the following to the Makefile: `docker_check` and `go_doc`
- Updated unit tests outside persistence to use testify (part of a larger find/replace operation)

**Minor discussion items**
1. Denomination - Denomination of `balance` is not obvious to new readers.
    1. Can we consider renaming it?
    2. Where/how do we document why it’s a string?
2. Height - What are `height` and `end_height`? Someone looking at `account.go` for the first time should be able to figure this out
    1. Can we consider renaming it?
    2. Where/how do we document why it’s a string?
3. More semantic naming for `DefaultEndHeight`.
    1. Perhaps `NoEndHeight`?
4. Any reasons why we’re using `func (p PostgresContext)` instead of `func (p *PostgresContext)`?
5. What should the default for `height` be in various tables?
6. Run `grep -r "DISCUSS" persistence`
7. Used fancy regex to replace unit tests and such. Everything passes but be aware of potential weird formatting.

**Important discussion items**
1. Potentially missing tests:
    1. block_test.go
    2. fisherman_test.go
2. Lots of unit tests but important bugs were not caught
    1. We have different approaches to testing (unit vs end-to-end)
    2. Missing inputs (nullify) , missing indecencies (in tables), etc… This is an indicator the tests are not enough
3. Fuzzing
    1. We should do this for everything
    2. We need tests where we modify the height between executions
    3. I’m not comfortable until we add these tests for everything
4. We need to have a “tech debt” month (related to the point above)
    1. I made some improvements but feel a little bit uncomfortable with the implementation. Need more time to think of a good approach, but not blocking anything now.
5. Cannot manage multiple constraints. E.g. See DISCUSS item in applicaiton.go
6. We need to move to Go 1.18 and start using generics (e.g. see gov.go)
7. Potential issues with copy-pasted code: fisherman insert query was using the (there were several similar occasions):

Olshansk:
@derrandz Can you post that idea in https://github.com/pokt-network/pocket/issues/77?

It's the issue we're using to track a proper migration strategy.

Olshansk:
@andrewnguyen22 Feel free to ignore my previous comment. Providing a cleaned-up version here that should be easier to follow in terms of next steps.

# Concerns

I have a few concerns with rushing this PR, so I'm highlighting a few of the reason below. The left-hand-side is before my changes and the right-hand-side is after.

1. Lack of code reuse (and the need to copy-paste after a bug / improvement is done)

Screen Shot 2022-05-17 at 3 00 52 PM

2. Core logic issues that were not caught by tests

Screen Shot 2022-05-17 at 3 02 35 PM

3. Lack of constraints, the issues with constraints, and the lack of unit tests to catch these issues

Screen Shot 2022-05-17 at 3 04 08 PM

4. Lack of testing or logic to catch issues like missing parameters when they are needed

Screen Shot 2022-05-17 at 3 06 11 PM

5. Lack of code simplicity when the fruit is low hanging and more generalizable, while also being easy to read.

Screen Shot 2022-05-17 at 3 08 32 PM

6. More code simplicity. Note that this can be further improve even more with generics now that Go 1.18 is stable

Screen Shot 2022-05-17 at 3 10 54 PM

7. Increasing Query complexity

Screen Shot 2022-05-17 at 3 13 17 PM

# Work Done 

I tried to update as much of it as I can in the time given, but did not manage to finish. Moreso, I'm just concerned that there are more major issues I did not catch.

**Out of scope changes added**
- Added the following to the Makefile: `docker_check` and `go_doc`
- Updated unit tests outside persistence to use testify (part of a larger find/replace operation)

# Next Steps

## Help Needed

Some of the files which need "copy-pasted" / transferred changes include:
- `fisherman*.go` (e.g. similar to `application*.go`)
- `service_node_*.go`
- `validator_*.go`

@andrewnguyen22  I was wondering if this is something you could help with?

## Topics of Discussion

### Minor discussion items
1. **Denomination** of `balance` is not obvious to new readers. Can we consider renaming it and/or documenting why it's a string?
2. **Height** - What are `height` and `end_height`? Someone looking at `account.go` for the first time should be able to figure this out. We should document this somehwere.
3. More semantic naming for `DefaultEndHeight`. Perhaps `NoEndHeight`?
4. Any reasons why we’re using `func (p PostgresContext)` instead of `func (p *PostgresContext)`?
5. What should the default for `height` be in various tables?
6. Run `grep -r "DISCUSS" persistence`

### Major discussion items
**Important discussion items**
1. Missing tests for `block_test.go`?
2. Lots of unit tests but IMPORTANT bugs were not caught. We should consider the fuzzing approach in `account_test.go` for everything. Wdyt?
3. How do we manage multiple constraints?

### Action Items

In this PR:
- [ ] We "copy-paste" the fixes from the files I changed to those that I didn't finish to unblock this PR from moving forward. Would appreciate your help on this?

Follow up PRs in order to unblock further persistence modules implementation:
- [ ] Upgrade to Go 1.18
- [ ] Add fuzzing tests for each persistence actor

Add the following `tech debt` Github issues, and discuss how we should prioritize it:
- [ ] Using Go generics where possible
- [ ] Removing verbosity & copy-pasted code (function reuse, SQL query reuse, etc...)

Work that we can add in other Github issues that will be done in parallel (by us, new team members or interviewees):
- [ ] Move to a proper query builder system (with the `infra tag`)
- [ ] Use proper SQL migration methodologies (github issue already exists)

Olshansk:
@andrewnguyen22 PR swap.

I made some progress on the PR but looking for your general thoughts/feedback before we move forward because I'm finding myself adding a lot of questions and TODOs throughout aside from the details in https://github.com/pokt-network/pocket/pull/73#issuecomment-1129381032.

Note the following:
1. **Partial implementation**:I haven't transferred the learnings from `account` and `application` to `fisherman`, `service_node` and `validator` yet. Assuming those changes make sense to you, and we answer the remaining questions, we can either generalize or copy over the changes.
2. **Specification**: We'll need to iterate on the spec to fill in the missing gaps in the clarity of how it should be translated to the definition.
3. **Testing**: I couldn't figure out how to add fuzzing tests in `application_test.go` similar to `account_test.go` so let me know if you have any ideas.

Olshansk:
@andrewnguyen22 I've attached the txt file with the discussion and user stories from our pow-wow today.

In terms of next steps, what we need to do is:
- [ ] Re-implement the persistence layer to avoid the use of `end_height` altogether
- [ ] Implement a generic protocol level actor interface + set of functions for the CRUD SQL queries
- [ ] Add support for one actor in this PR
- [ ] Add unit tests, various "user story" or "user lifecycle" tests and fuzzed tests
- [ ] Update the spec to remove the need / use of `deleted_at`

[persistence_debug_session.txt](https://github.com/pokt-network/pocket/files/8866612/persistence_debug_session.txt)

Olshansk:
@andrewnguyen22 I will go over some of the highlights of my review/updates with the team tomorrow. Meeting notes (not public at the moment) can be found here: https://www.notion.so/07-01-2022-6b27c7087c254d72862567adfd5bfdd5

Olshansk:
Next steps for this PR are:

1. @andrewnguyen22 re-reviews all the recently pushed changes and makes the appropriate changes
2. @Olshansk updates the changelog, persistence/README and GitHub PR description
3. @luyzdeleon reviews the PR (possibly after merging) to track follow-up work
4. Follow-up Github issues are created for the team & community based on the TODOs in the README

andrewnguyen22:
Pushed the changes and approve

Olshansk:
@andrewnguyen22 I updated the README, github description and changelog. 

The only change I wanted to bring to your attention is commit https://github.com/pokt-network/pocket/pull/73/commits/1b17b0c34605c89e7a2d4792414f063fdbabdad8 which I had to do to fix some tests. PTAL.

@derrandz This review is blocked since you requested changes to please review and/or approve when you have a chance.

Olshansk (approved):

derrandz (approved):
I have not reviewed, I am approving due to my last changes request to unblock merging, but this looked really good in the protocol hour, so keep it coming guys.

PR #81, opened 2022-05-10
Add VictoriaMetrics for local dev + initialize scraping

okdas:
Adding victoria metrics container, plus service discovery configuration that assumes the exporter is going to be running on `:9000/metrics`.

Also some little tweaks in docs/Makefile due to issues I ran into during onboarding.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ build:
 .PHONY: build_and_watch
 ## Continous build Pocket's main entrypoint as files change
 build_and_watch:
-	/bin/sh ${PWD}/scripts/watch_build.sh
+	/bin/sh ${PWD}/build/scripts/watch_build.sh
 
 .PHONY: client_start
 ## Run a client daemon which is only used for debugging purposes
--- a/build/deployments/docker-compose.yaml
+++ b/build/deployments/docker-compose.yaml
@@ -141,6 +141,23 @@ services:
     ports:
       - "5050:5050"
 
+  vm:
+    container_name: victoria-metrics
+    image: victoriametrics/victoria-metrics
+    restart: always
+    ports:
+      - "8428:8428"
+    volumes:
+      - vm:/storage
+      - ./victoria-metrics:/configs
+      - /var/run/docker.sock:/var/run/docker.sock
+    command:
+      - "--storageDataPath=/storage"
+      - "--httpListenAddr=:8428"
+      - "--promscrape.config=/configs/scrape.yml"
+
 volumes:
   db:
+    driver: local
+  vm:
     driver: local
--- a/build/deployments/victoria-metrics/scrape.yml
+++ b/build/deployments/victoria-metrics/scrape.yml
@@ -7,13 +7,24 @@ scrape_configs:
   - job_name: "victoria-metrics"
     static_configs:
       - targets: ["localhost:8428"]
-  - job_name: "docker-containers"
+  - job_name: "pocket-nodes"
     docker_sd_configs:
       - host: unix:///var/run/docker.sock # You can also use http/https to connect to the Docker daemon.
-        port: 9000
     relabel_configs:
-      # directory where docker-compose.yml file resides called `deployments` so we can configure SD to only pick up containers related to this docker-compose.
+      # only keep node\d.consensus containers in the discovery.
       - source_labels:
           [__meta_docker_container_label_com_docker_compose_project]
         regex: deployments
         action: keep
+      - source_labels: [__meta_docker_container_name]
+        regex: /node\d?.consensus
+        action: keep
+      - source_labels: ["__address__"]
+        separator: ":"
+        regex: '(.*):(\d*)'
+        target_label: "__address__"
+        replacement: "${1}:9000"
+      - source_labels:
+          [__meta_docker_container_label_com_docker_compose_service]
+        target_label: instance
+        replacement: "${1}"
--- a/build/deployments/victoria-metrics/scrape.yml
+++ b/build/deployments/victoria-metrics/scrape.yml
@@ -1,0 +1,19 @@
+# You can follow prometheus config syntax here.
+
+global:
+  scrape_interval: 5s
+
+scrape_configs:
+  - job_name: "victoria-metrics"
+    static_configs:
+      - targets: ["localhost:8428"]
+  - job_name: "docker-containers"
+    docker_sd_configs:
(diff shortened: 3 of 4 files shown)


derrandz:
Added the telemetry module to the mix, we are able to extract metrics properly. 

image

Let's discuss what metrics to track and what graphs are we interested in producing.

Thanks for setting this up so quickly!

derrandz:
TODO(@derrandz): Update the PR description to respect the contribution guide

Olshansk:
@derrandz @okdas Is it okay to convert this to a draft PR?

derrandz:
@Olshansk Done

derrandz:
As of the latest update, we now have the following charts:

Screen Shot 2022-05-19 at 17 46 10
Screen Shot 2022-05-19 at 17 49 46

p2p/pre2p/module.go
```diff
@@ -68,25 +70,51 @@ func (m *p2pModule) SetBus(bus modules.Bus) {
...
 		}
 	}()
 
+	// TODO(tema):
+	// No way to know if the node has actually properly started as of the current implementation of m.listener
+	// Make sure to do this after the node has started if the implementation allowed for this in the future.
+	m.
+		GetBus().
+		GetTelemetryModule().
+		IncCounter("blockchain_nodes_connected")
```
okdas, line 116: It is considered a decent practice to maintain gauge with a value of 1 that includes metadata in labels. E.g. port the service is running on can be a label in this case.

The name should be different though! The metric should carry a name from the node point of view, not from external point of view. So I'd suggest `p2p_listening`. Lmk if that doesn't make sense.

derrandz:
@Olshansk and @okdas I've updated the description to provide more important details on what's been added and how to run it locally.

p2p/pre2p/module.go
derrandz: I think this makes sense @okdas. Will include these changes in the next iteration.

derrandz:
> I don't think we should be committing binary files such as sqlite db ([98329e5#diff-53f57ca4c5a855f76cca893c50418204304876b6d1f419432d8d427cef778d09](https://github.com/pokt-network/pocket/commit/98329e5de9506be58c08f41a6886bb1294219aee#diff-53f57ca4c5a855f76cca893c50418204304876b6d1f419432d8d427cef778d09)).
> 
> As well as grafana plugins source code into the code base - there is a way to automatically provision plugins, datasources and dashboards: [this](https://github.com/pokt-network/pocket/blob/b14a5ef0642565a20bfa6451aa1fe1a937e5fc10/build/deployments/docker-compose.yaml#L348) is how provisioning configuration passed into Grafana, and dashboards that are being [mounted](https://github.com/pokt-network/pocket/blob/b14a5ef0642565a20bfa6451aa1fe1a937e5fc10/build/deployments/docker-compose.yaml#L349) so they can get created on start of grafana container.
> 
> We'll need to remove the excess code and squash commits so binaries won't make it into the master branch. Let me try to make this dashboard work with automatic provisioning so other team members would be able to get set up going and hopefully contribute.

Good work removing this 👍🏻

consensus/hotstuff_leader.go
```diff
@@ -200,6 +201,17 @@ func (handler *HotstuffLeaderMessageHandler) HandleCommitMessage(m *consensusMod
 	// There is no "replica behavior" to imitate here
 
 	m.paceMaker.NewHeight()
+
+	m.
+		GetBus().
+		GetTelemetryModule().
+		IncCounter("consensus_blockchain_height")
```
okdas, line 208: IIRC this should be gauge, not a counter, because we can reset the node to 0 height. Counters can only go up, however.

consensus/hotstuff_leader.go
```diff
@@ -200,6 +201,17 @@ func (handler *HotstuffLeaderMessageHandler) HandleCommitMessage(m *consensusMod
...
+
+	fmt.Println("Incremented blockchain height counter")
+	m.
+		GetBus().
+		GetTelemetryModule().
+		SetGauge("p2p_broadcast_msg_received_total_per_block", 0)
```
okdas, line 214: If we are bouncing gauge to zero on every new block, then we likely never will see a correct number in prometheus/vm/grafana. We are currently gathering the metrics with "pull" model - meaning an agent periodically scrapes the metrics from the exporter endpoint. Currently, the agent scrapes the telemetry every 5 second. So we almost guaranteed will lose the real number of `msg_received_total_per_block` because it is most likely going to be set to 0 in between scrapes. Does that make sense?

The potential solutions I see are:
1. Make this a counter `msg_received`, and do not care about per block part - we could calculate approximate number of messages with somewhat sophisticated query (I am not yet sure how to write, but can look into that) - something like for every increase of `blockchain_height` metric, calculate an increase of `msg_received`.
2. If we need a 100% correct number of msgs per block, not an approximate value, then we can make this a log entry with count per block (e.g. `msg="recieved messages" block=10 msg_rec=300`) when new block introduced. 
3. We look/switch into push model and trigger flush not only periodically, but also when some sort of event happens, such as new block introduction. This would work with VictoriaMetrics, probably will work with Prometheus, but can't tell for sure about some other data stores.

okdas: Seems like Summary metric type is the way to go: https://prometheus.io/docs/concepts/metric_types/#summary

p2p/pre2p/raintree/network.go
```diff
@@ -196,6 +202,19 @@ func (n *rainTreeNetwork) HandleNetworkData(data []byte) ([]byte, error) {
 	}
 
 	if rainTreeMsg.Level > 0 {
+		log.Println("Message hash=", rainTreeMsg.Hash)
+		hashString := fmt.Sprintf("%x", rainTreeMsg.Hash)
```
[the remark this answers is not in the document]
okdas: Summary (basically one-bucket histogram) might be a good candidate here as well, but we should not add randomized labels such as block hash anyway.

derrandz:
Screen Shot 2022-06-21 at 04 44 03
With my last push, we've succeeded in achieving event-metrics powered dashboard relying on ingested logs from loki.

Opening the PR for review once conflicts are resolved. Cheers 👋🏻

derrandz:
In order for us to an efficient review process, we have decided to split this PR up into multiple succinct scoped PRs.

Closing this draft PR in favor of the first smaller PR: #95 

@okdas @Olshansk

PR #82, opened 2022-05-12
Add metrics to P2P

derrandz:
Adding telementry to the p2p module. Don't merge. The base is `pre2p/raintree-metrics` for comparison reasons only.


derrandz:
Following our decision to deprecate p2p and reiterate on pre2p to transfer the learnings from p2p, we are closing this PR.

PR #91, opened 2022-06-10
Build & Release versioning documentation update

Olshansk:
## Description
Documentation update on how to manage build & release versions & tags.

## Type of change
- [ ] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## How Has This Been Tested?
NA

## Checklist:
NA

### DIFF
--- a/docs/build/README.md
+++ b/docs/build/README.md
@@ -1,17 +1,21 @@
-# Build System
+# Building & Versioning
 
-The build system we're using, mage, has been introduced with four targets; testing it involves invoking them and looking at their output.
+## Release Tag Versioning
 
-## Versioning
+We follow Go's [Module Version Numbering](https://go.dev/doc/modules/version-numbers) for software releases along with typical [Software release life cycles](https://en.wikipedia.org/wiki/Software_release_life_cycle).
 
-All the targets that produce a binary should inject version information on it, which can be inspected by running the executable with the `-version` flag.
+For example, `v0.0.1-pre-alpha.1` is the tag used for the first milestone merge.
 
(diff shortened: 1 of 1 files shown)


luyzdeleon (commented):
I wonder if our versioning should take into consideration which version of the specs this implementation reflects: https://github.com/pokt-network/pocket-network-protocol ?

iajrz (changes requested):
I noticed the proposed use of `X.Y.Z-pre-alpha`; bear in mind that because of semver's rules this is a bigger version than `X.Y.Z-alpha`, with an opposite result than intended.

Given closer analysis of semver 2, the build and hash information should not be separated by a '/' but a '+'.

Furthermore, because the intention is for branch and hash be informative, those should be separated with "+" not a "-", otherwise the branch name becomes meaningful when comparing versions.

So a potential version number should look like:

`X.Y.Z[-<pre_release_tag][+branch_name][+short_hash][+dirty]`

With the understanding that the tag is prefixed with `v` as usual with go projects, and the tag name contains the `X.Y.Z[-<pre_release_tag]` portion of the version pattern.

luyzdeleon:
@iajrz I believe this PR proposes to move away from semver to https://go.dev/doc/modules/version-numbers

Olshansk:
@iajrz I tried to account for your explanation in this commit: https://github.com/pokt-network/pocket/pull/91/commits/047b85b98dee772d67ceeef98d4e7ca5e28ab3c3

I also changed the first release to `v0.0.1-alpha.pre.1` per your comment.

Lmk if all of this makes sense.

iajrz (approved):

luyzdeleon (approved):

PR #96, opened 2022-06-22
[Telemetry] Add event metrics for RainTree

derrandz:
## Description

This PR adds event-metrics through to Pre2P RainTree implementation using the Telemetry module.

## Changes

- [x] Added instrumentation to record the reception of a RainTree message as an event for each block, this event has the following attributes:
     - namespace `pre2p`
     - event_name `raintree_messages`
     - labels: [height]

- [x] Added instrumentation to record the reception of redundant RainTree messages as an event for each new block, the event has the following attributes:
     - namespace `pre2p`
     - event_name `broadcast_message_redundancy_per_block`
     - labels: [height, message_hash] 

- [x] Added `logql` query to process the recorded events and fed the results to a numerical series graph

## How Has This Been Tested?
__

1. Spin up the stack
```zsh
$ make compose_and_watch
```

2. Make sure you spin the client up, and that kick the blockchain off:
```zsh
$ make client_start && make client_connect
```

3. Choose:
```
✔ TogglePacemakerMode
✔ TriggerNextView
```

4. Open Grafana at [http://localhost:3000](http://localhost:3000)
5. Open the `RainTree Telemetry` graph

6. Verify the results on the dashboard

### Additional items for review

The query that was used to generate the graph is the following:

1. Total # RainTree Msgs / Block
```sql
sum by (blockheight) (
   count_over_time({host="docker-desktop"}
   |= "[EVENT] consensus"
   | pattern `<datetime> <_> <time> <type> <_> <metric_name> <_> <blockheight> <step> <validator_role>`[6s]
   )
),
```

2. Avg # Raintree Msgs / Block

```sql
avg by (blockheight, instance) (
  sum by (blockheight) (
    count_over_time({host="docker-desktop"}
    |= "[EVENT] p2p raintree_messages"
    | pattern `<datetime> <_> <time> <type> <namespace> <metric> <_> <blockheight>`[6s])
  )
)
```

3. Avg. Redundancy / Block
```sql
avg by (blockheight, instance) (
  sum by (blockheight) (
    count_over_time({host="docker-desktop"}
    |= "[EVENT] p2p broadcast_message_redundancy_per_block"
    | pattern `<datetime> <_> <time> <type> <_> <redundancy> <_> <hashstring> <_> <blockheight>`[6s])
   )
)
```

Owners: @derrandz 
Co-owners: @okdas @Olshansk


consensus/module.go
```diff
@@ -201,3 +201,7 @@ func (m *consensusModule) handleHotstuffMessage(msg *typesCons.HotstuffMessage)
 	// Note that the leader also acts as a replica, but this logic is implemented in the underlying code.
 	leaderHandlers[msg.Step](m, msg)
 }
+
+func (m *consensusModule) GetBlockHeight() uint64 {
```
Olshansk, line 205: Any chance we could use the `GetLatestBlockHeight` from persistence instead?

p2p/pre2p/module.go
```diff
@@ -77,6 +77,17 @@ func (m *p2pModule) GetBus() modules.Bus {
 func (m *p2pModule) Start() error {
 	log.Println("Starting network module")
 
+	m.
+		GetBus().
```
Olshansk, line 81: I feel like `m.GetBus().` should be the whole first line.

Looks like too much whitespace.

p2p/pre2p/module.go
```diff
@@ -77,6 +77,17 @@ func (m *p2pModule) GetBus() modules.Bus {
...
+		GetTelemetryModule().
+		GetTimeSeriesAgent().
+		CounterRegister(
+			"p2p_nodes_online",
```
Olshansk, line 85: `p2p_nodes_started` - Online could mean something different.

p2p/pre2p/module.go
```diff
@@ -88,6 +99,12 @@ func (m *p2pModule) Start() error {
 		}
 	}()
 
+	m.
+		GetBus().
+		GetTelemetryModule().
+		GetTimeSeriesAgent().
+		CounterIncrement("p2p_nodes_online")
```
Olshansk, line 106: Thoughts on moving the counter / gauge names into some sort of shared file with constants?

p2p/pre2p/raintree/network.go
```diff
@@ -33,6 +37,9 @@ type rainTreeNetwork struct {
 
 	// TECHDEBT(drewsky): What should we use for de-duping messages within P2P?
 	mempool types.Mempool
+
+	// TODO(team): think about proper way to keep track of p2p seen msgs hashes
+	msgspool map[string]bool
```
Olshansk, line 42: Have you looked at how `mempool` defined above is used?

Curious to understand why it can't be reused / is not enough.

p2p/pre2p/raintree/network.go
```diff
@@ -139,6 +148,15 @@ func (n *rainTreeNetwork) networkSendInternal(data []byte, address cryptoPocket.
 }
 
 func (n *rainTreeNetwork) HandleNetworkData(data []byte) ([]byte, error) {
+	blockHeightInt := n.GetBus().GetConsensusModule().GetBlockHeight()
```
Olshansk, line 151: Persistence should be the source of the current height.

p2p/pre2p/raintree/network.go
```diff
@@ -150,6 +168,17 @@ func (n *rainTreeNetwork) HandleNetworkData(data []byte) ([]byte, error) {
 		return nil, err
 	}
 
+	msgHash := fmt.Sprintf("%x", rainTreeMsg.Hash)
```
Olshansk, line 171: This seems to dupliacte the logic below.

p2p/pre2p/raintree/utils.go
```diff
@@ -0,0 +1,11 @@
+package raintree
+
+import (
+	"math/rand"
+	"time"
+)
+
+func GenerateRandInt() int32 {
```
Olshansk, line 8: `network.go` has a util called `getNonce` that can be reused.

p2p/pre2p/stdnetwork/network.go
```diff
@@ -66,3 +67,8 @@ func (n *network) AddPeerToAddrBook(peer *typesPre2P.NetworkPeer) error {
 func (n *network) RemovePeerToAddrBook(peer *typesPre2P.NetworkPeer) error {
 	panic("RemovePeerToAddrBook not implemented")
 }
+
+func (n *network) GetBus() modules.Bus  { return nil }
```
Olshansk, line 71: I understand that you made this a module so that you can get the block height via the bus.

Needs aside, from an overall design standpoint, do you think it makes sense for this to be a module that is started and stopped? Can you think of other approaches?

Olshansk (changes requested):
I think this PR needs a little more thought / work.

1. Please provide reasoning for the addition of the `msgpool` given the existence of `mempool`. I know that adding code is both more fun and easier than reading and understanding existing code, but can lead to a lot of code duplication and things can get messy.

2. Aside from the immediate need to the bus, please reconsider make the `network` struct (i.e. a helper library to the network module) a module of its own. I haven't though this through in detail myself, but lmk if there's no other way.

3. Please add all the appropriate labels to the github PR.

4. See simpler comments inline.

p2p/pre2p/module.go
derrandz: That's the vision, hopefully json

derrandz (commented):
@Olshansk 

I have ended up using the mempool for p2p msg de-dupping. I used it as is, but I was so tempted to refactor it, since it seems to be transaction focused, and `mempool.AddTransaction(p2pMsgHashString)` is not semantic.

A suggestion for a refactor is to:

1. Create a `Pool` interface with methods: `Contains`, `AddElements` and `DeleteElement`

2. Create a `TxMemPool` interface that embeds the `Pool` interface and provides syntax sugar to imply: `Transaction`

```
type TxMemPool interface {
  Pool
  AddTx() // signature is ignored ofc
  DelTx()
}
```

3. Create a `MsgMemPool` and do as done with `TxMemPool`.

I think using the `mempool` for de-dupping as used in this PR is sufficient for now until we can afford such refactor. Let me know what you think.

p2p/pre2p/raintree/network.go
derrandz: I have tried using persistence now but left you a few open questions in discord instead. Once answered we can those changes in.

consensus/module.go
derrandz: Like discussed over discord, not until we have that exposed somehow. I've left a comment in `shared/modules/consensus_module.go` to keep track of this as you suggested.

p2p/pre2p/stdnetwork/network.go
derrandz: Good point. Maybe we can split the `Module` interface into `Runnable` and `Accessible` such that:

1. Runnable is:
```go
type Runnable interface {
    Start()
    Stop()
}
```

and 2. Accessible is:
```go
type Accessible interface {
    GetBus()
    SetBus()
}
```

such that the `Network` in this case can do without with the runnable functionality but still access (_and be accessed by, if necessary?_) other modules.

Olshansk (changes requested):
I like the idea of the `MemPool` interface. Can you create a separate github issue for it?

p2p/pre2p/module.go
derrandz: Not in this commit, I think these metrics might be subject to change and it wouldn't make sense to delay the PR for this addition at the moment.

p2p/pre2p/raintree/network.go
```diff
@@ -139,6 +142,14 @@ func (n *rainTreeNetwork) networkSendInternal(data []byte, address cryptoPocket.
 }
 
 func (n *rainTreeNetwork) HandleNetworkData(data []byte) ([]byte, error) {
+	blockHeightInt := n.GetBus().GetConsensusModule().GetBlockHeight()
+	blockHeight := fmt.Sprintf("%d", blockHeightInt)
```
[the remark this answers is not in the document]
derrandz: I wouldn't want to do two operations in one line for ease of debugging purposes.

derrandz (commented):
@Olshansk I've addressed most of your comments, namely:

1. Eschewing the `Start` and `Stop` definitions for the sub-components where it does not make sense:

To achieve this, I've split the `Module` interface like we discussed and it alleviated the problem. This is a sufficient solution for the moment.

2. Aggregated metric name definitions into a single file for the moment.

If we agree about these changes, we can take them to the consensus telemetry branch post merge.

Cheers,

p2p/pre2p/telemetry/metrics.go
```diff
@@ -0,0 +1,20 @@
+package pre2p_telemetry
+
+const (
+	// How to define metrics in this file?
+	//
+	// Must define metrics using following template: MetricName_MetricType_Attribute
+	// e.g: type= timeseries, name=nodes_alive_counter, attribute=name
+	// will result in: NODES_COUNT_COUNTER_TIME_SERIES_NAME="nodes_alive_counter"
```
Olshansk, line 8: Given that you describe the template above as `MetricName_MetricType_Attribute`, should this not be `NodesAliveCounter_Timeseries_Name`?

p2p/pre2p/telemetry/metrics.go
```diff
@@ -0,0 +1,20 @@
...
+	// you can also add: NODES_COUNT_COUNTER_TIME_SERIES_DESCRIPTION="for your description"
+
+	// Time Series Metrics
+	P2P_NODE_STARTED_TIMESERIES_METRIC_NAME        = "p2p_nodes_started_counter"
```
Olshansk, line 12: Do we need `METRIC` in here or do you think it's redundant

p2p/pre2p/stdnetwork/network.go
```diff
@@ -66,3 +68,6 @@ func (n *network) AddPeerToAddrBook(peer *typesPre2P.NetworkPeer) error {
...
+func (n *network) SetBus(_ modules.Bus) {}
```
Olshansk, line 73: Why even make it an integrable module if we don't need this?

Olshansk (changes requested):
@derrandz The split of the module LGTM. Make sure to let the rest of the team know about this change.

Also, please see some of the outstanding comments from my previous along with the two others I left here.

p2p/pre2p/telemetry/metrics.go
derrandz: I meant that template for the variable name, and not the actual metric. The value of the metric name should just be semantic in my opinion. Wdyt @Olshansk

p2p/pre2p/telemetry/metrics.go
derrandz: I think it's semantic and helps with clarity although verbose, let's keep it as is.

p2p/pre2p/stdnetwork/network.go
derrandz: This is just to comply with the interface as this is struct that implements the `Network` interface and this interface is composed of `IntegrateableModule`, as is the case for `RainTreeNetwork`. It's for compliance sake.

Olshansk (approved):

p2p/pre2p/raintree/utils.go
```diff
@@ -0,0 +1,12 @@
+package raintree
+
+import (
+	hex "encoding/hex"
+
+	cryptoPocket "github.com/pokt-network/pocket/shared/crypto"
+)
+
+func GetHashStringFromBytes(b []byte) string {
```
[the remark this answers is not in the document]
derrandz: I must've missed this. No worries

PR #122, opened 2022-07-25
BLS Signature aggregation

phthan0:
## Description
__
Please include a summary of the change and which issue it addresses. Please also include relevant motivation and context. List any dependencies that are required for this change.

**Fixes # (issue)**

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?
__
_[Not needed for integration PR]_
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

go test --bench=. -v

- [ ] Test A
- [ ] Test B

For bugs, please link to the related issue and provide the following information to the extent available:
**How to reproduce**
- Reporter Name: 
- Reported Date:  
- Steps to reproduce (as specific as possible):
- Release/Build Version: 
- Environment (OS, framework version):

## Checklist:
__
- [ ] My code follows the style guidelines of this project
- [ ] I have made corresponding changes to the documentation
- [ ] No new warnings are generated by my changes
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules


Olshansk:
@phthan0 Should this still be a `draft` PR in the meantime?

phthan0:
Correct, changing @Olshansk

consensus/module.go
```diff
@@ -27,8 +26,15 @@ var _ modules.ConsensusModule = &consensusModule{}
 type consensusModule struct {
 	bus        modules.Bus
 	privateKey cryptoPocket.Ed25519PrivateKey
+        // For BLS signature aggregation
+        var sec bls.SecretKey
```
Olshansk, line 30: Does this compile?

I don't think variable declaration and variable assignment inside a struct definition works.

consensus/module.go
```diff
@@ -1,18 +1,17 @@
-package consensus
+package  consensus
```
Olshansk, line 1: Let me know if you need help integrating this with the actual consensus module so we can get it working end to end with LocalNet.

Olshansk (changes requested):
Noticed that `phthan0:main` is in your forked repo (https://github.com/phthan0/pocket), but you should be able to create a branch directly in the main repo now.

consensus/module.go
```diff
@@ -1,263 +0,0 @@
-package consensus
```
Olshansk, line 1: I think we should keep this file.

phthan0: I had intended to delete the changes not the file hold on

Olshansk:
@phthan0 Other than the deleted file, should I review the or is it still a WIP?

phthan0:
It's still a WIP. Tests are coming

phthan0:
@Olshansk please proceed

Olshansk:
Request: if it's ready for review, can you please:
1. Remove (wip) from the title
2. Make this a non-draft by clicking the "Ready for review button" (see image)
3. Update the Assigness / Labels / Project / Milestone on the RHS (probably should do this even for drafts)

Screen Shot 2022-08-03 at 9 48 34 AM

consensus/module.go
phthan0: I think it may have not, restructured since.

shared/crypto/bls.go
```diff
@@ -0,0 +1,135 @@
...
+
+// NewKeyPair creates a new BLS signing key pair. The private key x is a scalar
+// and the public key X is a point on curve G2.
+func NewKeyPair(suite pairing.Suite, random cipher.Stream) (kyber.Scalar, kyber.Point) {
+	x := suite.G2().Scalar().Pick(random)
+	X := suite.G2().Point().Mul(x, nil)
+	return x, X
+}
+
+// Sign creates a BLS signature S = x * H(m) on a message m using the private
```
Olshansk, line 24: NIT: `message m` even though the parameter name is `msg`

shared/crypto/bls.go
```diff
@@ -0,0 +1,135 @@
...
+// key x. The signature S is a point on curve G1.
+func Sign(suite pairing.Suite, x kyber.Scalar, msg []byte) ([]byte, error) {
+	hashable, ok := suite.G1().Point().(hashablePoint)
+	if !ok {
+		return nil, errors.New("point needs to implement hashablePoint")
+	}
+	HM := hashable.Hash(msg)
```
Olshansk, line 31: NIT: use lowerCamelCase for local vars

shared/crypto/bls_test.go
```diff
@@ -0,0 +1,252 @@
...
+	sig, err := Sign(suite, private, msg)
+	require.Nil(t, err)
+	_, public := NewKeyPair(suite, random.New())
+	if Verify(suite, public, msg, sig) == nil {
+		t.Fatal("bls: verification succeeded unexpectedly")
+	}
+}
+
+func TestBLSAggregateSignatures(t *testing.T) {
+	msg := []byte("Hello Boneh-Lynn-Shacham")
```
Olshansk, line 47: Howdy

shared/crypto/bls_test.go
```diff
@@ -0,0 +1,252 @@
...
+	suite := bn256.NewSuite()
+	private1, public1 := NewKeyPair(suite, random.New())
+	private2, public2 := NewKeyPair(suite, random.New())
+	sig1, err := Sign(suite, private1, msg)
+	require.Nil(t, err)
+	sig2, err := Sign(suite, private2, msg)
+	require.Nil(t, err)
+	aggregatedSig, err := AggregateSignatures(suite, sig1, sig2)
+	require.Nil(t, err)
```
Olshansk, line 56: Can you update the test to also verify that the `sizeof(aggregatedSignature)=sizeof(sig1)` and `sizeof(aggregatedSignature)=sizeof(sig2)`

shared/crypto/bls_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+	aggregatedKey := AggregatePublicKeys(suite, public1, public2)
```
Olshansk, line 58: TOL: This is good to know. We'll probably want to precompute this and store it on chain anytime the validator set changes. I think that for threshold signatures, you only need one public key (not ana ggregted public key) to verify k-of-n, but my memory be wrong.

shared/crypto/bls.go
```diff
@@ -0,0 +1,135 @@
...
+	return aggregated
+}
+
+// BatchVerify verifies a large number of publicKey/msg pairings with a single aggregated signature.
+// Since aggregation is generally much faster than verification, this can be a speed enhancement.
+// Benchmarks show a roughly 50% performance increase over individual signature verification
+// Every msg must be unique or there is the possibility to accept an invalid signature
+// see: https://crypto.stackexchange.com/questions/56288/is-bls-signature-scheme-strongly-unforgeable/56290
+// for a description of why each message must be unique.
+func BatchVerify(suite pairing.Suite, publics []kyber.Point, msgs [][]byte, sig []byte) error {
```
Olshansk, line 70: I read this as "publix"

shared/crypto/bls.go
```diff
@@ -0,0 +1,135 @@
...
+		return err
+	}	
+	right := suite.Pair(s, suite.G2().Point().Base())
+	if !left.Equal(right) {
+		return errors.New ("bls: invalid signature")
+	}
+	return nil
+}
+func distinct(msgs [][]byte) bool {
+	m := make(map[[32]byte]bool)
```
Olshansk, line 125: This is a micro optimization but figured I'd share just in case you're not aware: https://stackoverflow.com/a/37320521/768439

Gophers seem to use empty structs to implement the set pattern, but honestly, it's a matter of preference so I don't think it mattesr

shared/crypto/bls.go
```diff
@@ -0,0 +1,135 @@
...
+		return errors.New("bls: invalid signature")
+	}
+	return nil
+}
+
+// Verify checks the given BLS signature S on the message m using the public
+// key X by verifying that the equality e(H(m), X) == e(H(m), x*B2) ==
+// e(x*H(m), B2) == e(S, B2) holds where e is the pairing operation and B2 is
+// the base point from curve G2.
+func Verify(suite pairing.Suite, X kyber.Point, msg, sig []byte) error {
```
Olshansk, line 107: NIT: rename vars to match comment?

Olshansk (changes requested):
@phthan0 I started reviewing this and leaving comments but then realized that most of it is taken from the OG library: https://github.com/dedis/kyber/tree/master/sign/bls.

My main question is:
1. Should I continue reviewing the core crypto logic in this PR?
2. Are there any deltas/changes I should focus on in my review?
3. Do you think there's a need to copy-pasta this or can we just use them as a dependency?

Other overhead/infra requests:
4. Can you update the PR attributes/labels/etc?
5. Can you update `docs/development/README.md`?

Questions:
6. Will the integration with the actual consensus module happen in this PR?

consensus/module.go
```diff
@@ -30,6 +29,7 @@ type consensusModule struct {
 	privateKey cryptoPocket.Ed25519PrivateKey
 	consCfg    *config.ConsensusConfig
 
+
```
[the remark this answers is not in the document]
phthan0: 1. Yes, I played with it and slightly modified with it, but I thought any addition of core library functionality still needs to go through a review, because just because it's a library doesn't mean it doesn't have problems.
2. Core logic. Play with it if you want, I have fairly extensively.
3. I put it in there because I thought we would modify it more extensively, and I'm concerned about not having a versioned instance of that.

shared/crypto/bls.go
phthan0: "Verify" is what they call this in the IETF spec for BLS so I want to keep that for that reason, so I'll change the comment

consensus/module.go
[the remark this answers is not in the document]
phthan0: 4. Yes
5. Yes
6. No, I thought we wanted to move away from larger PRs and this was the smallest I was willing to make a PR. Having primitives to actually do the thing you're trying to accomplish seems like a prerequisite.

consensus/module.go
```diff
@@ -1,19 +1,18 @@
...
+package  consensus
 
 import (
 	"encoding/hex"
 	"fmt"
 	"log"
-
 	"github.com/pokt-network/pocket/shared/types"
-
+        "fmt"
```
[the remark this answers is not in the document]
phthan0: I'm using vscode but apparently it snuck through. I'll fix it.

shared/crypto/bls.go
phthan0: https://www.ietf.org/id/draft-irtf-cfrg-bls-signature-05.html

Olshansk:
1. Regarding have a versioned copy of the code - agreed. As @andrewnguyen22 knows, I have a personal "allergy" against copy-pasting, so we can potentially fork their repo into the pocket codebase and make changes then if we need to?

I'm really hoping we'll never need to touch the actual crypto logic.

We can discuss this but it would be my personal preference

2. Regarding using this package, I found this comment at the top of the other file:

```
// Deprecated: This version is vulnerable to rogue public-key attack and the
// new version of the protocol should be used to make sure a signature
// aggregate cannot be verified by a forged key. You can find the protocol
// in kyber/sign/bdn. Note that only the aggregation is broken against the
// attack and a later version will merge bls and asmbls.
```

Are we sure we want to use this instead of `kyber/sign/bdn`? Again, I don't have much cryptographic context/knowledge here, but it seems scary.

phthan0:
> 1. Acceptable. I really don't want to make changes to it.

2. If we make sure the messages are distinct, that specific rogue key attack is not a problem. It's also a BSN signature instead of a BLS signature if we don't use BLS signatures, and the requirement I was given was to use BLS signatures.

Issue #128, opened 2022-07-27
[Persistence] Enable multiple persistence contexts on PostgresContext


andrewnguyen22:
# Objective
Enable the lifecycle of block processing and production by creating a transactional architecture around persistence contexts instead of writing directly to the database on every operation.

# Origin Document
### Background
Multiple persistence contexts enable a proper lifecycle for validators and fullnodes as they maintain 'ephemeral' states in situations like block validation/processing and block production.

# Goals / Deliverables
- [ ] Deliver completed code
- [ ] Document the architecture and intended use
- [ ] Test the transactional nature within the persistence module

## General milestone checklist
- [ ] Update all the relevant CHANGELOGs
- [ ] Update all the relevant READMEs
- [ ] Update the source code tree explanation
- [ ] Add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)
- [ ] Create followup milestones + issues
- [ ] Document small TODO along the way

---

Creator: @andrewnguyen22

PR #139, opened 2022-08-01
[Persistence] Optimize gov params schema  (Issue #76)

### LINKED ISSUE #76: [Persistence] Optimize gov params schema
Optimize params schema for better de-duplication of data while maintaining fast lookups

NOTE: The fundamental issue with the current 'naive' design is that it recreates the entire parameters list everytime we have an edit to a single parameter. However, it is also important to note, parameters are expected to be updated at a low frequency.

Per a suggestion from @iajrz offline, this schema was suggested

```
param values table
|------------+------------|
| field name | field type |
|------------+------------|
| height     | bigint     |
| param_id   | int        |
| value      | string     |
|------------+------------|


params table
|------------+------------|
| field name | field type |
|------------+------------|
| id         | int        |
| value_type | int        |
| param_name | string     |
|------------+------------|
```

My modification to it would be:

```

CREATE TYPE val_type AS ENUM ('STRING', 'BIGINT', 'SMALLINT');

CREATE TABLE params (
  name    string     PRIMARY_KEY,
  height  bigint     PRIMARY_KEY,
  enabled bool      NOT NULL,
  type    val_type  NOT NULL,
  value   string     NOT NULL,
)

CREATE TABLE flags (
  name    string     PRIMARY_KEY,
  height  bigint     PRIMARY_KEY,
  enabled bool      NOT NULL,
  type    val_type  NOT NULL,
  value   string     NOT NULL,
)
```

We can have very similar tables for flags and params and not how the primary key is a composite key of the name of the flag/param and the height, along with an enabled/disabled bool, so we can automatically (for example) in one snapshot of the state, predefined the behaviour of a flag only being applied for some number of blocks.

deblasis:
## Description

Currently WIP and paused because of https://github.com/pokt-network/pocket/issues/76#issuecomment-1200311076

This PR changes the schema for `params`, adds `flags` (to be implemented) and aims at improving the way we update the values.

It should also improve maintainability by:
- centralizing Gov schema related metadata into the protobuf by attaching metadata to it. (see gov.proto)
- using generics for get/set methods

Fixes [issue/76](https://github.com/pokt-network/pocket/issues/76)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] Other (<_REPLACE_ME_WITH_DETAILS_>)

## How Has This Been Tested?
_[Not needed for integration PR]_

_Describe the tests and that you ran to verify your changes. If applicable, provide steps to reproduce. Bonus points for images and videos or gifs._

- [x] `make test_all`
- [x] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md)

## Checklist
- [x] I have documented `flags` and their differences with `params` as per [comment](https://github.com/pokt-network/pocket/pull/139#issuecomment-1215602689)


persistence/db.go
```diff
@@ -108,32 +116,48 @@ func InitializeAllTables(ctx context.Context, db *pgx.Conn) error {
...
 	}
 	return nil
 }
 
 func InitializeGovTables(ctx context.Context, db *pgx.Conn) error {
-	_, err := db.Exec(ctx, fmt.Sprintf(`%s %s %s`, CreateTableIfNotExists, schema.ParamsTableName, schema.ParamsTableSchema))
+
+	//TODO (@deblasis): val_type enum
+	_, err := db.Exec(ctx, fmt.Sprintf(`%s %s`, fmt.Sprintf(CreateEnumType, schema.ValTypeName), schema.ValTypeEnumTypes))
+	if err != nil {
```
deblasis, line 144: As far as I know, CREATE TYPE ... doesn't support IF EXISTS semantics so I resorted to this graceful failure handling for the specific error

shared/types/genesis/proto/gov.proto
```diff
@@ -1,124 +1,232 @@
 syntax = "proto3";
+
 package genesis;
 
 option go_package = "github.com/pokt-network/pocket/shared/types/genesis";
 
 // DISCUSS(drewskey): Explore a more general purpose "feature flag" like approach for this.
 message Params {
+  //@gotags: pokt:"val_type=BIGINT"
```
deblasis, line 9: Here I am attaching "metadata" to the fields so that the generated structs have them in the form of `struct tags` that we can parse similarly to what JSON marshaller/unmarshaller do basically everywhere.

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
-		blocks_per_session_owner TEXT NOT NULL,
-		app_minimum_stake_owner TEXT NOT NULL,
-		app_max_chains_owner TEXT NOT NULL,
-		app_baseline_stake_rate_owner TEXT NOT NULL,
-		app_staking_adjustment_owner TEXT NOT NULL,
-		app_unstaking_blocks_owner TEXT NOT NULL,
-		app_minimum_pause_blocks_owner TEXT NOT NULL,
-		app_max_paused_blocks_owner TEXT NOT NULL,
+func init() {
+	govParamMetadataMap = parseGovProto()
```
deblasis, line 14: Here I am parsing the proto file since it's not going to change at runtime.

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
-		params.FishermanUnstakingBlocks,
-		params.FishermanMinimumPauseBlocks,
-		params.FishermanMaxPauseBlocks,
-
-		params.ValidatorMinimumStake,
-		params.ValidatorUnstakingBlocks,
-		params.ValidatorMinimumPauseBlocks,
-		params.ValidatorMaxPauseBlocks,
-		params.ValidatorMaximumMissedBlocks,
+	val := reflect.ValueOf(params)
```
deblasis, line 59: Here we use the metadata we obtained from the proto file along with reflection to construct our SQL

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
+	return fmt.Sprintf(`INSERT INTO %s VALUES %s`, ParamsTableName, subQuery[:len(subQuery)-1])
+}
 
-		params.ValidatorMaxEvidenceAgeInBlocks,
-		params.ProposerPercentageOfFees,
-		params.MissedBlocksBurnPercentage,
-		params.DoubleSignBurnPercentage,
+func GetParamQuery(paramName string) string {
+	//TODO (@deblasis): clarify if and how we should use `enabled` here
+	return fmt.Sprintf(`SELECT value FROM %s WHERE name=%s AND height=%d and enabled=true`, ParamsTableName, paramName, DefaultBigInt)
```
deblasis, line 79: I think I need to better understand your train of thought around the `enabled` flag. Not sure if this is correct,

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
-		hex.EncodeToString(params.MessageEditStakeServiceNodeFeeOwner),
-		hex.EncodeToString(params.MessageUnstakeServiceNodeFeeOwner),
-		hex.EncodeToString(params.MessagePauseServiceNodeFeeOwner),
-		hex.EncodeToString(params.MessageUnpauseServiceNodeFeeOwner),
-		hex.EncodeToString(params.MessageChangeParameterFeeOwner),
-		DefaultBigInt,
-	)
-}
+	subQuery := fmt.Sprintf(`SELECT %s`, fields)
+	//TODO (@deblasis): clarify if and how we should use `enabled` here
```
deblasis, line 96: I think I need to better understand your train of thought around the enabled flag. Not sure if this is correct,

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
 
-func SetParam(paramName string, paramValue interface{}, height int64) string {
-	paramNames := GetParamNames()
-	pNamesLen := len(paramNames)
-	var index int
-	// TODO (Team) optimize linear search
-	for i, s := range paramNames {
-		if s == paramName {
-			index = i
+func parseGovProto() (govParamMetadataMap map[string]govParamMetadata) {
```
deblasis, line 111: here we basically extract our custom tag `val_type` that's gonna dictate he enum we use for that specific field and also the "snake case" name that's gonna represent the "keys" we previously had defined as column names in the legacy implementation.

shared/types/genesis/proto/gov.proto
Olshansk: So does this mean we can use `protoc-go-inject-tag  ` to remove this hack I introduced? https://github.com/pokt-network/pocket/blob/main/shared/types/genesis/validator.go#L10

If so, we should definitely fix my hacks 😅

persistence/db.go
```diff
@@ -108,32 +116,48 @@ func InitializeAllTables(ctx context.Context, db *pgx.Conn) error {
...
 		return err
 	}
 	return nil
 }
 
 func InitializeGovTables(ctx context.Context, db *pgx.Conn) error {
-	_, err := db.Exec(ctx, fmt.Sprintf(`%s %s %s`, CreateTableIfNotExists, schema.ParamsTableName, schema.ParamsTableSchema))
+
+	//TODO (@deblasis): val_type enum
+	_, err := db.Exec(ctx, fmt.Sprintf(`%s %s`, fmt.Sprintf(CreateEnumType, schema.ValTypeName), schema.ValTypeEnumTypes))
```
Olshansk, line 143: if we're not going to use the first value, you can just inline `; err !=nil` (see above for an example).

Ditto below.

persistence/db.go
```diff
@@ -108,32 +116,48 @@ func InitializeAllTables(ctx context.Context, db *pgx.Conn) error {
...
+		if errors.As(err, &pgErr) && pgErr.Code != DuplicateObjectErrorCode {
+			return err
+		}
+	}
+	//TODO (@deblasis): params table
+	_, err = db.Exec(ctx, fmt.Sprintf(`%s %s %s %s`, CreateTable, IfNotExists, schema.ParamsTableName, schema.ParamsTableSchema))
+	if err != nil {
+		return err
+	}
+	//TODO (@deblasis): flags table
```
Olshansk, line 155: What exactly is the TODO if you're already creating the table?

persistence/gov.go
```diff
@@ -917,29 +916,11 @@ func (p PostgresContext) SetParam(paramName string, paramValue interface{}) erro
...
 	ctx, conn, err := p.DB.GetCtxAndConnection()
 	if err != nil {
-		return 0, err
+		return i, err
 	}
 	err = conn.QueryRow(ctx, schema.GetParamQuery(paramName)).Scan(&i)
 	return
 }
-
-func (p PostgresContext) GetStringParam(paramName string) (s string, err error) {
```
Olshansk, line 929: Yes!

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
-		params.ValidatorMinimumPauseBlocks,
-		params.ValidatorMaxPauseBlocks,
-		params.ValidatorMaximumMissedBlocks,
+	val := reflect.ValueOf(params)
+	var subQuery string
+	for _, k := range govParamMetadataKeys {
+		pnt := govParamMetadataMap[k]
+		pVal := val.Elem().FieldByName(pnt.PropertyName)
+
+		subQuery += `(`
```
Olshansk, line 65: NIT: I'd use a proper string builder: https://stackoverflow.com/a/1766304/768439

persistence/schema/gov.go
Olshansk, line 59: NIT: Please add a 'WARNING: reflections in prod' comment.

It's always a bit scary but I like this approach much more

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
+		switch govParamMetadataMap[k].PoktValType {
+		case ValTypeString:
+			subQuery += fmt.Sprintf("'%s', %d, true, '%s', '%s'", k, DefaultBigInt, pnt.PoktValType, pVal.Interface())
+		case ValTypeSmallInt, ValTypeBigInt:
+			subQuery += fmt.Sprintf("'%s', %d, true, '%s', %d", k, DefaultBigInt, pnt.PoktValType, pVal.Interface())
```
Olshansk, line 70: add a default case that logs a fatal error (i.e. panic)

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
+
+	return govParamMetadataMap
 }
 
-func ClearAllGovQuery() string {
-	return fmt.Sprintf(`DELETE FROM %s`, ParamsTableName)
+func extractStructTag(structTag string, key string) string {
+	for len(structTag) > 0 {
+		i := strings.IndexByte(structTag, ',')
+		if i < 0 {
```
Olshansk, line 135: s/`if < 0 {`/`if < 0 { // not found`

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
+			i = len(structTag)
+		}
+		switch s := structTag[:i]; {
```
Olshansk, line 138: Why even have the switch here?

persistence/schema/gov.go
Olshansk: Re height - I think we'll also need to do a `where height<=%d AND enabled=true order by height desc limit 1`.

CC @andrewnguyen22  for confirmation.

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
-		params.MessageUnstakeValidatorFee,
-		params.MessagePauseValidatorFee,
-		params.MessageUnpauseValidatorFee,
-		params.MessageStakeServiceNodeFee,
-		params.MessageEditStakeServiceNodeFee,
-		params.MessageUnstakeServiceNodeFee,
-		params.MessagePauseServiceNodeFee,
-		params.MessageUnpauseServiceNodeFee,
-		params.MessageChangeParameterFee,
+type ParamTypes interface {
```
Olshansk, line 87: s/ParamTypes/ProtobufAttributeTypes/>

persistence/schema/gov.go
```diff
@@ -1,442 +1,145 @@
...
-	copy(paramNames, SQLColumnNames)
-	return
+func ClearAllGovQuery() string {
+	return fmt.Sprintf(`DELETE FROM %s`, ParamsTableName)
 }
 
-func NullifyParamsQuery(height int64) string {
-	return fmt.Sprintf(`UPDATE %s SET end_height=%d WHERE end_height=%d`, ParamsTableName, height, DefaultBigInt)
+type govParamMetadata struct {
+	PoktValType  string
```
Olshansk, line 107: We use `PoktVal` and `Property`, but should we be consistent here like `PropertyType` and `PropertyName`, for example?

Olshansk (changes requested):

Olshansk:
I did a quick first review (will do an in-depth one next time) but this is frikkin awesome!

I think we should continue moving forward with it and the changes are well contained enough that I don't think the conflicts with deprecating persistence will be that bad. CC @andrewnguyen22

persistence/schema/gov.go
deblasis: `ParamTypes` represents the types that the generic can accept:
```go
type ParamTypes interface {
	int | int32 | int64 | []byte | string
}
```
technically it has nothing to do with the protobuf.
Agreed that `ParamTypes` sounds weird though

persistence/schema/gov.go
deblasis: Right, I simplified it. It was meant to host additional properties inside the tag but sounds like premature optimization since we might never need more than the current one here

persistence/db.go
deblasis: I forgot the `//TODO remove the //TODO` 🤣

shared/types/genesis/proto/gov.proto
deblasis: I'll have a look but in my comment above I wasn't referring to anything specific in the codebase but in general what the standard library (I think `encoding/json`) does, ergo: reading the struct tags that declaratively tell what to do.

:)

Also about "hacks": 
- they sent us to the moon and more.
- everything starts with a "hack"
- perfection is the enemy of action
- 🚀

deblasis:
Thanks @Olshansk! I have caught up with `main` and updated a few things but I will dive deeper and also test the parts I didn't test yet (Get/Set) after your next in-depth look and which should give me more clarity around how to use `enabled` and snapshotting 🙂.

persistence/schema/gov.go
Olshansk: `SupportedParamTypes` or `GovParamTypes`?

persistence/schema/gov.go
```diff
@@ -3,440 +3,151 @@ package schema
...
-		message_pause_service_node_fee_owner TEXT NOT NULL,
-		message_unpause_service_node_fee_owner TEXT NOT NULL,
-		message_change_parameter_fee_owner TEXT NOT NULL,
-		height BIGINT NOT NULL
-	)`
+	ParamsTableName   = "params"
+	ParamsTableSchema = `(
+		name VARCHAR(64) NOT NULL,
+		height BIGINT NOT NULL,
+		enabled BOOLEAN NOT NULL,
```
Olshansk, line 28: Let's remove `enabled` for params and only keep it for `flags`

persistence/schema/gov.go
```diff
@@ -3,440 +3,151 @@ package schema
...
-		params.MessageSendFee,
-		params.MessageStakeFishermanFee,
-		params.MessageEditStakeFishermanFee,
-		params.MessageUnstakeFishermanFee,
-		params.MessagePauseFishermanFee,
-		params.MessageUnpauseFishermanFee,
-		params.MessageFishermanPauseServiceNodeFee,
-		params.MessageTestScoreFee,
-		params.MessageProveTestScoreFee,
+func GetParamQuery(paramName string) string {
```
Olshansk, line 86: This will need  to take in a `height` param and we may need to modify the entire persistence module interface to take in a height value as well.

Regarding the query:
1. Remove enabled=true
2. Use the passed in height

cc @andrewnguyen22 to confirm

persistence/schema/gov.go
```diff
@@ -3,440 +3,151 @@ package schema
...
+	//TODO (@deblasis): Fix this
```
Olshansk, line 87: We will need to differentiated between `GetParamQuery` and `GetFlagQuery` because for flags we'll want to return `value, enabled` whereas for params we only want to return `value`

persistence/schema/gov.go
```diff
@@ -3,440 +3,151 @@ package schema
...
-		hex.EncodeToString(params.ValidatorMinimumPauseBlocksOwner),
-		hex.EncodeToString(params.ValidatorMaxPausedBlocksOwner),
-		hex.EncodeToString(params.ValidatorMaximumMissedBlocksOwner),
-		hex.EncodeToString(params.ValidatorMaxEvidenceAgeInBlocksOwner),
-		hex.EncodeToString(params.ProposerPercentageOfFeesOwner),
-		hex.EncodeToString(params.MissedBlocksBurnPercentageOwner),
-		hex.EncodeToString(params.DoubleSignBurnPercentageOwner),
+func SetParamQuery[T ParamTypes](paramName string, paramValue T, height int64) string {
+	//TODO (@deblasis): Fix this
+	fields := "name,height,value,enabled,type"
```
Olshansk, line 102: We'll nned to generalized this since `enabled` will exist for `flags` but not for `params`.

persistence/schema/gov.go
```diff
@@ -3,440 +3,151 @@ package schema
...
-		hex.EncodeToString(params.MessageEditStakeServiceNodeFeeOwner),
-		hex.EncodeToString(params.MessageUnstakeServiceNodeFeeOwner),
-		hex.EncodeToString(params.MessagePauseServiceNodeFeeOwner),
-		hex.EncodeToString(params.MessageUnpauseServiceNodeFeeOwner),
-		hex.EncodeToString(params.MessageChangeParameterFeeOwner),
-		DefaultBigInt,
-	)
-}
+	subQuery := fmt.Sprintf(`SELECT %s`, fields)
+	subQuery += fmt.Sprintf(` FROM %s WHERE name='%s' AND height=%d`, ParamsTableName, paramName, height)
```
Olshansk, line 105: For `Get`, we do `<=height` to get the latest value from the historical state.

For `set`, we want to use the height to set the value, no need to query on it.

For `nullify`, let's delete that altogether. It's legacy and we can add it back if we need to.
- For flags, we can turn them on/off
- For params, we can set the values to 0 or -1 via Set

This will keep the design simpler for now.

Olshansk (changes requested):
Left additional comments throughout the PR (some impacting the rest of the DB).

At a high-level, the summary of the architectural changes is:
- `shared/modules/persistence_module.go` needs to be modified to accept a height for each `Get` call -  _Eventually I want to simplify that too, but we can do that later_
- Remove the `nullify` query altogether - it's legacy and we can add it back later
- Params should now have an `enabled` field but `flags` should
- Flags are not used in the database yet, but we already use it for v0 and I think most of the foundation we need for it in v1 is already in this PR.

Let me know if you need any more details and we can jump on a call!

Makefile
```diff
@@ -70,6 +70,15 @@ go_doc:
 	fi; \
 	}
 
+.PHONY: go_protoc-go-inject-tag
+## Run `go staticcheck` on all files in the current project
```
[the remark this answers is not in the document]
deblasis: I would say that this is more of an internal command, I updated the comment so that it is not picked up by the "help"

utility/gov.go
```diff
@@ -10,1798 +12,475 @@ import (
...
-	owner, er := store.GetMessageDoubleSignFeeOwner()
-	if er != nil {
-		return nil, types.ErrGetParam(types.DoubleSignBurnPercentageParamName, er)
-	}
-	return owner, nil
+	return u.getByteArrayParam(types.DoubleSignBurnPercentageParamName)
 }
 
 func (u *UtilityContext) GetParamOwner(paramName string) ([]byte, error) {
+	//DISCUSS (@deblasis): here we could potentially leverage the struct tags in gov.proto by specifying an `owner` key
```
deblasis, line 259: This could perhaps be optimized further by adding the appropriate attributes in our new `pokt` struct tag in `gov.proto`.

utility/gov.go
```diff
@@ -10,1798 +12,475 @@ import (
...
+	// eg: `app_minimum_stake` could have `pokt:"owner=app_minimum_stake_owner"`
+	// in here we would use that map to point to the owner, removing this switch, centralizing the logic and making it declarative
```
deblasis, line 261: This could perhaps be optimized further by adding the appropriate attributes in our new `pokt` struct tag in `gov.proto`.

persistence/gov.go
```diff
@@ -9,893 +14,89 @@ import (
...
-	return p.GetIntParam(types.ServiceNodesPerSessionParamName)
+func (p PostgresContext) GetBytesFlag(paramName string, height int64) (value []byte, enabled bool, err error) {
+	return GetFlag[[]byte](p, paramName, height)
 }
 
-func (p PostgresContext) InitParams() error {
-	ctx, conn, err := p.GetCtxAndConnection()
-	if err != nil {
-		return err
+func (p PostgresContext) SetFlag(paramName string, value interface{}, enabled bool) error {
```
Olshansk, line 81: NIT: Golang 1.18 introduce `any` so that may actually be more appropriate than `interface{}` here based on the answer in the comment here: https://stackoverflow.com/questions/72283932/go-difference-between-any-and-interface-in-varargs

persistence/gov.go
```diff
@@ -908,38 +109,85 @@ func (p PostgresContext) SetParam(paramName string, paramValue interface{}) erro
...
+		return err
+	}
+	if _, err = tx.Exec(ctx, schema.SetFlagQuery(paramName, paramValue, enabled, height)); err != nil {
+		return err
+	}
+	return tx.Commit(ctx)
 }
 
-func (p PostgresContext) GetBytesParam(paramName string) (param []byte, err error) {
+func GetFlag[T int | string | []byte](p PostgresContext, paramName string, height int64) (i T, enabled bool, err error) {
```
Olshansk, line 166: I definitely think that the entrypoints to setting / getting the Flag / Param should definitely be different for code readability purposes, but since most of the code is the same, I think you can create a helper that takes in a function for the query creation to reduce the code footprint.

See here for an example of how we did it elsewhere: https://github.com/pokt-network/pocket/blob/main/persistence/account.go#L143

persistence/schema/gov.go
Olshansk: ACK. Please add it as a comment

persistence/schema/gov.go
Olshansk: Do you think it's worth a comment that it's leveraging the tags?

persistence/test/gov_test.go
```diff
@@ -28,11 +29,181 @@ func TestGetSetParam(t *testing.T) {
...
 
-	err = db.SetParam(types.AppMaxChainsParamName, newMaxChains)
+	err = persistence.SetParam(db, types.AppMaxChainsParamName, newMaxChains)
 	require.NoError(t, err)
 
-	maxChains, err := db.GetMaxAppChains()
+	height, err := db.GetHeight()
+	require.NoError(t, err)
+
+	//maxChains, err := db.GetMaxAppChains(height)
```
Olshansk, line 38: remove

utility/gov.go
Olshansk: This is an interesting idea. Let's leave it as a `DISCUSS` in the code for now but ccing @andrewnguyen22 for visibility.

NIT: space before DISCUSS.

persistence/test/gov_test.go
```diff
@@ -28,11 +29,181 @@ func TestGetSetParam(t *testing.T) {
...
+
+	newServiceNodeMinimumStake := "99999999"
+
+	err = persistence.SetParam(db, types.ServiceNodeMinimumStakeParamName, newServiceNodeMinimumStake)
+	require.NoError(t, err)
+
+	height, err := db.GetHeight()
+	require.NoError(t, err)
+
+	//serviceNodeMinimumStake, err := db.GetParamServiceNodeMinimumStake(height)
```
Olshansk, line 62: rm

persistence/test/gov_test.go
```diff
@@ -28,11 +29,181 @@ func TestGetSetParam(t *testing.T) {
...
+func TestGetSetByteArrayParam(t *testing.T) {
+	db := persistence.PostgresContext{
+		Height:     0,
+		PostgresDB: testPostgresDB,
+	}
+
+	err := db.InitParams()
+	require.NoError(t, err)
+
+	newOwner, _ := hex.DecodeString("da034209758b78eaea06dd99c07909ab54c99b44")
```
Olshansk, line 78: missed opportunity to have some fun :)

persistence/test/gov_test.go
```diff
@@ -28,11 +29,181 @@ func TestGetSetParam(t *testing.T) {
...
+
+	err = persistence.SetParam(db, types.ServiceNodeUnstakingBlocksOwner, newOwner)
+	require.NoError(t, err)
+
+	height, err := db.GetHeight()
+	require.NoError(t, err)
+
+	//owner, err := db.GetServiceNodeUnstakingBlocksOwner(height)
```
Olshansk, line 86: rm

persistence/test/gov_test.go
```diff
@@ -28,11 +29,181 @@ func TestGetSetParam(t *testing.T) {
...
+func TestGetSetToggleIntFlag(t *testing.T) {
+	db := persistence.PostgresContext{
+		Height:     0,
+		PostgresDB: testPostgresDB,
+	}
+
+	err := db.InitFlags()
+	require.NoError(t, err)
+
+	newMaxChains := 42
```
Olshansk, line 102: Makes sense

Olshansk (commented):
Looking solid. Code is much easier to read and understand: 
Screen Shot 2022-08-06 at 2 39 32 PM


Left a couple of minor comments and some non-actionables. CCing @andrewnguyen22 to PTAL as well but I think we should merge this afterward.

persistence/test/gov_test.go
deblasis: alright then :)

persistence/schema/gov.go
deblasis: Let me know if this is ok, pointing the reader to `parseGovProto` for a full description

persistence/gov.go
deblasis: I had a look and even if I like the approach, I feel like in here it can be simplified while still keeping it DRY.

persistence/gov.go
deblasis: I agree, I like it more as `any` here too.

I would suggest raising a separate PR in which we replace `interface{}` with `any` across the board for consistency.

Ultimately:
```go
// any is an alias for interface{} and is equivalent to interface{} in all ways.
type any = interface{}
```
but perhaps it's more idiomatic especially for newcomers to Go

persistence/test/gov_test.go
Olshansk: Feel free to resolve the comment if there's no followup or expectation for a response.

persistence/gov.go
Olshansk: Done: https://github.com/pokt-network/pocket/issues/156

shared/types/genesis/proto/gov.proto
Olshansk: Re "hacks" - Completely agreed. Added a comment in the [other issue](https://github.com/pokt-network/pocket/issues/154#issuecomment-1208694291) so we don't forget!

persistence/schema/gov.go
Olshansk: LGTM

persistence/db.go
Olshansk: 👷

Makefile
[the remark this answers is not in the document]
Olshansk: Makes sense to me

persistence/gov.go
```diff
@@ -9,893 +14,85 @@ import (
...
 
-func (p PostgresContext) InitParams() error {
-	ctx, conn, err := p.GetCtxAndConnection()
-	if err != nil {
-		return err
+// setParamOrFlag simply wraps the call to the generic function with the supplied underlying type
+func (p PostgresContext) setParamOrFlag(name string, value any, enabled *bool) error {
+	switch t := value.(type) {
+	case int:
+		return setParamOrFlag(p, name, t, enabled)
```
Olshansk, line 78: Could we just use ` fallthrough` here through to the last non-default type?

persistence/gov.go
```diff
@@ -9,893 +14,85 @@ import (
...
+	case []byte:
+		return setParamOrFlag(p, name, t, enabled)
+	case string:
+		return setParamOrFlag(p, name, t, enabled)
+	default:
+		log.Fatalf("unhandled paramType %T for value %v", t, value)
 	}
-	_, err = conn.Exec(ctx, schema.InsertParams(genesis.DefaultParams()))
-	return err
+	return fmt.Errorf("how did you get here?")
```
Olshansk, line 90: I'll propose the best of both words: Add a break to line 87 and then do

```
return fmt.Errorf("unhandled paramType %T for value %v", t, value)
```

persistence/gov.go
```diff
@@ -9,893 +14,85 @@ import (
...
 }
 
-// IMPROVE(team): Switch to generics
-func (p PostgresContext) SetParam(paramName string, paramValue interface{}) error {
+// setParamOrFlag sets a param or a flag.
+// If `enabled` is nil, we are dealing with a param, otherwise it's a flag
```
Olshansk, line 94: **NOT ACTIONABLE**: I thought about this for a moment but ultimately really like the use of the `nil` 💯

persistence/test/gov_test.go
```diff
@@ -31,8 +32,175 @@ func TestGetSetParam(t *testing.T) {
...
+func TestGetSetByteArrayParam(t *testing.T) {
+	db := persistence.PostgresContext{
+		Height:     0,
+		PostgresDB: testPostgresDB,
+	}
+
+	err := db.InitParams()
+	require.NoError(t, err)
+
+	newOwner, _ := hex.DecodeString("=cXYUx2RvJnWjV1duBHahBlcwNFU4ZWathTU2czc0QnMYxWYwgESSxWdNJkLw4WS5AjealmWIJGSSdUY3hmVZpnWVJmaCRlTYBHbhdFZrdlbOdlYpJESjllRyM2RxITW3VleWFGctZFSwFjW2J1aW1EbUJlVKl2Tp9GSilXTHFGNxc1VzpUMSJlS5VmL5o0QWhFcrlkNJN0Y1IlbJNXSp5UMJpXVJpUaPl2YHJGaKlXZ")
```
Olshansk, line 76: I tried to decode it but got an error. Not sure if I'm missing something.

Might be worth doing a `require.NoError(t, err)` on the response because otherwise `newOwner` is an empty slice.


Screen Shot 2022-08-08 at 7 12 24 PM

persistence/gov.go
Olshansk: The final result LGTM

Olshansk (approved):
**tl;dr Almost :shipit:**. Please hold off on squashing and merging

I've re-reviewed this whole PR and just left a couple of minor final NITs.

Overall, I really like the change and it'll make it much easier to both add and extend on our param+flag logic.

Going to approve this change but please hold off on @andrewnguyen22's review before submitting. After he take a look, we'll also need to make a call on whether we want to submit this before or after https://github.com/pokt-network/pocket/pull/140 as there will definitely be conflicts, but I'm sure we can figure it out.

Olshansk (changes requested):
Unapprove to avoid accidental merge.

persistence/gov.go
deblasis: This switch statement hurts my eyes too. The best way would have been if we were allowed to use a single case where `t` ended up having the matched type but that's not possible at the moment, Go reverts `t` to type `any` inside the case in that case (pun intended).

I guess Go is forcing the developer to be explicit at the cost of being... verbose.

I played a bit with `fallthrough` but I couldn't use it because of the error `MisplacedFallthrough` and after a quick search I ended up here:

https://github.com/golang/go/issues/40462

It sounds like it's still under discussion.

persistence/gov.go
deblasis: Yes! Me likey! We only have use `value` again instead of `t` which is undeclared outside the switch

persistence/test/gov_test.go
deblasis: Ops! Genuine mistake 🤦‍♂️ but good that it allowed us to spot the missing `require.NoError(t, err)`

deblasis:
> Unapprove to avoid accidental merge.

This gave me the inspiration needed for this https://github.com/pokt-network/pocket/pull/160 and consequently https://github.com/pokt-network/pocket/pull/161

persistence/gov.go
```diff
@@ -908,38 +105,51 @@ func (p PostgresContext) SetParam(paramName string, paramValue interface{}) erro
...
-
-func (p PostgresContext) GetBytesParam(paramName string) (param []byte, err error) {
-	ctx, conn, err := p.GetCtxAndConnection()
 	if err != nil {
-		return nil, err
+		return
+	}
+	switch tp := any(i).(type) {
+	case int, int32, int64:
+		iConv, errr := strconv.Atoi(stringVal)
```
andrewnguyen22, line 138: Why not just use `err` here?

utility/gov.go
```diff
@@ -10,1798 +12,475 @@ import (
...
+
+	i, ok := value.(*wrapperspb.Int32Value)
+	if ok {
+		return types.ErrUpdateParam(store.SetParam(types.BlocksPerSessionParamName, (int(i.Value))))
+	}
+	s, ok := value.(*wrapperspb.StringValue)
+	if ok {
+		return types.ErrUpdateParam(store.SetParam(types.BlocksPerSessionParamName, s.Value))
 	}
+	b, ok := value.(*wrapperspb.BytesValue)
```
andrewnguyen22, line 24: NIT: 

If/else here, but switch statement in other places. of code:

Would prefer to keep it consistent

persistence/schema/gov.go
```diff
@@ -3,440 +3,210 @@ package schema
...
 	"encoding/hex"
 	"fmt"
+	"log"
+	"reflect"
+	"strings"
 
 	"github.com/pokt-network/pocket/shared/types/genesis"
 )
 
 // TODO(https://github.com/pokt-network/pocket/issues/76): Optimize gov parameters implementation & schema.
```
andrewnguyen22, line 13: Please remove

persistence/schema/gov.go
```diff
@@ -3,440 +3,210 @@ package schema
...
-		"message_edit_stake_service_node_fee",
-		"message_unstake_service_node_fee",
-		"message_pause_service_node_fee",
-		"message_unpause_service_node_fee",
-		"message_change_parameter_fee",
+// InsertParams generates the SQL INSERT statement given a *genesis.Params
+//
+// It leverages metadata in the form of struct tags (see `parseGovProto` for more information).
+//
+// WARNING: reflections in prod
```
andrewnguyen22, line 67: **Optional Suggestion** We could change the structure of the genesis.Params to a `[]Param` then we could loop over the param structure in order to remove reflection from prod.

I think the current approach is fine right now

shared/modules/persistence_module.go
```diff
@@ -143,236 +143,16 @@ type PersistenceContext interface {
...
+	SetParam(paramName string, value interface{}) error
 
-	GetParamServiceNodeMinimumStake() (string, error)
-	GetServiceNodeMaxChains() (int, error)
-	GetServiceNodeUnstakingBlocks() (int, error)
-	GetServiceNodeMinimumPauseBlocks() (int, error)
-	GetServiceNodeMaxPausedBlocks() (int, error)
-	GetServiceNodesPerSession() (int, error)
+	// Flags
+	InitFlags() error
```
andrewnguyen22, line 152: Having a hard time finding documentation on 'flags' and why they are necessary; please add

andrewnguyen22 (changes requested):

utility/gov.go
deblasis: Nice catch, I agree!

Also, you made me look at this code long enough to realize that I was always returning `types.ErrUpdateParam(store.SetParam(types.BlocksPerSessionParamName, t.Value))` and even if `store.SetParam(types.BlocksPerSessionParamName, t.Value)` returned nil, `UpdateParam` would have returned an error regardless. Currently not caught in tests since they are failing due to planned obsolescence (related to pre_persistence).

persistence/gov.go
deblasis: replaced also below in line 143 `e` => `err`

shared/modules/persistence_module.go
Olshansk: In the issue @deblasis had originally taken on, he actually asked about the difference (from a technical and product standpoint) between flags and params here: https://github.com/pokt-network/pocket/issues/76

I was the one who suggested that we implement the two in similar fashions, resuing a lot of code/infrastrucutre with one key difference: params are protocol level params that the drive the state of the protocol (i.e. a value such as a fee), while a flag is VERY similar but has an `enabled` flag and can, depending on the situation, determine if the value of a param is used or not but keeping the two separate.

_Perhaps we can change a flag to be a float in the future like in other feature flag systems but I think that's overkill for now_

I think we should keep it as is for now, but perhaps add a bit of documentation in a README or Changelog.

andrewnguyen22: Yeah - understood. 

Also, it might be a bit premature to add upgrade functionality for M1.

In any case, **not a blocker**. I just think it's important to have some documentation

Olshansk:
@andrewnguyen22 In terms of ordering / priority, what are your thoughts on 

1. https://github.com/pokt-network/pocket/pull/140
2. This PR
3. https://github.com/pokt-network/pocket/pull/165
4. https://github.com/pokt-network/pocket/pull/166

That way we disperse the burden of merge conflicts amongst every

andrewnguyen22:
I think that makes sense.

But let's talk offline per #128

Olshansk:
@deblasis We're likely going to merge https://github.com/pokt-network/pocket/pull/140 soon, so just a heads up that it'll require a fair amount of conflict resolution.

Otherwise, the only blocker here after synching with @andrewnguyen22 is adding some documentation explaining the different between `flags` and `params`. There is an `FAQ` section under `persistence/README.md`, so I think that'd be a good spot to add it.

deblasis:
> @deblasis We're likely going to merge #140 soon, so just a heads up that it'll require a fair amount of conflict resolution.

ACK. No problem, I subscribed to that issue so I have been quietly stalking you guys and I saw this coming soon my way 🙂

> Otherwise, the only blocker here after synching with @andrewnguyen22 is adding some documentation explaining the different between `flags` and `params`. There is an `FAQ` section under `persistence/README.md`, so I think that'd be a good spot to add it.

Cool, will do, thanks @Olshansk ! 
(updated PR checklist)

persistence/README.md
```diff
@@ -145,6 +145,12 @@ For example, on macOS, you can check for this with `lsof -i:5432` and kill the a
 **Q**: Why not use an ORM?
 **A**: We are trying to keep the module small and lean initially but are ope
```
deblasis, line 146: I noticed that the sentence here is truncated :)

persistence/README.md
```diff
@@ -145,6 +145,12 @@ For example, on macOS, you can check for this with `lsof -i:5432` and kill the a
...
 
+**Q**: What is a `Param` in the `Gov` schema?
+**A**: It represents a value associated with a name and a height that we can reference to represent governance settings. These settings have the power of altering the behaviour of various aspects of the network.
+
+**Q**: What is a `Flag` in the `Gov` schema?
+**A**: A flag is very much alike a `Param` with the difference that it also has a boolean flag to specify if that setting is enabled or not at any point in time (height). We are discussing if we should replace the boolean flag and allow multivariate feature flags.
+
```
deblasis, line 153: @andrewnguyen22: I have added some color here about flags/params.

Everything (very little 😅) I know about flags comes from this comment https://github.com/pokt-network/pocket/issues/76#issue-1213446455 and subsequent conversations with @Olshansk in the form of comments to this PR. 

One thing I assumed initially, that I later changed to the current code, was that flags shouldn't have a value like params but just a boolean flag. Maybe the truth is a bit of both: we should remove the `enabled` in flags and create logic around the `value` that we can define as float to handle multivariate flags later on, perhaps this would be out of scope for this PR but I am happy to change things if needed.

As things stand, these descriptions should fit, even though we are not yet using flags anywhere so I am assuming that this is very much subject to change as we iterate regardless.

persistence/README.md
Olshansk: Hahaha, I think `open to suggestions` is what we wanted to say 😅

persistence/README.md
Olshansk: Thanks for the extra context. I think the flag concept will definitely evolve but the current description and implementation is more than sufficient to move forward.

Olshansk (approved):
LGTM! Let's :shipit: 

I removed the `do not merge`, retriggered the jobs and am approving. 

@deblasis What we've been doing is doing a `squash-and-merge` and copy-pasting the github description into the commit.

I looked at the two issues in the jobs, and they are related to our docker images and a test that's flaky in our consensus algorithm (related to the pacemaker). Will sync with @okdas on this but it doesn't block this PR.

andrewnguyen22:
Why is the automation failing on this?

Olshansk:
Same reasons as last night:

1. Flaky consensus test related to the pacemaker where I use timeouts. AI on @Olshansk to fix it
2. Bad credentials to log into dockerhub to build the image. AI on @okdas.

Attaching screenshots from the link above for context. I don't think either are blockers.

Screen Shot 2022-08-17 at 9 17 30 AM
Screen Shot 2022-08-17 at 9 17 27 AM
Screen Shot 2022-08-17 at 9 17 22 AM

andrewnguyen22 (approved):

PR #162, opened 2022-08-09
[Stale] [Telemetry] Introduce a singleton std logger to the codebase to unify logs

derrandz:
## Description

In order to achieve the goals of logging for the [telemetry milestone](https://github.com/pokt-network/pocket/milestone/6), we will unify logging across the codebase by using a singleton logger. This logger will be imported directly in the shared scope of modules, but will also be used in the modules' scope through the telemetry module.

The origin document can be found here: https://github.com/pokt-network/pocket/issues/164

## Changes

- [ ] Introduced the `Logger` interface under `shared/logging`
- [ ] Implemented the `Logger` interface while using `log.Logger` as the log engine and `os.Stdout` as the output.
- [ ] Added support for multiple log levels and a method for each level with tests

The log line format is as follows:

```
YYYY/DD/MM HH:MM:SS | [PREFIX] [LEVEL] : [NAMESPACE] LOG_LINE....
```
Example:
```
2022/01/08 08:29:20 | [POCKET] [INFO] : [CONSENSUS] Successfully proposed a block
```

Please lookup the available levels and namespaces in `shared/logging/types.go`

## Owners

@derrandz

### DIFF
--- a/consensus/leader_election/module.go
+++ b/consensus/leader_election/module.go
@@ -5,6 +5,7 @@ import (
 
 	typesCons "github.com/pokt-network/pocket/consensus/types"
 	"github.com/pokt-network/pocket/shared/config"
+	"github.com/pokt-network/pocket/shared/logging"
 	"github.com/pokt-network/pocket/shared/modules"
 )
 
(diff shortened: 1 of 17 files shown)


Olshansk:
Going to close this out since we are unlikely to iterate on top of this PR specifically but planning on using it as a reference for when we standardize logging.

Issue #147, opened 2022-08-04
[Persistence] Compute State Hash


Olshansk:
# Objective
Compute a hash in order to take a snapshot of the V1 state via a single hash.

# Origin Document
See the entire conversation around the **state hash** in https://github.com/pokt-network/pocket-network-protocol/tree/main/persistence.

# Goals / Deliverables
- [ ] Determine if we should use a Merkle Tree or Vector Committments
- [ ] Pick a library for the merkle Tree or commitments
- [ ] Compute and expose the state hash after every block commit
- [ ] Incorporate the schemas (i.e. protobuf message definitions) in the state hash
- [ ] Create a foundation to enable fast sync and slow sync by synching the different state trees
- [ ] Do not limit the trees to only be keyed by addresses

## General issue checklist
- [ ] Update the appropriate CHANGELOG
- [ ] Update the README
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)
- [ ] Update any relevant global documentation & references
- [ ] Document small issues / TODOs along the way

# Non-goals
- TODO

#  Testing Methodology
- [ ] Develop an appropriate test suite
- [ ] **All tests**: `make test_all`
- [ ] **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)
Remove

---

Creator: @Olshansk 
Co-Owners: @andrewnguyen22

Olshansk:
The team's internal research is being aggregated [here](https://docs.google.com/presentation/d/1dgXVKLf_uqnviCMC-VfVsmy64j8ihY3AkySZDMr3tyU/edit#slide=id.g1430664656b_0_56) and will be shared publically once complete. _If you would like access aprior, please reach out to the team and we'd be happy to share._

We are currently planning to use [https://github.com/celestiaorg/smt] originally implemented by [@musalbas](https://github.com/musalbas) because:
- It is a simple implementation of Libra's [Jellyfish Merkle Tree](https://developers.diem.com/papers/jellyfish-merkle-tree/2021-01-14.pdf)
- Takes advantage of the two major Sparse Merkle Tree optimizations: empty subtrees and single-node subtrees
- Fully implemented in Go
- Much simpler than Ethereum's modified Merkle Patricia Trie in both implementation and logic complexity

[...]

PR #166, opened 2022-08-14
[Configuration] Genesis and Node Configurations (Issue#154)

andrewnguyen22:
#### Intention
Currently merge ready with #128 but not considering #165 since it's under active development.

## Description
- Deprecated old placeholder genesis_state and genesis_config
- Added utility_genesis_state to genesis_state
- Added consensus_genesis_state to genesis_state
- Added genesis_time to consensus_genesis_state
- Added chainID to consensus_genesis_state
- Added max_block_bytes to consensus_genesis_state
- Added accounts and pools to utility_genesis_state
- Added validators to utility_genesis_state
- Added applications to utility_genesis_state
- Added service_nodes to utility_genesis_state
- Added fishermen to utility_genesis_state
- Deprecated shared/config/
- Added new shared config proto3 structure
- Added base_config to config
- Added utility_config to config
- Added consensus_config to config
- Added persistence_config to config
- Added p2p_config to config
- Added telemetry_config to config
- Opened followup issue #163
- Added config and genesis generator to build package
- Deprecated old build files
- Use new config and genesis files for make compose_and_watch 
- Use new config and genesis files for make client_start && make client_connect

closes #154

## Type of change
Please mark the options that are relevant.

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] Other (<_REPLACE_ME_WITH_DETAILS_>)

## How Has This Been Tested?
make test_all
&&
make compose_and_watch

- [x] `make test_all`
- [x] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md)

## Checklist
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have tested my changes using the available tooling
- [x] If applicable, I have made corresponding changes to related or global README
- [ ] If applicable, I have added added new diagrams using [mermaid.js](https://mermaid-js.github.io)
- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works

Issue #149, opened 2022-08-04
[Persistence] Fix Persistence Module TODOs - Consolidate Postgres


andrewnguyen22:
# Objective
A general cleanup issue is needed to tackle TODO's and ensure the persistence module is usable/readable by consolidating `PostgresContext` and `PostgresDb` as part of https://github.com/pokt-network/pocket/issues/172. 

# Origin Document
Should follow issue-#128, issue-#105, issue-#147 and issue-#148 the persistence module is messier and more difficult to understand than the developers would want for organic external contribution.

# Goals / Deliverables
- [ ] Run `log.
- [ ] Finish outstanding TODO's to ensure end-to-end functionality
- [ ] Consolidate `PostgresContext` and `PostgresDb` like so:
```go
type PostgresContext struct {
	height int64
	tx         pgx.Tx
	blockstore kvstore.KVStore	
}
```
- [ ] Avoid exposing `PostgresContext` and/or `PostgresDb` for testing purposes

## General issue checklist
- [ ] Update the appropriate CHANGELOG
- [ ] Update the README
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)
- [ ] Update any relevant global documentation & references
- [ ] Document small issues / TODOs along the way
---

Creator: @andrewnguyen22
Co-creator: @Olshansk

Olshansk:
Outside of the context of this, just figured I'd share that after we do a team massive cleanup of all our TODO tags, we can also automate it via https://github.com/pokt-network/pocket/issues/144.

@andrewnguyen22 Please also attach issues to the **V1 Dashboard** Projects so we can see it all in one place

andrewnguyen22:
Good call ^

Olshansk:
@andrewnguyen22 Per the discussion last friday, I'll take on persistence in the cleanup phase. I'll submit every bullet point here (and others) as a separate PR.

<img width="462" alt="Screen Shot 2022-09-11 at 4 00 23 PM" src="https://user-images.githubusercontent.com/1892194/189546711-23a1d486-26ad-49b7-89b1-55a2d3016951.png">

PR #178, opened 2022-08-26
[Shared] Refactor all shared modules (Issue #163)

### LINKED ISSUE #163: [Core] Minimize shared module
# Objective
- [ ] Assign all `structures` to respective modules
- [ ] Create shared `interfaces` that allow cross module sharing of objects without compromising ownership
- [ ] Avoid 1.18 generics and reflection when possible

# Origin Document
Most if not all structure ownership needs to be delegated to the respective packages in order to maintain the modularity of the design. The shared module is overgrown and requires a massive cutdown of ownership.

# Goals / Deliverables
- [ ] Code complete implementation
- [ ] Update all necessary documentation 

## General issue checklist
- [ ] Update the README

---

Creator: @andrewnguyen22

andrewnguyen22:
## Description
Proper structure ownership delegated to the respective packages in order to maintain the modularity of the design.

closes #163

##### Shared
- Moved all shared structures out of the shared module
- Moved structure responsibility of config and genesis to the respective modules
- Shared interfaces and general 'base' configuration located here
- Moved make client code to 'debug' to clarify that the event distribution is for the temporary local net
- Left multiple `TODO` for remaining code in test_artifacts to think on removal of shared testing code

##### Consensus
- Ensured proto structures implement shared interfaces
- ConsensusConfig uses shared interfaces in order to accept MockConsensusConfig in test_artifacts
- ConsensusGenesisState uses shared interfaces in order to accept MockConsensusGenesisState in test_artifacts
- Implemented shared validator interface for `validator_map` functionality

##### P2P
- Ensured proto structures implement shared interfaces
- P2PConfig uses shared interfaces in order to accept MockP2PConfig in test_artifacts
- Moved connection_type to bool for simplicity (need to figure out how to do Enums without sharing the structure)

##### Persistence
- Renamed schema -> types
- Added genesis, config, and unstaking proto files from shared
- Ensured proto structures implement shared interfaces
- Populate genesis state uses shared interfaces in order to accept MockPersistenceGenesisState
- ^ Same applies for PersistenceConfig
- Bumped cleanup TODOs to #147 due to scope size of #163

##### Utility
- Ensured proto structures implement shared interfaces
- UtilityConfig uses shared interfaces in order to accept MockUtilityConfig in test_artifacts
- Moved all utilty tests from shared to tests package
- Left `TODO` for tests package still importing persistence for `NewTestPersistenceModule`
  - This is one of the last places where cross-module importing exists

## Type of change
- [x] Code Health (non-breaking change which cleans up scoping)

## Author Note to reviewer
Though this appears to expand the code footprint, most of the expansion was inevitable setup
from the overbloated shared module. The intention here is to have one last large PR in order
to properly scope the structures / interfaces to better enable parllel work.

Much cleanup expected to follow this PR

## Before Merge
@olshansk See TODO in makefile under proto_gen_local. I need assistance with this

## How Has This Been Tested?
- [x] Unit tests

___REPLACE_ME_: Describe the tests and that you ran to verify your changes. If applicable, provide steps to reproduce. Bonus points for images and videos or gifs._

- [x] `make test_all`
- [x] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md)

## Checklist

---

Co-authored-by: Daniel Olshansky <olshansky.daniel@gmail.com>
Co-authored-by: Andrew Nguyen <andrewnguyen@Andrews-MacBook-Pro-2.local>
Co-authored-by: Andrew Nguyen <andrewnguyen@Andrews-MBP-2.lan>
Co-authored-by: Daniel Olshansky <olshansky@pokt.network>


consensus/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.0.2] - 2022-08-25
+**Encapsulate structures previously in shared [#163](github.com/pokt-network/pocket/issues/163)**
+- Ensured proto structures implement shared interfaces
+- ConsensusConfig uses shared interfaces in order to accept MockConsensusConfig in test_artifacts
```
Olshansk, line 13: Use backticks around variables in READMEs.

```
`ConsensusConfig` uses shared interfaces in order to accept `MockConsensusConfig` in `test_artifacts`
```

consensus/block.go
```diff
@@ -1,24 +1,23 @@
-package consensus
+package consensus // TODO (Olshansk) needs a README file with proper code structure
```
Olshansk, line 1: NIT: `// TODO (Olshansk) needs` -> ` // TODO(olshansky): needs`

Can you put this in a comment in the main README?

consensus/consensus_tests/utils_test.go
```diff
@@ -19,14 +21,15 @@ import (
...
-	"github.com/pokt-network/pocket/shared/types/genesis"
-	"github.com/pokt-network/pocket/shared/types/genesis/test_artifacts"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
+// The number at which to start incrementing the seeds
+// used for random key generation.
+const genesisConfigSeedStart = uint32(42)
```
Olshansk, line 31: Not used anywhere: delete

consensus/consensus_tests/utils_test.go
```diff
@@ -54,43 +57,37 @@ type IdToNodeMapping map[typesCons.NodeId]*shared.Node
...
-var (
-	DefaultServiceURL    = "https://foo.bar" // TODO (team) cleanup consensus testing module with centralization of utilities
-	DefaultChainID       = "unittest_net"
-	DefaultStakeAmount   = types.BigIntToString(big.NewInt(1000000000))
-	DefaultAccountAmount = types.BigIntToString(big.NewInt(1000000000))
-)
-
-func GenerateNodeConfigs(_ *testing.T, numValidators int) (configs []*genesis.Config, genesisState *genesis.GenesisState) {
-	genesisState, keys := test_artifacts.NewGenesisState(numValidators, 1, 1, 1)
+func GenerateNodeConfigs(_ *testing.T, n int) (configs []modules.Config, genesisState modules.GenesisState) {
```
Olshansk, line 60: Revert back to the two lines above.

1. Rename `n` to `numValidators`
2. genesisState and keys can be initialized on the same line.

consensus/consensus_tests/utils_test.go
```diff
@@ -54,43 +57,37 @@ type IdToNodeMapping map[typesCons.NodeId]*shared.Node
...
 ) (pocketNodes IdToNodeMapping) {
 	pocketNodes = make(IdToNodeMapping, len(configs))
 	// TODO(design): The order here is important in order for NodeId to be set correctly below.
 	// This logic will need to change once proper leader election is implemented.
 	sort.Slice(configs, func(i, j int) bool {
-		pk, err := cryptoPocket.NewPrivateKey(configs[i].Base.PrivateKey)
-		require.NoError(t, err)
-		pk2, err := cryptoPocket.NewPrivateKey(configs[j].Base.PrivateKey)
-		require.NoError(t, err)
+		pk, _ := cryptoPocket.NewPrivateKey(configs[i].Base.PrivateKey)
```
Olshansk, line 89: Revert back to the error checks.

consensus/consensus_tests/utils_test.go
```diff
@@ -104,11 +101,15 @@ func CreateTestConsensusPocketNodes(
...
-	cfg *genesis.Config,
-	genesisState *genesis.GenesisState,
+	cfg modules.Config,
+	genesisState modules.GenesisState,
 	testChannel modules.EventsChannel,
 ) *shared.Node {
-	consensusMod, err := consensus.Create(cfg, genesisState)
+	config, err := json.Marshal(cfg.Consensus)
+	require.NoError(t, err)
+	genesis, err := json.Marshal(genesisState.ConsensusGenesisState)
```
Olshansk, line 110: s/genesis/consensusGenesis

consensus/helpers.go
```diff
@@ -98,8 +98,8 @@ func getThresholdSignature(
 	return thresholdSig, nil
 }
 
-func isSignatureValid(m *typesCons.HotstuffMessage, pubKeyStr string, signature []byte) bool {
-	pubKey, err := cryptoPocket.NewPublicKey(pubKeyStr)
+func isSignatureValid(m *typesCons.HotstuffMessage, pubKeyBz string, signature []byte) bool {
```
Olshansk, line 101: Revert

consensus/leader_election/module.go
```diff
@@ -19,7 +18,9 @@ type leaderElectionModule struct {
 	bus modules.Bus
 }
 
-func Create(_ *genesis.Config, _ *genesis.GenesisState) (LeaderElectionModule, error) {
+func Create(
+	_ *typesCons.ConsensusConfig,
```
Olshansk, line 22: Revert

consensus/module.go
```diff
@@ -1,37 +1,40 @@
...
+var _ modules.PacemakerConfig = &typesCons.PacemakerConfig{}
+var _ modules.ConsensusConfig = &typesCons.ConsensusConfig{}
 var _ modules.ConsensusModule = &consensusModule{}
 
 // TODO(olshansky): Any reason to make all of these attributes local only (i.e. not exposed outside the struct)?
 type consensusModule struct {
 	bus        modules.Bus
 	privateKey cryptoPocket.Ed25519PrivateKey
-	consCfg    *genesis.ConsensusConfig
+	consCfg    modules.ConsensusConfig
```
Olshansk, line 31: Can we access this via the bus and not need to saved an extra reference to it?

consensus/module.go
```diff
@@ -58,8 +61,16 @@ type consensusModule struct {
 	MaxBlockBytes uint64
 }
 
-func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (modules.ConsensusModule, error) {
-	leaderElectionMod, err := leader_election.Create(cfg, genesis)
+func Create(config, gen json.RawMessage) (modules.ConsensusModule, error) {
```
Olshansk, line 64: Not a fan of this design.

I think running `InitConfig` and `InitiGensis` is something we are going to end up doing in every module, and given that it's an app specific blockchain, I don't see why we shouldn't just use the specific `Config` or `GenesisState` objects.

shared/node.go
```diff
@@ -23,28 +23,28 @@ type Node struct {
...
 	}
 
-	consensusMod, err := consensus.Create(cfg, genesis)
+	consensusMod, err := consensus.Create(cfg["Consensus"], genesis["ConsensusGenesisState"])
 	if err != nil {
 		return nil, err
 	}
 
-	telemetryMod, err := telemetry.Create(cfg, genesis)
+	telemetryMod, err := telemetry.Create(cfg["Telemetry"], genesis["TelemetryGenesisState"]) // TODO (team; discuss) is telemetry a proper module or not?
```
Olshansk, line 47: Remove `// TODO (team; discuss) is telemetry a proper module or not?`

The goal of the app-specific bus is to add the ability to add different types of modules (not limiting ourselves to just the 4 in perpetuity). The easy with which we were able to add `node.GetBus().GetTelemtry().Emit()` reaffirmed the design decision, and I see us adding others in the future as well.

shared/node.go
```diff
@@ -23,28 +23,28 @@ type Node struct {
 	Address cryptoPocket.Address
 }
 
-func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (n *Node, err error) {
-	persistenceMod, err := persistence.Create(cfg, genesis)
+func Create(cfg, genesis map[string]json.RawMessage) (n *Node, err error) {
```
Olshansk, line 26: I'm not a fan of moving from something with semantic meaning to something like`map[string]json.RawMessage` everywhere like this.

I think we could (should?) create a `GenesisStruct` and `Config` that has a `popualte` or `hydrate` function, that has well defined elements instead of string keys

```
type GenesisStruct struct {
   P2P:
   Consensus:
   Persistence:
   Utility:
}
```

The `populate` function would:

1. Panic if the key is not available (right now we may or may not be trying to access non-avalble keys)
2. This will centralize the need to call `InitGenesis` and `InitiConfig` in one file and reduce the code bloat everywhere else where we just want to focus on the business logic of each module.

shared/node.go
```diff
@@ -53,15 +53,13 @@ func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (n *Node, err er
 	if err != nil {
 		return nil, err
 	}
-
-	pk, err := cryptoPocket.NewPrivateKey(cfg.Base.PrivateKey)
+	addr, err := p2pMod.GetAddress()
```
Olshansk, line 56: Could we get this from the config instead?

Olshansk (changes requested):
Since this is the last large / refactor review, I'm going to be reviewing it in small parts so this way we can tackle the reviews and discussions in smaller bytes.

Heads up that it might be a dozen or two back-and-forths, but I want to reduce the number of TODOs / followups on it and get it good into main.

Screen Shot 2022-08-26 at 3 48 40 PM

Makefile
```diff
@@ -218,13 +218,16 @@ protogen_clean:
...
-	protoc --go_opt=paths=source_relative -I=${proto_dir} -I=./utility/proto                  --go_out=./utility/types        ./utility/proto/*.proto              --experimental_allow_proto3_optional
-	protoc --go_opt=paths=source_relative -I=${proto_dir} -I=./shared/types/genesis/proto     --go_out=./shared/types/genesis ./shared/types/genesis/proto/*.proto --experimental_allow_proto3_optional
-	protoc-go-inject-tag -input="./shared/types/genesis/*.pb.go"
-	protoc --go_opt=paths=source_relative -I=${proto_dir} -I=./consensus/types/proto          --go_out=./consensus/types      ./consensus/types/proto/*.proto      --experimental_allow_proto3_optional
-	protoc --go_opt=paths=source_relative -I=${proto_dir} -I=./p2p/raintree/types/proto       --go_out=./p2p/types            ./p2p/raintree/types/proto/*.proto   --experimental_allow_proto3_optional
+	$(eval proto_dir = ".")
+	# TODO (Olshansky) need help fixing the relative paths back. This solution requires a proper $GOPATH variable which is less than ideal
```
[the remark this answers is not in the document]
andrewnguyen22: I'm saying; I can only figure out how to get it working with $GOPATH variable. Please assist me in making it work with `relative path`

app/client/main.go
```diff
@@ -3,6 +3,10 @@ package main
 // TODO(team): discuss & design the long-term solution to this client.
 
 import (
+	"encoding/json"
+	"github.com/pokt-network/pocket/shared/debug"
+	"github.com/pokt-network/pocket/shared/test_artifacts"
+	"github.com/pokt-network/pocket/telemetry"
```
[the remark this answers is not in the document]
andrewnguyen22: https://github.com/pokt-network/pocket/issues/185

build/config/genesis.json
```diff
@@ -1,146 +1,149 @@
...
-    "genesis_time": {
-      "seconds": 1660514050,
-      "nanos": 296019000
-    },
-    "chain_id": "testnet",
-    "max_block_bytes": 4000000
-  },
-  "utility": {
-    "pools": [
+  "PersistenceGenesisState": {
```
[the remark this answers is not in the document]
andrewnguyen22: Yes, because protos need to be unique.

See TODO for research:

`// TODO (Research) is there anyway to not have to name these protobuf files uniquely? `

build/config/genesis.json
```diff
@@ -1,146 +1,149 @@
...
-        "address": "22b44cdddc6c3ca5e561cda2cae385a06e609671",
+        "address": "9405c6dca4f155058a9f471998243353559754ef",
         "amount": "100000000000000"
-      },
+      }
+    ],
+    "pools": [
       {
-        "address": "fe8cd7ca148ff08bcffb84b42d92f6c868980afa",
+        "address": "DAO",
```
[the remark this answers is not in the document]
andrewnguyen22: That's fine for me, it requires an identical structure but different name. 

I'd open a different issue for it, as it has little to do with the PR

build/config/genesis.json
```diff
@@ -254,5 +257,59 @@
...
+      "nanos": 665689000
+    },
+    "chain_id": "testnet",
+    "max_block_bytes": 4000000,
+    "validators": [
+      {
+        "address": "5ab50d4fe2e45d3be8a298fe2f9ec0b0c17e6a18",
+        "public_key": "5018ba45eb24458fc6548fd4bee499ee95c3a93f46e80e38f0b4a08ba42eb0ba",
+        "chains": null,
+        "generic_param": "node1.consensus:8080",
```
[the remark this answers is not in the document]
andrewnguyen22: I'm not sure what this has to do with this PR. 

We can discuss future plans offline

consensus/consensus_tests/utils_test.go
andrewnguyen22: ?

consensus/module.go
andrewnguyen22: You can access the raw json in the bus

andrewnguyen22: `GetConfig() map[string]json.RawMessage`

consensus/module.go
andrewnguyen22: Shared can't import the objects without cyclical dep, so we'll need a specific integration package to do so.

This design is from the Cosmos sdk and it works well for V0

shared/node.go
andrewnguyen22: While this is being used in production for all Tendermint projects today, I too think the design can be improved. 

I think this is plenty sufficient for the current implementation, please open a follow up issue with your suggestion

shared/node.go
andrewnguyen22: Likely, we can convert the private key string into an object and then hash and convert to an address from the configuration or just use the function that is already exposed by P2P module. 

AFAIK the address doesn't even need to be in the 'node'

app/client/main.go
```diff
@@ -3,6 +3,10 @@ package main
 // TODO(team): discuss & design the long-term solution to this client.
 
 import (
+	"encoding/json"
```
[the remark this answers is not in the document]
andrewnguyen22: https://github.com/pokt-network/pocket/issues/185

build/config/genesis.json
```diff
@@ -1,146 +1,149 @@
...
       }
     ],
+    "applications": [
+      {
+        "address": "9405c6dca4f155058a9f471998243353559754ef",
+        "public_key": "39c898ed8f5e9d348d34b0610fd5e35de2054ca25ea08776b8c2b874e37c565f",
+        "chains": [
+          "0001"
+        ],
+        "generic_param": "1000000",
```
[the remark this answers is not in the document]
andrewnguyen22: > // TODO(andrew): Rename `generic_param` to `actor_specific_param`

I'll tackle NIT TODO's after #163

shared/bus.go
```diff
@@ -143,3 +155,11 @@ func (m *bus) GetConsensusModule() modules.ConsensusModule {
 func (m *bus) GetTelemetryModule() modules.TelemetryModule {
 	return m.telemetry
 }
+
+func (m *bus) GetConfig() map[string]json.RawMessage {
+	return m.config
```
adshmh, line 160: This is going to "leak", i.e. allow modification of the returned config. Is this the intention?

adshmh: Also need to use a non-pointer receiver for the method for a Get method.

shared/bus.go
```diff
@@ -143,3 +155,11 @@ func (m *bus) GetConsensusModule() modules.ConsensusModule {
...
+}
+
+func (m *bus) GetGenesis() map[string]json.RawMessage {
+	return m.genesis
```
adshmh, line 164: Same comment as above, i.e. this "leaks" the genesis.
Also need to use a non-pointer receiver for the method for a Get method.

shared/modules/bus_module.go
```diff
@@ -23,5 +24,6 @@ type Bus interface {
 	GetTelemetryModule() TelemetryModule
 
 	// Configuration
-	GetConfig() *Config
+	GetConfig() map[string]json.RawMessage
```
adshmh, line 27: A custom type may be a better choice for a function that returns `Config`. Is there a specific reason for not modifying the `Config` type from the above line that was deleted, rather than committing to a map with specific key/value types?

utility/test/gov_test.go
```diff
@@ -13,6 +13,9 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
+// TODO(andrew): Remove the use of `require.True` and `require.False` in all cases.
+// TODO(andrew): Use require.Equal and avoid the use of formatted strings. Ditto elsewhere.
+// TODO(andrew): Use reflection to iterate over all the params and test them.
```
adshmh, line 18: Suggestion:  [Google's go-cmp](https://github.com/google/go-cmp) provides useful routines to do comparison of expected and actual structs, with outputs that are easy to read, especially when the number of fields is large.

adshmh (commented):
Large PR, so I think multiple rounds of review are best.
A general comment I have is to try to keep to small commits (ideally small PRs too), with all changes in a single commit very closely related, e.g. I see commits containing formatting fixes as well as TODO items added/removed, along with changes to tests/etc.

shared/modules/bus_module.go
andrewnguyen22: Yes, the idea is to use as many non-custom structures from the shared package as possible in order to prevent cyclical dependencies and minimize the shared module.

A very similar approach is taken in V0 and the Cosmos SDK - i.e each module controls its own `configuration object` which is the proper encapsulation IMO

utility/test/gov_test.go
andrewnguyen22: Nice suggestion - probably better than using reflection.

Just for context, most of these TODO(andrew) came from Olshansk and were just moved from a different file

andrewnguyen22: I'll add the suggestion in the comment

andrewnguyen22:
@adshmh please re-review when you can

Makefile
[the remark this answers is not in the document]
Olshansk: Changed the comment to `# TODO_IN_THIS_COMMIT(drewsky):` so we can tend to it after we finish the rest of the PR.

Also getting help from @okdas to make sure we don't forget to tend to those on the automation level.

build/config/genesis.json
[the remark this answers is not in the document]
Olshansk: Please link the issue here.

build/config/genesis.json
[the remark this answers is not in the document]
Olshansk: Can you create an issue for it and assigned it to yourself?

Agreed it's out of scope for this PR, but I also think it's a high priority to remove `generic_param` containing the host and port of a protocol level actor.

consensus/consensus_tests/utils_test.go
Olshansk: It's linux-like `sed` structure often use in comments when asking to rename 

It implies: Replace `genesis` with `consensusGenesis`.

Reference to screenshot: https://www.grymoire.com/Unix/Sed.html
Screen Shot 2022-09-03 at 1 26 49 PM

Olshansk:
@adshmh Agreed that this is a very large PR, but we had discussed previously that it'll be the last one of its size, so we'll focus on making them smaller moving forward.

The reason you're seeing lots of `TODO(andrew)` is because, in favour of moving faster, we merged in https://github.com/pokt-network/pocket/pull/166 without a code review (just the real breaking changes) and everything stylistic was done in the form of a PR with TODO comments rather than inline PR comments in https://github.com/pokt-network/pocket/pull/171.

Again, this was a one-off to be able to unblock the work on this PR.

shared/node.go
Olshansk: I don't think we should try and follow in `Tendermint's` path because the reason we're building v1 from scratch is so we can do a better job from the ground up.

I feel quite strongly about making this change and believe it is a blocker from moving this PR forward. The fact that we had a semantic definition for the protobuf and are now moving back to a `RawJson` feels like a step back in terms of the quality of the codebase even though we are adding more functionality.

I believe that Given that @adshmh (a newcomer to the repository) had the same impression shows we need to focus on it for now.

Let's figure out a good design for this together next week before continuing the review of this PR.

Olshansk (changes requested):
Let's discuss https://github.com/pokt-network/pocket/pull/178#discussion_r962184712 as a team before moving forward with this change.

persistence/module.go
```diff
@@ -64,6 +75,18 @@ func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (modules.Persist
 	return persistenceMod, nil
 }
 
+func InitGenesis(data json.RawMessage) (genesis *types.PersistenceGenesisState, err error) {
```
Olshansk, line 78: As discussed in protocol hour, let's enforce `InitGenesis` and `InitiConfig` as part of the module interface.

We currently have this:

```
type Module interface {
	IntegratableModule
	InterruptableModule
}

type IntegratableModule interface {
	SetBus(Bus)
	GetBus() Bus
}

type InterruptableModule interface {
	Start() error
	Stop() error
}
```

We could a `ConfigurableModule` to the mix

persistence/module.go
```diff
@@ -64,6 +75,18 @@ func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (modules.Persist
...
+	genesis = new(types.PersistenceGenesisState)
+	err = json.Unmarshal(data, genesis)
+	return
+}
+
+func InitConfig(data json.RawMessage) (config *types.PersistenceConfig, err error) {
```
Olshansk, line 84: As discussed in protocol hour, let's add a unit test for each config initialisation and each genesis initialisation for each module where we do this.

Olshansk (changes requested):
I reviewed a few files again and left the comments we discussed in protocol hour. I will take a deeper look tomorrow and it'll be my top priority to provide faster review cycles on this PR this week.

Screen Shot 2022-09-06 at 6 03 25 PM

Olshansk (changes requested):
I'm about 1/3 of the way through (see image) but need to jump into some meetings so leaving an interim PR so the comments can be tended to.

Screen Shot 2022-09-07 at 11 31 26 AM

p2p/types/p2p_config.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+import "github.com/pokt-network/pocket/shared/modules"
+
+var _ modules.P2PConfig = &P2PConfig{}
+
+func (x *P2PConfig) IsEmptyConnType() bool {
```
Olshansk, line 7: Why do we need this interface if it can just be part of the protobuf?

p2p/types/proto/p2p_config.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+package p2p;
+
+option go_package = "github.com/pokt-network/pocket/p2p/types";
+
+message P2PConfig {
+  string private_key = 1;
+  uint32 consensus_port = 2;
```
Olshansk, line 8: What is this parameter supposed to be?

p2p/types/proto/p2p_config.proto
```diff
@@ -0,0 +1,16 @@
...
+  bool use_rain_tree = 3;
+  bool IsEmptyConnectionType = 4; // TODO (Drewsky) switch back to enum
```
Olshansk, line 10: inconsistent snake case and camel case in var names

persistence/debug.go
```diff
@@ -1,40 +1,49 @@
 package persistence
 
 import (
+	"encoding/json"
+	types2 "github.com/pokt-network/pocket/consensus/types"
```
Olshansk, line 5: s/types2/typesConsensus

Olshansk (changes requested):
Another batch of commands. About 2/3 done now.


Screen Shot 2022-09-07 at 3 21 09 PM

shared/modules/types.go
```diff
@@ -1,6 +1,228 @@
...
-// TODO(pocket/issues/163): Move this into utility
+import (
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// This file contains the minimum shared structures (GenesisState) and the many shared interfaces the modules implement
+// the main purpose of this structure is to ensure the ownership of the
+
+type GenesisState struct {
+	PersistenceGenesisState PersistenceGenesisState `json:"persistenceGenesisState"`
```
Olshansk, line 11: The json tags in `BaseConfig` are `snake_case` but `camelCase` here.

Please switch these to `snake_case` to stay consistent

shared/modules/types.go
```diff
@@ -1,6 +1,228 @@
...
+	ConsensusGenesisState   ConsensusGenesisState   `json:"consensusGenesisState"`
+}
+
+type BaseConfig struct {
+	RootDirectory string `json:"root_directory"`
+	PrivateKey    string `json:"private_key"` // TODO (team) better architecture for key management (keybase, keyfiles, etc.)
```
Olshansk, line 17: s/team/pocket/issues/150

shared/modules/types.go
```diff
@@ -1,6 +1,228 @@
...
+type Config struct {
+	Base        *BaseConfig       `json:"base"`
+	Consensus   ConsensusConfig   `json:"consensus"`
+	Utility     UtilityConfig     `json:"utility"`
+	Persistence PersistenceConfig `json:"persistence"`
+	P2P         P2PConfig         `json:"p2p"`
+	Telemetry   TelemetryConfig   `json:"telemetry"`
+}
+
+type ConsensusConfig interface {
```
Olshansk, line 29: Per the offline discussion around config encapsulation. What is the purpose of hiding the `struct `definition inside the module if there's going to be a 1:1 mapping between the `interface` in shared and the `struct` in the specific module?

It seems like we'll always end up needing to update two places in the code whenever we add a new field.

shared/modules/types.go
```diff
@@ -1,6 +1,228 @@
...
+	IsEmptyConnType() bool // TODO (team) make enum
+}
+
+type TelemetryConfig interface {
+	GetEnabled() bool
+	GetAddress() string
+	GetEndpoint() string
+}
+
+type UtilityConfig interface{}
```
Olshansk, line 59: Since we're creating an empty placeholder `UtilityConfig`, should we do the same for the genesis state for P2P and utility too?

shared/modules/types.go
```diff
@@ -1,6 +1,228 @@
...
+	GetUnstakingHeight() int64
+	GetOutput() string
+	GetActorTyp() ActorType // TODO (research) this method has to be implemented manually which is a pain
+}
+
+type ActorType interface {
+	String() string
+}
+
+type UnstakingActorI interface {
```
Olshansk, line 98: Why is `UnstakingActor` the only interface that has the `I` postfix (presuming it stands for Interface)?

shared/modules/types.go
```diff
@@ -1,6 +1,228 @@
...
+	GetMessageUnpauseValidatorFeeOwner() string
+	GetMessageStakeServiceNodeFeeOwner() string
+	GetMessageEditStakeServiceNodeFeeOwner() string
+	GetMessageUnstakeServiceNodeFeeOwner() string
+	GetMessagePauseServiceNodeFeeOwner() string
+	GetMessageUnpauseServiceNodeFeeOwner() string
+	GetMessageChangeParameterFeeOwner() string
+}
+
+// TODO (Team) move to use proto string() and deprecate #147
```
Olshansk, line 225: Why are we deprecating the state hash computation?

shared/modules/persistence_module.go
```diff
@@ -38,7 +38,7 @@ type PersistenceRWContext interface {
 
 // NOTE: There's not really a use case for a write only interface,
 // but it abstracts and contrasts nicely against the read only context
-// TODO(andrew): convert address and public key to string not bytes; pocket/issues/163
+// TODO (andrew) convert address and public key to string not bytes #147
```
Olshansk, line 41: I don't think 147 is the appropriate issue for it. `convert address and public key to string` should be a sepeate issue

shared/modules/persistence_module.go
```diff
@@ -70,8 +70,7 @@ type PersistenceWriteContext interface {
 	SubtractPoolAmount(name string, amount string) error
 	SetPoolAmount(name string, amount string) error
 
-	// TODO(andrew): remove address from pool; pocket/issues/163
-	InsertPool(name string, address []byte, amount string) error
+	InsertPool(name string, address []byte, amount string) error // TODO (Andrew) remove address from pool #147
```
Olshansk, line 73: Ditto about 147. We should scope down the issues

shared/modules/p2p_module.go
```diff
@@ -2,12 +2,13 @@ package modules
...
-	"github.com/pokt-network/pocket/shared/types"
+	"github.com/pokt-network/pocket/shared/debug"
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
 type P2PModule interface {
 	Module
-	Broadcast(msg *anypb.Any, topic types.PocketTopic) error                       // TODO(derrandz): get rid of topic
-	Send(addr cryptoPocket.Address, msg *anypb.Any, topic types.PocketTopic) error // TODO(derrandz): get rid of topic
+	Broadcast(msg *anypb.Any, topic debug.PocketTopic) error                       // TODO(derrandz): get rid of topic
```
Olshansk, line 11: s/TODO(derrandz)/TECHDEBT

shared/modules/consensus_module.go
```diff
@@ -1,19 +1,18 @@
...
 
 import (
-	"github.com/pokt-network/pocket/shared/types"
-	typesGenesis "github.com/pokt-network/pocket/shared/types/genesis"
+	"github.com/pokt-network/pocket/shared/debug"
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
-type ValidatorMap map[string]*typesGenesis.Actor
+type ValidatorMap map[string]Actor // TODO (Drewsky) deprecate Validator map or populate from persistence module
```
Olshansk, line 8: No action. Agreed.

shared/modules/bus_module.go
Olshansk: Discussed offline. See other comments.

shared/bus.go
```diff
@@ -124,43 +123,43 @@ func CreateBusWithOptionalModules(
 	return bus
 }
 
-func (m *bus) PublishEventToBus(e *types.PocketEvent) {
+func (m bus) PublishEventToBus(e *debug.PocketEvent) {
```
Olshansk, line 126: Why is `bus` no longer a pointer. Here and bellow?

shared/modules/doc/CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
...
+
+All notable changes to this module will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [0.0.0.1] - 2022-08-21
+- Minimized shared module with #163
```
Olshansk, line 11: NIT (in here and/or for the future): Please add a hyperlink to the issue like so:

```
- Minimized shared module with [#163](https://github.com/pokt-network/pocket/issues/163)
```

newcomers may not know what 163 means.

Olshansk (changes requested):
Mostly done!

Going to wait for the next commit for review and also looking into the asks for help made in the commit in parallel.


Screen Shot 2022-09-08 at 4 52 41 PM

Olshansk:
@andrewnguyen22 I address the GOPATH issue you mentioned in [`2f16888` (#178)](https://github.com/pokt-network/pocket/pull/178/commits/2f168887665ac0ee56a6ca1360d0145797a1b98e). Make sure you do a `git pull --rebase` after you finish with your local changes.

utility/types/error.go
```diff
@@ -2,33 +2,78 @@ package types
...
+}
 
-	CodeTransactionSignError Code = 36
+func (se StdErr) Code() Code {
+	return se.CodeError
+}
 
-	CodeInterfaceConversionError Code = 38
-	CodeGetAccountAmountError    Code = 39
+func NewError(code Code, msg string) Error {
```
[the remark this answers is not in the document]
andrewnguyen22: Errors with codes is only for the Utility module. It's not used anywhere else. If other modules decide to use error codes or if that's a global standard that can be scoped in another telemetry like issue

shared/test_artifacts/mock.go
```diff
@@ -0,0 +1,909 @@
+package test_artifacts
```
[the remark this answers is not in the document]
andrewnguyen22: Only if we can produce proper genesis and config json files for the debug client. https://github.com/pokt-network/pocket/issues/202

telemetry/README.md
```diff
@@ -114,11 +114,11 @@ See the images below for an example of how to use the `Explore` functionality.
 
 Go to the explore page:
 
-![](./docs/explore-loki-on-grafana-pt-1.gif)
+![](docs/explore-loki-on-grafana-pt-1.gif)
```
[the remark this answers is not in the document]
andrewnguyen22: Yes

build/config/config2.json
```diff
@@ -1,15 +1,16 @@
 {
```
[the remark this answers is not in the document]
andrewnguyen22: You assume that the same private key will be used for all modules. Perhaps this isn't true (different P2P identity then validator identity for security)

build/config/genesis.json
```diff
@@ -254,5 +251,59 @@
       "message_unpause_service_node_fee_owner": "da034209758b78eaea06dd99c07909ab54c99b45",
       "message_change_parameter_fee_owner": "da034209758b78eaea06dd99c07909ab54c99b45"
     }
+  },
+  "consensusGenesisState": {
+    "genesis_time": {
+      "seconds": 1662143679,
+      "nanos": 362513000
+    },
+    "chain_id": "testnet",
```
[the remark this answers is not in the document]
andrewnguyen22: Yes, most blockchains support chain_id so that anyone may spin a fork of their network. `Pocket` is just the software that enables `Pocket Network`

consensus/types/proto/consensus_config.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
andrewnguyen22: Already asked and answered above

utility/test/account_test.go
```diff
@@ -168,76 +167,82 @@ func TestUtilityContext_SetAccountWithAmountString(t *testing.T) {
...
+func GetAllTestingAccounts(t *testing.T, ctx utility.UtilityContext) (acs []modules.Account) {
+	accs, err := (ctx.Context.PersistenceRWContext).GetAllAccounts(0)
+	require.NoError(t, err)
 	sort.Slice(accs, func(i, j int) bool {
-		return accs[i].Address < accs[j].Address
+		return accs[i].GetAddress() < accs[j].GetAddress()
 	})
-	require.NoError(t, err)
-	return accs
+	for _, acc := range accs {
```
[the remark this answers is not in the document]
andrewnguyen22: See dup below

consensus/types/types.go
```diff
@@ -38,3 +37,37 @@ func GetValAddrToIdMap(validatorMap map[string]*typesGenesis.Actor) (ValAddrToId
 
 	return valToIdMap, idToValMap
 }
+
+func (x *ConsensusConfig) GetPaceMakerConfig() modules.PacemakerConfig {
```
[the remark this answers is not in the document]
andrewnguyen22: Protobuf can't return an interface in a language agnostic way. I'm adding a method to the structure to satisfy the interface

consensus/types/types.go
```diff
@@ -38,3 +37,37 @@ func GetValAddrToIdMap(validatorMap map[string]*typesGenesis.Actor) (ValAddrToId
...
+	return x.GetPacemakerConfig()
+}
+
+func (x *PacemakerConfig) SetTimeoutMsec(u uint64) {
+	x.TimeoutMsec = u
+}
+
+type ValidatorMap map[string]modules.Actor
+
+func ValidatorMapToModulesValidatorMap(validatorMap ValidatorMap) (vm modules.ValidatorMap) {
```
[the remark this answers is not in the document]
andrewnguyen22: The `ValidatorMap` design is (AFAIK) a temporary hack and should be encapsulated to the consensus module

I do not see the production app sharing a map of Validators from the consensus module. If it does then we should make it an interface and not a structure.

consensus/types/types.go
```diff
@@ -38,3 +37,37 @@ func GetValAddrToIdMap(validatorMap map[string]*typesGenesis.Actor) (ValAddrToId
...
+	m = make(ValidatorMap, len(validators))
+	for _, v := range validators {
+		m[v.GetAddress()] = v
+	}
+	return
+}
+
+var _ modules.Actor = &Validator{}
+
+func (x *Validator) GetPausedHeight() int64         { panic("not implemented on consensus validator") }
```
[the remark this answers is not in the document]
andrewnguyen22: They are not part of the protobuf - ask the question with more detail and I can answer

p2p/raintree/network.go
```diff
@@ -33,7 +32,7 @@ type rainTreeNetwork struct {
 	maxNumLevels uint32
 
 	// TECHDEBT(drewsky): What should we use for de-duping messages within P2P?
-	mempool types.Mempool
+	mempool map[uint64]struct{} // TODO (drewsky) replace map implementation (can grow unbounded)
```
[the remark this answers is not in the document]
andrewnguyen22: Mempool is such a bad implementation for the purpose of dedup, it is a specialized structure for transaction reaping. If a semantic type is desired a separate issue should be opened for it.

p2p/transport.go
```diff
@@ -2,36 +2,35 @@ package p2p
...
 	TCPNetworkLayerProtocol = "tcp4"
 )
 
-func CreateListener(cfg *genesis.P2PConfig) (typesP2P.Transport, error) {
-	switch cfg.ConnectionType {
-	case genesis.ConnectionType_TCPConnection:
-		return createTCPListener(cfg)
-	case genesis.ConnectionType_EmptyConnection:
+func CreateListener(cfg modules.P2PConfig) (typesP2P.Transport, error) {
+	switch cfg.IsEmptyConnType() { // TODO (team) kept in switch format because this should be an enum not a bool
```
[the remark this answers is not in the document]
andrewnguyen22: The original designer of the module left it as an `enum` I believe `quik` might be another option

p2p/types/p2p_config.go
andrewnguyen22: Protobuf doesn't automatically generate this custom logic. This is used to support the migration from shared. There's a TODO to convert this back to an enum

p2p/types/proto/p2p_config.proto
andrewnguyen22: `bool use_rain_tree = 3;  // whether or not we use rain tree`
`bool is_empty_connection_type = 4; // if the connection type is using tcp or empty`

andrewnguyen22: NOTE: I didn't introduce new configurations, so keep in mind that any contention with the configuration might be out of scope

persistence/genesis.go
```diff
@@ -308,17 +309,18 @@ func (p PostgresContext) GetAllFishermen(height int64) (f []*genesis.Actor, err
...
 		}
-		f = append(f, p.BaseActorToActor(actor, genesis.ActorType_Fish))
+		f = append(f, p.BaseActorToActor(actor, types.ActorType_Fish))
 	}
 	return
 }
 
-func (p PostgresContext) BaseActorToActor(ba schema.BaseActor, actorType genesis.ActorType) *genesis.Actor { // TODO (Team) deprecate with interface #163
-	actor := new(genesis.Actor)
+// TODO (Team) deprecate with interface #163 <Bumped to #149> as #163 is getting large
```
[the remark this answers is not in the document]
andrewnguyen22: Convert BaseActor to shared.Actor <interface> 

It's much more semantic to use the `actor` interface in `shared` module + it allows for multiple struct implementations without changing the function

persistence/shared_sql.go
```diff
@@ -5,13 +5,14 @@ import (
...
-	"github.com/pokt-network/pocket/persistence/schema"
-	"github.com/pokt-network/pocket/shared/types"
+	"github.com/pokt-network/pocket/persistence/types"
+	"github.com/pokt-network/pocket/shared/modules"
+	types2 "github.com/pokt-network/pocket/utility/types"
 )
 
 // IMPROVE(team): Move this into a proto enum. We are not using `iota` for the time being
 // for the purpose of being explicit: https://github.com/pokt-network/pocket/pull/140#discussion_r939731342
+// TODO Cleanup with #149
```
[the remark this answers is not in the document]
andrewnguyen22: The comment above is the cleanup

persistence/test/setup_test.go
```diff
@@ -219,8 +213,8 @@ func fuzzSingleProtocolActor(
 			if originalActor.UnstakingHeight != db.Height { // Not ready to unstake
 				require.Nil(t, unstakingActors)
 			} else {
-				idx := slices.IndexFunc(unstakingActors, func(a *types.UnstakingActor) bool {
-					return originalActor.Address == hex.EncodeToString(a.Address)
+				idx := slices.IndexFunc(unstakingActors, func(a modules.UnstakingActorI) bool {
```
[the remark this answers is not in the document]
andrewnguyen22: One is a `proto` structure used only in Persistence

The other is a shared interface that other modules may accept without having a shared `structure`

persistence/types/base_actor.go
```diff
@@ -141,3 +143,12 @@ func (actor *BaseProtocolActorSchema) ClearAllQuery() string {
 func (actor *BaseProtocolActorSchema) ClearAllChainsQuery() string {
 	return ClearAll(actor.chainsTableName)
 }
+
+var _ modules.Actor = &Actor{}
+
+func (x *Actor) GetActorTyp() modules.ActorType {
```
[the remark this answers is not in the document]
andrewnguyen22: Not typo, it's a method to allow the proto structure to interact with our shared interface. Duplicate of above

shared/modules/types.go
andrewnguyen22: If you want to swap `Olshansk's HotPocket Consensus Module` for `Andrew's TendermintBFT Consensus Module` my configuration can satisfy the interface without needing to rely on the structure in shared.

shared/modules/types.go
andrewnguyen22: Because there is a structure inside of `persistence mod` that is called `UnstakingActor` I thought that might be confusing

shared/modules/types.go
andrewnguyen22: What? These are `param name` shared constants that should not live in `shared`

shared/modules/persistence_module.go
andrewnguyen22: A module by module cleanup issue may be scoped down, let that be tackled with #147 kindof counter intuitive to upscope this issue just to down scope others

shared/bus.go
andrewnguyen22: See Allesandro's `leak comment` above

consensus/module.go
Olshansk: Discussed offline and outdated, resolving.

consensus/module.go
```diff
@@ -1,37 +1,44 @@
...
 
-var _ modules.ConsensusModule = &consensusModule{}
+var _ modules.ConsensusGenesisState = &typesCons.ConsensusGenesisState{}
+var _ modules.PacemakerConfig = &typesCons.PacemakerConfig{}
+var _ modules.ConsensusConfig = &typesCons.ConsensusConfig{}
+var _ modules.ConsensusModule = &ConsensusModule{}
 
 // TODO(olshansky): Any reason to make all of these attributes local only (i.e. not exposed outside the struct)?
-type consensusModule struct {
+type ConsensusModule struct {
```
Olshansk, line 32: Why did you capitalize this?

consensus/module.go
```diff
@@ -1,37 +1,44 @@
...
 	"google.golang.org/protobuf/types/known/anypb"
+
+	consensusTelemetry "github.com/pokt-network/pocket/consensus/telemetry"
+	"github.com/pokt-network/pocket/shared/modules"
 )
 
 const (
-	DefaultLogPrefix string = "NODE" // Just a default that'll be replaced during consensus operations.
+	DefaultLogPrefix    string = "NODE" // Just a default that'll be replaced during consensus operations.
+	ConsensusModuleName        = "consensus"
```
Olshansk, line 21: The goal of creating constants is to be able to easily use the same variable in different / multiple locations, but this is only being used in one place as far as I can tell. Why not simply do this:

```
func (m *ConsensusModule) GetModuleName() string {
	return "consensus"
}
```

consensus/module.go
```diff
@@ -1,37 +1,44 @@
...
+	PacemakerModuleName        = "pacemaker"
```
Olshansk, line 22: I think we should put this in `pacemaker.go`. Doesn't make sense for it to be here

consensus/consensus_tests/utils_test.go
```diff
@@ -54,33 +53,29 @@ type IdToNodeMapping map[typesCons.NodeId]*shared.Node
...
-var (
-	DefaultServiceURL    = "https://foo.bar" // TODO (team) cleanup consensus testing module with centralization of utilities
-	DefaultChainID       = "unittest_net"
-	DefaultStakeAmount   = types.BigIntToString(big.NewInt(1000000000))
-	DefaultAccountAmount = types.BigIntToString(big.NewInt(1000000000))
-)
-
-func GenerateNodeConfigs(_ *testing.T, numValidators int) (configs []*genesis.Config, genesisState *genesis.GenesisState) {
-	genesisState, keys := test_artifacts.NewGenesisState(numValidators, 1, 1, 1)
+func GenerateNodeConfigs(_ *testing.T, numOfVals int) (configs []modules.Config, genesisState modules.GenesisState) {
```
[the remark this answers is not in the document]
Olshansk: I want to make sure we:
1. Stay consistent
2. Agree on something now and use it moving forward
3. Go with industry standards & make it clear to future redeers

I went through a few threads and the general consensus seems to be `numValidators` or `validatorCount`.

Screen Shot 2022-09-11 at 2 48 40 PM


Source: https://stackoverflow.com/questions/3742650/naming-conventions-for-number-of-foos-variables

shared/node.go
Olshansk: `AFAIK the address doesn't even need to be in the 'node'`

My thinking is currently this:
1. Getting the address is something every module is likely to have to do multiple times
2. We want to avoid doing the hash + conversion more than once
3. What should we do given the architecture of different modules having different keys, and in turn, addresses?

AI: Should we just make this easily accessible via the node?

utility/types/error.go
```diff
@@ -2,33 +2,78 @@ package types
...
 
-const (
-	CodeOK                         Code = 0
-	CodeInvalidSignerError         Code = 3
-	CodeDecodeMessageError         Code = 4
-	CodeUnmarshalTransaction       Code = 5
-	CodeUnknownMessageError        Code = 6
-	CodeAppHashError               Code = 7
-	CodeNewPublicKeyFromBytesError Code = 8
+// TODO (Team) move errors to respective modules #163
```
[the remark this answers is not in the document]
Olshansk: Seems outdated.

utility/types/actor.go
```diff
@@ -13,30 +12,30 @@ const (
...
-	switch actorType {
-	case ActorType_App:
-		return genesis.Pool_Names_AppStakePool.String()
-	case ActorType_Val:
-		return genesis.Pool_Names_ValidatorStakePool.String()
-	case ActorType_Fish:
-		return genesis.Pool_Names_FishermanStakePool.String()
-	case ActorType_Node:
-		return genesis.Pool_Names_ServiceNodeStakePool.String()
+func (x UtilActorType) GetActorPoolName() string {
```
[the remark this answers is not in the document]
Olshansk: @andrewnguyen22 resurfacing this comment.

shared/bus.go
Olshansk: Noting that you probably meant @adshmh rather than @deblasis for diligence, but understood.

shared/modules/persistence_module.go
Olshansk: `let that be tackled with https://github.com/pokt-network/pocket/issues/147 kindof counter intuitive to upscope this issue just to down scope others`

Do you want me to resolve this as part of https://github.com/pokt-network/pocket/issues/147?

shared/modules/persistence_module.go
Olshansk: Do you want me to resolve this as part of https://github.com/pokt-network/pocket/issues/147?

shared/modules/types.go
Olshansk: I'm confused as well. Do you want me to do this as part of https://github.com/pokt-network/pocket/issues/147?

consensus/module.go
```diff
@@ -116,7 +138,39 @@ func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (modules.Consens
...
+	data, err := ioutil.ReadFile(pathToConfigJSON)
+	if err != nil {
+		return
+	}
+	// over arching configuration file
+	rawJSON := make(map[string]json.RawMessage)
+	if err = json.Unmarshal(data, &rawJSON); err != nil {
+		log.Fatalf("[ERROR] an error occurred unmarshalling the config.json file: %v", err.Error())
+	}
+	// persistence specific configuration file
```
Olshansk, line 151: persistence - is this a copy-paste error?

If so, it's what I'm generally afraid of if we keep copy-pasting everywhere.

consensus/module.go
```diff
@@ -116,7 +138,39 @@ func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (modules.Consens
 	return m, nil
 }
 
-func (m *consensusModule) Start() error {
+func (m *ConsensusModule) InitConfig(pathToConfigJSON string) (config modules.ConfigI, err error) {
+	data, err := ioutil.ReadFile(pathToConfigJSON)
```
Olshansk, line 142: Lines 142 - 150 (8 lines) will need to be copy-pasted twice for every single module that we create. This triggers my "allergies" that we're doing something wrong.

I think we should create a ticket to improve this but wanted to see what you think.

consensus/module.go
```diff
@@ -58,7 +65,18 @@ type consensusModule struct {
 	MaxBlockBytes uint64
 }
 
-func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (modules.ConsensusModule, error) {
+func Create(configPath, genesisPath string, useRandomPK bool) (modules.ConsensusModule, error) {
```
Olshansk, line 68: I would like for us to discuss the addition of `useRandomPK` before moving forward? Seems like a big step back in our modular architecture.

In the previous approach where we could "inject" a random private key without hacks like this shows that it was a good design from a "separation of concerns" an good for testing/mocking/development. I understand it doesn't satisfy the "encapsulation requirement", but just wanted to point that the whenever we need to start adding params like "useRandomPk", it's often a sign that there might be a better way of doing something.

Ccing @adshmh and @deblasis for context so we can discuss it as a group and ideate.

consensus/module.go
```diff
@@ -116,7 +138,39 @@ func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (modules.Consens
...
-func (m *consensusModule) Start() error {
+func (m *ConsensusModule) InitConfig(pathToConfigJSON string) (config modules.ConfigI, err error) {
+	data, err := ioutil.ReadFile(pathToConfigJSON)
+	if err != nil {
+		return
+	}
+	// over arching configuration file
+	rawJSON := make(map[string]json.RawMessage)
+	if err = json.Unmarshal(data, &rawJSON); err != nil {
+		log.Fatalf("[ERROR] an error occurred unmarshalling the config.json file: %v", err.Error())
```
Olshansk, line 149: Replace with: `log.Fatalf("[ERROR] an error occurred unmarshalling the %s file: %v", pathToConfigJSON , err.Error())`

Ditto for every other error in config and JSON unmarshalling, but only leaving this one comment

consensus/module.go
```diff
@@ -116,7 +138,39 @@ func Create(cfg *genesis.Config, genesis *genesis.GenesisState) (modules.Consens
...
+	data, err := ioutil.ReadFile(pathToGenesisJSON)
+	if err != nil {
+		return
+	}
+	// over arching configuration file
+	rawJSON := make(map[string]json.RawMessage)
+	if err = json.Unmarshal(data, &rawJSON); err != nil {
+		log.Fatalf("[ERROR] an error occurred unmarshalling the gensis.json file: %v", err.Error())
+	}
+	// persistence specific configuration file
```
Olshansk, line 167: s/persistence/consensus?

consensus/module.go
```diff
@@ -58,7 +65,18 @@ type consensusModule struct {
...
+	cm := new(ConsensusModule)
+	c, err := cm.InitConfig(configPath)
+	if err != nil {
+		return nil, err
+	}
+	g, err := cm.InitGenesis(genesisPath)
+	if err != nil {
+		return nil, err
+	}
+	cfg := c.(*typesCons.ConsensusConfig)
```
Olshansk, line 78: You can do the type assertion/casting directing in `InitConfig` and `InitiGenesis`. See this code snippet:

```go
package main

import (
	"fmt"
)

type ConfigI interface {
  GetConfigName() string  
}

var _ ConfigI = &configA{}
var _ ConfigI = &configB{}

type configA struct {}
type configB struct {}

func (c *configA) GetConfigName() string { return "A" }
func (c *configB) GetConfigName() string { return "B" }

func GetConfigA() ConfigI {
  return &configA{}
}

func GetConfigB() ConfigI {
  return &configB{}
}

func main() {
	fmt.Println("Hello, World!")
        fmt.Println(GetConfigA().GetConfigName())
       fmt.Println(GetConfigB().GetConfigName())
}
```

consensus/types/types.go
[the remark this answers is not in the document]
Olshansk: https://github.com/pokt-network/pocket/issues/203

p2p/raintree/network.go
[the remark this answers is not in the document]
Olshansk: Makes sense to me

p2p/transport.go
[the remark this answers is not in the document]
Olshansk: Makes sense. Resolving.

persistence/shared_sql.go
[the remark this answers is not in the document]
Olshansk: Noted. I'll handle it in the coming weeks.

shared/modules/types.go
Olshansk: Got it. Given how you phrased that, it makes me fear that there might be something wrong with the consensus module.

Do you see us potentially needing to revert to a TendermintBFT consensus module at some point?

shared/modules/types.go
Olshansk: Mainly I was just going for conssitency, but if I had to think of examples:

`p2p genesis state` - initial seeders for the addrbook
`utility genesis state` - initial servicers and the DAO permissioned fisherman

Olshansk (changes requested):
I'm not fully done the re-review, but am 118/151 files in:
Screen Shot 2022-09-11 at 4 15 34 PM

I left some comments in line but also added these discussion items to make sure they are more easily surfaced:

- [ ] Action Item: When I run `make develop_test`, I get `undefined: TelemetryConfig`. Can you update the appropriate targets to streamline this?
- [ ] DISCUSS_CODE_ARCHITECTURE:  Do we really want per-module private keys?
    - https://github.com/pokt-network/pocket/pull/178/files#r965100054
- [ ] DISCUSS_CODE_ARCHITECTURE:  Where should we store/access the the address?
    - See this discussion: https://github.com/pokt-network/pocket/pull/178#discussion_r956397637
- [ ] DISCUSS_CODE_ARCHITECTURE: The need to introduce `useRandomPK` in the modules’ `Create` function?
    - The previous “functional” design enable “injecting” a random private key which was a good source of logic encapsulation
    - See the comment in `module.go`
- [ ] DISCUSS_CODE_SYNTAX: When do we choose to capitalize vars/constants vs keeping it lowerCamelCase?
    - E.g. look at constants not accessed outside package or how `consensusModule` was renamed to `ConsensusModule`.
- [ ] DISCUSS_CODE_SYNTAX: What naming convention should we use for “number of something”?
    - E.g.`numOfVals` vs `numValidators` in this discussion: https://github.com/pokt-network/pocket/pull/178/files#r965144108.
    - Important to pick something now that we’ll use moving forward.
    - GetModuleName
- [ ] DISCUSS_CODE_SYNTAX: Avoiding `func (x` where `x` is just a catchall.
- [ ] DOCUMENT: why do we need `chain_id`?
- [ ] TICKET_TO_CREATE: Create a ticket for refactor `generic_param` to something more appropriate
- [ ] TICKET_TO_CREATE: Align on how we want to create, track and organize errors. Research -> align -> design -> refactor.
- [x] TICKET_TO_CREATE: Encapsulate ValidatorMap to the consensus module only
- [ ] Look through all open remaining comment.

utility/types/actor.go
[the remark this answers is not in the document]
andrewnguyen22: I'm not sure that I agree with this for the following reasons:

- Protogen creates `func(x`
- It's more semantic for all receivers to have the same name (Jetbrains Goland complains when they have a different name)

build/config/config2.json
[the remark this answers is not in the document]
andrewnguyen22: Yeah - this is a standard in the Tendermint ecosystem. 

I'm fine with using the same key for all, but I think allowing this configurability is also nice

shared/modules/types.go
andrewnguyen22: It's an example of modularity - not a comment about ownership or algorithm

shared/modules/types.go
andrewnguyen22: Peers / seeds are not part of the `state` so this seems counter intuitive to me.

> initial servicers and the DAO permissioned fisherman

^ Is already in the persistence genesis state

shared/modules/types.go
andrewnguyen22: I'm fine for you to do it as long as you understand that it's just about removing these shared variables and utilizing the `protobuf.String()` function so that the naming of the parameters may be defined by the language agnostic generator

shared/modules/persistence_module.go
andrewnguyen22: I think it's supposed to say #149

shared/modules/persistence_module.go
andrewnguyen22: Sorry - I understand the confusion now. I meant #149

build/config/genesis.json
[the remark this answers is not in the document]
andrewnguyen22: https://github.com/pokt-network/pocket/issues/206

consensus/module.go
andrewnguyen22: We can move the ioutil to a shared utility file if you'd prefer - but I don't think this is a big deal

consensus/module.go
andrewnguyen22: Injecting a private key is a `hack` the real solution is to use a different configuration file for the `debug_client`

consensus/module.go
andrewnguyen22: Yes you can, but the argument for the creation of each module is not the `interface` it's the `structure` yet the return argument from the `InitConfig` and `InitGenesis` is the interface. This is due to the ask to add the `initConfig` and `initGenesis` to the shared interface module

andrewnguyen22:
- [x] Action Item: When I run `make develop_test`, I get `undefined: TelemetryConfig`. Can you update the appropriate targets to streamline this?
- [x] DISCUSS_CODE_ARCHITECTURE:  Do we really want per-module private keys?
    - https://github.com/pokt-network/pocket/pull/178/files#r965100054 
    - **Responded**
- [x] DISCUSS_CODE_ARCHITECTURE:  Where should we store/access the the address?
    - See this discussion: https://github.com/pokt-network/pocket/pull/178#discussion_r956397637
    - **Responded**
- [x] DISCUSS_CODE_ARCHITECTURE: The need to introduce `useRandomPK` in the modules’ `Create` function?
    - The previous “functional” design enable “injecting” a random private key which was a good source of logic encapsulation
    - See the comment in `module.go`
    - **Responded**
- [x] DISCUSS_CODE_SYNTAX: When do we choose to capitalize vars/constants vs keeping it lowerCamelCase?
    - E.g. look at constants not accessed outside package or how `consensusModule` was renamed to `ConsensusModule`.
    - **Responded**
- [x] DISCUSS_CODE_SYNTAX: What naming convention should we use for “number of something”?
    - E.g.`numOfVals` vs `numValidators` in this discussion: https://github.com/pokt-network/pocket/pull/178/files#r965144108.
    - Important to pick something now that we’ll use moving forward.
    - GetModuleName
    - **Fixed**
- [x] DISCUSS_CODE_SYNTAX: Avoiding `func (x` where `x` is just a catchall.
- **Responded**
- [x] DOCUMENT: why do we need `chain_id`?
- **Responded**
- [x] TICKET_TO_CREATE: Create a ticket for refactor `generic_param` to something more appropriate
- **Done**
- [ ] TICKET_TO_CREATE: Align on how we want to create, track and organize errors. Research -> align -> design -> refactor.
- *Unclear*
- [x] TICKET_TO_CREATE: Encapsulate ValidatorMap to the consensus module only

consensus/module.go
Olshansk: Makes sense.

consensus/types/types.go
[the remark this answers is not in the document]
Olshansk: Discussed offline.

consensus/types/types.go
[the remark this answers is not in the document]
Olshansk: Discussed offline.

persistence/genesis.go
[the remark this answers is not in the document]
Olshansk: Makes sense

shared/node.go
Olshansk: Leaving this discussion for the future.

utility/types/actor.go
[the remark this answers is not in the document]
Olshansk: https://stackoverflow.com/questions/73696125/golang-naming-conventions-for-proto-generated-types

build/config/config2.json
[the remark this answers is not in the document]
Olshansk: Makes sense

utility/test/account_test.go
```diff
@@ -168,76 +167,82 @@ func TestUtilityContext_SetAccountWithAmountString(t *testing.T) {
...
+func GetAllTestingPools(t *testing.T, ctx utility.UtilityContext) (acs []modules.Account) {
+	accs, err := (ctx.Context.PersistenceRWContext).GetAllPools(0)
+	require.NoError(t, err)
 	sort.Slice(accs, func(i, j int) bool {
-		return accs[i].Address < accs[j].Address
+		return accs[i].GetAddress() < accs[j].GetAddress()
 	})
-	require.NoError(t, err)
-	return accs
+	for _, acc := range accs {
```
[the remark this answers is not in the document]
Olshansk: I made this change and all the tests still passed. Not pushing yet because I wanted to double check I'm not missing something obvious:

Screen Shot 2022-09-12 at 7 20 33 PM

Olshansk:
@andrewnguyen22 I pushed the changes I made during our pair session today. 

Below is a list of some items I made while we were on the call (I will create tickets for it later), and I left a few conversations unresolved for potential future references/discussions.

Unit tests all pass, and my LocalNet setup ran but I have tested it yet. My asks are:
1. Can you verify that it works (unit tests and LocalNet flow) on your machine
2. Review the changes I pushed
3. See the comment I left in `utility/test/account_test.go`

--- 
- [ ] Search and replace/remove all `func (x…)`
    - Context: added wrappers to return interfaces where structs are the default
    - https://stackoverflow.com/questions/73696125/golang-naming-conventions-for-proto-generated-types
- [ ] Search for `testingConfigFilePath+strconv.Itoa(n)+jsonPosfix` and cleanup/standardize
- [ ] Search for `os.Remove` and cleanup/standardize
- [ ] Global replace for `log.Fatal(fmt.Sprintf(`
- [ ] TICKET_TO_CREATE: Global nits by Olshansky
- [ ] DISCUSS: s/testing/test for test variable naming

andrewnguyen22:
- [x] Can you verify that it works (unit tests and LocalNet flow) on your machine
- [x] Review the changes I pushed
- [x] See the comment I left in utility/test/account_test.go

For the last item all I see is:
```
// TODO(andrew): Remove all `require.True` in test files unless we're checking the value of a boolean
// TODO(andrew): Remove all `fmt.Sprintf(`
```

I can do that during TECHDEBT cleanup. Please advise

Olshansk:
Sorry, I meant PR comment... This isn't a blocker since we can tend to it later, but just felt like I was missing something obvious and was unsure.

Screen Shot 2022-09-13 at 6 32 20 PM

Olshansk (approved):
Just pushed the final change, ran tests locally and verified LocalNet.

Let's squash and merge this 💩!

andrewnguyen22:
Roger!

PR #198, opened 2022-09-08
[Tooling] Flaky tests - Issue #192

### LINKED ISSUE #192: [Tooling] Flaky tests
## Objective

Tests should be deterministic

## Origin Document

It looks like that some of our tests are flaky, just logging this for now, I need to dive deeper to understand the root cause(s), it might be something very easy.

A couple of examples:

https://github.com/pokt-network/pocket/runs/8227176834?check_suite_focus=true
https://github.com/pokt-network/pocket/runs/8187453124?check_suite_focus=true

But it can be verified locally with multiple executions of 
```bash
make test_all
```
### First run 👍
image
### Second run 💩
image

## Goals

- [ ] Identify the flaky tests
- [ ] Fix them
- [ ] Document (if you think it's necessary) the root cause for future reference

## Deliverable

- [ ] Deterministic test suite

- ...

## [Optional] Testing Methodology

---

**Creator**: @deblasis 
**Co-Owners**: @Olshansk

deblasis:
## Description

This PR addresses test flakiness due to:
- race conditions
- use of timers that causes non-determinism

#### **TestHotstuff4Nodes1BlockHappyPath**

```bash
for i in {1..10}; do go test -timeout 30s -count=1 -run ^TestHotstuff4Nodes1BlockHappyPath$ github.com/pokt-network/pocket/consensus/consensus_tests; done | egrep  '^\--- FAIL:|ok '
```
image

#### **TestTinyPacemakerTimeouts**

```bash
for i in {1..10}; do go test -timeout 30s -count=1 -run ^TestTinyPacemakerTimeouts$ github.com/pokt-network/pocket/consensus/consensus_tests; done | egrep  '^\--- FAIL:|ok '
```

image

### Time mocking
Following the input in this [comment](https://github.com/pokt-network/pocket/issues/192#issuecomment-1240033668) I proceeded implementing time mocking.

The idea is that the nodes will have a reference that can proxy calls to either the real time package or a mocked version.
I am using [github.com/benbjohnson/clock](https://github.com/benbjohnson/clock) and I am starting to build utility functions around it that could become a wrapper at some point but it's early to say.

Something like this:
image

I believe this functionality is going to be quite useful in general especially since I believe we'll have to deal with timeouts and time-bound stuff quite a bit.

### Race conditions

Mostly due to the fact that structs are accessed by multiple goroutines at the same time.

Addressed using `mutexes` and wrapping the getters/setters into methods that also handle lock/unlock.

## Issue

Fixes [issue/192](https://github.com/pokt-network/pocket/issues/192)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Code health or cleanup
- [x] Other: Investigation

## List of changes

- Introduced time mocking
- Fixed race conditions in `Consensus` module

## Testing

- [x] `make test_all`

## Checklist

- [x] If applicable, I have made corresponding changes to related local or global README


deblasis:
getting somehere 
image

deblasis:
10 runs with and without the `-race` flag and no flakiness detected 🙌

image

deblasis:
Hey @okdas! Can you please review this specific commit?

https://github.com/pokt-network/pocket/pull/198/commits/6e0b00fda4e31ca759aad540bb1af5f5d7919358

I made sure we skip the dockerhub steps that currently are simply failing because not fully setup.

This way we don't get false negatives in CI runs.

consensus/pacemaker.go
```diff
@@ -36,6 +37,8 @@ var _ PacemakerDebug = &paceMaker{}
 type paceMaker struct {
 	bus modules.Bus
 
+	m sync.RWMutex
```
Olshansk, line 40: Can you add a comment about why this is needed and when it should/shouldn't be used.

p2p/module_raintree_test.go
```diff
@@ -281,7 +281,8 @@ func generateKeys(_ *testing.T, numValidators int) []cryptoPocket.PrivateKey {
 
 // A mock of the application specific to know if a message was sent to be handled by the application
 // INVESTIGATE(olshansky): Double check that how the expected calls are counted is accurate per the
-//                         expectation with RainTree by comparing with Telemetry after updating specs.
+//
+//	expectation with RainTree by comparing with Telemetry after updating specs.
```
Olshansk, line 285: Should probably `git checkout main` edits to this file since its out of scope

shared/bus.go
```diff
@@ -21,6 +23,8 @@ type bus struct {
 	utility     modules.UtilityModule
 	consensus   modules.ConsensusModule
 	telemetry   modules.TelemetryModule
+
+	clock clock.Clock
```
Olshansk, line 27: I really like the ability and design of injecting a clock into the node and bus. 

Any chance you could add some detail (short & concise + diagram additions/updates) to `shared/README.md`?

consensus/module.go
```diff
@@ -34,6 +35,8 @@ type ConsensusModule struct {
 	privateKey cryptoPocket.Ed25519PrivateKey
 	consCfg    modules.ConsensusConfig
 
+	m sync.RWMutex
```
Olshansk, line 38: Can you add a comment about why its needed and where/how it should be used?

Olshansk (changes requested):
**General**: I did a first pass at the review and think this is super interesting. It has the foundation for important infrastructure we'll really need in the future when we do more extensive testing. 🧑‍🔬 🤔 🧪 

Will do a second pass after we tend to the first set of comments.

**Random Ideas**: Thoughts on creating `PocketClock` module. Not creating an issue for it (yet) since it's just an idea, but wanted to see if you think it'd be overkill.

**Primary concern**: I did some googling and it seems that Setters and Getters in go are idiomatic, but I'm concerned that:
1. We need a way to enforce/encourage this because of the mutex requirements (you can imagine the issues as we have more contributors)
2. Grep & update all existing sets/gets

okdas:
> I made sure we skip the dockerhub steps that currently are simply failing because not fully setup.

This should really not happen because this part is set up for sure and usually works, but then sometimes docker hub does not authenticate and throws an error. I'll check why that might happen. Thank you for calling this out.

consensus/consensus_tests/hotstuff_test.go
```diff
@@ -15,24 +17,33 @@ func TestHotstuff4Nodes1BlockHappyPath(t *testing.T) {
 	numNodes := 4
 	configs, genesisStates := GenerateNodeConfigs(t, numNodes)
 
+	clockMock := clock.NewMock()
+	go timeReminder(clockMock, 100*time.Millisecond)
+
 	// Create & start test pocket nodes
 	testChannel := make(modules.EventsChannel, 100)
-	pocketNodes := CreateTestConsensusPocketNodes(t, configs, genesisStates, testChannel)
+	pocketNodes := CreateTestConsensusPocketNodes(t, clockMock, configs, genesisStates, testChannel)
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/198/commits/ef6d01e5e2b93293b33763c1070961d0a6399b43

consensus/helpers.go
```diff
@@ -199,23 +221,102 @@ func (m *ConsensusModule) electNextLeader(message *typesCons.HotstuffMessage) {
...
+	m.LeaderId = &leaderId
+}
+
+func (m *ConsensusModule) currentNodeIsLeader() bool {
+	m.m.RLock()
+	defer m.m.RUnlock()
+	return m.LeaderId != nil && *m.LeaderId == m.NodeId
+}
+
+/*** General Purpose Getters / Setters ***/
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/198/commits/cb755f98a228b038f7b217416905764696d30e33

consensus/hotstuff_leader.go
```diff
@@ -98,9 +98,9 @@ func (handler *HotstuffLeaderMessageHandler) HandlePrepareMessage(m *ConsensusMo
 		return // TODO(olshansky): Should we interrupt the round here?
 	}
 
-	m.Step = PreCommit
-	m.HighPrepareQC = prepareQC
-	m.MessagePool[Prepare] = nil
+	m.setStep(PreCommit)
```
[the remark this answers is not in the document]
deblasis: I would say that what we should really wrap in Read / Write methods is the atomic Read / Write operations, maybe in more domain specific methods. Obviously getters/setters are a special case of these, with a single Read / Write. 

For example:
```go
func (m *ConsensusModule) resetForNewHeight() {
	m.m.Lock()
	defer m.m.Unlock()
	m.Round = 0
	m.Block = nil
	m.HighPrepareQC = nil
	m.LockedQC = nil
}
``` 
is better than 4 separate calls to setters for each of the involved properties.

The reason is the fact that the scope of the lock has to be the smallest possible and at the same time we want to minimize the amount of locking/unlocking.

This first pass is to show the problem (race conditions) and what I think it's the solution (mutex) but I agree, a little refactoring is needed to also include parts that are not causing issues with the tests at the moment but it's likely that they will as code coverage grows unless we pre-emptively manage concurrency applying this getter/setter pattern across the board.

I was trying to be frugal and I didn't want to increase the scope of this specific PR and instead push it ASAP in order to have feedback from CI about the changes we are making (TECHDEBT week touches a lot of stuff, we might just break things without knowing we did).

If you feel strongly about this, I can review this whole file / module.

p2p/module_raintree_test.go
deblasis: Agreed, but with the occasion, a few words 🙂

I need to edit this file every time I change branch. `gofmt` automatically changes that comment spacing and I included it in the changeset by mistake.

With this I want to say that there's probably something we can improve here in terms of developer experience.

- switching branch and running `make develop_test` is not smooth as you would expect. I need to exclude / comment a line in this file to make it work (`mocksP2P "github.com/pokt-network/pocket/p2p/types/mocks"`)
- making any change to this file and saving it changes the formatting of that comment in many default VSCode setups, mine included. It sounds a bit weird to have to tell every new community dev that they have to ignore the automatic formatting / disable `format on save` and use our special spacing.
   - related: if we want to use automatic linters / formatters this problem will surface there as well because they would do what I did basically (changing the spacing in that comment) so if there's a strong need for a specific formatting, we should enforce it programmatically somehow. 🤔

consensus/consensus_tests/pacemaker_test.go
```diff
@@ -14,108 +17,139 @@ import (
...
+				Step:   uint8(consensus.NewRound),
+				Round:  1,
+			},
+			GetConsensusNodeState(pocketNode))
 	}
 
-	// Cause the pacemaker to timeout
-	time.Sleep(paceMakerTimeout)
-
+	go func() {
```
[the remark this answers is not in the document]
deblasis: Makes sense, easier to read and the implementation details are one click away.

Addressed in https://github.com/pokt-network/pocket/pull/198/commits/4fd70c01eff3f4f6689c486d033d01e2f4ed70c8

consensus/helpers.go
```diff
@@ -199,23 +221,102 @@ func (m *ConsensusModule) electNextLeader(message *typesCons.HotstuffMessage) {
...
+
+}
+
+func (m *ConsensusModule) setLeaderId(leaderId typesCons.NodeId) {
+	m.m.Lock()
+	defer m.m.Unlock()
+	m.LeaderId = &leaderId
+}
+
+func (m *ConsensusModule) currentNodeIsLeader() bool {
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/198/commits/8eb426d612259a9ef2e38b0bab4985f54238bb3a

shared/bus.go
deblasis: Maybe addressed in https://github.com/pokt-network/pocket/pull/198/commits/e1d579283f0476e87a17708dfc9963f3e48bf345

consensus/consensus_tests/hotstuff_test.go
```diff
@@ -15,24 +17,33 @@ func TestHotstuff4Nodes1BlockHappyPath(t *testing.T) {
...
-	newRoundMessages, err := WaitForNetworkConsensusMessages(t, testChannel, consensus.NewRound, consensus.Propose, numNodes, 1000)
+	newRoundMessages, err := WaitForNetworkConsensusMessages(t, clockMock, testChannel, consensus.NewRound, consensus.Propose, numNodes, 1000)
 	require.NoError(t, err)
-	for _, pocketNode := range pocketNodes {
+	for nodeId, pocketNode := range pocketNodes {
 		nodeState := GetConsensusNodeState(pocketNode)
-		require.Equal(t, uint64(1), nodeState.Height)
-		require.Equal(t, uint8(consensus.NewRound), nodeState.Step)
-		require.Equal(t, uint8(0), nodeState.Round)
+		assertConsensusState(t, nodeId,
```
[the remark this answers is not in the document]
deblasis: I thought about it but there are properties that we don't want to check for equality in that assertion.
I didn't want to alter the previous test logic.

Here we want to make sure that `Height`, `Step` and `Round` match but we "don't care" about `LeaderId`, `IsLeader` and `NodeId` which are either ignored or handled in specific assertions.

consensus/module.go
deblasis: Maybe addressed in https://github.com/pokt-network/pocket/pull/198/commits/5c87b153042b43b85152faeccde815e02f9748d5

consensus/helpers.go
```diff
@@ -124,6 +125,15 @@ func (m *ConsensusModule) isOptimisticThresholdMet(n int) error {
 	return nil
 }
 
+func (m *ConsensusModule) resetForNewHeight() {
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/198/commits/97cee6385ab82b82bb978dca2b3cda0503e38bde

consensus/pacemaker.go
deblasis: Removed as not used anymore in https://github.com/pokt-network/pocket/pull/198/commits/37a0bd8be657b48c365b20c424513121679f0e49

consensus/hotstuff_leader.go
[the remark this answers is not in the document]
deblasis: Your comment made me think thoroughly about the whole locking thing and I think I have simplified it to the main points which are really the message handlers (on both leader and replica) that ultimately perform **writes** and the `GetNodeState` method that is used to **read** the state.

I have simplified the whole thing a lot and removed the `accessors.go` file that I previously included.

consensus/consensus_tests/hotstuff_test.go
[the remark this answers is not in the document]
Olshansk: Got it. In that case, I think we should call it `assertConsensusView` which is just a `view` into the (height, step, round) tuple. Terminology taken from https://arxiv.org/pdf/1803.05069.pdf.

consensus/consensus_tests/utils_test.go
```diff
@@ -401,3 +406,51 @@ func baseTelemetryEventMetricsAgentMock(t *testing.T) *modulesMock.MockEventMetr
...
+// sleep pauses the goroutine for the given duration on the mock clock and logs what just happened.
+//
+// Note: time has to be moved forward in a separate goroutine, see `advanceTime`.
+func sleep(clock *clock.Mock, duration time.Duration) {
+	log.Printf("[⌚ CLOCK 💤] sleeping for %v", duration)
+	clock.Sleep(duration)
+}
+
+// timeReminder simply prints, at a given interval and in a separate goroutine, the current mocked time to help with debug.
+func timeReminder(clock *clock.Mock, frequency time.Duration) {
```
Olshansk, line 430: Not actionable: this is probably the most idiomatic go function I've seen

consensus/module.go
Olshansk: I have some other related questions, but this definitely addresses anything a reader should need to know. No more & no less.

shared/README.md
```diff
@@ -45,7 +45,7 @@ The key things to keep in mind are:
   - Receive asynchronous events from the **main events channel**
 - The **Persistence** module is the only module that communicates with the local database
 - The **P2P** module is the only one that communicates with the outside world
-
+- **Clock** is a drop-in replacement for some of the features offered by the `time` package, it acts as an injectable clock implementation used to provide time manipulation while testing.
```
Olshansk, line 48: make sure to update the changelog as well

shared/bus.go
Olshansk: LGTM

consensus/pacemaker.go
```diff
@@ -103,15 +103,17 @@ func (m *paceMaker) SetConsensusModule(c *ConsensusModule) {
 }
 
 func (p *paceMaker) ValidateMessage(m *typesCons.HotstuffMessage) error {
+	currentHeight := p.consensusMod.Height
```
Olshansk, line 106: I really like the new simpler implementation, but this function is a good example of trying to figure out when I could/should use the lock or not.

Intuitively, I'd assume that I should acquire a read lock at the beginning of this function before getting `currentHeight := p.consensusMod.Height`, and I don't think it'll hurt, but the question is: should I?

Per this comment: `and a RLock when you want to make sure that no other goroutine is changing the values you are trying to read concurrently`, I think the answer is yes but I think that not acquiring a lock here was an explicit decision.

Olshansk (changes requested):

consensus/consensus_tests/hotstuff_test.go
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/198/commits/1367042997c0d30be1d365c4d8d4fd17b9ee9869

consensus/pacemaker.go
deblasis: Great question that perhaps is going to make us both, and anyone reading us in the future really, how to go about this.

Basically we want to acquire lock as a way of stopping time and therefore other goroutine for changing state of the things we are trying to access/manipulate.

If we lock in the middle of a non-atomic operation we risk finding ourselves in an inconsistent state aka 🤷‍♂️

For example here, if we lock inside `ValidateMessage` and another goroutine changes the state right after, the actions that follow the message validation could not be coherent anymore with that state we just validated and cause all sorts of errors.

Your question made me think about it even further and I realized that I was locking inside the message handlers but really I had to include also the validation and that change is visible here https://github.com/pokt-network/pocket/pull/198/commits/7bb04c3fae2c6d016c198f826058b51446dd64ca 

So, to answer your question: I believe that locking context should be the smallest possible but not smaller than a single "unit of work".
This PR allowed me to dive deeper into the code and better understand its moving parts where I started naively with the smallest locks possible (getters/setters) but then with tests and questions I started zooming out until I found what I believe is now the right balance.

My confidence however increased by running this whole thing every once in a while :smile: :
```bash
for i in {1..100}; do go test -timeout 2s -count=1 -run ^TestTinyPacemakerTimeouts$ github.com/pokt-network/pocket/consensus/consensus_tests; done > test_runpacemakertimeouts.txt 
for i in {1..100}; do go test -timeout 2s -count=1 -run ^TestHotstuff4Nodes1BlockHappyPath$ github.com/pokt-network/pocket/consensus/consensus_tests; done > testrun_happypath.txt 
for i in {1..100}; do go test -timeout 2s -count=1 -race -run ^TestTinyPacemakerTimeouts$ github.com/pokt-network/pocket/consensus/consensus_tests; done > test_runpacemakertimeouts_race.txt 
for i in {1..100}; do go test -timeout 2s -count=1 -race -run ^TestHotstuff4Nodes1BlockHappyPath$ github.com/pokt-network/pocket/consensus/consensus_tests; done > testrun_happypath_race.txt
```

shared/README.md
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/198/commits/09f1914ba2e85be46dbac330f48e48b403fe93db

consensus/hotstuff_leader.go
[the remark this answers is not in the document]
Olshansk: Lack of `accessors.go` makes it much simpler & clearer so going to resolve this.

p2p/module_raintree_test.go
Olshansk: I have some ideas with regarding to this, but haven't fully fleshed them out yet.

# Regard `make develop_test`

The core issue is:
1. We add a new dependency in branch A
2. We sitch to branch B where dependency does not exist
3. We can't `mockgen` because of inconsistent direcotires
4. We can't `go mod tidy` because we need to run `mockgen`

Screen Shot 2022-09-21 at 8 03 10 PM

I think something with the `mod build flag` might help (see https://github.com/golang/mock#debugging-errors) , but I haven't quite figured it out yet...

Screen Shot 2022-09-21 at 8 05 26 PM


# Regarding linters

I recently came by https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint, where makrdown linting can be defined via a YAML file. I'm sure something similar exists for Go, so we can make it part of the developer guide to set up their IDE to import these configs.

--- 

Will look into this tomorrow when I need a break from thinking :)

consensus/pacemaker.go
Olshansk: > Your question made me think about it even further

Lol, I think my questions have caused you a lot of thinking this week :)

> My confidence however increased by running this whole thing every once in a while

Put it in a make target! Something like `consensus_concurrenct_tests`

> So, to answer your question: I believe that locking context should be the smallest possible but not smaller than a single "unit of work".

Can you add this comment next to the mutex declaration in `consensus/module.go`

consensus/pacemaker.go
```diff
@@ -154,42 +156,41 @@ func (p *paceMaker) RestartTimer() {
...
-		case <-time.After(stepTimeout + 30*time.Millisecond): // Adding 30ms to the context timeout to avoid race condition.
+		case <-clock.After(stepTimeout + 30*timePkg.Millisecond): // Adding 30ms to the context timeout to avoid race condition.
 			return
 		}
 	}()
 }
 
 func (p *paceMaker) InterruptRound() {
-	p.consensusMod.nodeLog(typesCons.PacemakerInterrupt(p.consensusMod.Height, p.consensusMod.Step, p.consensusMod.Round))
+	p.consensusMod.nodeLog(typesCons.PacemakerInterrupt(p.consensusMod.CurrentHeight(), p.consensusMod.Step, p.consensusMod.Round))
```
Olshansk, line 183: I did a grep for `.Lock()` and can see that literally we've isolated it to two places:
1. HandleMessage
2. HandleDebugMessage

This makes sense and I like because it is the entrypoint to all the other business logic. 

The only piece I believe it's missing (though I can't say I'm 100% confident) is asynchronous events, whereas the `HandleMessage` and `HandleDebugMessage` are synchronous in their downstream operations.

What I did is grep for `go ` in the rest of the consensusModule and `RestartTimer` is the only thing that came up.

tl;dr I think we need to acquire a lock at the beginning of `InterruptRound` as well.

consensus/module.go
```diff
@@ -34,6 +35,11 @@ type ConsensusModule struct {
 	privateKey cryptoPocket.Ed25519PrivateKey
 	consCfg    modules.ConsensusConfig
 
+	// m is a mutex used to control synchronization when multiple goroutines are accessing the struct and its fields / properties.
```
Olshansk, line 38: I think I figured out the answer to my own question based on your "smallest unit of work" idea.

Every method that implements a function from the consensus interface (i.e. `type ConsensusModule interface`) needs to acquire & defer a lock.

I think we should put this as a comment in `shared/modules/consensus_module.go`. Something like: "NOTE: Consensus is the core of the replicated state machine and is driven by various asynchronous events. Consider adding a mutex to your implementation that is acquired at the beginning of each entrypoint/function implemented in this interface"

Olshansk (changes requested):
I think we're almost on the home stretch!!

Also, for future reference, you should be able to create branches directly from the main repo; `deblasis:issue/192-tooling-flaky-tests`.

p2p/module_raintree_test.go
Olshansk: Your change to mockgen helped solve one of the problems: https://github.com/pokt-network/pocket/pull/240/files

consensus/pacemaker.go
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/198/commits/8a30ef9d2911eba6fbf4a34cfb4f15494121d546 and https://github.com/pokt-network/pocket/pull/198/commits/2a7283ecc375cb9520cc1ce8590268fa14c9dffd

consensus/pacemaker.go
deblasis: `InterruptRound` is called internally when we handle messages, locking in there, even if it sounds correct, would cause deadlocks.

I have added some info about this here https://github.com/pokt-network/pocket/pull/198/commits/46b1d648b52e445c09d228760df10363a8fff1c0

consensus/module.go
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/198/commits/46b1d648b52e445c09d228760df10363a8fff1c0

p2p/module_raintree_test.go
Olshansk: 1. @okdas is helping with streamlining configs
2. Created #257 to track this work
3. We resolved the mockgen piece

consensus/pacemaker.go
Olshansk: LGTM!

Olshansk (approved):

PR #220, opened 2022-09-15
[TECHDEBT] Consolidate `GetHashStringFromBytes`

deblasis:
Tending to the following TODO:
```golang
// CLEANUP: Consolidate this with other similar functions in the codebase.
```
found in `p2p/raintree/utils.go` 

Heads up for @andrewnguyen22 since it touches a couple of files in `utility` as well.


Olshansk (approved):

Olshansk:
@deblasis I think its overkill to add a project/milestone/etc to all these small tickets, but let's just use the `code health` label

PR #190, opened 2022-09-02
[Automation] Add CI linters/checks

okdas:
## Description

Adding linters and static checkers to CI rotation.

## Issue

CI part of https://github.com/pokt-network/pocket/issues/175

## Type of change

Please mark the options that are relevant.

- [ ] New feature, functionality or library
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other: <REPLACE_ME>

## List of changes

- Adding more checks for CI

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have tested my changes using the available tooling
- [x] If applicable, I have made corresponding changes to related local or global README
- [ ] If applicable, I have added new diagrams using [mermaid.js](https://mermaid-js.github.io)
- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works

PR #235, opened 2022-09-20
[Shared] Config and genesis handling refactoring idea

deblasis:
## Description

This PR is a way to convey a design idea for the refactoring of the way we handle `config` and `genesis` as highlighted in a couple of Protocol Hours and also summarized, along with more context that should help us making the right decisions, into the discussion https://github.com/pokt-network/pocket/discussions/234

## Issue

There's no issue per-se, this is basically TECHDEBT handling I guess. Let me know if I should create one.

## Type of change

- [x] Code health or cleanup

## List of changes

- Abstracted config and genesis handling
- Mockable runtime
- Refactored all modules to use RuntimeMgr
- Updated RuntimeMgr to manage clock as well
- Modules now accept `interfaces` instead of paths.
- Unmarshalling is done in a new `runtime` package (runtime because what we do in there affects the runtime of the application)
- We are now able to accept configuration via environment variables (thanks to @okdas for inspiration and [sp13 for Viper]("github.com/spf13/viper"))

## Testing

- [x] `make test_all`

## Checklist

- [x] If applicable, I have made corresponding changes to related local or global README


p2p/module_raintree_test.go
```diff
@@ -281,7 +283,8 @@ func generateKeys(_ *testing.T, numValidators int) []cryptoPocket.PrivateKey {
 
 // A mock of the application specific to know if a message was sent to be handled by the application
 // INVESTIGATE(olshansky): Double check that how the expected calls are counted is accurate per the
-//                         expectation with RainTree by comparing with Telemetry after updating specs.
+//
```
andrewnguyen22, line 286: Getting some failed tests:

```
module_raintree_test.go:386
        	            				module_raintree_test.go:209
        	            				module_raintree_test.go:180
        	Error:      	Received unexpected error:
        	            	Config File "config" Not Found in "[/Users/andrewnguyen/go/src/github.com/pokt-network/pocket/p2p]"
        	Test:       	TestRainTreeCompleteTwentySevenNodes
```

runtime/config.go
```diff
@@ -0,0 +1,89 @@
+package runtime
```
andrewnguyen22, line 1: My general issue with this package as described in https://github.com/pokt-network/pocket/discussions/234

Is that we essentially allow for other modules to access `module specific` configuration and genesis states.

This is not at all the intention

This seems like an error prone architecture to me

shared/modules/module.go
```diff
@@ -18,9 +19,3 @@ type InterruptableModule interface {
 	Start() error
 	Stop() error
 }
-
-type InitializableModule interface {
```
andrewnguyen22, line 22: When we talk about `code clarity` ; here's another example where we loosen interfaces. It's difficult for me to say that the `V1 - Structs` design is definitely `clearer`

shared/node.go
```diff
@@ -28,32 +29,38 @@ type Node struct {
...
+		return nil, err
+	}
+
+	persistenceMod, err := persistence.Create(cfg.Persistence, genesis.PersistenceGenesisState)
 	if err != nil {
 		return nil, err
 	}
 
-	p2pMod, err := p2p.Create(configPath, genesisPath, false)
+	p2pMod, err := p2p.Create(cfg.P2P, false)
```
andrewnguyen22, line 43: The design pattern for module creation is loosened here

shared/test_artifacts/generator.go
```diff
@@ -34,64 +41,64 @@ var (
 // TODO (Team) this is meant to be a **temporary** replacement for the recently deprecated
 // 'genesis config' option. We need to implement a real suite soon!
 func NewGenesisState(numValidators, numServiceNodes, numApplications, numFisherman int) (genesisState modules.GenesisState, validatorPrivateKeys []string) {
-	apps, appsPrivateKeys := NewActors(MockActorType_App, numApplications)
-	vals, validatorPrivateKeys := NewActors(MockActorType_Val, numValidators)
-	serviceNodes, snPrivateKeys := NewActors(MockActorType_Node, numServiceNodes)
-	fish, fishPrivateKeys := NewActors(MockActorType_Fish, numFisherman)
+	apps, appsPrivateKeys := NewActors(types.UtilActorType_App, numApplications)
```
andrewnguyen22, line 44: I like that we deleted the mocks!

I don't think this change is exclusive to either `v1 strings or structs` implementation however

utility/test/module_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"os"
 	"testing"
 
+	typesPers "github.com/pokt-network/pocket/persistence/types"
```
andrewnguyen22, line 11: `Utility mod` added import from `Persistence mod`. 

We're trying to get rid of this cross module sharing

andrewnguyen22 (changes requested):

shared/node.go
deblasis: Maybe this is better
image

maybe builder is not the best naming here

image

image

image

shared/test_artifacts/generator.go
deblasis: I tried to cater to both point of views and hacked some code to see if something in between could be feasible of not.

utility/test/module_test.go
deblasis: Perhaps we could move module creation for testing purposes under test_artifacts to hide away the implementation details of the other modules when we are testing one

runtime/config.go
deblasis: I see what you mean but if we think about configurability in general isn't the environment (and therefore environment variables, configuration files, etc) something that's accessible everywhere within the app one way or another? 

What we should guard against is not the mere access but whenever for example inside the P2P module I use the PostgresUrl (part of persistence config) to build some logic on top of. Obviously, that's fragile and spaghetti code and bug prone because someone might just change that in persistence and consequently break P2P for example.

The environment is a `os.GetEnv()` call away and therefore any config value as well potentially.
The point is: will it make it to `main`?
Same applies for accessing configuration/genesis files.

If we want the modules to be completely isolated from each other and to bring to zero the shared code I think they should be running as separate binaries on their own isolation context (Docker container or whatnot) and communicate with the other modules via some form of well defined RPC. But we are still in monolith land here and everything is a tradeoff. Anyway, I am glad I can pick your brain about this kind of stuff. 🙏

p2p/module_raintree_test.go
andrewnguyen22: Awesome :)

shared/node.go
andrewnguyen22: Yes, the `getRandomPK` should go away soon, but I like this a log better if you guys do

++ I renaming NewBuilder though

Olshansk (changes requested):
I think my review is a bit outdated but figured I'd publish the comments I have as an interim iteration. Looking great so far

app/client/main.go
```diff
@@ -47,24 +48,32 @@ var consensusMod modules.ConsensusModule
 
 func main() {
 	var err error
-	consensusMod, err = consensus.Create(defaultConfigPath, defaultGenesisPath, true) // TECHDEBT: extra param required for injecting private key hack for debug client
+
+	runtime := runtime.New(defaultConfigPath, defaultGenesisPath, runtime.WithRandomPK())
+	cfg := runtime.GetConfig()
+	genesis := runtime.GetGenesis()
+
+	cons, err := consensus.Create(runtime) // TECHDEBT: extra param required for injecting private key hack for debug client
```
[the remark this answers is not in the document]
deblasis: Addressed in 2fec88b

app/client/main.go
```diff
@@ -47,24 +48,32 @@ var consensusMod modules.ConsensusModule
...
+	consensusMod := cons.(modules.ConsensusModule)
 	if err != nil {
 		log.Fatalf("[ERROR] Failed to create consensus module: %v", err.Error())
 	}
-	p2pMod, err = p2p.Create(defaultConfigPath, defaultGenesisPath, true) // TECHDEBT: extra param required for injecting private key hack for debug client
+	p2pm, err := p2p.Create(runtime) // TECHDEBT: extra param required for injecting private key hack for debug client
```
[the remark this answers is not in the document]
deblasis: Addressed in 2fec88b

app/client/main.go
```diff
@@ -47,24 +48,32 @@ var consensusMod modules.ConsensusModule
...
 	if err != nil {
 		log.Fatalf("[ERROR] Failed to create p2p module: %v", err.Error())
 	}
+	p2pMod := p2pm.(modules.P2PModule)
 	// This telemetry module instance is a NOOP because the 'enable_telemetry' flag in the `cfg` above is set to false.
 	// Since this client mimics partial - networking only - functionality of a full node, some of the telemetry-related
 	// code paths are executed. To avoid those messages interfering with the telemetry data collected, a non-nil telemetry
 	// module that NOOPs (per the configs above) is injected.
-	telemetryMod, err := telemetry.Create(defaultConfigPath, defaultGenesisPath)
+	telemetryM, err := telemetry.Create(runtime)
```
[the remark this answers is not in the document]
deblasis: Addressed in 2fec88b

consensus/module.go
```diff
@@ -65,35 +64,35 @@ type ConsensusModule struct {
...
+func Create(runtime modules.Runtime) (modules.Module, error) {
+	var m ConsensusModule
+	return m.Create(runtime)
+}
+
+func (*ConsensusModule) Create(runtime modules.Runtime) (modules.Module, error) {
+	cfg := runtime.GetConfig()
+	genesis := runtime.GetGenesis()
+
+	moduleCfg := cfg.Consensus.(*typesCons.ConsensusConfig)
```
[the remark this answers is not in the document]
deblasis: The problem is that when we return `any` we would still have to cast to the module specific config. Validation is something that we can add a signature for but casting unfortunately has to be done on the calling site I think.

I took a stab at the concept, please have a look at https://github.com/pokt-network/pocket/pull/235/commits/25979e351e0da2f5b43ead67245afa97e773c9be

If you likey I am gonna apply the same concept to all the other modules as well, also what's nice is that we would know if a module needs genesis or config or both with these interfaces. Naming is hard so suggestions are welcome :)

consensus/module.go
```diff
@@ -65,35 +64,35 @@ type ConsensusModule struct {
...
+	paceMaker, err := CreatePacemaker(moduleCfg)
 	if err != nil {
 		return nil, err
 	}
 
-	valMap := typesCons.ValidatorListToMap(genesis.Validators)
+	valMap := typesCons.ValidatorListToMap(moduleGenesis.Validators)
 	var privateKey cryptoPocket.PrivateKey
-	if useRandomPK {
+	if runtime.ShouldUseRandomPK() {
```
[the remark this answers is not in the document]
deblasis: Have a look at https://github.com/pokt-network/pocket/pull/235/commits/7c5b94399f48b7587336b621274edc41c20c406a, if you like the concept I'll apply elsewhere too. 

We could argue that this line https://github.com/pokt-network/pocket/commit/7c5b94399f48b7587336b621274edc41c20c406a#diff-eba95a8b904f8186683d2f9c106a3dc021f014918420e6a074d0a2b4e59c261dR210

could sound a bit "cryptic", pun intended, but we are good citizens and config has been already validated when we are trying to extract stuff from it, eg: `PrivateKey`

runtime/runtime.go
```diff
@@ -0,0 +1,108 @@
...
+	genesisPath string
+
+	config  *Config
+	genesis *Genesis
+
+	useRandomPK bool
+}
+
+func New(configPath, genesisPath string, options ...func(*RuntimeConfig)) *RuntimeConfig {
+	b := &RuntimeConfig{
```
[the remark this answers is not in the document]
deblasis: the struct used to be called Builder in its infancy but now it's a teen RuntimeConfig even if some folks still call it "b" :) 

Enough, time to grow up!

runtime/runtime.go
```diff
@@ -0,0 +1,108 @@
...
+
+	cfg, genesis, err := b.init()
+	if err != nil {
+		log.Fatalf("[ERROR] Failed to initialize runtime builder: %v", err)
+	}
+	b.config = cfg
+	b.genesis = genesis
+
+	for _, o := range options {
+		o(b)
```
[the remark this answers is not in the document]
deblasis: Correct, I would say, and correct me if I am wrong: this allow for flexibility, can it be improved? Maybe. Can this be used maliciously, of course. 

In Go you are the owner of the code and you can do anything with it. In other languages you can enforce immutability and there are well-known patterns, in Go I have seen something like wrapping values in a struct but there are walkarounds and it makes the code hard to understand IMHO. 

Open to suggestions here, happy to learn something new

runtime/runtime.go
```diff
@@ -0,0 +1,108 @@
...
+	viper.SetConfigType("json")
+
+	// The lines below allow for environment variables configuration (12 factor app)
+	// Eg: POCKET_CONSENSUS_PRIVATE_KEY=somekey would override `consensus.private_key` in config
+	viper.SetEnvPrefix("POCKET")
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
+
+	err = viper.ReadInConfig()
+	if err != nil {
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/69af7f16829f0be9f49c83b6ea547cc09937b5fd

runtime/runtime.go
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/69af7f16829f0be9f49c83b6ea547cc09937b5fd

runtime/runtime.go
```diff
@@ -0,0 +1,108 @@
...
+		return
+	}
+
+	if config.Base == nil {
+		config.Base = &BaseConfig{}
+	}
+	config.Base.ConfigPath = b.configPath
+	config.Base.GenesisPath = b.genesisPath
+
+	genesis, err = ParseGenesisJSON(b.genesisPath)
```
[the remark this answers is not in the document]
deblasis: Yes, currently accessed by persistence in HandleDebugMessage
https://github.com/deblasis/pocket/blob/69af7f16829f0be9f49c83b6ea547cc09937b5fd/persistence/debug.go#L18-L19

shared/modules/runtime_module.go
```diff
@@ -0,0 +1,7 @@
+package modules
+
+type Runtime interface {
+	GetConfig() Config
+	GetGenesis() GenesisState
+	ShouldUseRandomPK() bool
```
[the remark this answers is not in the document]
deblasis: I put it there exactly because there's a semantic conflict between that flag and the privateKey in config.

If this was only used in tests, yeah, we can mock that but currently the client needs a random private key to work for debugging purposes. I concur, something to think about.

persistence/test/setup_test.go
```diff
@@ -107,11 +108,14 @@ func newTestPersistenceModule(databaseUrl string) modules.PersistenceModule {
 	}
 	genesisState, _ := test_artifacts.NewGenesisState(5, 1, 1, 1)
 	createTestingGenesisAndConfigFiles(cfg, genesisState)
-	persistenceMod, err := persistence.Create(testingConfigFilePath, testingGenesisFilePath)
+
+	runtime := runtime.New(testingConfigFilePath, testingGenesisFilePath)
```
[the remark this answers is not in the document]
deblasis: Nothing stops us from having a `runtime.NewFromSomething(something Something)` 

If you can provide practical examples / an issue, we can totally do that

p2p/module.go
```diff
@@ -42,55 +42,39 @@ func (m *p2pModule) GetAddress() (cryptoPocket.Address, error) {
...
 	log.Println("Creating network module")
-	c, err := new(p2pModule).InitConfig(configPath)
-	if err != nil {
-		return nil, err
-	}
-	cfg := c.(*typesP2P.P2PConfig)
-	l, err := CreateListener(cfg)
+
+	cfg := runtime.GetConfig()
+	moduleCfg := cfg.P2P.(*typesP2P.P2PConfig)
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/56be6750b57fa284eb86f92e6b58445faa2738fc for all modules that used moduleCfg

p2p/module.go
```diff
@@ -22,6 +20,8 @@ import (
 )
 
 var _ modules.P2PModule = &p2pModule{}
+var _ modules.Module = &p2pModule{}
+var _ modules.InitializableModule = &p2pModule{}
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/49980dc8a92dbe8a7f80302f9ce2e130001ed8ac

shared/modules/runtime_module.go
Olshansk, line 6:
```suggestion
	ShouldUseRandomPK() bool // INVESTIGATE: look into how we can remove this from the runtime interface
```

runtime/config.go
```diff
@@ -0,0 +1,106 @@
...
+	return c.Consensus.MaxMempoolBytes
+}
+
+// modules.P2PConfig
+
+func (c *Config) GetConsensusPort() uint32 {
+	return c.P2P.ConsensusPort
+}
+
+func (c *Config) GetUseRainTree() bool {
```
Olshansk, line 72: Similar to how you hid the pacemaker config from here, can you also do the same thing for RainTree?

I think one can argue that if someone implements a `kademila` or `gemini` P2P layer, it shouldn't know of `RainTree`.

utility/test/module_test.go
Olshansk: Created https://github.com/pokt-network/pocket/issues/239

Olshansk (changes requested):
I think one more round of reviews and we'll get over the 🏁

runtime/runtime.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/pokt-network/pocket/consensus/types"
+	"github.com/pokt-network/pocket/shared/modules"
+	"github.com/spf13/viper"
+)
+
+var _ modules.Runtime = &RuntimeConfig{}
+
+type RuntimeConfig struct {
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/c3ac3cc55bd9d59f79c1918cb5a88caa24e6dfa9

runtime/runtime.go
```diff
@@ -0,0 +1,108 @@
...
+	configPath  string
+	genesisPath string
+
+	config  *Config
+	genesis *Genesis
+
+	useRandomPK bool
```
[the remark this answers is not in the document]
deblasis: I think I bypassed the problem via https://github.com/pokt-network/pocket/pull/235/commits/66b4806adb2aebea3fbbedcfa7b067921656b9f4

runtime/runtime.go
```diff
@@ -0,0 +1,108 @@
...
+	rc := &RuntimeConfig{
+		configPath:  configPath,
+		genesisPath: genesisPath,
+	}
+
+	cfg, genesis, err := rc.init()
+	if err != nil {
+		log.Fatalf("[ERROR] Failed to initialize runtime builder: %v", err)
+	}
+	rc.config = cfg
```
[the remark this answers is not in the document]
deblasis: I wouldn't put side-effects in `init`, we are accessing the filesystem and we want to have control also on -when- that happens.

shared/modules/module.go
```diff
@@ -21,6 +23,17 @@ type InterruptableModule interface {
 
 type InitializableModule interface {
 	GetModuleName() string
-	InitConfig(pathToConfigJSON string) (IConfig, error)
-	InitGenesis(pathToGenesisJSON string) (IGenesis, error)
+	Create(runtime Runtime) (Module, error)
+}
+
+type ConfigurableModule interface {
```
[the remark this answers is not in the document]
deblasis: My Ouija board sometimes delivers :)

shared/node.go
```diff
@@ -27,37 +28,52 @@ type Node struct {
 	Address cryptoPocket.Address
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/d663287b23063f9d9d0c701d54087942a0874bb8

consensus/module.go
```diff
@@ -22,10 +19,17 @@ const (
...
 
-var _ modules.ConsensusGenesisState = &typesCons.ConsensusGenesisState{}
-var _ modules.PacemakerConfig = &typesCons.PacemakerConfig{}
-var _ modules.ConsensusConfig = &typesCons.ConsensusConfig{}
-var _ modules.ConsensusModule = &ConsensusModule{}
+var (
+	_ modules.ConsensusGenesisState = &typesCons.ConsensusGenesisState{}
+
+	_ modules.PacemakerConfig = &typesCons.PacemakerConfig{}
+	_ modules.ConsensusConfig = &typesCons.ConsensusConfig{}
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/9e5602b0fed0c066875072df67c9767d1af515fd

consensus/module.go
```diff
@@ -65,47 +69,53 @@ type ConsensusModule struct {
...
+	consensusCfg := cfg.Consensus.(*typesCons.ConsensusConfig)
+
+	genesis := runtime.GetGenesis()
+	if err := m.ValidateGenesis(genesis); err != nil {
+		log.Fatalf("genesis validation failed: %v", err)
 	}
-	cfg := c.(*typesCons.ConsensusConfig)
-	genesis := g.(*typesCons.ConsensusGenesisState)
-	leaderElectionMod, err := leader_election.Create(cfg, genesis)
+	moduleGenesis := genesis.ConsensusGenesisState.(*typesCons.ConsensusGenesisState)
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/fb873b6fead51aef65b563791fc41b2c1ace8af5

persistence/module.go
```diff
@@ -36,19 +34,19 @@ const (
...
+	var m PersistenceModule
+	return m.Create(runtime)
+}
+
+func (*PersistenceModule) Create(runtime modules.Runtime) (modules.Module, error) {
+	cfg := runtime.GetConfig()
+	genesis := runtime.GetGenesis()
+
+	persistenceCfg := cfg.Persistence.(*types.PersistenceConfig)
+	moduleGenesis := genesis.PersistenceGenesisState.(*types.PersistenceGenesisState)
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/8d26fd389701e23295921b6de5a709722011329b

utility/test/module_test.go
okdas: @Olshansk that is likely going to be a somewhat sophisticated AST rule but should be doable! 👍

runtime/config.go
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/dd78b2ead26dc5ec6373e412a75e5ee1837a132e

persistence/debug.go
```diff
@@ -15,11 +15,11 @@ func (m *PersistenceModule) HandleDebugMessage(debugMessage *debug.DebugMessage)
 		m.showLatestBlockInStore(debugMessage)
 	case debug.DebugMessageAction_DEBUG_CLEAR_STATE:
 		m.clearState(debugMessage)
-		g, err := m.InitGenesis(m.genesisPath)
+		g, err := runtime.ParseGenesisJSON(m.GetBus().GetConfig().Base.GenesisPath)
```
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/c6c95dd570d9b9cdb0cf0008af3f8d248d16da5f

consensus/module.go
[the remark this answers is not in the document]
Olshansk: > could sound a bit "cryptic"

( ͡° ͜ʖ ͡°)

In general, I don't **love** this solution but I think it's more than good enough to get us going.

persistence/test/setup_test.go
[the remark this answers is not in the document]
Olshansk: Resolving!

runtime/runtime.go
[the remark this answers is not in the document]
Olshansk: I do like this approach and was simply confirming. +💯

consensus/module.go
```diff
@@ -216,6 +188,24 @@ func (m *ConsensusModule) SetBus(pocketBus modules.Bus) {
...
+}
+
+func (*ConsensusModule) ValidateGenesis(genesis modules.GenesisState) error {
+	if _, ok := genesis.ConsensusGenesisState.(*typesCons.ConsensusGenesisState); !ok {
+		return fmt.Errorf("cannot cast to ConsensusGenesisState")
+	}
+	return nil
+}
+
+func (*ConsensusModule) GetPrivateKey(runtime modules.Runtime) (cryptoPocket.PrivateKey, error) {
```
Olshansk, line 205: Do we not need the `sholdUseRandomPk` part here?

consensus/module.go
```diff
@@ -22,10 +19,12 @@ const (
...
 )
 
-var _ modules.ConsensusGenesisState = &typesCons.ConsensusGenesisState{}
-var _ modules.PacemakerConfig = &typesCons.PacemakerConfig{}
-var _ modules.ConsensusConfig = &typesCons.ConsensusConfig{}
-var _ modules.ConsensusModule = &ConsensusModule{}
+var (
+	_ modules.ConsensusModule = &ConsensusModule{}
+
+	_ modules.ConsensusConfig       = &typesCons.ConsensusConfig{}
```
Olshansk, line 25:
```suggestion
	_ modules.ConsensusModule = &ConsensusModule{}
	_ modules.ConsensusConfig       = &typesCons.ConsensusConfig{}
```

p2p/module.go
```diff
@@ -125,7 +107,7 @@ func (m *p2pModule) Start() error {
 		return err
 	}
 
-	if m.p2pConfig.GetUseRainTree() {
+	if m.p2pConfig.(*typesP2P.P2PConfig).UseRainTree {
```
Olshansk, line 110: I feel like we can locally store the casted type on line 31.

Olshansk (approved):
Going to call out I'm pretty sure we're all aware we'll keep iterating on this in main over time, but I really like where it's going so LFG!

consensus/module.go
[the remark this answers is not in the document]
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/f5ea7e47c9d7c1b2e84b08cb40a0ca70c61299b5

p2p/module.go
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/c60206fc430c0b2a8c93ec473d9626aaebadef20

Also applied the same pattern to the other modules for consistency and also removed config related fields from the modules to use the config instead.

consensus/module.go
deblasis: not anymore

image


😉

deblasis:
> Going to call out I'm pretty sure we're all aware we'll keep iterating on this in main over time, but I really like where it's going so LFG!

I think this needs a nod from @andrewnguyen22 as well

Olshansk:
> I think this needs a nod from @andrewnguyen22 as well

Agreed. LFG after we get a 👍 from @andrewnguyen22, or not go yet 😅

app/client/main.go
```diff
@@ -47,24 +48,32 @@ var consensusMod modules.ConsensusModule
 
 func main() {
 	var err error
-	consensusMod, err = consensus.Create(defaultConfigPath, defaultGenesisPath, true) // TECHDEBT: extra param required for injecting private key hack for debug client
+
+	runtime := runtime.New(defaultConfigPath, defaultGenesisPath, runtime.WithRandomPK())
+	cfg := runtime.GetConfig()
```
andrewnguyen22, line 53: I don't understand why all modules get passed `runtime` object (which contains config and genesis) and yet, the creation of the `Bus` takes the config and genesis objects separately. Please pass the runtime object to the `bus creation`

app/client/main.go
```diff
@@ -47,24 +48,32 @@ var consensusMod modules.ConsensusModule
...
+	genesis := runtime.GetGenesis()
+
+	consM, err := consensus.Create(runtime)
+	consensusMod := consM.(modules.ConsensusModule)
```
andrewnguyen22, line 57: Error handle before casting ^

consensus/consensus_tests/pacemaker_test.go
```diff
@@ -133,7 +133,7 @@ func TestPacemakerCatchupSameStepDifferentRounds(t *testing.T) {
 		Hash:              hex.EncodeToString(appHash),
 		NumTxs:            0,
 		LastBlockHash:     "",
-		ProposerAddress:   leader.Address.Bytes(),
+		ProposerAddress:   leader.GetP2PAddress().Bytes(),
```
andrewnguyen22, line 136: The `P2PAddress` should not be the address contained in the block header as it's just an ephemeral P2P identity. It should be the `Consensus Address`

consensus/consensus_tests/utils_test.go
```diff
@@ -127,21 +133,20 @@ func CreateTestConsensusPocketNode(
...
+	bus, err := shared.CreateBus(config, genesis, persistenceMock, p2pMock, utilityMock, consensusMod.(modules.ConsensusModule), telemetryMock)
 	require.NoError(t, err)
 
 	pk, err := cryptoPocket.NewPrivateKey(cfg.Base.PrivateKey)
 	require.NoError(t, err)
 
-	pocketNode := &shared.Node{
-		Address: pk.Address(),
-	}
+	pocketNode := shared.NewWithAddress(pk.Address())
```
andrewnguyen22, line 142: This function name is very non-descript. Please make it clearer what object is being returned from `shared` package

consensus/consensus_tests/utils_test.go
```diff
@@ -176,21 +181,21 @@ func StartAllTestPocketNodes(t *testing.T, pocketNodes IdToNodeMapping) {
 // TODO(discuss): Should we use reflections inside the testing module as being done here or explicitly
 // define the interfaces used for debug/development. The latter will probably scale more but will
 // require more effort and pollute the source code with debugging information.
-func GetConsensusNodeState(node *shared.Node) typesCons.ConsensusNodeState {
+func GetConsensusNodeState(node modules.NodeModule) typesCons.ConsensusNodeState {
```
andrewnguyen22, line 184: I don't think this change in naming convention made it clearer (I think the opposite)

consensus/leader_election/module.go
```diff
@@ -16,13 +16,18 @@ type LeaderElectionModule interface {
...
 
-var _ leaderElectionModule = leaderElectionModule{}
+var _ LeaderElectionModule = &leaderElectionModule{}
 
 type leaderElectionModule struct {
 	bus modules.Bus
 }
 
-func Create(_ *typesCons.ConsensusConfig, _ *typesCons.ConsensusGenesisState) (LeaderElectionModule, error) {
+func Create(runtime modules.Runtime) (modules.Module, error) {
```
andrewnguyen22, line 25: This create wrapper seems like a clunkier design, considering the function signatures are the same.

Have we considered `new(consensusModule).Create(runtime)` instead of using a wrapper for each?

consensus/module.go
```diff
@@ -65,47 +64,53 @@ type ConsensusModule struct {
...
-	c, err := cm.InitConfig(configPath)
-	if err != nil {
-		return nil, err
+func Create(runtime modules.Runtime) (modules.Module, error) {
+	var m ConsensusModule
+	return m.Create(runtime)
+}
+
+func (*ConsensusModule) Create(runtime modules.Runtime) (modules.Module, error) {
+	var m *ConsensusModule
```
andrewnguyen22, line 73: It doesn't make sense to me that you have a receiver variable that is ignored, but then the line below you declare another consensus module `var m *ConsensusModule`

consensus/module.go
```diff
@@ -124,53 +129,20 @@ func Create(configPath, genesisPath string, useRandomPK bool) (modules.Consensus
 		validatorMap: valMap,
 
 		utilityContext:    nil,
-		paceMaker:         paceMaker,
-		leaderElectionMod: leaderElectionMod,
+		paceMaker:         pacemakerMod,
```
andrewnguyen22, line 132: You renamed the variable `pacemakerMod` but kept the structure field the legacy `pacemaker` let's be consistent with that

p2p/module_raintree_test.go
```diff
@@ -15,12 +15,14 @@ import (
 	"testing"
```
andrewnguyen22, line 15:
```golang
for i := 0; i < numAddressessToBeAdded; i++ {
				newAddr, _ := crypto.GenerateAddress()
				network.AddPeerToAddrBook(&types.NetworkPeer{Address: newAddr})
			}
```

Catch the error in `newAddr, _ := crypto.GenerateAddress()`

p2p/module_raintree_test.go
```diff
@@ -15,12 +15,14 @@ import (
...
 	"time"
 
+	"github.com/pokt-network/pocket/runtime"
 	"github.com/pokt-network/pocket/shared/debug"
 	"github.com/pokt-network/pocket/shared/test_artifacts"
 
 	"github.com/golang/mock/gomock"
 	typesP2P "github.com/pokt-network/pocket/p2p/types"
 	mocksP2P "github.com/pokt-network/pocket/p2p/types/mocks"
+	typesPers "github.com/pokt-network/pocket/persistence/types"
```
andrewnguyen22, line 25: This is a step backwards when it comes to encapsulation. You've introduced cross module imports. Please revert / fix to use `test_artifacts`

persistence/types/converters.go
```diff
@@ -0,0 +1,149 @@
...
+func toPersistenceActor(actor modules.Actor) *Actor {
+	return &Actor{
+		Address:      actor.GetAddress(),
+		PublicKey:    actor.GetPublicKey(),
+		StakedAmount: actor.GetStakedAmount(),
+		GenericParam: actor.GetGenericParam(),
+	}
+}
+
+func ToPersistenceActors(actors []modules.Actor) []*Actor {
```
andrewnguyen22, line 14: Another case where cross module importing is introduced. (Used by p2p)

telemetry/prometheus_module.go
```diff
@@ -33,22 +34,31 @@ const (
...
-func CreatePrometheusTelemetryModule(cfg *TelemetryConfig) (*PrometheusTelemetryModule, error) {
+func CreatePrometheusTelemetryModule(runtime modules.Runtime) (modules.Module, error) {
+	var m PrometheusTelemetryModule
+	return m.Create(runtime)
+}
+
+func (m *PrometheusTelemetryModule) Create(runtime modules.Runtime) (modules.Module, error) {
+	cfg := runtime.GetConfig()
+	if err := m.ValidateConfig(cfg); err != nil {
+		log.Fatalf("config validation failed: %v", err)
```
andrewnguyen22, line 45: I don't really understand the pattern of not returning the error and logging fatal instead. (**Ditto for all module creation**)

runtime/runtime.go
```diff
@@ -0,0 +1,116 @@
+package runtime
+
+import (
+	"log"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/pokt-network/pocket/consensus/types"
```
andrewnguyen22, line 10: Let's be consistent with import naming `consensus/types` to `consTypes`

runtime/config.go
```diff
@@ -0,0 +1,102 @@
...
+	typesUtil "github.com/pokt-network/pocket/utility/types"
+)
+
+var _ modules.ConsensusConfig = &Config{}
+var _ modules.P2PConfig = &Config{}
+var _ modules.PersistenceConfig = &Config{}
+var _ modules.TelemetryConfig = &Config{}
+var _ modules.UtilityConfig = &Config{}
+
+type Config struct {
```
andrewnguyen22, line 21: There is a very similar structure in `shared/modules/types.go` that should be consolidated.

andrewnguyen22, line 21: I really prefer if this was an exported interface and not an exported structure

shared/modules/consensus_module.go
```diff
@@ -11,6 +11,9 @@ type ValidatorMap map[string]Actor // TODO (Drewsky) deprecate Validator map or
 
 type ConsensusModule interface {
 	Module
+	ConfigurableModule
```
andrewnguyen22, line 14: Question: Are there other `modules` that are not configurable and genesis dependent?

I would like to see if we can move some of these interfaces to `Module`

shared/modules/types.go
```diff
@@ -15,6 +15,8 @@ type GenesisState struct {
 type BaseConfig struct {
 	RootDirectory string `json:"root_directory"`
 	PrivateKey    string `json:"private_key"` // TODO (pocket/issues/150) better architecture for key management (keybase, keyfiles, etc.)
+	ConfigPath    string `json:"config_path"`
```
andrewnguyen22, line 18: These fields that are introduced are not used anywhere. What's the reasoning?

Also, we have a very similar structure under `runtime` package. What's the plan there?

shared/modules/bus_module.go
```diff
@@ -27,6 +25,6 @@ type Bus interface {
 	GetTelemetryModule() TelemetryModule
 
 	// Configuration
-	GetConfig() map[string]json.RawMessage
-	GetGenesis() map[string]json.RawMessage
+	GetConfig() Config
```
andrewnguyen22, line 28: Why are these methods re-introduced?

utility/test/module_test.go
```diff
@@ -63,7 +65,7 @@ func NewTestingUtilityContext(t *testing.T, height int64) utility.UtilityContext
 // TODO(andrew): Take in `t` and fail the test if there's an error
 func newTestPersistenceModule(databaseUrl string) modules.PersistenceModule {
 	cfg := modules.Config{
-		Persistence: &test_artifacts.MockPersistenceConfig{
+		Persistence: &typesPers.PersistenceConfig{
```
andrewnguyen22, line 68: Cross module import

utility/test/module_test.go
andrewnguyen22: I'm not sure how this was resolved. We just added an issue to automatically check

andrewnguyen22 (changes requested):

consensus/consensus_tests/utils_test.go
deblasis: I am not sure I understand what you mean. 

Is the fact that we are accepting the interface or the name of the interface that sounds odd?

utility/test/module_test.go
deblasis: Maybe addressed in https://github.com/pokt-network/pocket/pull/235/commits/1924e0e7e3c6988d13cb0b4a2a4edc5c07c64f2b#diff-e31d62c553295c0f69210e9d518a19d448a13f20bb057ad8d94b58a13bfcfa05L11

shared/modules/bus_module.go
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/2837b5aff91973a67988c6d5db9ae893a6445437

shared/modules/types.go
deblasis: Refactored mostly in https://github.com/pokt-network/pocket/pull/235/commits/1924e0e7e3c6988d13cb0b4a2a4edc5c07c64f2b

Relying more on abstractions

runtime/config.go
deblasis: the renamed `runtimeConfig` now implements `modules.Config`.

runtime package exports the constructor and the `BaseConfig`

https://github.com/pokt-network/pocket/pull/235/commits/1924e0e7e3c6988d13cb0b4a2a4edc5c07c64f2b#diff-22e2eb4a42003f34d95bacbe4bf7dfa8d3ddd6ec3eb66c5ed611b5b6b72de26bR14-R23

runtime/runtime.go
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/a8530d2ab589c248d53352a8df2b37d351b362ac

telemetry/prometheus_module.go
deblasis: Addressed in https://github.com/pokt-network/pocket/pull/235/commits/2a686db94104405ac0ff6b2320c740bbf32a11ae

app/client/main.go
deblasis: Updated.

Since we removed `GetConfig()` and `GetGenesis()` from the `bus`, I am currently not passing in the `runtimeMgr` in 
https://github.com/pokt-network/pocket/pull/235/files#diff-91e861a0bd67a2298f7a42d2ad6f060e8b59153fab4519a9f6bfd4fa293919c4R75 
and also in the `CreateBus()`.

Please let me know if you think it could be beneficial have a reference to the runtime accessible via the bus. But anyway we can always consider adding it when/if needed

shared/modules/consensus_module.go
deblasis: I'd rather keep interfaces separated to be honest.

It's true that the majority of modules are `ConfigurableModule` but only a couple are `GenesisDependentModule`.

I'd like to keep `Module` to the bare minimum.

persistence/types/converters.go
deblasis: This doesn't happen anymore

p2p/module_raintree_test.go
deblasis: Not happening anymore 

https://github.com/pokt-network/pocket/pull/235/files#diff-a8351346d15469e6fb8938609c72b7b4387e5d77fd634fc64ed82327ad40fb5dR4-R25

p2p/module_raintree_test.go
deblasis: Addressed here

https://github.com/pokt-network/pocket/pull/235/files#diff-2a2651c6ad9e1f7d34a394f137a8da69f8204d60e0d22ba7b304cb48d72e5d77R118

consensus/leader_election/module.go
deblasis: 🤦‍♂️ you are totally right

https://github.com/pokt-network/pocket/pull/235/files#diff-eba95a8b904f8186683d2f9c106a3dc021f014918420e6a074d0a2b4e59c261dR66-R68

consensus/module.go
deblasis: 🤦‍♂️ you are totally right

https://github.com/pokt-network/pocket/pull/235/files#diff-eba95a8b904f8186683d2f9c106a3dc021f014918420e6a074d0a2b4e59c261dR66-R68

consensus/leader_election/module.go
deblasis: Without the wrapper we would have to expose all the structs and actually in this round of changes, I made all modules structs unexported

consensus/consensus_tests/utils_test.go
deblasis: Agreed 

https://github.com/pokt-network/pocket/pull/235/files#diff-9f97d711d54398bf6257c266207519c11eb1d254c798d578786d0e94713cb164R151

consensus/consensus_tests/pacemaker_test.go
deblasis: This should fix it

https://github.com/pokt-network/pocket/pull/235/files#diff-c474ba0a5fb5b813e523913230483776eb1344705fb79db6d267ced0566c85f8R132-R141

deblasis:
Clarification about the fact that cross-module imports were still present when I requested review.

I mistakenly thought that the issue (https://github.com/pokt-network/pocket/issues/239) mentioned in response to this https://github.com/pokt-network/pocket/pull/235#discussion_r975789014 was a form of de-scoping from the current issue and instead was to track via automation if we have them.

My bad, sorry guys. 

Also I have noticed that it should really be (removed quotes for exclude-dir):
```bash
grep --exclude-dir=persistence -r "github.com/pokt-network/pocket/persistence"
grep --exclude-dir=consensus -r "github.com/pokt-network/pocket/consensus"
...
```

Makefile
```diff
@@ -383,3 +383,24 @@ gen_genesis_and_config:
 ## Clear the genesis and config files for LocalNet
 clear_genesis_and_config:
 	rm build/config/gen.*.json
+
+.PHONY: check_cross_module_imports
```
Olshansk, line 387: No-op.

In @okdas's check on linters, we have the ability to create custom linters in go: https://github.com/pokt-network/pocket/pull/190

I think we might be able to port this there.

deblasis:
I would like to know what's the plan of action with the current cross-module imports. 
Not sure if they are in scope for this PR since there is also other/older TECHDEBT in the mix.

I have updated/improved `make check_cross_module_imports` in https://github.com/pokt-network/pocket/pull/235/commits/72f6c90272ac9f901450fb077e06020c04c6fd98 to ignore stuff that's obviously dependent on these packages. 
Things like the `Makefile` itself, `shared`, `app`, `runtime`.

Result:

image

I am sure @okdas can make it even better of point me to further improvements 🙏

Olshansk:
@deblasis I'm going to do another look through this PR from start-to-finish and prioritize it for the beginning of this week.

> I would like to know what's the plan of action with the current cross-module imports.
> Not sure if they are in scope for this PR since there is also other/older TECHDEBT in the mix.

I believe this PR was a net improvement in terms of configuration management but we have cross-module imports both before and after.

If there are any "extremely offensive" cross-module imports, we should remove them here, but otherwise we should create a followup ticket and enable automation to prevent this once they're all gone.

@andrewnguyen22 After another round of reviews from both of us, does that sound like a reasonable plan?

consensus/types/consensus_genesis.go
```diff
@@ -0,0 +1,18 @@
...
+
+import "github.com/pokt-network/pocket/shared/modules"
+
+var _ modules.ConsensusGenesisState = &ConsensusGenesisState{}
+
+func (x *ConsensusGenesisState) GetVals() []modules.Actor {
+	return ActorsToActorsInterface(x.GetValidators())
+}
+
+func ActorsToActorsInterface(a []*Validator) (actorI []modules.Actor) {
```
Olshansk, line 11: s/a/vals

consensus/types/converters.go
```diff
@@ -0,0 +1,20 @@
...
+		GenericParam: actor.GetGenericParam(),
+	}
+}
+
+func ToConsensusValidators(actors []modules.Actor) []*Validator {
+	r := make([]*Validator, 0)
+	for _, a := range actors {
+		r = append(r, actorToValidator(a))
+	}
+	return r
```
Olshansk, line 19:
```suggestion
	vals = make([]*Validator, len(actors))
	for i, actor := range actors {
		vals[i] = actor
	}
	return
```

p2p/module.go
```diff
@@ -28,7 +27,7 @@ const (
 
 type p2pModule struct {
 	bus       modules.Bus
-	p2pConfig modules.P2PConfig // TODO (olshansky): to remove this since it'll be available via the bus
+	p2pConfig *typesP2P.P2PConfig // TODO (olshansky): to remove this since it'll be available via the bus
```
Olshansk, line 30: In consensus, we renamed `consConfig` to `config` but get `p2pConfig` here. 

My preference is to keep explicit naming for the explicit struct (p2pConfig), but moreso to just be consistent between the two regardless of the approach.

persistence/types/converters.go
```diff
@@ -0,0 +1,151 @@
...
+	r := make([]*Actor, 0)
+	for _, a := range actors {
+		r = append(r, toPersistenceActor(a))
+	}
+	return r
+}
```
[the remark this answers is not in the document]
Olshansk: I think if you search for `, 0)` in another couple of places, the same approach can be adopted.

telemetry/prometheus_module.go
Olshansk: Need to update this one as well

utility/test/module_test.go
```diff
@@ -67,58 +65,27 @@ func NewTestingUtilityContext(t *testing.T, height int64) utility.UtilityContext
...
+
+	genesisState, _ := test_artifacts.NewGenesisState(5, 1, 1, 1)
+	mockRuntimeMgr.EXPECT().GetGenesis().Return(genesisState).AnyTimes()
+
+	persistenceMod, err := persistence.Create(mockRuntimeMgr)
+	require.NoError(t, err)
+
+	err = persistenceMod.Start()
+	require.NoError(t, err)
+	return persistenceMod.(modules.PersistenceModule)
```
Olshansk, line 90:
```suggestion

	return persistenceMod.(modules.PersistenceModule)
```

Olshansk (changes requested):

Makefile
```diff
@@ -392,3 +392,24 @@ gen_genesis_and_config:
...
+## Lists cross-module imports
+check_cross_module_imports:
```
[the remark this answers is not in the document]
deblasis: I previously tried to do it with eval, I spent like 5 mins on it, I "dropped the ball" and moved on to something else.

Now with renewed motivation it worked immediately, the culprit is always in *escaping* quotes 🤣

consensus/types/consensus_genesis.go
```diff
@@ -0,0 +1,18 @@
+package types
+
+import "github.com/pokt-network/pocket/shared/modules"
+
+var _ modules.ConsensusGenesisState = &ConsensusGenesisState{}
+
+func (x *ConsensusGenesisState) GetVals() []modules.Actor {
```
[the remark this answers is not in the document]
deblasis: We can't because there's already a `func (x *ConsensusGenesisState) GetValidators() []*Validator `

shared/README.md
```diff
@@ -54,11 +55,11 @@ flowchart TD
         subgraph Pocket's Application Specific Bus
             B("Bus")
             E("Main Events Channel")
-            Clock("Clock")
             B <-.-> E
-            B <-.-> Clock
         end
         subgraph Pocket's Core Modules
+            RMGR(RuntimeMgr)
```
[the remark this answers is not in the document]
deblasis: I don't think it's relevant at this level, sounds like an implementation detail. 
It would make the diagram quite "busy" IMHO, also I could argue that `config` and `genesis` are probably more important than clock in the context of `runtimeMgr`

shared/modules/types.go
```diff
@@ -72,6 +73,7 @@ type ConsensusGenesisState interface {
 	GetGenesisTime() *timestamppb.Timestamp
 	GetChainId() string
 	GetMaxBlockBytes() uint64
+	GetVals() []Actor
```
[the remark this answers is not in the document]
deblasis: Can't do. There's already a GetValidator

shared/modules/node_module.go
```diff
@@ -0,0 +1,8 @@
+package modules
```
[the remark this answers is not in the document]
deblasis: The idea was something we could mock later on. since I see I have embedded `P2PAddressableModule` in there, I think I was thinking about P2P at the time of writing.

shared/node.go
```diff
@@ -16,56 +17,69 @@ import (
...
 )
 
-var _ modules.Module = &Node{}
+var _ modules.NodeModule = &Node{}
 
 const (
 	MainModuleName = "main"
 )
 
 type Node struct {
```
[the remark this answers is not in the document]
deblasis: https://github.com/pokt-network/pocket/pull/235#discussion_r986192819

p2p/module_raintree_test.go
```diff
@@ -371,85 +367,75 @@ func prepareConnMock(t *testing.T, expectedNumNetworkReads, expectedNumNetworkWr
...
-				PrivateKey:            valKeys[i].String(),
-				ConsensusPort:         8080,
-				UseRainTree:           true,
-				IsEmptyConnectionType: true,
-			},
-		}
+	mockGenesisState := createMockGenesisState(t, valKeys)
+	for i := range mockRuntimeMgrs {
+		mockConfig := modulesMock.NewMockConfig(ctrl)
+		mockConfig.EXPECT().GetBaseConfig().Return(&runtime.BaseConfig{
```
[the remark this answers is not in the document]
deblasis: I think the scope of this PR has been stretched already quite a bit, it's going to become harder and harder to maintain if we keep adding stuff :)

Nothing is overkill but time kills

consensus/module.go
Olshansk: Nice!

Olshansk:
The only blocker from my end is fixing the unit tests. Some of it might be resolved after we merge in #282, so let's aim to get that in first.

Screen Shot 2022-10-06 at 10 02 35 AM

----

I took another look through the PR. With the goal of unblocking downstream work. the 3 biggest action items remaining IMO are (see screenshots below):

1. Consider renaming `NodeModules`
2. Figuring out how to implement `ValidateConfig` while allowing injection of mocks
3. Renaming `GetVals` to `GetValidators` because `Vals` can be misinterpreted

@deblasis Can you create a "235 follow ups" to track it?

@andrewnguyen22 Assuming we merge in #282 and create the follow-up for that work, can we merge this in?

---

Screen Shot 2022-10-06 at 10 00 49 AM
Screen Shot 2022-10-06 at 10 00 35 AM
Screen Shot 2022-10-06 at 10 00 27 AM

andrewnguyen22:
I will review it again @Olshansk

deblasis:
I have created the issues to track the open items as requested. They are not assigned to me and I put them in QoL, please change that as you see fit.

> The only blocker from my end is fixing the unit tests. Some of it might be resolved after we merge in #282, so let's aim to get that in first.

I have to point out that tests are not failing in this PR... what's failing is the CI because of the new linting: 

image

and also for the build / push part:

image

Olshansk (approved):
Thanks for creating the follow up work.

👍  from my POV. Not perfect but I was to prevent the scope of the PR from increasing and it's already a big improvement over what we had.

Let's wait to hear if @andrewnguyen22 has any blockers though

consensus/module.go
```diff
@@ -75,48 +73,53 @@ type ConsensusModule struct {
...
+	paceMaker := paceMakerMod.(Pacemaker)
+
+	m = &consensusModule{
 		bus: nil,
 
 		privateKey:  privateKey.(cryptoPocket.Ed25519PrivateKey),
-		consCfg:     cfg,
-		consGenesis: genesis,
+		consCfg:     cfg.GetConsensusConfig().(*typesCons.ConsensusConfig),
+		consGenesis: genesis.GetConsensusGenesisState().(*typesCons.ConsensusGenesisState),
```
andrewnguyen22, line 122: Non-Blocker: 

Change this to safe catch error:

```

casted, ok := cfg.GetConsensusConfig().(*typesCons.ConsensusConfig)
if !ok {
  // handle error
}
```

consensus/hotstuff_leader.go
```diff
@@ -306,10 +315,11 @@ func (m *ConsensusModule) validatePartialSignature(msg *typesCons.HotstuffMessag
 }
 
 // TODO: This is just a placeholder at the moment for indexing hotstuff messages ONLY.
-//       It doesn't actually work because SizeOf returns the size of the map pointer,
-//       and does not recursively determine the size of all the underlying elements
-//       Add proper tests and implementation once the mempool is implemented.
-func (m *ConsensusModule) tempIndexHotstuffMessage(msg *typesCons.HotstuffMessage) {
+//
+//	It doesn't actually work because SizeOf returns the size of the map pointer,
```
andrewnguyen22, line 319: Non-Blocker: 

Revert, Goland doesn't recognize this as part of the TODO anymore... Maybe that was the intention

consensus/module.go
andrewnguyen22: Same with LeaderElectionModule casting below (and all other casting)

consensus/module.go
```diff
@@ -205,28 +175,48 @@ func (m *ConsensusModule) Start() error {
...
+func (*consensusModule) ValidateConfig(cfg modules.Config) error {
+	// DISCUSS (team): we cannot cast if we want to use mocks and rely on interfaces
+	// if _, ok := cfg.GetConsensusConfig().(*typesCons.ConsensusConfig); !ok {
+	// 	return fmt.Errorf("cannot cast to ConsensusConfig")
+	// }
+	return nil
+}
+
+func (*consensusModule) ValidateGenesis(genesis modules.GenesisState) error {
+	// DISCUSS (team): we cannot cast if we want to use mocks and rely on interfaces
```
andrewnguyen22, line 208: Exactly, see comment above

consensus/module.go
```diff
@@ -75,48 +73,53 @@ type ConsensusModule struct {
...
-	m := &ConsensusModule{
+	paceMaker := paceMakerMod.(Pacemaker)
+
+	m = &consensusModule{
 		bus: nil,
 
 		privateKey:  privateKey.(cryptoPocket.Ed25519PrivateKey),
-		consCfg:     cfg,
-		consGenesis: genesis,
+		consCfg:     cfg.GetConsensusConfig().(*typesCons.ConsensusConfig),
```
andrewnguyen22, line 121: Blocker:

We should not cast here because this prevents us from using mocks etc. 

We should use interfaces only

p2p/module.go
```diff
@@ -27,8 +26,8 @@ const (
 )
 
 type p2pModule struct {
-	bus       modules.Bus
-	p2pConfig modules.P2PConfig // TODO (olshansky): to remove this since it'll be available via the bus
+	bus    modules.Bus
+	p2pCfg *typesP2P.P2PConfig // TODO (olshansky): to remove this since it'll be available via the bus
```
andrewnguyen22, line 30: Blocker:

Another example of the casting, this should be `modules.P2PConfig`

p2p/module.go
```diff
@@ -119,12 +100,12 @@ func (m *p2pModule) Start() error {
...
 		)
 
-	addrBook, err := ValidatorMapToAddrBook(m.p2pConfig, m.bus.GetConsensusModule().ValidatorMap())
+	addrBook, err := ValidatorMapToAddrBook(m.p2pCfg, m.bus.GetConsensusModule().ValidatorMap())
 	if err != nil {
 		return err
 	}
 
-	if m.p2pConfig.GetUseRainTree() {
+	if m.p2pCfg.UseRainTree {
```
andrewnguyen22, line 108: Blocker:

Revert once using interface

p2p/module_raintree_test.go
```diff
@@ -367,85 +363,75 @@ func prepareConnMock(t *testing.T, expectedNumNetworkReads, expectedNumNetworkWr
...
-
-	p2pFileBz, err := json.MarshalIndent(configFile, "", "    ")
-	require.NoError(t, err)
-	require.NoError(t, ioutil.WriteFile(testingGenesisFilePath+jsonPosfix, genesisFileBz, 0777))
-	require.NoError(t, ioutil.WriteFile(testingConfigFilePath+strconv.Itoa(n)+jsonPosfix, p2pFileBz, 0777))
-}
-
-func createConfigs(t *testing.T, numValidators int) (configs []modules.Config, genesisState modules.GenesisState) {
-	configs = make([]modules.Config, numValidators)
+func createMockRuntimeMgrs(t *testing.T, numValidators int) []modules.RuntimeMgr {
```
andrewnguyen22, line 376: Blocker:

Please add comment around any `Mock` creation, see the other mocks in this file for example around expected behaviors

p2p/module_raintree_test.go
```diff
@@ -367,85 +363,75 @@ func prepareConnMock(t *testing.T, expectedNumNetworkReads, expectedNumNetworkWr
...
-	return
+	return mockRuntimeMgrs
 }
 
 func validatorId(_ *testing.T, i int) string {
 	return fmt.Sprintf(serviceUrlFormat, i)
 }
 
-func createGenesisState(t *testing.T, valKeys []cryptoPocket.PrivateKey) modules.GenesisState {
+func createMockGenesisState(t *testing.T, valKeys []cryptoPocket.PrivateKey) modules.GenesisState {
```
andrewnguyen22, line 407: Blocker:

Please add comment around any Mock creation, see the other mocks in this file for example around expected behaviors

runtime/manager.go
```diff
@@ -0,0 +1,174 @@
...
+	mgr.genesis = genesis
+
+	for _, o := range options {
+		o(mgr)
+	}
+
+	return mgr
+}
+
+func NewManagerFromReaders(configReader, genesisReader io.Reader, options ...func(*Manager)) *Manager {
```
andrewnguyen22, line 48: Blocker:

No usage function with no comments or tests. Please add one of the two

shared/modules/node_module.go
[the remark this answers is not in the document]
andrewnguyen22: Blocker:

We need comments here to describe this thing.

I agree with Olshansk

> I don't fully understand the logic behind this interface. Can you expand on the thought process? I think we might need to remove it.

shared/modules/module.go
```diff
@@ -21,6 +24,21 @@ type InterruptableModule interface {
 
 type InitializableModule interface {
 	GetModuleName() string
-	InitConfig(pathToConfigJSON string) (IConfig, error)
-	InitGenesis(pathToGenesisJSON string) (IGenesis, error)
+	Create(runtime RuntimeMgr) (Module, error)
```
andrewnguyen22, line 27: Non-Blocker:

The granularity here is getting a bit sticky for me. Consider mental load vs. verbosity

shared/test_artifacts/generator.go
```diff
@@ -34,28 +40,31 @@ var (
...
-	fish, fishPrivateKeys := NewActors(MockActorType_Fish, numFisherman)
-	return modules.GenesisState{
-		ConsensusGenesisState: &MockConsensusGenesisState{
+func NewGenesisState(numValidators, numServiceNodes, numApplications, numFisherman int) (modules.GenesisState, []string) {
+	apps, appsPrivateKeys := NewActors(types.ActorType_App, numApplications)
+	vals, validatorPrivateKeys := NewActors(types.ActorType_Validator, numValidators)
+	serviceNodes, snPrivateKeys := NewActors(types.ActorType_ServiceNode, numServiceNodes)
+	fish, fishPrivateKeys := NewActors(types.ActorType_Fisherman, numFisherman)
+
+	genesisState := runtime.NewGenesis(
```
andrewnguyen22, line 49: Non-Blocker:

Is there no way to use `runtime` for this?

shared/test_artifacts/gov.go
```diff
@@ -1,9 +1,11 @@
 package test_artifacts
```
andrewnguyen22, line 1: Non-Blocker:

Can we move this to the runtime package?

telemetry/noop_module.go
```diff
@@ -62,6 +59,14 @@ func (m *NoopTelemetryModule) GetBus() modules.Bus {
 	return m.bus
 }
 
+func (*NoopTelemetryModule) ValidateConfig(cfg modules.Config) error {
+	// DISCUSS (team): we cannot cast if we want to use mocks and rely on interfaces
```
andrewnguyen22, line 63: Agree, see applicable comment above

utility/module.go
```diff
@@ -1,78 +1,75 @@
...
-func (u *UtilityModule) GetBus() modules.Bus {
+func (u *utilityModule) GetBus() modules.Bus {
 	if u.bus == nil {
 		log.Fatalf("Bus is not initialized")
 	}
 	return u.bus
 }
+
+func (*utilityModule) ValidateConfig(cfg modules.Config) error {
+	// DISCUSS (team): we cannot cast if we want to use mocks and rely on interfaces
```
andrewnguyen22, line 70: Agree see applicable comment above

andrewnguyen22 (changes requested):
Pretty much there, a few blockers and a few discussion items. 

Resolve conflicts and blockers and I'll approve

runtime/manager.go
deblasis, line 48: We could potentially remove this. I have used during development and forgot to clean after myself

consensus/module.go
deblasis: We are now using interfaces and I think I found a solution for the problem: "What if there are Consensus modules that have a Pacemaker and others that don't?" (and similar use-cases).

consensus/module.go
deblasis: Maybe addressed with https://github.com/pokt-network/pocket/pull/235#discussion_r993986279

consensus/module.go
deblasis: Maybe addressed with https://github.com/pokt-network/pocket/pull/235#discussion_r993986279

consensus/hotstuff_leader.go
deblasis: Done, VSCode formats files on save automatically, I need to configure it to mimic Goland I suppose. I'll look into it

p2p/module.go
deblasis: Maybe addressed with https://github.com/pokt-network/pocket/pull/235#discussion_r993986279

p2p/module.go
deblasis: Maybe addressed with https://github.com/pokt-network/pocket/pull/235#discussion_r993986279

shared/test_artifacts/gov.go
deblasis: Done

shared/test_artifacts/generator.go
deblasis: If you mean to move test_artifacts under runtime. I have done it.

telemetry/noop_module.go
deblasis: Maybe addressed with https://github.com/pokt-network/pocket/pull/235#discussion_r993986279

utility/module.go
deblasis: Maybe addressed with https://github.com/pokt-network/pocket/pull/235#discussion_r993986279

shared/modules/module.go
deblasis: I see. If it sounds like over-optimization or premature optimization let's consider the fact that soon we'll have modules that are very simple and only implement a subset of the Module interface. Logging for example, which is coming soon. Keybase as well I guess.

Maybe my .Net background surfaces here. I am used to very small and targeted interfaces.

I am open to suggestions.

deblasis:
@andrewnguyen22 thank you for your time and feedback reviewing this  :) 

I have addressed all the comments that I have marked with a 👍 and left comments otherwise. There are a couple of questions from me but nothing major, I guess.

Regarding `NodeModule` I think that since both of you flagged it, its short life had to come to an end. I have axed it and replaced with the struct. Perhaps I was over-abstracting.

andrewnguyen22 (approved):
Great work @deblasis

PR #236, opened 2022-09-21
[TECHDEBT] Utility Module, remove `DELETE` *actor* 'No-op' code

### LINKED ISSUE #226: [TECHDEBT] Utility Module, remove `DELETE` *actor* 'No-op' code
## Objective

Remove all `Delete Actor` related 'no-op' code.

## Origin Document

In the world of blockchain, nothing is ever truly deleted from the `world state`. In this case, the `Delete` operation is essentially a `no-op` because the structure always exists in history. 

Since it's clear that we don't have a use for `Delete` currently, the code should be removed until we have sufficient reason to have it.

## Goals

- Reduce code footprint 
- Reduce cognitive load
- 
## Deliverable

- Code complete implementation
- Updated CHANGELOG documentation

## General issue deliverables

---

**Creator**: @andrewnguyen22

andrewnguyen22:
## Description

In the world of blockchain, nothing is ever truly deleted from the `world state`. In this case, the `Delete` operation is essentially a `no-op` because the structure always exists in history. 

Since it's clear that we don't have a use for `Delete` currently, the code should be removed until we have sufficient reason to have it.

## Issue

Fixes #226 

## Type of change

- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other: <REPLACE_ME>

## List of changes

- Remove all `Delete Actor` related 'no-op' code.

## Testing

- [x] `make test_all`

## Checklist

- [ ] If applicable, I have made corresponding changes to related local or global README
- [ ] If applicable, I have added new diagrams using [mermaid.js](https://mermaid-js.github.io)


Olshansk (changes requested):
Can you also remove:

```go
func ErrDelete(err error) Error {
	return NewError(CodeDeleteError, fmt.Sprintf("%s: %s", DeleteError, err.Error()))
}
```

And the corresponding `CodeDeleteError` and `DeleteError`

Olshansk:
I cleaned up https://github.com/pokt-network/pocket/pull/238 off of this.

utility/types/error.go
```diff
@@ -34,66 +34,66 @@ func NewError(code Code, msg string) Error {
...
+	CodeInvalidStatusError               Code = 54
+	CodeAddPoolAmountError               Code = 55
+	CodeSubPoolAmountError               Code = 56
+	CodeGetStatusError                   Code = 57
+	CodeSetUnstakingHeightAndStatusError Code = 58
+	CodeGetReadyToUnstakeError           Code = 59
+	CodeAlreadyExistsError               Code = 60
+	CodeGetExistsError                   Code = 61
+	CodeGetLatestHeightError             Code = 62
+
```
Olshansk, line 96: Should we care about the missing "63" which is deprecated even before we move to production?

I know it's now a big issue, but I think we should either:
1. Switch to using iota. We don't have to worry about keeping numbers but it would have to be changed once we mo e to prod.

2. Leave a comment or change the code to `Deprecated_CodeDeleteError Code = 63`. This is what I've seen in production at larger companies that use protobuf enums.

3. Just curious if you think this even matters at all?

Olshansk (approved):

PR #240, opened 2022-09-23
[One line change] Fix order in `make develop_test`

Olshansk:
Fix the order of `clean_deps` and `mockgen` in `make develop_test`


andrewnguyen22:
Does this prevent you from deleting the `vendor` folder?

Olshansk:
Combined with the change that @deblasis submitted recently regarding mock generation, it should.

Olshansk:
@andrewnguyen22 Along with Alessandro's change for mock generation, it should.

deblasis (approved):
Tested after switching back and forth between branches and it seems to be working

PR #250, opened 2022-09-26
[TECHDEBT][Consensus] Consensus module techdebt consolidation

### LINKED ISSUE #249: [Consensus][TECHDEBT] Consensus module techdebt consolidation
## Objective

Consolidate, remove and tend to minor TODOs across the consensus module.

## Origin Document

A lot of learnings and changes have been made to our process since the first implementation of the consensus module, so there is room to do some general 🧹 at a module-level perspective before we change or improve the business logic.

## Goals

- Prepare the consensus module for a proper consensus <-> utility <-> persistence state hash computation
- Remove unneeded TODOs
- Consolidate repetitive TODOs

## Deliverable

- [ ] A PR tending to the goals above only touching the consensus module (not pacemaker, persistence, utility etc...)
- [ ] A small sequence diagram showing cross-module context initialization

- Changing any business logic in the consensus module
- Changing cross-module interaction in the consensus module
- Changes tending to major TODOs across the module

## General issue deliverables

- [ ] Update the source code tree explanation
- [ ] Updated sequence diagram using [mermaid](https://mermaid-js.github.io/mermaid/) for basic components

## Testing Methodology

---

**Creator**: @Olshansk

Olshansk:
## Description

Consolidate & update techdebt in the consensus module (excluding the pacemaker module which is being modified by #198) in preparation for an end-to-end state hash flow explored in #237.

## Issue

Fixes #249

## Type of change

- [X] Code health or cleanup
- [ ] Major breaking change
- [X] Documentation
- [ ] Other: NA

## List of changes

**Consensus logic**
- Pass in a list of messages to `findHighQC` instead of a hotstuff step
- Made `CreateProposeMessage` and `CreateVotemessage` accept explicit values, identifying some bugs along the way
- Made sure to call `applyBlock` when using `highQC` from previous round
- Moved business logic for `prepareAndApplyBlock` into `hotstuff_leader.go`
- Removed `MaxBlockBytes` and storing the consensus genesis type locally as is

**Consensus cleanup**
- Using appropriate getters for protocol types in the hotstuff lifecycle
- Replaced `proto.Marshal` with `codec.GetCodec().Marshal`
- Reorganized and cleaned up the code in `consensus/block.go`
- Consolidated & removed a few `TODO`s throughout the consensus module
- Added TECHDEBT and TODOs that will be require for a real block lifecycle
- Fixed typo in `hotstuff_types.proto`
- Moved the hotstuff handler interface to `consensus/hotstuff_handler.go` 

**Consensus testing**
- Improved mock module initialization in `consensus/consensus_tests/utils_test.go`

**General**
- Added a diagram for `AppHash` related `ContextInitialization`
- Added `Makefile` keywords for `TODO`

## Testing

- [x] `make test_all`

## Checklist

- [ ] If applicable, I have made corresponding changes to related local or global README


shared/docs/flows/README.md
```diff
@@ -0,0 +1,3 @@
+# Flows
```
deblasis, line 1: I think this is a great addition and we can put in here more flows as well. We just need to keep this stuff updated.

On the latter, not sure if we should enforce this via PR reviews or if we should add to the PR template. 
The current wording *could* include this already but a new contributor wouldn't know where to look at.

```markdown
- [ ] If applicable, I have added new diagrams using [mermaid.js](https://mermaid-js.github.io)
```

consensus/hotstuff_handler.go
```diff
@@ -0,0 +1,47 @@
...
+		// and continues to receive NewRound messages, we avoid logging the "message discard"
+		// because it creates unnecessary spam.
+		if !(m.LeaderId != nil && !m.isLeader() && step == NewRound) {
+			m.nodeLog(typesCons.WarnDiscardHotstuffMessage(msg, err.Error()))
+		}
+		return err
+	}
+
+	// Leader Election - Need to execute leader election if there is no leader and we are in a new round.
+	if m.Step == NewRound && m.LeaderId == nil {
```
deblasis, line 33: I feel drawn to the idea of codifying this condition into its own function.

Something like `m.shouldElectNextLeader() bool`

consensus/module.go
```diff
@@ -27,42 +27,43 @@ var _ modules.PacemakerConfig = &typesCons.PacemakerConfig{}
...
+	IdToValAddrMap typesCons.IdToValAddrMap // TODO: This needs to be updated every time the ValMap is modified
 
 	// Consensus State
-	appHash      string
+	lastAppHash  string // TODO: Always retrieve this variable from the persistence module and simplify this struct
 	validatorMap typesCons.ValidatorMap
 
 	// Module Dependencies
-	utilityContext    modules.UtilityContext
+	UtilityContext    modules.UtilityContext
```
deblasis, line 58: Why is this public now? As far as I can tell it's only used within consensus

consensus/messages.go
```diff
@@ -42,54 +41,60 @@ func CreateProposeMessage(
...
+		log.Printf("[WARN] error signing message: %v\n", err)
 		return nil
 	}
+
 	return signature
 }
 
-// Signature should only be over a subset of the fields in a HotstuffMessage
-func getSignableBytes(m *typesCons.HotstuffMessage) ([]byte, error) {
+// Signature only over subset of fields in HotstuffMessage
```
deblasis, line 91: noop: I am sure there's a reason but I am gonna ask why offline. 

What happens if I am a node that signs and sends different blocks with different justifications?

consensus/messages.go
```diff
@@ -4,35 +4,34 @@ import (
 	"log"
 
 	typesCons "github.com/pokt-network/pocket/consensus/types"
+	"github.com/pokt-network/pocket/shared/codec"
 	"github.com/pokt-network/pocket/shared/crypto"
-	"google.golang.org/protobuf/proto"
 )
 
 func CreateProposeMessage(
```
deblasis, line 11: noop: nice not depending on consensusModule anymore for these

consensus/messages.go
```diff
@@ -4,35 +4,34 @@ import (
...
+	round uint64,
+	step typesCons.HotstuffStep,
+	block *typesCons.Block,
 	qc *typesCons.QuorumCertificate,
 ) (*typesCons.HotstuffMessage, error) {
-	if m.Block == nil {
-		return nil, typesCons.ErrNilBlockProposal
-	}
-
 	msg := &typesCons.HotstuffMessage{
```
deblasis, line 18: shouldn't we still guard for `block == nil` like before?

```go
	if block == nil {
		return nil, typesCons.ErrNilBlockProposal
	}
```

consensus/hotstuff_replica.go
```diff
@@ -25,45 +25,52 @@ var (
 
 func (handler *HotstuffReplicaMessageHandler) HandleNewRoundMessage(m *ConsensusModule, msg *typesCons.HotstuffMessage) {
 	handler.emitTelemetryEvent(m, msg)
+	defer m.paceMaker.RestartTimer()
```
deblasis, line 28: I feel a bit battled about inverting the order of these two lines (and below as well).

consensus/hotstuff_replica.go
```diff
@@ -195,20 +208,40 @@ func (m *ConsensusModule) validateProposal(msg *typesCons.HotstuffMessage) error
...
+	// The equivalent of `lockedQC.Block.ExtendsFrom(justifyQC.Block)` in the hotstuff whitepaper is done in `applyBlock` below.
+	if protoHash(lockedQC.GetBlock()) == protoHash(justifyQC.Block) {
 		m.nodeLog(typesCons.ProposalBlockExtends)
 		return nil
 	}
 
-	// Liveness: node is locked on a QC from the past. [TODO]: Do we want to set `m.LockedQC = nil` here or something else?
+	// Liveness: node is locked on a QC from the past.
+	// DISCUSS: Where should additional logic be added to unlock the node?
 	if justifyQC.Height > lockedQC.Height || (justifyQC.Height == lockedQC.Height && justifyQC.Round > lockedQC.Round) {
```
deblasis, line 219: noop: just an idea

What if we had something like:

```go
if nodeIsLockedOnPastQC(justifyQC, lockedQC) {
```

and then decompose `nodeIsLockedOnPastQC` further into functions that return the logical pieces we need. I think this is valid especially if we are considering the idea of adding additional logic here that could make this `if` very long and hard to understand IMHO.

consensus/hotstuff_replica.go
```diff
@@ -25,45 +25,52 @@ var (
...
 	if err := handler.anteHandle(m, msg); err != nil {
 		m.nodeLogError(typesCons.ErrHotstuffValidation.Error(), err)
 		return
 	}
-	// TODO(olshansky): add step specific validation
-	m.paceMaker.RestartTimer()
+
+	// Clear the previous utility context, if it exists, and create a new one
+	if err := m.refreshUtilityContext(); err != nil {
+		return
```
deblasis, line 37: should we log the `err` here as well? Otherwise we'd simply return with no memory of what went wrong

consensus/hotstuff_leader.go
```diff
@@ -18,48 +21,59 @@ var (
...
 		m.nodeLog(typesCons.OptimisticVoteCountWaiting(NewRound, err.Error()))
 		return
 	}
-
-	// TODO(olshansky): Do we need to pause for `MinBlockFreqMSec` here to let more transactions come in?
 	m.nodeLog(typesCons.OptimisticVoteCountPassed(NewRound))
 
+	// Clear the previous utility context, if it exists, and create a new one
+	if err := m.refreshUtilityContext(); err != nil {
+		return
```
deblasis, line 45: should we log the err here as well? Otherwise we'd simply return with no memory of what went wrong

consensus/hotstuff_leader.go
```diff
@@ -18,48 +21,59 @@ var (
 	}
 )
 
-type HotstuffLeaderMessageHandler struct{}
-
 /*** Prepare Step ***/
 
 func (handler *HotstuffLeaderMessageHandler) HandleNewRoundMessage(m *ConsensusModule, msg *typesCons.HotstuffMessage) {
 	handler.emitTelemetryEvent(m, msg)
+	defer m.paceMaker.RestartTimer()
```
deblasis, line 28: I feel a bit battled about inverting the order of these two lines (and below as well).

consensus/helpers.go
```diff
@@ -1,58 +1,58 @@
...
 	var pss []*typesCons.PartialSignature
 	for _, msg := range m.MessagePool[step] {
-		// TODO(olshansky): Add tests for this
 		if msg.GetPartialSignature() == nil {
 			m.nodeLog(typesCons.WarnMissingPartialSig(msg))
 			continue
 		}
-		// TODO(olshansky): Add tests for this
-		if msg.Height != height || msg.Step != step || msg.Round != round {
+		if msg.GetHeight() != height || msg.GetStep() != step || msg.GetRound() != round {
```
deblasis, line 50: noop: again, I would love a function here that returns a bool.

I am about to go on a tangent... I'd love your thoughts on this.

The more I see these inside the consensus module and the more I come to understand/believe that we should probably have some sort of "state machine" that given the current "state" exposes some validation logic that could be made quite generic.

This state machine or `consensusMessageValidatorThinghy`™️ would be injected in the consensus module and the validation itself would be an interface method for which the inputs would be the hotstuff message and the current "state" tuple made of height, step, round and whatever else is needed to perform these validations.

I guess this would be an issue on itself.

shared/docs/flows/AppHash.md
```diff
@@ -0,0 +1,36 @@
...
+    end
+    Note over N, P2P: Hotstuff lifecycle
+    N-->>C: HandleMessage(anypb.Any)
+    critical Decide Message
+        Note over C, PM: See 'Block Commit'
+    end
+```
+
+## Block Application
+
```
deblasis, line 35: Should we add a TODO here and below? I know it's kinda obvious but since we leverage `make todo_list` I think it would be nice to have a reminder

consensus/module.go
```diff
@@ -27,42 +27,43 @@ var _ modules.PacemakerConfig = &typesCons.PacemakerConfig{}
...
-	logPrefix   string                                                  // TODO(design): Remove later when we build a shared/proper/injected logger
-	MessagePool map[typesCons.HotstuffStep][]*typesCons.HotstuffMessage // TODO(design): Move this over to the persistence module or elsewhere?
-	// TODO(andrew): Explain (or remove) why have an explicit `MaxBlockBytes` if we are already storing a reference to `consCfg` above?
-	// TODO(andrew): This needs to be updated every time the utility module changes this value. It can be accessed via the "application specific bus" (mimicking the intermodule interface in ABCI)
-	MaxBlockBytes uint64
+	// DEPRECATE: Remove later when we build a shared/proper/injected logger
+	logPrefix string
+
+	// TECHDEBT: Move this over to use the txIndexer
+	MessagePool map[typesCons.HotstuffStep][]*typesCons.HotstuffMessage
```
deblasis, line 66: I think this should not be exposed outside of consensus as well but I understand we are sunsetting/moving it

consensus/hotstuff_replica.go
```diff
@@ -195,20 +208,40 @@ func (m *ConsensusModule) validateProposal(msg *typesCons.HotstuffMessage) error
...
+	// TECHDEBT: Retrieve this from persistence
+	lastByzValidators := make([][]byte, 0)
+
+	// Apply all the transactions in the block and get the appHash
+	appHash, err := m.UtilityContext.ApplyBlock(int64(m.Height), block.BlockHeader.ProposerAddress, block.Transactions, lastByzValidators)
+	if err != nil {
+		return err
+	}
+
+	// CONSOLIDATE: Terminology of `blockHash`, `appHash` and `stateHash`
```
deblasis, line 237: noop: 💯

consensus/hotstuff_replica.go
```diff
@@ -195,20 +208,40 @@ func (m *ConsensusModule) validateProposal(msg *typesCons.HotstuffMessage) error
...
 	return typesCons.ErrUnhandledProposalCase
 }
 
+// This helper applies the block metadata to the utility & persistence layers
+func (m *ConsensusModule) applyBlock(block *typesCons.Block) error {
+	// TECHDEBT: Retrieve this from persistence
+	lastByzValidators := make([][]byte, 0)
+
+	// Apply all the transactions in the block and get the appHash
+	appHash, err := m.UtilityContext.ApplyBlock(int64(m.Height), block.BlockHeader.ProposerAddress, block.Transactions, lastByzValidators)
```
deblasis, line 232: I have a couple of "problems"/questions with this and other instances where we access `m.UtilityContext`:

1. `m.UtilityContext` -could- be `nil` and we would simply panic here.

   I understand that we call `refreshUtilityContext` when we start a new round, I was wondering if we can just safely rely on that. 

2. If a replica receives a `Prepare` message that passes validations and the handler is invoked down the line, we would find ourselves here trying to apply a block but with which `m.UtilityContext`? What if it was started with a different height and what if is `nil` instead?

shared/docs/flows/README.md
Olshansk: Short-term solution: https://github.com/pokt-network/pocket/commit/245cb6e5481bed56bcbc2c5b1a0a2e9d0675314c

Long-term solution: https://github.com/pokt-network/pocket/issues/255

consensus/module.go
Olshansk: Because of the following line in `consensus/consensus_tests/pacemaker_test.go`

```go
consensusModImpl.FieldByName("UtilityContext").Set(reflect.ValueOf(utilityContext))
```

Reflect doesn't let us set the values of unexported fields and I'm utilizing the inner values of the consensus implementation to "pre-set it" to a certain state.

Some solutions to this involve using unsafe pointers, which I want to avoid, but the better solution is likely to either:
1. Move the tests into the appropriate package
2. Create an interface so we can mock the values of the consensus module

Going to think about this some more...

consensus/messages.go
Olshansk: Added this to the comment:
```
// For reference, see section 4.3 of the the hotstuff whitepaper, partial signatures are
// computed over `tsignr(hm.type, m.viewNumber , m.nodei)`. https://arxiv.org/pdf/1803.05069.pdf
```

Does this make more sense?
- Same block
- It's a partial signature by 1 node

consensus/messages.go
Olshansk: Yessir. Ty!

consensus/hotstuff_replica.go
Olshansk: Don't hesitate to bring stuff like this up. As a reader, you have a completely different view on it and this sort of things pays dividends.

If I had an explicit decision as to why I had a specific order, I'd push back, but I like this order better as well.

Plus, it took 10 seconds:

Screen Shot 2022-09-28 at 5 43 11 PM

shared/docs/flows/AppHash.md
Olshansk: Agreed!

consensus/hotstuff_replica.go
Olshansk: Checkout https://github.com/pokt-network/pocket/pull/250/commits/a8e5f3dae04d54da3745c502e14ab27801578f69

consensus/hotstuff_leader.go
Olshansk: Update all-the-things!

consensus/hotstuff_leader.go
```diff
@@ -18,48 +21,59 @@ var (
...
+	}
+
 	// Likely to be `nil` if blockchain is progressing well.
-	highPrepareQC := m.findHighQC(NewRound)
+	// TECHDEBT: How do we properly validate `highPrepareQC` here?
+	highPrepareQC := m.findHighQC(m.MessagePool[NewRound])
 
-	// TODO(olshansky): Add more unit tests for these checks...
+	// TODO: Add more unit tests for these checks...
 	if highPrepareQC == nil || highPrepareQC.Height < m.Height || highPrepareQC.Round < m.Round {
```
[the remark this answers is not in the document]
Olshansk: See https://github.com/pokt-network/pocket/pull/250/commits/8a41361cb93007896bfd8222782819bc415b2f37

consensus/module.go
Olshansk: I think this is going to be a non-trivial amount of work so I created #256.

consensus/hotstuff_replica.go
Olshansk: I see what you mean.

1. Re nil - I think we can easily check everywhere, but it's just not pretty. I'm holding off on doing that until we make a decision.

2. Regarding "race conditions" - we can definitely get into weird "stateful" situations where we have contexts and messages at different heights.

3. I'm thinking of modifying the UtilityModule interface to have a "InitializeContext(height)" or "PrepareContext(height)" functions to simplify everything and make it stateless. The business logic on block validation moves entirely to the utility context. However, I'd want to discuss that with @andrewnguyen22 prior.

Olshansk:
@deblasis I still didn't tend to all of your feedback, but please take another look didn't tend to all of your comments and the changes that I did make and the responses I left.

consensus/module.go
deblasis: I think this can be done by:
1. creating a `module_test.go` in consensus
2. inside of it have something like this:
```go
func (c *ConsensusModule) SetUtilityContext(utilityContext modules.UtilityContext) {
	c.utilityContext = utilityContext
}
```
3. in `pacemaker_test.go` use reflection to call the set method (that's available only during tests):

```go
consensusModImpl.MethodByName("SetUtilityContext").Call([]reflect.Value{reflect.ValueOf(utilityContext)})
```

this way we can keep the `utilityContext` private

consensus/messages.go
deblasis: Makes sense, thank you!

consensus/hotstuff_replica.go
deblasis: Since it requires some thinkering / discussion I'd say that for now it's just best to just track it in an issue and move forward.

deblasis:
> @deblasis I still didn't tend to all of your feedback, but please take another look didn't tend to all of your comments and the changes that I did make and the responses I left.

Nice one, most things are solved, I think there are only 2 AI:

- Create an issue / something to track this https://github.com/pokt-network/pocket/pull/250#discussion_r982505262
- Probably quick and easy fix for this: https://github.com/pokt-network/pocket/pull/250#discussion_r986088739

consensus/module.go
Olshansk: Haha, it worked!

I was thinking of doing the same thing for `Round`, `Height`, etc., but it feels wrong to create setters for all of them solely for testing purposes. 

Since the exposure of utility context was a step backwards, I definitely like this approach better, but for the other params, I want to keep the ticket open and figure out something better.

deblasis (approved):
LGTM

PR #253, opened 2022-09-27
[TECHDEBT] [P2P] Moved p2p telemetry metrics into telemetry module

deblasis:
Tending to the TODO item:

https://github.com/pokt-network/pocket/blob/5b958411177d1ae0a16da8354555d53fe6a7e306/p2p/telemetry/metrics.go#L3

it also renames the constants by making them have "P2P" as prefix so that constants from various modules can be easily identified while developing.

## Thoughts

Having metrics labels defined inside the `telemetry` module makes sense, however, that means that the `telemetry` module knows about some strings belonging to some other module that have some sort of meaning. `telemetry` is still pretty ignorant about the implementation details of the other modules but there's some interdependency. 
It also means that now `p2p` imports `telemetry`.

🤔

### DIFF
--- a/p2p/module.go
+++ b/p2p/module.go
@@ -9,14 +9,13 @@ import (
 	"io/ioutil"
 	"log"
 
-	"github.com/pokt-network/pocket/shared/debug"
-
 	"github.com/pokt-network/pocket/p2p/raintree"
 	"github.com/pokt-network/pocket/p2p/stdnetwork"
-	p2pTelemetry "github.com/pokt-network/pocket/p2p/telemetry"
 	typesP2P "github.com/pokt-network/pocket/p2p/types"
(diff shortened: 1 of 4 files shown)


Olshansk (approved):

Olshansk:
> Having metrics labels defined inside the telemetry module makes sense, however, that means that the telemetry module knows about some strings belonging to some other module that have some sort of meaning. telemetry is still pretty ignorant about the implementation details of the other modules but there's some interdependency.
It also means that now p2p imports telemetry.

I see your point. From my perspective, if someone were to implement a different p2p module, it would encourage the to reuse/update the existing telemetry metrics. That way:

1. Swap out the P2P module
2. Don't need to update any dashboard or telemetry ingusters 

Wdyt?

PR #281, opened 2022-10-04
[Consensus] LocalNet doesn't work properly: debug messages not processed

deblasis:
## Description

Localnet starts correctly but messages from the client are not handled. This PR fixes that.

### Explainer

We need the locks in the first place because the testing harness is simulating multiple nodes running at the same time and therefore we have race conditions (since goroutines are involved).

In the realm of LocalNet, where each node is its own environment with isolation (docker containers), when we handle a debug message, we don't have any contention because there's only one goroutine doing the work.

Specifically, the **deadlock** came from the fact that, for example, `printNodeState` internally calls `GetNodeState` which again locks the mutex.

## Type of change

- [x] Bug fix

## List of changes

- removed a misplaced `lock` that caused a deadlock.

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/consensus/debugging.go
+++ b/consensus/debugging.go
@@ -9,8 +9,6 @@ import (
 )
 
 func (m *ConsensusModule) HandleDebugMessage(debugMessage *debug.DebugMessage) error {
-	m.m.Lock()
-	defer m.m.Unlock()
 	switch debugMessage.Action {
 	case debug.DebugMessageAction_DEBUG_CONSENSUS_RESET_TO_GENESIS:
 		m.resetToGenesis(debugMessage)


Olshansk (commented):
I'm happy to approve this, but could you explain why this happens?

When we were adding the locks, I had several questions related to when one should or shouldn't use locks, and we had an issue here even though I thought it made sense to have a lock here.

Should we update the comment or PR description with more detail?

deblasis:
> I'm happy to approve this, but could you explain why this happens?
> 
> When we were adding the locks, I had several questions related to when one should or shouldn't use locks, and we had an issue here even though I thought it made sense to have a lock here.
> 
> Should we update the comment or PR description with more detail?

I have updated the description with my explanation:

We need the locks in the first place because the testing harness is simulating multiple nodes running at the same time and therefore we have race conditions (since goroutines are involved).

In the realm of LocalNet, where each node is its own environment with isolation (docker containers), when we handle a debug message, we don't have any contention because there's only one goroutine doing the work.

Specifically, the **deadlock** came from the fact that, for example, `printNodeState` internally calls `GetNodeState` which again locks the mutex.

Olshansk (approved):

Olshansk:
> because the testing harness is simulating multiple nodes running at the same time and therefore we have race conditions (since goroutines are involved).

Note to self: will keep in mind how we can improve this in the future. Not something we need to design for right now but likely in the future.

> Specifically, the deadlock came from the fact that, for example, printNodeState internally calls GetNodeState which again locks the mutex.

Got it. I think that once we have more automation of a Devnet or LocalNet via some sort of DSL, we'll be able to capture deadlocks preemptively

Issue #261, opened 2022-09-30
[Testing] Mock 10% Persistence Module in Utility Tests


andrewnguyen22:
## Objective

Mock the Persistence Module for all Utility Testing

## Origin Document

The Utility Module relies on the Persistence Module in order to execute all of it's testing:

See 
```golang
func newTestPersistenceModule(_ *testing.M, databaseUrl string) modules.PersistenceModule {
```

Under `utility/test`

## Goals

- Enabling utility tests for persistence business logic that doesn't exist
- Remove cross module imports in Utility
- Allow complete encapsulation of test suite in Utility
- Create a clear and understandable `Mock`

## Deliverable

- Code complete implementation
- Plenty of comments in the `Mock`
- Updated changelog

## Non-goals / Non-deliverables

- Little nits that have nothing to do with the issue

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)

## [Optional] Testing Methodology

- \__REPLACE_ME_: Make sure to update the testing methodology appropriately\_
- **Task specific tests**: `make ...`
- **All tests**: `make test_all`
- **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @andrewnguyen22

Olshansk:
Idea posted in #290 as a starting point for a discussion.

Olshansk:
I scrapped the first iteration of the idea (using a passthrough) and submitted a second one (using mocks only). Looking for preliminary feedback to see if it meets the requirements of the ticket before I put more effort into it

Olshansk:
See #290 as a reference for future iterations on this work

Olshansk:
We have been using the persistence module directly for utility tests, so am closing this out until we refactor the test suite.

PR #252, opened 2022-09-27
[Core] StateHash Interfaces & Diagrams

Olshansk:
## Description

Interface updates and diagrams to document the cross-module flow for stateHash implementation.

## Issue

Fixes part of #251

## Type of change

Please mark the relevant option(s):

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [x] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

The goal of this change is to show the interface changes and plan of action to implement the state hash implementation.

Documentation and external interface changes required for updating the state hash. Details related to the internals of the persistence module will be done in a follow up commit.

The major change is that the final `quorumCertificate` is only known at the time of committal, not proposal, so the corresponding changes were made as well as interface changes to make the terminology easier to follow with the corresponding documentation.

**General**
- Update some comments / TODOs throughout the code
- Handle errors where appropriate to support interface changes

**UtilityContext interface changes**
- Change `CommitPersistenceContext()` to `Commit(quorumCert)
- Change `ReleaseContext()` to `Release() error`

**PostgresContext interface changes**
- Removed `quorumCert` from the `SetProposalBlock` method signature
- Changed `Commit()` to `Commit(quorumCert)`
- Renamed `ResetContext` to `Release`
- Replace `AppHash` with `UpdateAppHash` 
- Added `ReleaseWriteContext` to the module level interface

**Persistence Module/Interface changes**
- Reduce the cope of the `StoreBlock` function
- Make `storeBlock` accept a `quorumCert`
- Remove `quorumCertificate` from local state and corresponding setters/getters
- Reduce the cope of the `IndexTransactions` function and remove from interface

**Consensus module changes**
- Add error handling where appropriate corresponding to the interface changes

## Testing

- [x] `make develop_test`
- [x] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md) w/ all of the steps outlined in the `README`

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have tested my changes using the available tooling
- [ ] If applicable, I have made corresponding changes to related local or global README
- [ ] If applicable, I have updated the corresponding CHANGELOG
- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works
- [ ] If applicable, I have added new diagrams using [mermaid.js](https://mermaid-js.github.io)

Issue #251, opened 2022-09-27
[Core] Update module interface required for end-to-end  StateHash implementation


Olshansk:
## Objective

Document the cross-module flow and update the interfaces required to compute a stateHash.

## Origin Document

The original ticket to implement the state hash (#147) was first tended to in (#152), but the scope grew and some exploratory work was done in #237.

Since the stateHash is a core component of any blockchain, it needs to be well understood and designed to make sure that it can be QAed against all possible scenarios.

## Goals

- Design & define the end-to-end flow to compute a state hash driving the task

## Deliverable

- [ ] A PR that updates the persistence, consensus and utility interfaces where needed
- [ ] A mermaid sequence diagram showing which components & interfaces are necessary & used in this flow
- [ ] Changes in the implementation that support the updates interfaces but without introducing new business logic

## Non-goals / Non-deliverables

- Implement the code necessary to compute a stateHash
- Design a testing suite / approach for the state hash

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)

## Testing Methodology

- **Task specific tests**: `make ...`
- **All tests**: `make test_all`
- **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @Olshansk 
**Co-Owners**: @andrewnguyen22

PR #285, opened 2022-10-05
[Persistence] First Implementation of the StateHash (#284)

### LINKED ISSUE #284: [Persistence] First Implementation of the StateHash
## Objective

Implement the first fully functional, tested state hash implementation

## Origin Document

Implement the interface and flow agreed on in #251. This replaces the original ticket (#147) whose scope was too large, whose implementation & discussion started in this PR #152 and whose further exploration was done in #237. 

## Goals

- Translate #251 into an implementation
- Make the state hash computation simple, efficient and easy to understand
- Create a foundation to add other objects/elements to the state hash in the future
- Ensure that persistence operations related to committing blocks are atomic

## Deliverable

- [ ] A PR with the first implementation of the state hash
- [ ] Unit tests for verifying the business logic used to implement state hash computation

- Modify any module interfaces
- E2E tests for cross-module state hash verification
- Provide the complete implementation of the state hash that will be used in production

## General issue deliverables

## Testing Methodology

- [ ] **Add**: Must add a suite of new unit tests
- [ ] **All tests**: `make develop_test`

---

**Creator**: @Olshansk

Olshansk:
## Description

The first implementation of the 

## Issue

Fixes #284 with follow up work in #361.

## Type of change

- [x] Bug fix
- [x] Code health or cleanup
- [x] Major breaking change
- [x] Documentation
- [ ] Other 

### Persistence - Core Changes for SMT
* Introduced & defined for `block_persistence.proto`
    * A persistence specific protobuf for the Block stored in the BlockStore
* On `Commit`, prepare and store a persistence block in the KV Store, SQL Store
* Replace `IndexTransactions` (plural) to `IndexTransaction` (singular)
* Maintaining a list of StateTrees using Celestia’s SMT and badger as the KV store to compute the state hash
* Implemented `ComputeStateHash` to update the global state based on:
    * Validators
    * Applications
    * Servicers
    * Fisherman
    * Accounts
    * Pools
    * Transactions
    * Added a placeholder for `params` and `flags`
* Added a benchmarking and a determinism test suite to validate this

### Persistence - General module changes
* Implemented `GetAccountsUpdated`, `GetPoolsUpdated` and `GetActorsUpdated` functions
* Removed `GetPrevAppHash` and `indexTransactions` functions
* Removed `blockProtoBytes` and `txResults` from the local state and added `quorumCert`
* Consolidate all `resetContext` related operations into a single function
* Implemented `ReleaseWriteContext`
* Implemented ability to `ClearAllState` and `ResetToGenesis` for debugging & testing purposes
* Added unit tests for all of the supporting SQL functions implemented
* Some improvements in unit test preparation & cleanup (limited to this PR's functionality)

### Persistence - KVStore changes
* Renamed `Put` to `Set`
* Embedded `smt.MapStore` in the interface containing `Get`, `Set` and `Delete`
* Implemented `Delete`
* Modified `GetAll` to return both `keys` and `values`
* Turned off badger logging options since it’s noisy

### Persistence - Module Interface changes
* Removed `GetPrevHash` and just using `GetBlockHash` instead
* Removed `blockProtoBz` from `SetProposalBlock` interface
* Removed `GetLatestBlockTxs` and `SetLatestTxResults` in exchange for `IndexTransaction`
* Removed `SetTxResults`
* Renamed `UpdateAppHash` to `ComputeStateHash`
* Removed some getters related to the proposal block (`GetBlockTxs`, `GetBlockHash`, etc…)

### Consensus
* Propagate `highPrepareQC` if available to the block being created
* Remove `blockProtoBytes` from propagation in `SetProposalBlock`
* Guarantee that `writeContext` is released when refreshing the `utilityContext`
* Use `GetBlockHash(height)` instead of `GetPrevAppHash` to be more explicit
* Use the real `quorumCert` when preparing a new block

### Configs
* Updated the test generator to produce deterministic keys
* Added `trees_store_dir` to persistence configs
* Updated `LocalNet` configs to have an empty `tx_indexer_path` and `trees_store_dir`

### Makefile changes
* Added `db_cli_node`
* Added `db_show_schemas`
* Added `test_persistence_state_hash`
* Added `benchmark_persistence_state_hash`

### Debug
* `ResetToGenesis` - Added the ability to reset the state to genesis
* `ClearState` - Added the ability to clear the state completely (height 0 without genesis data)
## Testing

**New:**
- [x] `make benchmark_persistence_state_hash`
- [x] `make test_persistence_state_hash`
- [x] Iteration 3 demo was done using this PR: https://drive.google.com/file/d/1IOrzq-XJP04BJjyqPPpPu873aSfwrnur/view?usp=sharing

**Existing:**
- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


Olshansk:
@andrewnguyen22 This is a preliminary (i.e. the 25% solution implementation of #252 after the discussion we had in #152).

In short, look at the tests in `persistence/test/state_test.go`, it works! I had to hack `shared/test_artifacts/generator.go` to get deterministic addresses, but otherwise would appreciate feedback before I add more tests, account for edge cases and finish it off.

The core files to look at are:

```bash
persistence/proto/block_persistence.proto
persistence/shared_sql.go
persistence/block.go
persistence/context.go
persistence/state.go
```

You can run the test with `go test -v -count=1 -v ./persistence/... -run TestStateHash`

andrewnguyen22:
This was de-prioritized in the 10/10/22 sync meeting

Olshansk:
@andrewnguyen22 Per the discussion this morning, please take a look at these files when you have a chance. Though it wasn't scoped in the previous iteration, I'm hoping to get a basic implementation of this done in the next iteration and looking for preliminary thoughts/feedback on the implementation approach before I invest more time in it.

Consider this to be a high-level review request just in terms of business logic (not code structure/format).

The initial test I have is in `persistence/test/state_test.go`, which can be run via `go test -v -count=1 -v ./persistence/... -run TestStateHash`, but I will add more in this PR.

Per my previous comment, the main files to focus review on are:

```
persistence/proto/block_persistence.proto
persistence/shared_sql.go
persistence/block.go
persistence/context.go
persistence/state.go
```

persistence/block.go
```diff
@@ -44,30 +45,56 @@ func (p PostgresContext) TransactionExists(transactionHash string) (bool, error)
 	return false, nil
 }
 
-func (p PostgresContext) StoreTransaction(transactionProtoBytes []byte) error {
-	log.Println("TODO: StoreTransaction not implemented")
+func (p *PostgresContext) StoreTransaction(transactionProtoBytes []byte) error {
+	p.currentBlockTxs = append(p.currentBlockTxs, transactionProtoBytes)
```
andrewnguyen22, line 49: This is confusing - at this part of the lifecycle, I'm not convince that the `currentBlock` wouldn't already be populated with `transactions`

andrewnguyen22: I'm also not excited that this is being worked on before https://github.com/pokt-network/pocket/pull/302

Notice this lifecycle and consider conforming to it

persistence/block.go
```diff
@@ -44,30 +45,56 @@ func (p PostgresContext) TransactionExists(transactionHash string) (bool, error)
...
+		var err error
+		prevHash, err = p.GetBlockHash(p.Height - 1)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		prevHash = []byte("HACK: get hash from genesis")
+	}
+
+	block := &types.Block{
```
andrewnguyen22, line 83: See #302 and how `storeTransaction` is meant to be used in the lifecycle

persistence/context.go
```diff
@@ -23,26 +15,38 @@ func (p PostgresContext) RollbackToSavePoint(bytes []byte) error {
...
-func (p PostgresContext) Commit() error {
+func (p PostgresContext) Commit(proposerAddr []byte, quorumCert []byte) error {
 	log.Printf("About to commit context at height %d.\n", p.Height)
 
+	block, err := p.prepareBlock(proposerAddr, quorumCert)
+	if err != nil {
+		return err
+	}
+
+	if err := p.insertBlock(block); err != nil {
```
andrewnguyen22, line 33: Add TODO to make atomic

persistence/debug.go
```diff
@@ -27,7 +27,7 @@ func (m *persistenceModule) HandleDebugMessage(debugMessage *debug.DebugMessage)
 func (m *persistenceModule) showLatestBlockInStore(_ *debug.DebugMessage) {
 	// TODO: Add an iterator to the `kvstore` and use that instead
 	height := m.GetBus().GetConsensusModule().CurrentHeight() - 1 // -1 because we want the latest committed height
-	blockBytes, err := m.GetBlockStore().Get(heightToBytes(int64(height)))
+	blockBytes, err := m.GetBlockStore().Get(HeightToBytes(int64(height)))
```
andrewnguyen22, line 30: Does this function need to be exported?

persistence/kvstore/kvstore.go
```diff
@@ -1,33 +1,47 @@
...
 	// TODO: Add a proper iterator interface
-	Put(key []byte, value []byte) error
-	Get(key []byte) ([]byte, error)
 	// TODO: Add pagination for `GetAll`
 	GetAll(prefixKey []byte, descending bool) ([][]byte, error)
+
 	Exists(key []byte) (bool, error)
 	ClearAll() error
+
+	// Same interface as in `smt.MapStore``
```
andrewnguyen22, line 23: Enforce by putting `smt.MapStore` inside the structure:

```golang
type KVStore interface {
  smt.MapStore
```

persistence/module.go
```diff
@@ -78,8 +83,17 @@ func (*persistenceModule) Create(runtimeMgr modules.RuntimeMgr) (modules.Module,
 		genesisState: persistenceGenesis,
 		blockStore:   blockStore,
 		writeContext: nil,
+		// contexts:     make(map[contextId]modules.PersistenceContext),
```
andrewnguyen22, line 86: What is this line for?

persistence/proto/block_persistence.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+package persistence;
+
+option go_package = "github.com/pokt-network/pocket/persistence/types";
+
+message Block {
```
andrewnguyen22, line 6: Transactions were planned to be stored in the `transaction indexer` is there a reason you think we should store them in a `Block`?

Consider `TransactionsHash` to represent them instead

persistence/shared_sql.go
```diff
@@ -44,6 +44,44 @@ func (p *PostgresContext) GetExists(actorSchema types.ProtocolActorSchema, addre
...
+
+	rows, err := tx.Query(ctx, actorSchema.GetUpdatedAtHeightQuery(height))
+	if err != nil {
+		return nil, err
+	}
+
+	addrs := make([][]byte, 0)
+	for rows.Next() {
+		var addr string
+		if err = rows.Scan(&addr); err != nil {
```
andrewnguyen22, line 61: Possible memory leak. `Returning` without closing rows

persistence/state.go
```diff
@@ -0,0 +1,187 @@
...
+	accountMerkleTree
+	poolMerkleTree
+
+	// Data / State Merkle Trees
+	blocksMerkleTree
+	paramsMerkleTree
+	flagsMerkleTree
+
+	// Used for iteration purposes only - see https://stackoverflow.com/a/64178235/768439
+	lastMerkleTree
```
andrewnguyen22, line 35: `Limit` is more of an appropriate naming convention - this sounds like the `last block merkle tree`

persistence/state.go
```diff
@@ -0,0 +1,187 @@
...
+	return trees, nil
+}
+
+func loadMerkleTrees(map[MerkleTree]*smt.SparseMerkleTree, error) {
+	log.Fatalf("loadMerkleTrees not implemented yet")
+}
+
+func (p *PostgresContext) updateStateHash() error {
+	// Update all the merkle trees
+	for treeType := MerkleTree(0); treeType < lastMerkleTree; treeType++ {
```
andrewnguyen22, line 79: The pattern of a `for loop` with a `switch` statement is not my favorite.

It makes me think there might be an opportunity to use a map like structure `merkleTrees[treeType]`.

persistence/state.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	// Get the root of each Merkle Tree
+	roots := make([][]byte, 0)
+	for treeType := MerkleTree(0); treeType < lastMerkleTree; treeType++ {
+		roots = append(roots, p.merkleTrees[treeType].Root())
+	}
+
+	// DISCUSS(drewsky): In #152, we discussed the ordering of the roots
```
andrewnguyen22, line 120: Consider we could add more trees in the future.

I would encourage to future proof the solution with `strict` ordering using the enum

From what I can tell - it seems like this solution does just that. So I `think` I am a fan of this solution... LMK if I'm missing anything

persistence/state.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// Get the state hash
+	rootsConcat := bytes.Join(roots, []byte{})
+	stateHash := sha256.Sum256(rootsConcat)
+
+	p.currentStateHash = stateHash[:]
+	return nil
+}
+
+func (p PostgresContext) updateActorsTree(actorType types.ActorType, actors []*types.Actor) error {
```
andrewnguyen22, line 135: Since this is a partial implementation - I won't dig into it - but wanted to **emphasize** that the `world state` is a lot more than just the `actors`.

persistence/test/state_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+		apps, err := db.GetAllApps(height)
+		require.NoError(t, err)
+		app := apps[0]
+
+		addrBz, err := hex.DecodeString(app.GetAddress())
+		require.NoError(t, err)
+
+		// Update the app's stake
+		newStakeAmount := types.BigIntToString(big.NewInt(height + int64(420000000000)))
```
andrewnguyen22, line 39: Make arbitrary value a constant and then make a comment on how it's arbitrary

runtime/test_artifacts/generator.go
```diff
@@ -132,7 +134,7 @@ func NewPools() (pools []modules.Account) { // TODO (Team) in the real testing s
 
 func NewAccounts(n int, privateKeys ...string) (accounts []modules.Account) {
 	for i := 0; i < n; i++ {
-		_, _, addr := GenerateNewKeysStrings()
+		_, _, addr := GenerateNewKeysDeterministic(69)
```
andrewnguyen22, line 137: Where did 69 come from? Move to constant and document

utility/block.go
```diff
@@ -37,22 +40,30 @@ func (u *UtilityContext) ApplyBlock(latestHeight int64, proposerAddress []byte,
...
+
 	// end block lifecycle phase
 	if err := u.EndBlock(proposerAddress); err != nil {
 		return nil, err
 	}
-	// return the app hash (consensus module will get the validator set directly
-	return u.GetAppHash()
+
+	// TODO: What if everything above succeeded but updating the app hash failed?
+	appHash, err := u.Context.UpdateAppHash()
```
andrewnguyen22, line 60: Fatal error maybe? What are some of the reasons it would fail?

andrewnguyen22 (changes requested):

persistence/block.go
Olshansk: I apoligize. Wasn't my intention.

Per my comment in https://github.com/pokt-network/pocket/pull/302#pullrequestreview-1146448248, I'm pausing work on this until we discuss the end-to-end flow and get #302 in.

Olshansk:
@andrewnguyen22 No need to look at this PR yet, but I found that I need to make a fair amount of changes to the unit tests to enable some potential bugs in state management. Partially related to #261.

In order to make everything easier to review and prevent this from getting out of hand, I'm going to split out the test only changes into a separate PR.

It might have a few conflicts with #315, but I believe it'll enable less conflicts holistically. 

No action item at the moment but just wanted to give you a heads up. I'll help with w/e conflict resolution we need in the future and we could do a PR code swap if needed as well.

persistence/state.go
Olshansk: I try a few different approaches but did not succeed. Mainly because the data (e.g. blocks), accounts (pools, accounts) and actors all have slightly different logic so it ended up making the code more complex than not.

Resolving for now, but will keep thinking.

Olshansk (commented):
Not re-requesting a review (yet) but just wanted to post an update. I will keep working on this in parallel knowing that we will have to rebase it on top of #315 afterwards. Since this will also get pretty large, it may require a "PR swap" or a "synchronous review" as well.

Made some good headway on this. I replied to and resolved the comments left in r1 of review, and we can dive deeper in r2. The scope is getting a bit bigger because I had to refactor some tests and utilities based on the changes but it's looking cleaner.

Since the core functionality is mostly complete, the next steps that I will focus are:
1. Add configurations to load the SMT from a previous state. Right now it creates in-mem SMT's (using in-memory badger). The fact that this works is already great IMO.
2. Utility functions to help visualize what's in the SMTs
3. Add benchmarking tests
4. Add fuzzing tests

Olshansk:
--- 
Note to the reviewer (i.e. @andrewnguyen22 ): The size of this PR grew very large in size, so I want to try and stop it from growing in size, while also getting feedback as soon as possible so we can unblock downstream work (e.g. state sync) while being able to finish off lose ends here. Specifically:
- I left several TODO_IN_NEXT_COMMIT, which means that I’ll create tickets for it before we finish this.
- I left several TODO_IN_THIS_COMMIT, which means that I’ll continue working on it while the first round of review starts
- I left several DO_NOT_REVIEW_THIS_YET, which means that I’ll do it in this commit but do not look at it yet
- I left several DISCUSS items, which I hesitate from implementing or changing because it interferes with the tx-indexer bridge integration, so it is something I’d prefer to discuss online/offline before changing it as I deem necessary
--- 

For the core logic changes, please see:
- `persistence/state.go`, `persistence/state_test.go` and `persistence/benchmark_state_test.go`

I’m open to splitting this into smaller commits if need be.

### Primary Implementation changes
- Added a persistence specific protobuf for the Block stored in the BlockStore
- Maintaining a list of StateTrees using Celestia’s SMT and badger as the KV store to compute the state hash
- Implemented `ComputeStateHash` to update the global state based on:
    - Validators
    - Applications
    - Servicers
    - Fisherpeople
    - Accounts
    - Polols
    - Transactions
    - Placeholder for flags & params created
- Added a benchmarking and a determinism test suite to validate this

### Persistence Module Interface
- Removed `blockProtoBz` from `SetProposalBlock`
- Removed `GetLatestBlockTxs`  and `SetLatestTxResults` in exchange for `IndexTransaction`

### Persistence Module State / Implementation
- Removed `blockProtoBz` from `SetProposalBlock`
- Removed `IndexTransactions` // plural
- Consolidate `resetContext` operations into a single function
- Implemented `ReleaseWriteContext`

### Configs
- Updated the test generator to produce deterministic keys
- Added `trees_store_dir` to persistence configs
- Updated LocalNet configs to have an empty `tx_indexer_path` and `trees_store_dir`

### Makefile changes
- Added `db_show_schemas`
- Added `test_persistence_state_hash`
- Added `benchmark_persistence_state_hash`

### Debug
- Updated debug commands to be able to `ClearState` and `ResetToGenesis`

### KV Store Changes
- Added a `Delete` function
- Updated `GetAll` to also return the keys
- Turned off `badger logging` so it’s not too noisy
- Renamed `Put` to `Set`

### Non core changes
- Added various comments, TODOs, etc throughout the PR
- `block_test

Olshansk:
Though it is still WIP, I believe it is ready for review.

persistence/test/benchmark_state_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+const (
+	maxStringAmount = 1000000000000000000
+)
+
+var isModifierRe = regexp.MustCompile(`^(Insert|Set|Add|Subtract)`) // Add Update?
+
+// INVESTIGATE: This benchmark can be used to experiment with different Merkle Tree implementations
+// and key-value stores.
+func BenchmarkStateHash(b *testing.B) {
```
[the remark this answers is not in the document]
andrewnguyen22: I'm not sure if `benchmark` is the most intuitive name.

persistence/genesis.go
```diff
@@ -10,9 +10,7 @@ import (
 	"github.com/pokt-network/pocket/shared/modules"
```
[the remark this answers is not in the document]
andrewnguyen22: Ditto for `all modules` affected

persistence/block.go
```diff
@@ -76,64 +73,86 @@ func (p PostgresContext) TransactionExists(transactionHash string) (bool, error)
...
+func (p PostgresContext) storeBlock(block *types.Block) error {
+	blockBz, err := codec.GetCodec().Marshal(block)
+	if err != nil {
+		return err
+	}
+	return p.blockStore.Set(heightToBytes(p.Height), blockBz)
+}
+
+func (p PostgresContext) getTxsHash(height int64) ([]byte, error) {
+	// The order (descending) is important here since it is used to comprise the hash in the block
```
andrewnguyen22, line 138: Either the comment or implementation is wrong:

Screen Shot 2022-11-09 at 5 22 01 PM

persistence/debug.go
```diff
@@ -1,58 +1,123 @@
...
-	block := &typesCons.Block{}
+	block := &types.Block{}
 	codec.Unmarshal(blockBytes, block)
 
-	log.Printf("Block at height %d with %d transactions: %+v \n", height, len(block.Transactions), block)
+	log.Printf("Block at height %d: %+v \n", height, block)
 }
 
-func (m *persistenceModule) clearState(_ *debug.DebugMessage) {
+// TODO: MAke sure this is atomic
```
andrewnguyen22, line 57:
```suggestion
// TODO: Make sure this is atomic
```

persistence/debug.go
```diff
@@ -1,58 +1,123 @@
...
 func (m *persistenceModule) showLatestBlockInStore(_ *debug.DebugMessage) {
 	// TODO: Add an iterator to the `kvstore` and use that instead
-	height := m.GetBus().GetConsensusModule().CurrentHeight() - 1 // -1 because we want the latest committed height
+	height := m.GetBus().GetConsensusModule().CurrentHeight()
 	blockBytes, err := m.GetBlockStore().Get(heightToBytes(int64(height)))
 	if err != nil {
 		log.Printf("Error getting block %d from block store: %s \n", height, err)
 		return
 	}
 	codec := codec.GetCodec()
```
andrewnguyen22, line 50: Change to:

```golang
        block := &types.Block{}
	codec.GetCodec().Unmarshal(blockBytes, block)
```

persistence/genesis.go
```diff
@@ -150,8 +144,22 @@ func (m *persistenceModule) populateGenesisState(state modules.PersistenceGenesi
...
-		log.Fatalf("an error occurred during commit() on genesis state %s ", err.Error())
+	// Updates all the merkle trees
+	appHash, err := rwContext.ComputeAppHash()
+	if err != nil {
+		log.Fatalf("an error occurred updating the app hash during genesis: %s", err.Error())
+	}
+
+	// TODO_IN_THIS_COMMIT: Figure out what these values for genesis should be
+	genesisQuorumCert := []byte("placeholderQuorumCert")
+	genesisProposer := []byte("placeholderProposer")
```
andrewnguyen22, line 155: DISCUSS:

Do these affect the appHash at genesis?

persistence/shared_sql.go
```diff
@@ -44,6 +44,44 @@ func (p *PostgresContext) GetExists(actorSchema types.ProtocolActorSchema, addre
...
+		if err != nil {
+			return nil, err
+		}
+		addrs = append(addrs, addrBz)
+	}
+	rows.Close()
+
+	actors = make([]*types.Actor, len(addrs))
+	for i, addr := range addrs {
+		actor, err := p.GetActor(actorSchema, []byte(addr), height)
```
andrewnguyen22, line 75:
```suggestion
		actor, err := p.GetActor(actorSchema, addr, height)
```

persistence/shared_sql.go
```diff
@@ -71,7 +114,8 @@ func (p *PostgresContext) GetChainsForActor(
 	tx pgx.Tx,
 	actorSchema types.ProtocolActorSchema,
 	actor *types.Actor,
-	height int64) (a *types.Actor, err error) {
+	height int64,
+) (a *types.Actor, err error) {
```
andrewnguyen22, line 118: New Olshansk style or accidental?

Recommend reverting

persistence/state.go
```diff
@@ -0,0 +1,330 @@
...
+
+	accountMerkleTree: "account",
+	poolMerkleTree:    "pool",
+
+	transactionsMerkleTree: "transactions",
+	paramsMerkleTree:       "params",
+	flagsMerkleTree:        "flags",
+}
+
+var actorTypeToMerkleTreeName map[types.ActorType]merkleTree = map[types.ActorType]merkleTree{
```
andrewnguyen22, line 66:
```suggestion
var actorTypeToMerkleTreeName = map[types.ActorType]merkleTree{
```

Ditto for below

persistence/state.go
```diff
@@ -0,0 +1,330 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Data Tree Helpers
+
+func (p *PostgresContext) updateTransactionsTree(height int64) error {
```
andrewnguyen22, line 303: Unused `height` param?

Can comment future usage

persistence/state.go
```diff
@@ -0,0 +1,330 @@
...
+	txResults, err := p.txIndexer.GetByHeight(p.Height, false)
```
andrewnguyen22, line 304: Should this really be `p.Height` or `height`?

persistence/state.go
```diff
@@ -0,0 +1,330 @@
...
+	// nodeStores & valueStore are part of the SMT, but references are kept below for convenience
+	// and debugging purposes
+	nodeStores  map[merkleTree]kvstore.KVStore
+	valueStores map[merkleTree]kvstore.KVStore
+}
+
+// A list of Merkle Trees used to maintain the state hash.
+const (
+	// VERY IMPORTANT: The order in which these trees are defined is important and strict. It implicitly
+	// defines the index of the the root hash each independent as they are concatenated together
```
andrewnguyen22, line 30:
```suggestion
	// defines the index of the root hash each independent as they are concatenated together
```

shared/modules/persistence_module.go
```diff
@@ -58,12 +58,10 @@ type PersistenceWriteContext interface {
...
-	// DISCUSS_IN_THIS_COMMIT: Can this function be removed ? If so, could we remove `TxResult` from the public facing interface given that we set transactions in `SetProposalBlock`?
-	SetLatestTxResults(txResults []TxResult)
-	// TODO(#284): Remove `blockProtoBytes`
-	SetProposalBlock(blockHash string, blockProtoBytes, proposerAddr []byte, transactions [][]byte) error
-	// Store the block into persistence
-	UpdateAppHash() ([]byte, error)
+	SetProposalBlock(blockHash string, proposerAddr []byte, quorumCert []byte, transactions [][]byte) error
+	GetLatestBlockTxs() [][]byte              // Returns the transactions set by `SetProposalBlock`
+	ComputeAppHash() ([]byte, error)          // Update the merkle trees, computes the new state hash, and returns in
+	IndexTransaction(txResult TxResult) error // DISCUSS_IN_THIS_COMMIT: How can we remove `TxResult` from the public interface?
```
andrewnguyen22, line 64: Great question. No answer yet - but will think

utility/block.go
```diff
@@ -93,31 +97,41 @@ func (u *UtilityContext) ApplyBlock() (appHash []byte, err error) {
 		if err := tx.ValidateBasic(); err != nil {
 			return nil, err
 		}
+		// DISCUSS(#315): Currently, the pattern is allowing nil err with an error transaction...
```
andrewnguyen22, line 100: Issue is closed. should we open another?

andrewnguyen22 (changes requested):
First pass - I know I'm missing some review items.

I also recognize that there's a big diff between this branch and `main` too - so will require another review then.

Will try again in round 2 for deeper understanding and flow.

app/client/cli/debug.go
```diff
@@ -21,12 +21,13 @@ import (
...
 )
 
+// TECHDEBT: Lowercase variables / constants that do not need to be exported.
 const (
-	PromptResetToGenesis         string = "ResetToGenesis"
-	PromptPrintNodeState         string = "PrintNodeState"
-	PromptTriggerNextView        string = "TriggerNextView"
-	PromptTogglePacemakerMode    string = "TogglePacemakerMode"
-	PromptShowLatestBlockInStore string = "ShowLatestBlockInStore"
+	PromptResetToGenesis         string = "ResetToGenesis (broadcast)"
```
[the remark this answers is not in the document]
Olshansk: I reverted this change (since it's out of scope anyhow) and created this ticket for a starter task in the future: #344

consensus/block.go
```diff
@@ -57,7 +57,8 @@ func (m *consensusModule) refreshUtilityContext() error {
 	// Ideally, this should not be called.
 	if m.utilityContext != nil {
 		m.nodeLog(typesCons.NilUtilityContextWarning)
-		if err := m.utilityContext.Release(); err != nil {
+		err := m.utilityContext.Release()
```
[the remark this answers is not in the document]
Olshansk: Reverted.

consensus/hotstuff_replica.go
```diff
@@ -244,8 +240,9 @@ func (m *consensusModule) applyBlock(block *typesCons.Block) error {
 	}
 
 	// CONSOLIDATE: Terminology of `appHash` and `stateHash`
-	if block.BlockHeader.Hash != hex.EncodeToString(appHash) {
-		return typesCons.ErrInvalidAppHash(block.BlockHeader.Hash, hex.EncodeToString(appHash))
+	appHashString := hex.EncodeToString(appHash)
+	if blockHeader.Hash != appHashString {
+		return typesCons.ErrInvalidAppHash(blockHeader.Hash, appHashString)
```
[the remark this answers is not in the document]
Olshansk: Nice

utility/block.go
Olshansk: I'm going to create a ticket explicitly for removing `TxResult` from the interface

Olshansk (commented):
Still not complete, but see some of my responses inline.

utility/block.go
Olshansk: #346

persistence/genesis.go
Olshansk: **Update:** I wasn't really happy with this answer so just wanted to provide an update that I'm looking deeper into this a bit.

Current thinking (not a proposal yet) is:

```
StateHash = what we have right now (the ordered concatination of all the state tries)
PrevAppHash = Hash(concat(StateHash, ProposerId, Hash(QC))
```

I still need more time to think through the different attacks that can take place though

---

In Cosmos, they have a variety of hahses, all mostly related to state:

Screen Shot 2022-11-14 at 11 06 19 AM

In flow (hostuff in Go), they only have a payload hash but seem to be missing the links to the previous proposer and quorum cert:

Screen Shot 2022-11-14 at 11 08 41 AM

In Aptos (a hotstuff algorithm but in rust), they have a lot of hashes mainly related to the VM execution output:

Screen Shot 2022-11-14 at 11 20 39 AM

Olshansk:
@andrewnguyen22 I want to focus on getting this in and creating any necessary follow up work in smaller follow up tasks. 

**The current "development" I see from my point of view (will aim to do this asap):**
1. Updating CHANGELOG
2. Updating the README explaining how the state hash is computed
3. Doing a self-review
4. Creating follow up tasks (some unit tests, atomicity, etc...)

**The open questions I have which may block this PR:**
1. See my use of `err := db.IndexTransaction(txResult)`. I change some of the flow related to `SetTxResults` and based on the discussion we're having in #251, my understanding is that this is not intended to be called until Commit

**The open questions I see which do not block this PR but may be high priority afterwards:**

1. The logic I implemented is related to `StateHash`, but I'm thinking we may also have a type called a `BlockHash` which incorporated the QC and proposer as well. Lmk your thoughts based don't the discussion here: https://github.com/pokt-network/pocket/pull/285#discussion_r1019828507

2. Idea around storing the "proposal block contents" in the `UtilityContext` which would avoid us needing to retrieve it during the `ApplyBlock` flow. For example, `blockTxs` is stored in the `PersistenceContext` but is only accessed by the `UtilityContext` and written to disk when we really commit.

persistence/context.go
```diff
@@ -1,75 +1,114 @@
...
-func (p *PostgresContext) Reset() error {
-	p.txResults = nil
-	p.blockHash = ""
-	p.proposerAddr = nil
-	p.blockProtoBytes = nil
-	p.blockTxs = nil
-	return nil
+func (p *PostgresContext) ComputeAppHash() ([]byte, error) {
+	// DISCUSS_IN_THIS_COMMIT:
+	// 1. Should we compare the `appHash` returned from `updateMerkleTrees`?
```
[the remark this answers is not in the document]
Olshansk: Going to resolve this and track follow-up work in #361. Will link to this thread in the comments as a reference.

As an extra reference, here are the comments I previously left for discussion purposes:

```
	// `PostgresContext` contains a `blockHash` in its state set by `SetProposalBlock`.
	// The caller of `ComputeAppHash` is responsible (given the context) to compare
	// the return hash with the proposed hash (if that is the case).

	// Situations:
	// 	1. What if `p.updateMerkleTrees()` != `p.blockHash` && `p.blockHash` != nil?
	// 	2. Do we set `p.blockHash` =`p.updateMerkleTrees()` if it's nil?
	// 	3. In `SetProposalBlock`,

	// Future work: Need to implement rollback of the trees (i.e. key-value stores) alongside the SQL DB transaction.

	// 1. Should we compare the `appHash` returned from `updateMerkleTrees`?
	// 2. Should this update the internal state of the context?
	// Idea: If `p.blockHash` is nil => update it
	//       If `p.blockHash` is nil => compare it with the return value of `updateMerkleTrees` and error if different
```

Olshansk (commented):
I've done several self-reviews now and, aside from the implementation, made sure to focus on:

1. Documentation (for business logic)
2. Comments (for hacks, bugs, follow up work, etc...)
3. Adding unit tests (where appropriate)

I've also created #361 to track & document follow up work, because there's a non-trivial amount.

I don't believe this PR is perfect by any means, but due to its growth in scope and given how long it's been open for, I believe it's good enough. As the title says, it is the "first implementation of the state hash."

Olshansk:
> I've done several self-reviews now and, aside from the implementation, made sure to focus on:
> 
> 1. Documentation (for business logic)
> 2. Comments (for hacks, bugs, follow up work, etc...)
> 3. Adding unit tests (where appropriate)
> 
> I've also created #361 to track & document follow up work, because there's a non-trivial amount.
> 
> I don't believe this PR is perfect by any means, but due to its growth in scope and given how long it's been open for, I believe it's good enough. As the title says, it is the "first implementation of the state hash."

@deblasis I merged in the interface change and don't really have expectations of anyone having the context of reviewing this. I've done a few rounds of self-review and a couple from @andrewnguyen22, so my ask is:

1. Please look at the PR description just to have context of the changes
2. Please review `shared/docs/PROTOCOL_STATE_HASH.md`

Hoping to merge in this shortly after.

persistence/docs/PROTOCOL_STATE_HASH.md
```diff
@@ -0,0 +1,131 @@
...
+- Parameters
+- Flags
+
+## Compute State Hash
+
+_Note: `GetRecordsUpdatedAtHeight` is an abstraction for retrieving all the records from the corresponding SQL tables depending on the type of record (Actors, Transactions, Params, etc...)_
+
+This flow shows the interaction between the `PostgresDB` and `MerkleTrees` listed above to compute the state hash. Assuming the process of applying a proposal block to the current context (i.e. the uncommited SQL state) is done, the following steps compute the hash of the new world state.
+
+1. Loop over all of the merkle tree types
```
deblasis, line 66: What happens if a new tree type gets introduced at some point? Do we enforce ordering deterministically?

persistence/docs/PROTOCOL_STATE_HASH.md
```diff
@@ -0,0 +1,131 @@
...
+2. Use `GetRecordsUpdatedAtHeight` to retrieve all the records updated at the context's height
+3. Serialize each record using the corresponding underlying protobuf
+4. Insert the serialized record into the corresponding tree (which is back by a key-value store)
+5. Compute the root hash of each tree
+6. Aggregate all the root hashes by concatenating them together
```
deblasis, line 71: Similarly to point 1. comment. Is ordering enforced/guaranteed somehow?

persistence/docs/PROTOCOL_STATE_HASH.md
```diff
@@ -0,0 +1,131 @@
...
+        P->>+PKV: GetRoot()
+        PKV->>-P: rootHash
+    end
+
+    P->>P: stateHash = hash(concat(rootHashes))
+    activate P
+    deactivate P
+```
+
+_IMPORTANT: The order in which the `rootHashes` are concatenated is based on the definition in which the trees are ordered in within `state.go`._
```
deblasis, line 96: ACK. Noted the iota loop. 

(I was scared when I initially saw maps)

persistence/docs/PROTOCOL_STATE_HASH.md
deblasis: Nevermind, answered myself below

persistence/docs/PROTOCOL_STATE_HASH.md
deblasis: Nevermind, answered myself below

persistence/docs/PROTOCOL_STATE_HASH.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Store Block (Commit)
+
+When the `Commit(quorumCert)` function is invoked, the current context is committed to disk. The `PersistenceContext` does the following:
+
+1. Read data from the persistence context's in-memory state
+2. Prepare a instance of the `Block` proto & serialize it
+3. Insert the `Block` into the `BlockStore`
+4. Insert the `Block` into the SQL Store
+5. Commit the context's SQL transaction to disk
```
deblasis, line 106: What about the unhappy path / rollback? Out of scope for now? Do we have issues/TODOs for it?

deblasis: LOL nevermind maybe answered myself with the below

persistence/docs/PROTOCOL_STATE_HASH.md
```diff
@@ -0,0 +1,131 @@
...
+    %% Insert into the Block Store (i.e. Key-Value store)
+    P->>+PKV: Put(height, block)
+    PKV->>-P: result, err_code
+
+    %% Commit the SQL transaction
+    P->>+PSQL: Commit(SQL Tx to disk)
+    PSQL->>-P: result, err_code
+```
+
+_TODO: If an error occurs at any step, all of the operations must be reverted in an atomic manner._
```
deblasis, line 131: 😂

persistence/proto/block_persistence.proto
```diff
@@ -0,0 +1,14 @@
...
+    uint64 height = 1;
+    string stateHash = 2;
+    string prevStateHash = 3; // The stateHash of the block at height-1
+    bytes proposerAddress = 4; // The proposer of this block
+    bytes quorumCertificate = 5; // The quorum certificate containing signature from 2/3+ validators at height
+    // INVESTIGATE(#361): Decide if we neeed `transactionsHash` given that it is capture in the `transactionsTree`.
+    bytes transactionsHash = 6; // The hash of all the translactions in the block
```
deblasis, line 13:
```suggestion
    bytes transactionsHash = 6; // The hash of all the transactions in the block
```

persistence/proto/block_persistence.proto
```diff
@@ -0,0 +1,14 @@
...
+
+option go_package = "github.com/pokt-network/pocket/persistence/types";
+
+message Block {
+    uint64 height = 1;
+    string stateHash = 2;
+    string prevStateHash = 3; // The stateHash of the block at height-1
+    bytes proposerAddress = 4; // The proposer of this block
+    bytes quorumCertificate = 5; // The quorum certificate containing signature from 2/3+ validators at height
+    // INVESTIGATE(#361): Decide if we neeed `transactionsHash` given that it is capture in the `transactionsTree`.
```
deblasis, line 12:
```suggestion
    // INVESTIGATE(#361): Decide if we need `transactionsHash` given that it is captured in the `transactionsTree`.
```

persistence/context.go
```diff
@@ -14,62 +18,100 @@ func (p PostgresContext) NewSavePoint(bytes []byte) error {
...
-func (p *PostgresContext) Reset() error {
-	p.txResults = nil
-	p.blockHash = ""
-	p.proposerAddr = nil
-	p.blockProtoBytes = nil
-	p.blockTxs = nil
-	return nil
+func (p *PostgresContext) ComputeAppHash() ([]byte, error) {
+	// IMPROVE(#361): Guarantee the integrity of the state
+	//                Full details in the thread from the PR review: https://github.com/pokt-network/pocket/pull/285/files?show-viewed-files=true&file-filters%5B%5D=#r1033002640
```
deblasis, line 29: This is not a permalink, it leads me to an empty files list since it has filters.

image

persistence/test/benchmark_state_test.go
```diff
@@ -0,0 +1,174 @@
...
+		{10, 1, 10},
+		{10, 10, 10},
+
+		// This takes a VERY long time to run
+		// {100, 1, 1},
+		// {100, 1, 100},
+		// {100, 100, 100},
+	}
+
+	for _, testCase := range testCases {
```
deblasis, line 58: no-op: nice, I missed some table-based tests :)

deblasis (changes requested):
Solid work from the looks of it. 

A couple of nits here and there and a good amount of confidence on the previous rounds of review (self or not 😉)

Olshansk:
> A couple of nits here and there and a good amount of confidence on the previous rounds of review (self or not 😉)

tenor

persistence/docs/PROTOCOL_STATE_HASH.md
Olshansk: I'm glad you brought this up though. I added this in #361 and will add a clarifying comment here too.

```
18. Consider specifying the order of the merkle trees in a .proto enum rather than in .go iota to separate state commitment from implementation.
```

persistence/docs/PROTOCOL_STATE_HASH.md
Olshansk: This is clearly not clear or evident which may mean it's not a good design. 

Added the following to the top of the tree section and will revisit in #361.

```
Note that the order in which the trees are defined (found in `persistence/state.go`) is important since it determines how the state hash is computed. _TODO(#361): Consider specifying the oder in a `.proto` `enum` rather than a `.go` `iota`._
```

persistence/docs/PROTOCOL_STATE_HASH.md
Olshansk: Added this section
```
## Failed Commitments

TODO: Failed commitments and the implementation of rollbacks is tracked in #327 and #329.
```

persistence/docs/PROTOCOL_STATE_HASH.md
Olshansk: 😅 It'll be a fun task.

persistence/test/benchmark_state_test.go
Olshansk: I'm personally quite proud of the test harness I built for this feature :)

persistence/context.go
Olshansk: Thank you. Updated here: https://github.com/pokt-network/pocket/pull/285#discussion_r1018471719

persistence/test/benchmark_state_test.go
deblasis: you should! However....... tests are failing I believe 🧐

deblasis:
Tests failing is the only thing that is holding me back from approving now

Olshansk:
> Tests failing is the only thing that is holding me back from approving now

Works on my machine, but retriggering the job as well.

Screenshot 2022-11-28 at 5 10 22 PM

Olshansk:
> > Tests failing is the only thing that is holding me back from approving now
> 
> Works on my machine, but retriggering the job as well.
> 
> Screenshot 2022-11-28 at 5 10 22 PM
> 
> Screenshot 2022-11-28 at 5 11 28 PM
> 
> Also, we changed `69` files with 2`42` commits and and the sha starts with `42`. What else do you need?

Going to sync with @okdas to figure out how to iterate faster on debugging remote errors. It's passing locally but we should definitely make sure it passes with our CI pipeline first.

Olshansk:
From the logs, it looked like utility is what was failing, so I try to reduce the scope to see if the interference from other modules was the source, and confirmed that it was not.

My gut tells me it's a race condition related to the states of the testing docker container, but I have yet to pinpoint the root cause.

Olshansk:
I'm able to repro locally by running: `go test -v -p=1 -json -count=1 ./... -run TestUtilityContext_HandleMessage | tee test_results.json`.

@deblasis Please don't spend time looking into it; finishing the tech debt so P2P can work properly is a higher priority.

deblasis:
> > I'm able to repro locally by running: `go test -v -p=1 -json -count=1 ./... -run TestUtilityContext_HandleMessage | tee test_results.json` and will focus on resolving this today.
> > @deblasis Please don't spend time looking into it; finishing the tech debt so P2P can work properly is a higher priority.
> 
> Update: The latest [job](https://github.com/pokt-network/pocket/actions/runs/3576344317/jobs/6014037249) outputted this error:
> 
> ```
> === RUN   TestUtilityContext_HandleMessageUnpause/ServiceNode.HandleMessageUnpause
> SIGQUIT: quit
> PC=0x4724e1 m=0 sigcode=0
> 
> goroutine 0 [idle]:
> runtime.futex()
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/sys_linux_amd64.s:559 +0x21 fp=0x7ffd191881f0 sp=0x7ffd191881e8 pc=0x4724e1
> runtime.futexsleep(0x0?, 0x0?, 0x0?)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/os_linux.go:69 +0x36 fp=0x7ffd19188240 sp=0x7ffd191881f0 pc=0x437f76
> runtime.notesleep(0x14f1a88)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/lock_futex.go:160 +0x87 fp=0x7ffd19188278 sp=0x7ffd19188240 pc=0x40df87
> runtime.mPark(...)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/proc.go:1457
> runtime.stopm()
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/proc.go:2247 +0x8c fp=0x7ffd191882a8 sp=0x7ffd19188278 pc=0x442a0c
> runtime.findRunnable()
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/proc.go:2874 +0x9e8 fp=0x7ffd19188398 sp=0x7ffd191882a8 pc=0x4440a8
> runtime.schedule()
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/proc.go:3214 +0xbe fp=0x7ffd191883d0 sp=0x7ffd19188398 pc=0x444ede
> runtime.park_m(0xc3a9d66340?)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/proc.go:3363 +0x14d fp=0x7ffd19188400 sp=0x7ffd191883d0 pc=0x44540d
> runtime.mcall()
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/asm_amd64.s:448 +0x43 fp=0x7ffd19188410 sp=0x7ffd19188400 pc=0x46e363
> 
> goroutine 1 [chan receive, 12 minutes]:
> runtime.gopark(0x1?, 0xc00014b9f8?, 0x7f?, 0xf7?, 0xc00014b9a8?)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/proc.go:363 +0xd6 fp=0xc00014b978 sp=0xc00014b958 pc=0x43ebf6
> runtime.chanrecv(0xc2c2fd2a80, 0xc00014ba77, 0x1)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/chan.go:583 +0x49b fp=0xc00014ba08 sp=0xc00014b978 pc=0x408b9b
> runtime.chanrecv1(0xe12a69?, 0xe12a69?)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/chan.go:442 +0x18 fp=0xc00014ba30 sp=0xc00014ba08 pc=0x408698
> testing.(*T).Run(0xc0003eeea0, {0xe12a69?, 0x524865?}, 0xe4c5f8)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1494 +0x37a fp=0xc00014bad0 sp=0xc00014ba30 pc=0x525c7a
> testing.runTests.func1(0xc000419830?)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:[184](https://github.com/pokt-network/pocket/actions/runs/3576344317/jobs/6014037249#step:11:185)6 +0x6e fp=0xc00014bb20 sp=0xc00014bad0 pc=0x527bce
> testing.tRunner(0xc0003eeea0, 0xc0002efc28)
> 	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1446 +*** Test killed: ran too long (11m0s).
> FAIL	github.com/pokt-network/pocket/utility/test	746.140s
> ```
> 
> I noticed that some tests that do not have a `PASS` condition fail when there are `subtests` involved. E.g. I see a `RUN` for `RUN TestUtilityContext_HandleMessageUnpause/ServiceNode.HandleMessageUnpause` but no `FAIL`, which is what led me down this path.
> 
> I found [this answer](https://stackoverflow.com/a/72724245/768439) on stack overflow which, again, lead me to believe it has something to do in the setup/teardown of our docker testing container + inter-test state + test race conditions:
> 
> Screenshot 2022-11-29 at 9 32 45 AM

Interesting. The test is timing out and gets killed by the testing harness. Look at the errors in the json file, there must be something wrong in there.

consensus/consensus_tests/utils_test.go
```diff
@@ -141,8 +141,8 @@ func CreateTestConsensusPocketNode(
 	utilityMock := baseUtilityMock(t, testChannel)
 	telemetryMock := baseTelemetryMock(t, testChannel)
 	loggerMock := baseLoggerMock(t, testChannel)
-  rpcMock := baseRpcMock(t, testChannel)
-  
+	rpcMock := baseRpcMock(t, testChannel)
```
okdas, line 144: Thank you for fixing this, apparently I committed this when I had a conflict.

Olshansk:
@deblasis Tests are passing! https://github.com/pokt-network/pocket/pull/285

tl;dr Loading state from disk (i.e. when we don't load genesis) was an issue in a remote environment. The `shouldHydrateGenesisDb` function is something we need to revisit, but all I did was make sure to clear the SQL DB if we need to load the genesis state.

Makefile
```diff
@@ -290,16 +305,10 @@ generate_cli_commands_docs:
...
-
-.PHONY: test_all_with_coverage
-## Run all go unit tests, output results & coverage into files
-test_all_with_coverage: generate_rpc_openapi # generate_mocks
-	go test -p 1 -v ./... -covermode=count -coverprofile=coverage.out
-	go tool cover -func=coverage.out -o=coverage.out
+.PHONY: test_all_with_json_coverage
+## Run all go unit tests, output results & coverage into json & coverage files
+test_all_with_json_coverage: generate_rpc_openapi # generate_mocks
+	go test -p 1 -json ./... -covermode=count -coverprofile=coverage.out -run TestUtilityContext_HandleMessage | tee test_results.json | jq
```
deblasis, line 311: It looks like this offending line is still here, please clarify if that's intended and if it is that we have a TODO/ticket to track the effort necessary to restore the "all" part of `test_all_with_json_coverage`

deblasis (changes requested):

Makefile
Olshansk: Thanks for catching this. I reduce the scope of the tests to iterate faster.

Removed and re-running now.

deblasis (approved):
LGTM, solid work

Anything else, let's iterate on it. 

I'd still like @okdas 👀 before merging

okdas (approved):
UPD: welp, should've taken more time looking.

.github/workflows/main.yml
```diff
@@ -50,27 +50,41 @@ jobs:
         run: make install_cli_deps
       - name: generate protobufs, RPC server, RPC client and mocks
         run: make protogen_local && make mockgen && make generate_rpc_openapi
-      - name: run all tests
-        run: make test_all_with_json
+      - name: Create coverage report and run tests
+        # Not utilizing makefile target here to make use of pipefail bash feature.
+        run: |
+          set -euo pipefail
+          make test_all_with_json_coverage
```
okdas, line 57: Actually, @Olshansk can we bring utilizing the CLI back here? I tried to make this work with `Makefile`, but the reason job succeeded is because `set -euo pipefail` didn't propagate into makefile target. Makefile doesn't utilize bash shell. The test still failed.

okdas (changes requested):

.github/workflows/main.yml
okdas: image

What happens is due to the fact we are utilizing `tee`, the `exit(1)` from `go test` never forces the CI job to fail:
```
test_all_with_json_coverage: generate_rpc_openapi # generate_mocks
	go test -p 1 -json ./... -covermode=count -coverprofile=coverage.out | tee test_results.json | jq
```
Due to the fact make targets do not run in bash shell, hence it does not recognize `set -euo pipefail` (the actual setting that forces non-zero exit code to propagate to the CLI execution down the pipe). 

I've tried to make this work on my local PR with different ways, but decided it is best to use go CLI directly, and I put this comment for that reason: `# Not utilizing makefile target here to make use of pipefail bash feature.`

Olshansk: [`6742174` (#285)](https://github.com/pokt-network/pocket/pull/285/commits/67421745f5810449ce3871eada61290c86bb4369)

Rerunning now

okdas (approved):
👍  sweet green them test results

Issue #288, opened 2022-10-05
[Telemetry] Apply shared logging module


jessicadaugherty:
# Objective
Use the foundations to unify and standardize the logging approach across the entire codebase created in https://github.com/pokt-network/pocket/issues/164 and apply to a subset of files to use this framework.

# Origin Document

See the github discussion for full details: https://github.com/pokt-network/pocket/discussions/214

The implementation of this was started by @derrandz here: https://github.com/pokt-network/pocket/pull/162.

This ticket can either build on top of the existing WIP or use those learnings and apply them to a different approach.

# Goals / Deliverables
- [ ] Select subset of files to use the same logging framework
- [ ] Add support for multiple log levels
- [ ] Add support for attaching labels to logs
- [ ] ~~Build a log panel in Grafana where logs can be toggled based on level or label~~ (covered in #189)

## General issue checklist
- [ ] Update the appropriate CHANGELOG
- [ ] Update the appropriate README on how to use and test this feature
- [ ] Update any relevant global documentation & references
- [ ] Add a video and/or gif on how to use feature
- [ ] Document small issues / TODOs along the way

# Non-goals
- Add new logs throughout the codebase
- Improve the explicitness or correctness of existing logs
- Add new business logic functionality

# Testing Methodology
- **All tests**: `make test_all`
- **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)
Remove

---

Creator: @Olshansk

Olshansk:
@jessicadaugherty I've udpated the milestone + labels

jessicadaugherty:
Note to self/team: revisit #301 after completing #288

Olshansk:
![Screenshot 2022-12-22 at 3 56 59 PM](https://user-images.githubusercontent.com/1892194/209227856-50ed4f37-2019-435e-9ca9-7c94ce451d48.png)

@Gustavobelfort As you go into this work and integrate it with Grafana (or other), I wanted to share what my dream feature here would be:

- Being able to filter by level `WARN`, `DEBUG`, etc..
- Filtering / toglging by stuff like `node1.consensus`, `node2.consensus`, etc...
- Filtering / toggling by `REPLICA` / `LEADER`
- Aggregating across `event_*`

I'm sure we'll have all of these over time, but wanted to drop my thoughts in case you have better ideas :)

jessicadaugherty:
[...]

Issue #258, opened 2022-09-29
[Code Structure] Follow protobuf enum best practices


Olshansk:
## Objective

Not following best practices for protobuf enum naming.

## Origin Document

In #245, the following change was made:

<img width="1228" alt="Screen Shot 2022-09-29 at 12 06 46 PM" src="https://user-images.githubusercontent.com/1892194/193120710-ce1e72ce-a522-420b-a9cc-85dd588a1df9.png">

In order to follow [best practices](https://developers.google.com/protocol-buffers/docs/style), we should do the following:

<img width="886" alt="Screen Shot 2022-09-29 at 12 08 20 PM" src="https://user-images.githubusercontent.com/1892194/193124479-86a0d829-9cb6-481b-b74e-f337f12c7ea4.png">

## Goals

- Change all existing protobuf enums to follow best practices
- Enforce best practices moving forward

## Deliverable

- [ ] A single PR that converts existing protobufs (and their references) to follow best practices
- [ ] A single PR to update to update workflow to enforce best practices

## Non-goals / Non-deliverables

- Changing any business logic in the codebase

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references

## Testing Methodology

- **All tests**: `make test_all`
- **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @Olshansk 
**Co-Owners**: @okdas + ?

nickatnight23:
Can I take this on?

Olshansk:
@nickatnight23 Definitely! Feel free to reach out either here or in discord if you need any help or clarifications.

The (dev guide](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md) is probably the best place to start, but if something is unclear or could be made better, please don't hesitate to point it out

nickatnight23:
Can I get some clarification on installing the dependencies? I am having trouble with this section- protoc-gen-go, protoc-go-inject-tag and mockgen by running make install_cli_deps. Not sure how to make the build first?

Olshansk:
Can you try running this command: https://github.com/pokt-network/pocket/blob/main/Makefile#L99

If it works, feel free to update the dev guide as well!

nickatnight23:
[...]

PR #301, opened 2022-10-11
Logger module #164

### LINKED ISSUE #164: [Telemetry] Design shared logging module
# Objective
Create the foundations to unify and standardize the logging approach across the entire codebase.

# Origin Document

See the github discussion for full details: https://github.com/pokt-network/pocket/discussions/214

The implementation of this was started by @derrandz here: https://github.com/pokt-network/pocket/pull/162.

This ticket can either build on top of the existing WIP or use those learnings and apply them to a different approach.

# Goals / Deliverables
- [x] Research and identify common logging frameworks available in Go and use in major codebases (e.g. Uber)
- [ ] Design a new shared logging module based on the results from the research above. This means it could be:
    - Created from scratch
    - Imports an existing framework
    - Extends an existing one with a lightweight wrapper

## General issue checklist
- [ ] Update the appropriate README on how to use and test this feature
- [ ] Add a video and/or gif on how to use feature

# Non-goals
- Add new logs throughout the codebase
- Improve the explicitness or correctness of existing logs
- Add new business logic functionality

# Testing Methodology
Remove

---

Creator: @Olshansk 
Co-Owners: @vlad-she @derrandz

okdas:
## Description

Logger module addition. The module will allow us to utilize structured logs for better observability.

## Issue

Fixes #164 

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- New module addition

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


build/config/config1.json
```diff
@@ -31,5 +31,9 @@
     "enabled": true,
     "address": "0.0.0.0:9000",
     "endpoint": "/metrics"
+  },
+  "logger": {
```
okdas, line 35: if we change the name of the module to "log" that would look prettier I think.

logger/proto/logger_config.proto
```diff
@@ -0,0 +1,23 @@
...
+package logger;
+
+option go_package = "github.com/pokt-network/pocket/logger";
+
+message LoggerConfig {
+  LogLevel level = 1;
+  LogFormat format = 2;
+}
+
+enum LogLevel {
```
okdas, line 11: I thought it's better to utilize enum since we know what potential values can be.

shared/modules/logger_module.go
```diff
@@ -0,0 +1,20 @@
...
+
+//go:generate mockgen -source=$GOFILE -destination=./mocks/logger_module_mock.go -aux_files=github.com/pokt-network/pocket/shared/modules=module.go
+import (
+	"github.com/rs/zerolog"
+)
+
+type Logger = zerolog.Logger
+
+type LoggerModule interface {
+	Module
```
deblasis, line 11: I guess `LoggerModule` doesn't have to embed all the interfaces from `Module` since most of them are unused/no-ops.

We can always revisit it once we merge in the config/genesis PR, as we speak, this is how `module.go` looks like https://github.com/pokt-network/pocket/pull/235/files#diff-ae3949fd3b550b6789e9ec9f0557c73dc0b3eb99b91048bf8ad4406a0b77f1abR9

deblasis (commented):
I like where this is going, just a few comments and happy to help if you have questions

Olshansk:
@okdas Let me know if/when you need another set of 👀 at this

app/client/main.go
```diff
@@ -45,26 +46,36 @@ var p2pMod modules.P2PModule
 // A consensus module is initialized in order to get a list of the validator network
 var consensusMod modules.ConsensusModule
 
+var globalLogger = logger.GlobalLogger()
```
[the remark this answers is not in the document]
okdas: @deblasis thank you, and that makes sense. Though I was under the impression it is better to have just one instance of `LoggerModule`, and all other modules would branch out of the that one instance (via bus), we certainly could have a separate instance of `LoggerModule` for global logger.

shared/modules/logger_module.go
okdas: If we do that, then this signature is going to be violated: `	Create(runtime RuntimeMgr) (Module, error)`

logger/proto/logger_config.proto
```diff
@@ -0,0 +1,25 @@
...
+  debug = 0;
+  info = 1;
+  warn = 2;
+  error = 3;
+  fatal = 4;
+  panic = 5;
+}
+
+enum LogFormat {
```
Olshansk, line 22: If we decide to keep enums, please follow best practices: https://github.com/pokt-network/pocket/issues/258

logger/proto/logger_config.proto
Olshansk: > Still keeping enums because they can be useful for config validation.

If we aren't going to be using them, we should remove it.

> with the new config management via runtimeMgr I don't think I can use protobof types anymore.

Could you explain why? Let's see if we can update runtimeMgr for it.

shared/modules/logger_module.go
```diff
@@ -0,0 +1,20 @@
...
+
+	// TODO(@okdas): move that to Module interface, so each module has its own logger (with context)
+	// https://github.com/pokt-network/pocket/issues/288
```
Olshansk, line 14:
```suggestion
	// TODO(#288): Embed `ObservableModule` inside of the `Module` interface so each module has its own context specific logger
```

shared/modules/logger_module.go
```diff
@@ -0,0 +1,20 @@
...
+	ObservableModule
+
+	// CreateLoggerForModule returns the logger with additional context (module name)
```
Olshansk, line 17:
```suggestion
	// `CreateLoggerForModule` returns the logger with additional context (e.g. module name)
```

shared/modules/module.go
```diff
@@ -41,3 +41,11 @@ type KeyholderModule interface {
...
 	GetP2PAddress() cryptoPocket.Address
 }
+
+// TODO(@okdas): this should also initialize metrics-related stuff, just the logger for now
+// InitMetrics()
+// InitTracing()
+type ObservableModule interface {
+	InitLogger()
+	GetLogger() Logger
+}
```
Olshansk, line 51:
```suggestion
// TODO(@okdas): this should also initialize metrics related configs, but only logging is implemented at the moment
type ObservableModule interface {
	InitLogger()
	GetLogger() Logger
	
	// InitMetrics()
        // InitTracing()
}
```

logger/module.go
```diff
@@ -0,0 +1,90 @@
...
+
+func Create(runtimeMgr modules.RuntimeMgr) (modules.Module, error) {
+	return new(loggerModule).Create(runtimeMgr)
+}
+
+func (*loggerModule) CreateLoggerForModule(moduleName string) modules.Logger {
+	return mainLogger.With().Str("module", moduleName).Logger()
+}
+
+func (*loggerModule) Create(runtimeMgr modules.RuntimeMgr) (modules.Module, error) {
```
Olshansk, line 38: When do we use this Create versus the one long line 30?

Olshansk (changes requested):
Left a couple small comments, but overall it's looking really clean!

logger/proto/logger_config.proto
okdas: @Olshansk I don't mind removing them, but is there a place I can document possible config values? If no, I can start a config file documentation page.

logger/proto/logger_config.proto
okdas: > Could you explain why? Let's see if we can update runtimeMgr for it.

Sure! So if we were to utilize enum types in the *Config types, this would trigger an import cycle, here's an example:
https://github.com/okdas/pocket/commit/3d83f223aec8ed96521ebe6da28262238bb565b4

Please let me know if you think of a better way of utilizing enums! :)

logger/module.go
okdas: `logger.Create` is used in https://github.com/pokt-network/pocket/pull/301/commits/ea447886932c4711de5ebba42af308f073858ccf#diff-bcff8b319bd7c6f37325379309e118b3310f7308755b8ba7fb6a5de101c4d9edR76 - just pushed this.

build/config/config1.json
Olshansk: @okdas Please resolve comments if no further feedback/discussion is required.

logger/proto/logger_config.proto
Olshansk: Got it. Keep the enum encapsulated to the `shared` repo and use a string in shared.

In shared, you can add a comment that it is an enum that's being externally represented as a string due to cyclic dependencies and that we'll look into a way to improve it in the future.

Olshansk (changes requested):

logger/proto/logger_config.proto
okdas: @Olshansk this is the part I've got most of my confusion from.
The reason I've suggested to use enums here for is because we know potential values for the configuration - and they cannot be any different from what we specify in enum. Think of this as of values validation/verification. If we specify string here, and we want to continue to use enums, I need to add a validation step on create to verify string value against our enums.

logger/proto/logger_config.proto
okdas: @Olshansk I've got familiar with the best practices doc, but I am not sure if they are suitable in this use-case.

My point of utilizing enubs for configuration is because we can specify potential static configuration values. No other parameters can be used but the ones defined in enum. If we follow the best practices document, then we need to do a couple of things:
1. rename the values: e.g., `debug` becomes `LOG_LEVEL_DEBUG`, which is not suitable as a value for that configuration parameter, we want `debug`. If we do rename `debug` to `LOG_LEVEL_DEBUG`, we'd need to then again do conversion from `LOG_LEVEL_DEBUG` to `debug` in `Create()`. Sounds very unnecesary.
2. introduce `UNSPECIFIED` value, which is not a valid value for configuration parameters.

Considering the values passed to the library - zerolog - I'd expect them to also do some verification so I'd ditch enums completely in case you want to enforce the enum best practices in that case. 

I totally agree - enums best practices make 100% sense when they are used for communication like grpc, but in this case this doesn't seem correct to me.

logger/module.go
Olshansk: Discussed offline. We are following the pattern elsewhere in the codebase. Resolving.

logger/proto/logger_config.proto
Olshansk: Discussed offline, we will go with an enum for Pocket specific logging configs along with a mapper that maps the enums to `zerolog` specific strings. The following is just example pseudocode:

```
var pocketToZeroLogMappingConfig = map[EnumType]string{
	EnumType_LOG_LEVEL_DBEUG: "debug",
}
```

logger/proto/logger_config.proto
Olshansk: Discussed offline. We will go with an enum + mapper approach and "let it fail" if the logging config is incorrect, which is valid.

shared/modules/logger_module.go
okdas: We discussed this offline. Create() can be changed to return `InitializableModule` instead of `Module`. Ticket for this change: https://github.com/pokt-network/pocket/issues/333

shared/modules/module.go
```diff
@@ -41,3 +41,12 @@ type KeyholderModule interface {
 type P2PAddressableModule interface {
 	GetP2PAddress() cryptoPocket.Address
 }
+
+// TODO(@okdas): this should also initialize metrics related configs, but only logging is implemented at the moment
+type ObservableModule interface {
+	InitLogger()
+	GetLogger() Logger
+	
+	// InitMetrics()
```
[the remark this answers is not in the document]
okdas: This is something I'll do/look at when I'll work on observability metrics anyway. I don't think we need a special ticket for this.

consensus/consensus_tests/utils_test.go
```diff
@@ -408,6 +409,16 @@ func baseTelemetryEventMetricsAgentMock(t *testing.T) *modulesMock.MockEventMetr
 	return eventMetricsAgentMock
 }
 
+func baseLoggerMock(t *testing.T, _ modules.EventsChannel) *modulesMock.MockLoggerModule {
+	ctrl := gomock.NewController(t)
+	loggerMock := modulesMock.NewMockLoggerModule(ctrl)
+
+	loggerMock.EXPECT().Start().Do(func() {}).AnyTimes()
+	loggerMock.EXPECT().SetBus(gomock.Any()).Do(func(modules.Bus) {}).AnyTimes()
```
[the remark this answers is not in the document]
okdas: Yep - I was following the pattern, but it doesn't seem we need to test for this since these functions are no-ops anyway.

app/client/cli/debug.go
```diff
@@ -189,21 +190,24 @@ func initDebug(remoteCLIURL string) {
 			log.Fatalf("[ERROR] Failed to create p2p module: %v", err.Error())
 		}
 		p2pMod = p2pM.(modules.P2PModule)
-		if err != nil {
```
okdas, line 192: removing since this would never happen (already checked few lines above)

shared/modules/types.go
```diff
@@ -60,6 +61,13 @@ type TelemetryConfig interface {
 	GetEndpoint() string
 }
 
+type LoggerConfig interface {
+	// We have protobuf enums for the following values, but they are represented as
```
Olshansk, line 65: > We have protobuf enums

Add a file name or package name in brackets.

shared/modules/types.go
```diff
@@ -60,6 +61,13 @@ type TelemetryConfig interface {
...
+	// `strings` to avoid circular dependencies.
```
Olshansk, line 66: Circular dependencies between which two packages?

runtime/config.go
```diff
@@ -77,3 +79,6 @@ func (c *runtimeConfig) GetP2PConfig() modules.P2PConfig {
 func (c *runtimeConfig) GetTelemetryConfig() modules.TelemetryConfig {
 	return c.Telemetry
 }
+func (c *runtimeConfig) GetLoggerConfig() modules.LoggerConfig {
```
Olshansk, line 82: NIT: Can you add newlines between these functions.

logger/proto/logger_config.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+package logger;
+
+option go_package = "github.com/pokt-network/pocket/logger";
+
+message LoggerConfig {
+  // LogLevel level = 1;
```
Olshansk, line 7: Remove/revert?

app/client/cli/debug.go
```diff
@@ -189,21 +190,24 @@ func initDebug(remoteCLIURL string) {
...
 		telemetryMod := telemetryM.(modules.TelemetryModule)
 
-		_ = shared.CreateBusWithOptionalModules(runtimeMgr, nil, p2pMod, nil, consensusMod, telemetryMod) // TODO: refactor using the `WithXXXModule()` pattern accepting a slice of IntegratableModule
+		loggerM, err := logger.Create(runtimeMgr)
+		if err != nil {
+			logger.Global.Fatal().Err(err).Msg("Failed to create logger module")
+		}
+		loggerMod := loggerM.(modules.LoggerModule)
+
+		_ = shared.CreateBusWithOptionalModules(runtimeMgr, nil, p2pMod, nil, consensusMod, telemetryMod, loggerMod) // TODO: refactor using the `WithXXXModule()` pattern accepting a slice of IntegratableModule
```
Olshansk, line 210: See https://goldin.io/blog/stop-using-todo ###

```suggestion
		_ = shared.CreateBusWithOptionalModules(runtimeMgr, nil, p2pMod, nil, consensusMod, telemetryMod, loggerMod) // REFACTOR: use the `WithXXXModule()` pattern accepting a slice of IntegratableModule
```

logger/module.go
```diff
@@ -0,0 +1,103 @@
...
+	cfg := runtimeMgr.GetConfig()
+	m := loggerModule{
+		config: cfg.GetLoggerConfig(),
+	}
+
+	m.InitLogger()
+
+	// Mapping config string value to the proto enum
+	pocketLogLevel := LogLevel(LogLevel_value[`LogLevel_LOG_LEVEL_`+strings.ToUpper(m.config.GetLevel())])
```
Olshansk, line 63: Can you do a map check and return an error if `!ok`: https://stackoverflow.com/questions/2050391/how-to-check-if-a-map-contains-a-key-in-go

Olshansk (changes requested):
@deblasis Could you take a look at this when you have a chance as well?

Olshansk:
@okdas Please re-request a review when this is ready again.

Also, please update the description of title in this PR since it's in review and not just a WIP anymore.

okdas:
@deblasis please take a look at this if you've got a moment. I want to make sure I'm utilizing the foundation you've built correctly.

shared/modules/logger_module.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/rs/zerolog"
+)
+
+type Logger = zerolog.Logger
+
+type LoggerModule interface {
+	Module
+
+	// TODO(#288): Embed `ObservableModule` inside of the `Module` interface so each module has its own context specific logger
+	ObservableModule
```
deblasis, line 14: I think that `LoggerModule` _maybe_ is not also an `ObservableModule`.

```suggestion
```

Where would you use `LoggerModule.GetLogger()` if in each module you are getting the `LoggerModule` from the bus and then you are calling `CreateLoggerForModule(someModule.GetModuleName())` in *their* `GetLogger`?

The other modules, yes, they are. And I also like the `Global` logger to log for all the other places. 

If you agree with me, I'd also remove these

```go
func (m *loggerModule) InitLogger() {
	m.logger = m.CreateLoggerForModule(m.GetModuleName())
}

func (m *loggerModule) GetLogger() modules.Logger {
	return m.logger
}
```
but I appreciate the fact that they show how you would use them in a module.

deblasis (approved):
Solid work, just a question that potentially is not an action item, please let me know if you think that I am getting it wrong.

Everything else, I believe that when the rubber hits the road in #288 there might be (or not) some small changes to this but the foundation looks pretty good to me.

shared/modules/logger_module.go
jessicadaugherty: Ty @deblasis left a comment on #288 just in case to capture:

> I believe that when the rubber hits the road in https://github.com/pokt-network/pocket/issues/288 there might be (or not) some small changes to this but the foundation looks pretty good to me.

Olshansk (approved):
👀 👌 

Really appreciate the scoping of this so we could extend it elsewhere afterwards!

PR #342, opened 2022-11-09
[Consensus] State Sync - Block by Block Design (issue-#125)

### LINKED ISSUE #125: [Consensus] State Sync - Documentation & Initial Block By Block Design
# Objective
Design an interface and outline a flow for a block by block state sync

# Origin Document
As described in the [V1 Pocket Network Consensus Specification](https://github.com/pokt-network/pocket-network-protocol/tree/main/consensus), we need to fully implement state sync to enable new and existing nodes to stay in sync with each other.

# Goals / Deliverables
- [ ] A sequence diagram showing the flow of blocks and data between different components during state sync using [mermaid flow charts](https://mermaid-js.github.io/mermaid/#/sequenceDiagram). For an example, see what we have in the [persistence module](https://github.com/pokt-network/pocket-network-protocol/tree/main/persistence#562-state-transition-sequence-diagram) architecture doc.
- [ ] A flow chart showing which components and modules are involved in this process using [mermaid flow charts](https://mermaid-js.github.io/mermaid/#/flowchart)
- [ ] The interface that each node will need to expose in order to request & receive blocks
- [ ] Follow up tasks of things that need to be design and implemented (e.g. fast sync design, slow sync implementation, etc...)

# Non-goals
- Implementing any part of state sync (i.e. no code)
- Fast sync design (i.e. no synching from a snapshot yet)

## Industry References

As a starting point, we can reference how other chains implement this mechanism.

### Tendermint

- https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-042-state-sync.md

### Flow

- https://github.com/onflow/flow-go/tree/master/cmd/consensus#synchronization

### Aptos
- https://github.com/aptos-labs/aptos-core/blob/30b385bf38d3dc8c4e8ee0ff045bc5d0d2f67a85/state-sync/README.md

---

Creator: @Olshansk 
Co-Owners: @andrewnguyen22

andrewnguyen22:
## Description

Design an interface and outline a flow for a block by block state sync

As described in the [V1 Pocket Network Consensus Specification](https://github.com/pokt-network/pocket-network-protocol/tree/main/consensus), we need to fully implement state sync to enable new and existing nodes to stay in sync with each other.

## Issue

Fixes #125 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

- [x] A sequence diagram showing the flow of blocks and data between different components during state sync using [mermaid flow charts](https://mermaid-js.github.io/mermaid/#/sequenceDiagram). For an example, see what we have in the [persistence module](https://github.com/pokt-network/pocket-network-protocol/tree/main/persistence#562-state-transition-sequence-diagram) architecture doc.
- [x] A flow chart showing which components and modules are involved in this process using [mermaid flow charts](https://mermaid-js.github.io/mermaid/#/flowchart)
- [x] The interface that each node will need to expose in order to request & receive blocks
- [x] Follow up tasks of things that need to be design and implemented (e.g. fast sync design, slow sync implementation, etc...)

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,106 @@
...
+    A[StateSync] -->|IsCaughtUp| B(Pacemaker Mode)
+    B --> |Consensus Messages| C(ConsensusModule.Pacemaker)
+    A -->|IsSyncing| E(Sync Mode)
+    E -->|Request Block| G[Peers]
+    G--> |ApplyBlock| A
+    A --> D[Server Mode]
+    D --> |Serve Blocks Upon Request| G
+```
+
+## Follow up tasks
```
Olshansk, line 82: Optional for all the bullet points below: Consider using `*` to make these lists.

consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,106 @@
...
+
+`Fast Sync Design` - Sync only the last `N` blocks from `Latest Network Height`
+
+`Parallel Sync Design` - Optimize the State Sync protocol by parallelizing requests and computation
+
+## Research Items
```
Olshansk, line 88: Thoughts on moving these into an GH issue (or GH discussion) titled `[Core] Evaluate & Prioritize State Sync Improvements`. That way we can add more (through edits or comments) ideas as they come up and have a discussion around them.

It can be an M5 ticket.

consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,106 @@
...
+    G--> |ApplyBlock| A
+    A --> D[Server Mode]
+    D --> |Serve Blocks Upon Request| G
+```
+
+## Follow up tasks
+
+`Fast Sync Design` - Sync only the last `N` blocks from `Latest Network Height`
+
+`Parallel Sync Design` - Optimize the State Sync protocol by parallelizing requests and computation
```
Olshansk, line 86: I think the canonical name for this might be `Optimistic Sync`: https://geth.ethereum.org/docs/interface/sync-modes.

```
Optimistic sync downloads blocks before the execution client has validated them. In optimistic sync the node assumes the data it receives from its peers is correct during the downloading phase but then retroactively verifies each downloaded block. Nodes are not allowed to attest or propose blocks while they are still ‘optimistic’ because they can’t yet guarantee their view of the head of the chain is correct.
```

consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,106 @@
...
+    A -->|IsSyncing| E(Sync Mode)
+    E -->|Request Block| G[Peers]
+    G--> |ApplyBlock| A
+    A --> D[Server Mode]
+    D --> |Serve Blocks Upon Request| G
+```
+
+## Follow up tasks
+
+`Fast Sync Design` - Sync only the last `N` blocks from `Latest Network Height`
```
Olshansk, line 84:
```suggestion
`Checkpoint Sync Design` - Sync only the last `N` blocks from a previously saved state or a trusted source
```

consensus/doc/PROTOCOL_STATE_SYNC.md
Olshansk, line 82: Consider Adding

`Block Chunk Design` - Update the `Block by Block` design to be able to receive and provide multiple blocks at a time.

Source for idea: `Snapshot chunks` from evmos: https://docs.evmos.org/validators/setup/statesync.html#check-logs

`Block Stream Design` - Update the `Block by Block` design to stream blocks via a WebSocket from a single connectable peer.

Source for idea: `Stremable` approach from Chia Network https://docs.chia.net/networking-protocol

consensus/state_sync/state_sync.go
```diff
@@ -0,0 +1,104 @@
...
+	GetPeers() []PeerSyncMeta
+	// returns ordered array of missing block heights
+	GetMissingBlockHeights() []int64
+	// set missing block height if an error occurred during processing
+	SetMissingBlockHeight(blockHeight int64) error
+}
+
+type BlockRequestMessage interface {
+	// the height the peer wants from the block store
+	GetRequestedHeight() int64
```
Olshansk, line 88: Given that the interface is named `BlockRequestMessage`, I think `GetHeight` is self explanatory

consensus/state_sync/state_sync.go
```diff
@@ -0,0 +1,104 @@
...
+	// latest local height
+	LatestHeight() int64
+	// latest network height (from the aggregation of Peer Sync Meta)
+	LatestNetworkHeight() int64
+	// retrieve peer meta (actively updated through churn management)
+	GetPeers() []PeerSyncMeta
+	// returns ordered array of missing block heights
+	GetMissingBlockHeights() []int64
+	// set missing block height if an error occurred during processing
+	SetMissingBlockHeight(blockHeight int64) error
```
Olshansk, line 83: What did you have in mind for this?

I feel like the internal state of the module should accurately reflect `LatestHeight` and `GetMissingBlockHeights` depending on how it operates so this shouldn't be necessary.

consensus/state_sync/state_sync.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// `HandleBlock` function contract:
+	// - Create a Utility Context
+	// - Block.ValidateBasic()
+	// - Consensus Module Replica Path
+	//   - Prepare Block:  utilityContext.SetProposalBlock(block)
+	//   - Apply Block:    utilityContext.ApplyBlock(block)
+	//   - Validate Block: utilityContext.AppHash == Block.AppHash
+	//   - Store Block:    consensusModule.CommitBlock()
+	SetHandleBlock(HandleBlock func(block *typesCons.Block) error)
```
Olshansk, line 22: The callback paradigm feels very `cosmosy` to me and I'd like to avoid it if possible.

The application specific bus currently supports two approaches of communication:
1. Synchronous (e.g. `m.GetBus().GetModule().CallFoo(...)`); currently used in most places
2. Asynchoronous (e.g. `node.GetBus().PublishEventToBus(e)`); currently used by P2P to send messages to consensus

#2 flows through the following code path:
```go
	// While loop lasting throughout the entire lifecycle of the node to handle asynchronous events
	for {
		event := node.GetBus().GetBusEvent()
		if err := node.handleEvent(event); err != nil {
			log.Println("Error handling event: ", err)
		}
	}
```

The idea here was that it's easy to log, capture errors, telemetry, etc for this that are async in a centralized manner.

I would prefer that this be changed to:

```
HandleBlock(*typesCons.Block) error
```

This way we have optionality on it being called through (1) OR (2) and not need to manage a bunch of async callbacks comes and going to different places.

consensus/state_sync/state_sync.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// `GetPeerSyncMeta` function contract:
+	// - Retrieve a list of active peers with their metadata (identified and retrieved through P2P's `Churn Management`)
+	SetGetPeerMetadata(GetPeerSyncMeta func() (peers []PeerSyncMeta, err error))
+
+	// `NetworkSend` function contract:
+	// - sends data to an address via P2P network
+	SetNetworkSend(NetworkSend func(data []byte, address cryptoPocket.Address) error)
+
+	// -- Sync modes --
```
Olshansk, line 32: Alternate design pattern for sync modes.

If you think of `mode` as an enum, we can do:

```go
StartMode(mode)
StopMode(mode)
```

This was we can (for example) turn off `ServerMode` if we need to based on some configs and a state the node enters.

Lmk what you think.

consensus/state_sync/state_sync.go
```diff
@@ -0,0 +1,104 @@
...
+	// the height the peer wants from the block store
+	GetRequestedHeight() int64
+}
+
+type BlockResponseMessage interface {
+	// the bytes of the requested block from the block store
+	GetBlockBytes() []byte
+}
+
+// TODO: needs to be shared between P2P as the Churn Manageent Process updates this information
```
Olshansk, line 96: Can you create a ticket for this TODO?

I think understanding what part of the interface will be exposed under `shared/modules/p2p_module.go` is important since it'll affect @deblasis work on peer discovery & churn.

Olshansk (changes requested):
@andrewnguyen22 I did a first round of review and covered about half the content.

About 90% of the comments I left are just optional NITs/edits/clarifications (from a reader's perspective), and about 10% have more implications on the actual design so am curious to get your initial thoughts.

consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,106 @@
...
+```golang
+type PeerSyncAggregate struct {
+  Peers []PeerSyncMeta // The current list of peers and the known metadata
+  MaxPeerHeight int64  // The maximum height associated with all known peers
+  ...
+}
+```
+Using the `PeerSyncAggregate`, a Node is able to understand its local `SyncState` against that of the Global Network.
+
+### Pacemaker Mode
```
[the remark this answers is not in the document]
andrewnguyen22: This is a really interesting point... 

If the pacemaker only works for Validators then it's inevitable that the State Sync protocol will have to distinguish between `Full Nodes` and `Validators`. Unless we **remove pacemaker mode** in favor of **SyncedMode**.

I'd much prefer the `State Sync` protocol be agnostic to the node type...

I think we should replace `Pacemaker Mode` with `Synced Mode` and limit the Pacemaker to just play with consensus but allow it to pass a block to the `State Sync` protocol with something like a 'self peer - block response message'...

WYT?

consensus/state_sync/state_sync.go
andrewnguyen22: Blocks are set as `!Missing` when received from a peer but before validation to not block the block download loop.

If a received block is invalid - it must `Set` a block height as `Missing` again

consensus/state_sync/state_sync.go
andrewnguyen22: Hmmm `triggered` by **Cosmosy** xd

Okay so I see your point, however - technically speaking - callbacks are more package agnostic and could be used in another application without requiring an `AppBus`. 

However, to your point, we are building for Pocket. 

WYT?

andrewnguyen22: Assuming you're for the bus here; 

How would you expect to reflect using the bus `In_This_Commit` when we just have interfaces.

Describe it in a comment?

consensus/state_sync/state_sync.go
andrewnguyen22: It might be better to implement `modes` as `modules` so they have **Service Like** interfaces like `module.Start() and module.Stop()` 

WYT?

andrewnguyen22:
@Olshansk good review; Responded to heavy hitting points. Will tackle `nits` tomorrow

consensus/state_sync/state_sync.go
Olshansk: tl;dr `StateSyncModule` would have a function called `HandleStateSyncMessage(typesCons.StateSyncMessage) error` that is expected to be called by `ConsensusModule.HandleMessage`, and `In_This_Commit`, we would have a protobuf called

```proto
enum StateSyncMessageType {
...
    STATE_SYNC_PROVIDE_BLOCK = N;
...
}

message StateStateProvideBlockMessage {
    Block block = 1;
}

message StateSyncMessage {
    StateSyncMessageType type = 1;
    google.protobuf.Any message = 2;
    ...
}
```

That gives us the ability to send/receive various types of state sync messages between nodes.

If we go with this approach, it might require changing the flow a bit, because there's room for making the type "Provide, Request, etc...", but also makes it easier to adapt in the future within a switch case statement we implement.

--- 

The consensus module has a function called `HandleMessage(*anypb.Any) error`.

That function is implemented like so:

```go
func (m *consensusModule) HandleMessage(message *anypb.Any) error {
	m.m.Lock()
	defer m.m.Unlock()
	switch message.MessageName() {
	case HotstuffMessage:
		msg, err := codec.GetCodec().FromAny(message)
		if err != nil {
			return err
		}
		hotstuffMessage, ok := msg.(*typesCons.HotstuffMessage)
		if !ok {
			return fmt.Errorf("failed to cast message to HotstuffMessage")
		}
		if err := m.handleHotstuffMessage(hotstuffMessage); err != nil {
			return err
		}
	case UtilityMessage:
		panic("[WARN] UtilityMessage handling is not implemented by consensus yet...")
	default:
		return typesCons.ErrUnknownConsensusMessageType(message.MessageName())
	}

	return nil
}
```

Which uses a protobuf like so:

```proto
message HotstuffMessage  {
    HotstuffMessageType type = 1;
    uint64 height = 2;
    ...
```

My assumption is that when we call `RunSyncMode`, it would do things like `m.GetBus().GetP2P().Send(<GiveMeBlockX>)` and eventually `StateSyncModule.HandleMessage(msg)` would get triggered asynchronously.

This is quite different from what we have right now, but I think might give the implementation more flexibility when we get there and more extendability in the future.

consensus/state_sync/state_sync.go
Olshansk: Yes! We simply stop synching once it's synched!!!

consensus/state_sync/state_sync.go
Olshansk: If we have a package (including a struct) implementing the `StateSyncModule`, I would assume it has an internal state that is modified based on some of the functions exposed (e.g. `HandleBlock`).

This way, the state internals of the implementation knows if a block is missing or not without having an external party modify it. The decision as to whether the module should discard additional `HandleBlock` for a height it has cached, but not processed, is left up to the implementor.

My goal here is to expose as few functions as necessary in the interface so the "contract" is simple, which gives more optionality to the implementor.

consensus/doc/PROTOCOL_STATE_SYNC.md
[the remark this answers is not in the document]
andrewnguyen22: Bump @Olshansk

consensus/state_sync/state_sync.go
andrewnguyen22: I'm not convinced this is self explanatory, but also - this would be the interface needed to be implemented, unless I'm missing something?

consensus/state_sync/state_sync.go
andrewnguyen22: So basically what you're saying is *this should be internal* - that's fine by me

consensus/state_sync/state_sync.go
andrewnguyen22: Disagree with the needs for structures or protobuf `IN_THIS_COMMIT` - but I understand the point to be:

TL;DR - 

```golang
// Replace
SetHandleBlock(HandleBlock func(block *typesCons.Block) error)

// With
HandleStateSyncMessage(msg BlockResponseMessage)
```

Also - any thoughts on my feedback:
> callbacks are more package agnostic and could be used in another application without requiring an AppBus.

consensus/state_sync/state_sync.go
andrewnguyen22: Perf - the interface implements `modules.Module`

consensus/state_sync/state_sync.go
andrewnguyen22: What's the AI here?

consensus/state_sync/state_sync.go
andrewnguyen22: https://github.com/pokt-network/pocket/issues/357

consensus/doc/PROTOCOL_STATE_SYNC.md
andrewnguyen22: For research items or for follow up tasks. In anycase are you suggesting that the PROTOCOL.md file is not the best place for these - as I'm not a fan of duplicating TODO's

consensus/state_sync/state_sync.go
Olshansk: > callbacks are more package agnostic and could be used in another application without requiring an AppBus.

I can see myself arguing both sides, but ultimately it's about tradeoffs and the overall design of the system.
Every system needs have the ability to send message synchronously and asynchronously.

**Synchronous**:
   * Passing data around manually
   * Having a Bus and calling functions similar to what we have
   * A dependency injection framework
   * A design pattern similar to react's reflex

**Asynchronous**:
   * Pub/sub
   * Callbacks
   * Promises
   * Async/await

Re callbacks:
- Pro: it's asynchronous, and there is no "central nervous system" (i.e. bus)
- Con: are all the callbacks going to be set and/or defined in one place or will it be dispersed?

Re dependency injection:
- Pro: a single place defines everything that we inject into the system
- Con: Customizing injections on a per module/submodule place can get difficult, and it's hard for different modules to be independent.

Re pub/sub:
- Pro: A really great design (in theory) that's completely independent from the core logic and everything is defined via topics and messages.
- Con: These things are really hard to debug, and come with tradeoffs (at least once, at most once, exactly once, etc...) and are often hard to debug


This is just a brought set of points / ideas, but I can go into more depth if you'd like.

My biggest concern, and the reason I feel strongly against callbacks, is because having too many different patterns can lead to requiring different types of tooling for debugging/visibility, different workflows, different types of test suites, different mindsets/approaches to different parts of the codebase, and overall disparity when trying to understand how the whole system works

The app bus in this case, is similar to some sort of object that sets the callback. Calling the function based on an event is a continuous runtime thing, whereas setting the callback is a one time event when the application starts up. My preference is the former is because I've found it to be much easier to debug. There is a single place where you can add logging, a breakpoint, etc, and take it from there.

---

I wanted to keep this "short" but happy to provide a more formulated argument around this. Lmk

Olshansk: > Disagree with the needs for structures or protobuf IN_THIS_COMMIT

That's fair. I don't think we have a great definition of what should & shouldn't go into the interface design PR (e.g. should we include protobufs or not), so it's up to the architect.

We could add a comment saying "Implementation may need to define a set of protobufs" and I think identifying them during implementation will also be easier and help reduce the scope of this PR independently.

consensus/state_sync/state_sync.go
Olshansk: I was thinking of replacing `GetRequestedHeight` with `GetHeight`. If these were a protobuf, the message would be:

```
message BlockRequestMessage {
  height int64 = 1;
  ...
}
```

as opposed to:

```
message BlockRequestMessage {
  height_request int64 = 1;
  ...
}
```

Olshansk (changes requested):

consensus/doc/PROTOCOL_STATE_SYNC.md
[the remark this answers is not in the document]
andrewnguyen22: State sync has little to do with `Proposals` 

It is an engine to process `committed blocks` - but yes. Your understanding is correct

consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,118 @@
...
+
+In the `StateSync` protocol, the Node fields valid `BlockRequests` from its peers to help them `CatchUp` to be `Synced`. This sub-protocol is continuous throughout the lifecycle of StateSync.
+
+```mermaid
+graph TD
+    A[StateSync] -->|IsCaughtUp| B(Pacemaker Mode)
+    B --> |Consensus Messages| C(ConsensusModule.Pacemaker)
+    A -->|IsSyncing| E(Sync Mode)
+    E -->|Request Block| G[Peers]
+    G--> |ApplyBlock| A
```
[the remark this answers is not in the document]
andrewnguyen22: It's showing the `requestLoop`

**Request Block** -> Peers -> **ApplyBlock** -> REPEAT StateSync

consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,119 @@
...
+If the Node is `Synced` or `localSyncState.Height == GlobalSyncMeta.Height` then the `StateSync` protocol is in `SyncedMode`.
+
+In `SyncedMode`, the Node is caught up to the latest block and relies on new blocks to be propagated via the P2P network after Validators reach consensus.
+
+### Sync Mode
+If the Node is `Syncing` or `localSyncState.Height < GlobalSyncMeta.Height` then the `StateSync` protocol is in `SyncMode`.
+
+In `SyncMode`, the Node is catching up to the latest block by making `BlockRequests` to its fellow eligible peers. A peer is eligible for a `BlockRequest` if `PeerMeta.MinHeight` <= `self.MaxBlockHeight` <= `PeerMeta.MaxHeight`.
+
+Though it is `unspecified` whether or not a Node may make `BlockRequests` in order or in parallel, the cryptographic restraints of block processing require the Node to call `ApplyingBlock` sequentially until it is `Synced`.
```
Olshansk, line 71:
```suggestion
Though it is `unspecified` whether or not a Node may make `BlockRequests` in order or in parallel, the cryptographic restraints of block processing require the Node to call `ApplyBlock` sequentially until it is `Synced`.
```

consensus/doc/PROTOCOL_STATE_SYNC.md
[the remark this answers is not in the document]
Olshansk: Got it!

consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Glossary
+
+`ApplyingBlock`: The process of playing block parts and its subsequent transactions against the Node's world state using the Utility Module and Validating the `AppHash` contained in the block against the produced `AppHash` from the local state.
+
+`BlockRequests`: A message from an active peer, requesting a block to sync the chain to the Global Network State.
+
+`Churn Management Protocol`: The protocol in Pocket's P2P Module that ensures the most updated and valid Network Peer list possible.
+
+`Network Peer`: A node that is directly connected and sharing resources without going through a third party server. Peers may start the connection through an `inbound` or `outbound` initialization.
```
Olshansk, line 117: I feel like `resources` can be interpreted as compute resources so I rewarded it a bit.

```suggestion
`Network Peer`: Another node on the network that this node can directly communicate with, without going through a third-party server. Peers may start the connection through an `inbound` or `outbound` initialization to share and transmit data.
```

consensus/doc/PROTOCOL_STATE_SYNC.md
```diff
@@ -0,0 +1,119 @@
...
+## Research Items
+TODO(M5): Create issues to track and discuss these work items.
+
+How the persistence layer design of `pruning` the Merkle Tree affects `StateSync`.
+
+How DB Pruning of the `SQL DB` might affect fast sync.
+
+How the Churn Management operations of RainTree might provide opportunities and obstacles with StateSync.
+
+## Glossary
```
Olshansk, line 109: Question: do we need to differentiate between protocol-level actors and simply full nodes?

In the context of RainTree, I know we've discussed that list, being the same across all actors, is a function of the world state. For state sync, I believe full nodes are fair game as well.

I think adding a comment about this could be a helpful distinction.

consensus/state_sync/state_sync.go
```diff
@@ -0,0 +1,102 @@
...
+	// Uses 'NetworkSend' to send a `BlockResponseMessage` to a specific peer
+	// This function is used in 'ServerMode()'
+	HandleBlockRequest(message BlockRequestMessage) error
+
+	// Uses `HandleBlock` to process retrieved blocks from peers
+	// Must update sync state using `SetMissingBlockHeight`
+	ProcessBlock(block *typesCons.Block) error
+}
+
+type SyncState interface {
```
Olshansk, line 73: In the `.md` file, we define `PeerSyncAggregate` and `PeerSyncMeta`, but neither is used here.

Are the former structs just for illustrative purposes (not implementation)?

consensus/state_sync/state_sync.go
```diff
@@ -0,0 +1,102 @@
+package state_sync
+
+import (
+	typesCons "github.com/pokt-network/pocket/consensus/types"
+	cryptoPocket "github.com/pokt-network/pocket/shared/crypto"
+	"github.com/pokt-network/pocket/shared/modules"
+)
+
+type StateSyncModule interface {
+	modules.Module
```
Olshansk, line 10: We have 3 `Run` functions here, each of which can be started/stopped independently; with the exception of synched and sync mode running concurrently.

I was envisioning that each of them would be an independent module, so something like `ServerMode` can be started and stopped separately.

Olshansk (changes requested):
@andrewnguyen22 The `.md` file LGTM, albeit a couple last minor NITS.

I left a few questions/comments in the the `state_sync.go` but do think it still needs some thought & work. In particular, I want to avoid callbacks in `GetPeerMetadata` and `NetworkSend` complextely, so that we don't mix how the bus is or is not used.

I'm thinking we could merge it as is (after fixing the .md) and I could take on #125 to submit a follow-up iteration of this. Wdyt?

andrewnguyen22:
> I left a few questions/comments in the the state_sync.go but do think it still needs some thought & work. In particular, I want to avoid callbacks in GetPeerMetadata and NetworkSend complextely, so that we don't mix how the bus is or is not used.

I'm thinking we could merge it as is (after fixing the .md) and I could take on https://github.com/pokt-network/pocket/issues/125 to submit a follow-up iteration of this. Wdyt?

Sounds like a plan to me

consensus/state_sync/state_sync.go
andrewnguyen22: > GetPeers() []PeerSyncMeta

> type PeerSyncMeta interface {

It is used a few times. I think it lines up

consensus/state_sync/state_sync.go
andrewnguyen22: I can see that as a real possibility

consensus/doc/PROTOCOL_STATE_SYNC.md
andrewnguyen22: I think we do not at this level - but not strong on this. If you think it's helpful - add a comment before the merge

consensus/state_sync/state_sync.go
jessicadaugherty: ea6955dd-4506-473d-9abb-806dd0464797_text - Dwight Schrute - Andrew Nguyen

consensus/state_sync/state_sync.go
```diff
@@ -0,0 +1,102 @@
+package state_sync
+
+import (
+	typesCons "github.com/pokt-network/pocket/consensus/types"
+	cryptoPocket "github.com/pokt-network/pocket/shared/crypto"
+	"github.com/pokt-network/pocket/shared/modules"
+)
+
+type StateSyncModule interface {
```
Olshansk, line 9:
```suggestion
TODO(#362): Update the interface so it can be easily integrated with the app specific bus
type StateSyncModule interface {
```

Olshansk (approved):

consensus/state_sync/state_sync.go
Olshansk: Should be good to merge after we submit this.

PR #137, opened 2022-07-29
[P2P]  RainTree Cleanup & Redundancy Layers (Issue#85)

Olshansk:
## Description
Add the redundancy layer and cleanup layer to the P2P module.

_NOTE: This was split out of https://github.com/pokt-network/pocket/pull/127 to focus on just the logic of P2P and not the code cleanup._

Fixes [issue/85](https://github.com/pokt-network/pocket/issues/85)

## Type of change
Please mark the options that are relevant.

- [X] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] Other

## How Has This Been Tested?
- [ ] New unit test being added to the P2P module
- [ ] `make test_all`
- [ ] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md)

## Checklist
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have tested my changes using the available tooling
- [ ] If applicable, I have made corresponding changes to related or global README
- [ ] If applicable, I have added added new diagrams using [mermaid.js](https://mermaid-js.github.io)
- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works

PR #127, opened 2022-07-27
Issue #123 - Deprecate Pre2P

andrewnguyen22:
## Description

Deprecates the Pre2P module and move it to P2P by closing #123.

## Type of change
Please mark the options that are relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [X] Code cleanup

## How Has This Been Tested?
- `make test_all`
- `LocalNet`

## Checklist:
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] No new warnings are generated by my changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

---

Co-authored-by: Andrew Nguyen <andrewnguyen@Andrews-MacBook-Pro-2.local>
Co-authored-by: Daniel Olshansky <olshansky.daniel@gmail.com>

Issue #85, opened 2022-06-02
[P2P] RainTree Redundancy Layer Implementation


Olshansk:
# Objective
Implement and test RainTree's redundancy layer on top of the Pre2P module.

# Origin Document
[The Pocket Network V1 P2P Specification](https://github.com/pokt-network/pocket-network-protocol/tree/main/p2p), supported by [this explanation of RainTree](https://drive.google.com/file/d/1PaAWTx47nPSrM798lxfKv0GnhLthZLYm/view?usp=sharing) is partially implemented in the `pre2p/raintree` branch atop the `Pre2P` module at the time of creating this issue, but without the redundancy and cleanup layers.

This implementation has already been started and is available in the `pre2p/raintree_redundancy` branch at the time of writing this thread in https://github.com/pokt-network/pocket/pull/80.

The initial implementation of the redundancy layer code is available in the following commit: https://github.com/pokt-network/pocket/commit/756b0a09cd7b33a72b46e1690e4a858c6033792b.

# Goals
## Deliverables
- [ ] Fully implement the redundancy layers of RainTree including as described in the docs referenced by the `Origin Document` section:
   - [ ] 1. ACK/Adjust/Resend
   - [ ] 2. Redundancy layer
   - [ ] 3. Daisy Chain clean-up
- [ ] Update `raintree_utils_test.go` to add support for
  - [ ] Dead / faulty nodes
  - [ ] Partial visibility of the network
- [ ] Update the `README.md` (to be added in the `pre2p/raintree` soon) describing the added / modified code layout of `Pre2P module` with the new modifications

## Testing Methodology

1. Use `make test_pre2p` to run existing tests
2. Update the test suite in `raintree_utils_test.go` and add `raintree_redundancy_layer_test.go` int he same package with new tests
3. [Optional] Use `LocalNet` as described in `docs/development/README.md`
4. [Optional] Using `Telemetry` (if ready) to validate the results from (2)

## Non-goals
- Scaling LocalNet to many nodes
- Resolving tech debt or optimizing existing code
- Replacing P2P with Pre2P

Creator: @Olshansk 
Co-Owners: @andrewnguyen22

andrewnguyen22:
I'm increasing the scope of this issue to add the cleanup layer

Olshansk:
This is being worked on in https://github.com/pokt-network/pocket/pull/137

jessicadaugherty:
[...]

PR #368, opened 2022-11-30
[P2P] RainTree Cleanup & Redundancy Layers (Issue#85) - handover

deblasis:
## Description
Add the redundancy layer and cleanup layer to the P2P module.

_NOTE: This was split out of https://github.com/pokt-network/pocket/pull/127 to focus on just the logic of P2P and not the code cleanup._

Fixes [issue/85](https://github.com/pokt-network/pocket/issues/85)

### NOTE

This PR started from previous work tracked in https://github.com/pokt-network/pocket/pull/137

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation
- [ ] Other

## How Has This Been Tested?
- [ ] New unit test being added to the P2P module
- [ ] `make test_all`
- [ ] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md)

## Checklist


p2p/raintree/network.go
```diff
@@ -78,8 +103,21 @@ func (n *rainTreeNetwork) networkBroadcastAtLevel(data []byte, level uint32, non
...
+// Cleanup layer is just another send left / right
+// This layer happens at level -1
+func (n *rainTreeNetwork) cleanupLayer() []target {
+	// TODO (Team) unhappy path where the left / right nodes are down
+	// 			   (continue to search left and right until you have a hit)
+	return n.getCleanupTargets()
+}
+
+// Redundancy layer is simply a redundant send to the original +1/3 && -1/3
+func (n *rainTreeNetwork) redundancyLayer() (level int32, msgLevel int32) {
```
Olshansk, line 115:
```suggestion
func (n *rainTreeNetwork) redundancyLayer() (maxNumLevels int32, msgLevel int32) {
```

p2p/raintree/network.go
```diff
@@ -25,46 +25,71 @@ type rainTreeNetwork struct {
...
 	msg := &typesP2P.RainTreeMessage{
 		Level: level,
 		Data:  data,
 		Nonce: nonce,
 	}
+
+	// Redundancy layer
+	if level == 0 {
+		if n.redundancyLayerEnabled {
+			level, msg.Level = n.redundancyLayer()
```
Olshansk, line 66: I want to rename`level` to something like `levelforTargetCompute`, but I think the better approach is to capture all the logic for redundancy and cleanup in separate functions and do `if.redundancyEnable { n.maybeTriggerRedundantLayer } `. Wdyt?

p2p/raintree/network.go
```diff
@@ -25,46 +25,71 @@ type rainTreeNetwork struct {
...
+				log.Println("Error demoting self during RainTree message propagation: ", err)
+			}
+			return nil
+		}
+	}
+
+	targets := n.getTargetsAtLevel(uint32(level))
+
+	// Cleanup layer
+	if level == -1 {
```
Olshansk, line 78: Screenshot 2022-12-01 at 7 20 40 PM

I understand your source of confusion. I believe the code is right and the text can be made clearer. 

`level == -1 ` only when we call `demo(level=0)` which calls `internalBroadcast(-1)`

It's weird to describe because the way the code flows, it'll work out, so i would suggest trying is changing the code a bit such that.

1. if level == 0:
- If cleanup is enabled => do cleanup
- if redundancy is anebled => do redundancy
2. Do not call `internalBroadcast` if level == 0

If my intuition & understanding is correct, the tests should pass after this change.

p2p/raintree/types/proto/raintree.proto
```diff
@@ -4,7 +4,7 @@ package raintree;
 option go_package = "github.com/pokt-network/pocket/p2p/types";
 
 message RainTreeMessage {
-  uint32 level = 1;
+  int32 level = 1;
```
Olshansk, line 7: Screenshot 2022-12-01 at 7 35 55 PM

Given my comment above, we should be able to revert this.

p2p/raintree/network.go
```diff
@@ -25,46 +25,71 @@ type rainTreeNetwork struct {
...
+		if !n.cleanupLayerEnabled {
+			return nil
+		}
+		targets = n.cleanupLayer()
```
Olshansk, line 82: Screenshot 2022-12-01 at 7 36 47 PM

Redundancy is like UDP, but at the application layer.
Cleanup is like TCP, but at the application layer.

Out of scope. Let's add a TODO.

p2p/raintree/network.go
```diff
@@ -78,8 +103,21 @@ func (n *rainTreeNetwork) networkBroadcastAtLevel(data []byte, level uint32, non
...
+// This layer happens at level -1
+func (n *rainTreeNetwork) cleanupLayer() []target {
+	// TODO (Team) unhappy path where the left / right nodes are down
+	// 			   (continue to search left and right until you have a hit)
+	return n.getCleanupTargets()
+}
+
+// Redundancy layer is simply a redundant send to the original +1/3 && -1/3
+func (n *rainTreeNetwork) redundancyLayer() (level int32, msgLevel int32) {
+	return int32(n.peersManager.maxNumLevels), 0
```
Olshansk, line 116: Screenshot 2022-12-01 at 7 38 38 PM

You are correct. I believe this was simply overlooked and would simply consider this incomplete.

The changes we need to make is:
1. Potentially make this is a UDP message (not TCP), but seems like that's a lot of added work.
2. Make sure this results in a send to the receiving node (not a broadcast)
3. Make the change to have this be O(N)

p2p/raintree/addrbook_utils.go
```diff
@@ -37,6 +37,57 @@ func (n *rainTreeNetwork) getTargetsAtLevel(level uint32) []target {
...
+	} else if len(addrBook) == 2 {
+		addrStr1 := peersManagerStateView.addrList[1]
+		np1, ok := peersManagerStateView.addrBookMap[addrStr1]
+		if !ok {
+			log.Printf("[DEBUG] addrStr %s not found in addrBookMap", addrStr1)
+			return nil
+		}
+		return []target{{
+			address:                np1.Address,
+			serviceUrl:             peersManagerStateView.addrBookMap[peersManagerStateView.addrList[1]].ServiceUrl,
```
Olshansk, line 55: Screenshot 2022-12-01 at 7 45 36 PM

+1

Olshansk (changes requested):

p2p/raintree/network.go
deblasis: If a basic implementation (happy paths) of redundancy and cleanup layer is not in scope, what is in scope then? 
Can we clarify that?

From my understanding, since this PR was in review and you guys mentioned that it was "very close" I am definitely missing something................. or maybe this is just an @Olshansk joke and I am the victim.... where's the camera? :) 

Seriously: **Correct me if I am mistaken.**

From the code I see that what was added was:
- feature flags
- tests calls with feature flags on/off
- some weird handling of the `level` 
- redundancy and clean-up essentially implemented to just spit out targets

I can't reconcile the code with the paper so which one of the two is wrong? 🤷‍♂️

p2p/raintree/network.go
deblasis: 🤯 you guys grilled my code for way less than this, come on... 

You say that the code is right but how? If `level` is `uint32` (casted or not) how can the if in line 78 ever be truthy?
Overlooked? OK, but you didn't flag that? 🤔

p2p/raintree/network.go
deblasis: I don't think the entire logic can be encapsulated in a single function. Cleanup for example has to contemplate for the IGYW part and that's "interactive" at a higher level because you need to pass messages, check their contents and act accordingly.

Olshansk:
@deblasis Thank you for calling out my sloppiness on this review. I didn't do the due diligence I should've.

I'm going to follow up with more detailed comments & reviews. My current intuition is telling me that we will likely rescope it (e.g. split out redundancy & cleanup into two tickets), but please hold off until I follow up.

Olshansk:
> @deblasis Thank you for calling out my sloppiness on this review. I didn't do the due diligence I should've.
> 
> I'm going to follow up with more detailed comments & reviews. My current intuition is telling me that we will likely rescope it (e.g. split out redundancy & cleanup into two tickets), but please hold off until I follow up.

After various discussions we've had this week around gossip, etc, the next steps here are:

0 . RainTree (i.e. structured gossip) will be used to make sure staked actors are up to sync as fast as possible (e.g. new block is committed) and so ephemeral data is disseminated as fast as possible (e.g. a transaction that enters the mempool)

1. We will need to update the specs to clarify how `redundancy` and `cleanup` work

2. A single issue to implement & test cleanup

3. A single issue to implement & test redundancy

4. Explore ideas related to adversarial nodes that may cause liveness issues when gossiping and whether (2) & (3) help mitigate if. More research.

@deblasis How does that sound in terms of next steps?

deblasis:
> @deblasis How does that sound in terms of next steps?

We are on the same page.

Olshansk:
I referenced this in #438 which will capture and defined next steps on the RainTree & Redundancy layer.

PR #372, opened 2022-12-01
[Quick CI fix] sanitize json test results before triggering annotation action

okdas:
## Description

As we're using `tee` to pipe results of the tests to stdout and to `test_results.json` we need to sanitize what actually gets into that file, because non-json text breaks actions that expect json (obviously).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,16 +55,15 @@ jobs:
         run: |
           set -euo pipefail
           go test -p 1 -json ./... -covermode=count -coverprofile=coverage.out 2>&1 | tee test_results.json
-      - name: Output test failures
-        # Makes it easier to find failed tests so no need to scroll through the whole log.
-        if: ${{ failure() && env.TARGET_GOLANG_VERSION == matrix.go }}
-        run: |
-          cat test_results.json | jq 'select(.Action == "fail")'
+
       - name: Sanitize test results
         # We're utilizing `tee` above which can capture non-json stdout output, we also need to remove non-json lines.
         if: ${{ always() && env.TARGET_GOLANG_VERSION == matrix.go }}
-        run: |
-          cat test_results.json | jq -c -R 'fromjson? | select(type == "object")' > tmp.json && mv tmp.json test_results.json
+        run: cat test_results.json | jq -c -R 'fromjson? | select(type == "object")' > tmp.json && mv tmp.json test_results.json
+      - name: Output test failures
+        # Makes it easier to find failed tests so no need to scroll through the whole log.
+        if: ${{ failure() && env.TARGET_GOLANG_VERSION == matrix.go }}
+        run: cat test_results.json | jq 'select(.Action == "fail")'
       - name: Upload test results
         if: ${{ always() && env.TARGET_GOLANG_VERSION == matrix.go }}
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -58,7 +58,13 @@ jobs:
       - name: Output test failures
         # Makes it easier to find failed tests so no need to scroll through the whole log.
         if: ${{ failure() && env.TARGET_GOLANG_VERSION == matrix.go }}
-        run: cat test_results.json | jq 'select(.Action == "fail")'
+        run: |
+          cat test_results.json | jq 'select(.Action == "fail")'
```
Olshansk, line 62: I don't fully understand this, what makes sure that we run `Sanitize test results` before `Output test failures`?

Olshansk (changes requested):

.github/workflows/main.yml
okdas: @Olshansk We wanted to make sure we sanitize JSON prior to submitting it to the `guyarb/golang-test-annotations` github action. As you mention it, it probably is a good idea to also sanitize prior to exicuting `jq`'s `select()` query.

Added a change! Thanks.

.github/workflows/main.yml
```diff
@@ -55,6 +55,10 @@ jobs:
         run: |
           set -euo pipefail
           go test -p 1 -json ./... -covermode=count -coverprofile=coverage.out 2>&1 | tee test_results.json
+      - name: Sanitize test results
+        # We're utilizing `tee` above which can capture non-json stdout output, we also need to remove non-json lines.
+        if: ${{ always() && env.TARGET_GOLANG_VERSION == matrix.go }}
+        run: cat test_results.json | jq -c -R 'fromjson? | select(type == "object")' > tmp.json && mv tmp.json test_results.json
```
Olshansk, line 61: Curious if this works if we do `> test_results.json` instead of `> tmp.json && mv tmp.json test_results.json`?

okdas: @Olshansk I didn't want to `cat` and forward `stdout` to the same file at the same time. To be frank, I don't even know if this will work. `jq` also didn't provide a way to put changes in place, so I resorted to output to the temporary file instead.

.github/workflows/main.yml
```diff
@@ -55,6 +55,10 @@ jobs:
         run: |
           set -euo pipefail
           go test -p 1 -json ./... -covermode=count -coverprofile=coverage.out 2>&1 | tee test_results.json
+      - name: Sanitize test results
+        # We're utilizing `tee` above which can capture non-json stdout output, we also need to remove non-json lines.
```
Olshansk, line 59:
```suggestion
        # We're utilizing `tee` above which can capture non-json stdout output so we need to remove non-json lines before parsing it with jq.
```

.github/workflows/main.yml
Olshansk: `To be frank, I don't even know if this will work`

You're not sure if `fromjson` will work?

okdas: @Olshansk `fromjson?` works, and I've validated it.

Your question was if `> test_results.json` could work instead of `> tmp.json && mv tmp.json test_results.json`. My answer is I don't know, but I wouldn't do it anyway—because we're doing `cat` against the same file: `cat test_results.json > test_results.json`, just additionally processing via pipe `| jq -c -R 'fromjson? | select(type == "object")'`.

.github/workflows/main.yml
okdas: @Olshansk rephrased it a bit to reflect this is necessary for the external github action.

Olshansk (approved):

Issue #331, opened 2022-11-01
[SPIKE][Consensus] Remove `validatorMap` from the `consensusModule` struct


andrewnguyen22:
## Objective

Remove `validatorMap` from the `consensusModule` struct state & access it from the persistence layer (added n the past) 

The `validatorMap` is just a placeholder for the global `ValidatorSet` which is stored at the persistence layer. 

*NOTE: this is a multi-module issue*

## Origin Document

Originally scoped in issue #315 - separate issue to downscope

## Goals
- [ ] Remove `validatorMap` from the `consensusModule` struct and port it to the persistence module's `GetActor(s)` (Validator) 
- [ ] Update Changelog.md

## Deliverable
- [ ] Code complete
- [ ] Supporting documentation

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)

**Creator**: @andrewnguyen22

deblasis:
@jessicadaugherty heads up I am a bit behind schedule with #331 and #203 (ValidatorMap related work). I was supposed to be done yesterday but I think I'll need another day or two to have them both under review.

Issue #203, opened 2022-09-11
[TECHDEBT] Consolidate & encapsulate `ValidatorMap` logic


Olshansk:
## Objective

Encapsulate the concept of a `ValidatorMap` only to the consensus module

## Origin Document

[This discussion](https://github.com/pokt-network/pocket/pull/178#discussion_r965167220) in https://github.com/pokt-network/pocket/pull/178:

<img width="899" alt="Screen Shot 2022-09-11 at 3 15 53 PM" src="https://user-images.githubusercontent.com/1892194/189545145-72ddde0a-6712-4f23-bcd6-1e27fd2c9f42.png">

## Goals

- Remove all occurrences/references to the `ValidatorMap` throughout the codebase
- Only maintain knowledge/understanding o the `ValidatorMap` in the consensus module

## Deliverable

- [ ] Do a grep for `ValidatorMap` throughout the whole codebase and it should only show up in the consensus module
- [ ] A single PR refactoring / cleaning up this goal

## Non-goals / Non-deliverables

- Adding new functionality

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references

## Testing Methodology

- **All tests**: `make test_all`
- **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @Olshansk

Olshansk:
This is related to #271 in the greater scope/vision of having dynamic validator sets as they change with each height.

#271 is scoped to only P2P and doing structured gossip amongst the validator nodes (for now). In general, P2P does care who the nodes are but simply gossips amongst the actors it is given.

This ticket is scoped to consensus (techdebt cleanup) to make sure the concept of a "validator set" only exists in consensus.

_As this is done (in isolation), the goal is to think how to implement dynamic additions of validators & P2P actors._

deblasis:
@jessicadaugherty heads up I am a bit behind schedule with #331 and #203 (ValidatorMap related work). I was supposed to be done yesterday but I think I'll need another day or two to have them both under review.

PR #370, opened 2022-11-30
[Core] Productionize End-to-end tx from iteration 3 demo

Olshansk:
## Description

Productionze the changes made to iteration 3 demo.

## Issue

Fixes #364 

## Type of change

Please mark the relevant option(s):

- [x] New feature, functionality or library
- [x] Bug fix
- [x] Code health or cleanup
- [x] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

In addition to make the CLI broadcast a transaction, two major breaking changes were made
- Changed the scope of `TransactionExists` from the `PostgresContext` to the `PersistenceModule`
- Changed the scope of `CheckTransaction` from the `UtilityContext` to the `UtilityModule`

The reason for this change is because the codebase was creating unnecessary contexts when the module directly could be used. In particular, access to the `mempool` and the `txIndexer` is not scope to a `context` but exists in the lifecycle of the entire node. Making these changes simplified the code, end-to-end flow and is also beginning to set us up for cleaner rollbacks where contexts can be just cleared & released.

#### Shared Changes
- Changed the scope of `TransactionExists` from the `PostgresContext` to the `PersistenceModule`
- Changed the scope of `CheckTransaction` from the `UtilityContext` to the `UtilityModule`
- Change the `bus` to be a pointer receiver rather than a value receiver in all the functions it implements

#### RPC Changes
- Updated `PostV1ClientBroadcastTxSync` to broadcast the transaction it receives
- Avoid creating an unnecessary utility context and use the utility module directly

#### Utility Module Changes
- Introduce a general purpose `HandleMessage` method at the utility level
- Move the scope of `CheckTransaction` from the context to the module level
- Add an `IsEmpty` function to the `Mempool`
- Rename `DeleteTransaction` to `RemoveTransaction` in the mempool
- Rename `LatestHeight` to `Height` in the `utilityContext`
- Add comments inside `CheckTransaction` so its functionality is clearer
- Add comments and cleanup the code in `mempool.go`

#### Consensus Changes
- Removed unused `consensus.UtilityMessage`

#### Persistence Changes
- Changed the scope of `TransactionExists` from the `PostgresContext` to the `PersistenceModule`

## Testing

- [x] `make develop_test`
- [ ] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md) w/ all of the steps outlined in the `README`
- [ ] Followed the instructions at `docs/demos/iteration_3_end_to_end_tx.md`

## Required Checklist

[...]

Issue #270, opened 2022-10-03
[TECHDEBT] [P2P] P2P integration with utility for addressBook lookups


deblasis:
## Objective

Tend for the TODO

```go
// CLEANUP(drewsky): These functions will turn into more of a "ActorToAddrBook" when we have a closer
// integration with utility.
```

## Origin Document

**Techdebt** week emphasizes code health. Particularly, removing all `placeholder` and `hacky` code. 

The past few months, `shared/modules` have built up quite a few TODOs. 

The PR should remove all `non-blocked` TODOs

## Goals

- [ ] Address tech debt

## Deliverable

- [ ] Rename ValidatorMapToNetworkPeer to ActorMapToNetworkPeer or ActorsToNetworkPeer
- [ ] Look into the changing the interface call from ConsensusModule.ValidatorMap() to PersistenceModule.GetActorsAtHeight() because all the staked protocol level actors can help in broadcasting messages.

## Non-goals / Non-deliverables

- 

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)

## [Optional] Testing Methodology

- \__REPLACE_ME_: Make sure to update the testing methodology appropriately\_
- **Task specific tests**: `make ...`
- **All tests**: `make test_all`
- **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @deblasis
**Co-Owners**: @Olshansk

Olshansk:
The action item for this ticket is:

1. Rename `ValidatorMapToNetworkPeer` to `ActorMapToNetworkPeer` or `ActorsToNetworkPeer`

2. Look into the changing the interface call from `ConsensusModule.ValidatorMap()` to `PersistenceModule.GetActorsAtHeight()` because all the staked protocol level actors can help in broadcasting messages.

@andrewnguyen22 Do you think we should only have validators participating in RainTree or any protocol level actor that stakes.

PR #374, opened 2022-12-05
[TECHDEBT] [P2P] integration with persistence for addressBook management - (Issue #271)

### LINKED ISSUE #271: [TECHDEBT] [P2P] integration with persistence for addressBook management
## Objective

Tend for the TODO

```go
// TODO(team): Need to integrate with persistence layer so we are storing this on a per height basis.
// We can easily hit an issue where we are propagating a message from an older height (e.g. before
// the addr book was updated), but we're using `maxNumLevels` associated with the number of
// validators at the current height.
```

Discuss with @Olshansk to better define the requirements and update below

Related to #222 and #270 

### Note to self:
Understand the commonalities between the P2P <-> Utility (#270) and P2P <-> Persistence integrations 

## Origin Document

[ Why? Issue justification and/or link to another document]

## Goals

- _\_*REPLACE_ME*: List of things that are high level ideas or goals driving the task_
- [ ] ...

## Deliverable

- _\_*REPLACE_ME*: List of things that are concrete deliverables_
- [ ] ...

- _\_*REPLACE_ME*: List of things that are out of scope_
- ...

## General issue deliverables

## [Optional] Testing Methodology

- \__REPLACE_ME_: Make sure to update the testing methodology appropriately\_
- **Task specific tests**: `make ...`

---

**Creator**: @deblasis
**Co-Owners**: @Olshansk

deblasis:
## Description

This PR is a stepping stone for the removal of the `ValidatorMap`.
Specifically it allows us to use per-Height addressBook and lays the ground for churn management.
Right before a new height is reached, the addressBook is propagated via a message on the `Bus` so that other interested modules can handle it. This part is out-of-scope for this PR and will be covered later on in a specific issue, unless differently agreed during code-review.

Also, this PR tends to the TODO:
https://github.com/pokt-network/pocket/blob/444c4242f8d165ba7702f160f6e94cd416fd0b8f/p2p/raintree/network.go#L29

## Issue

Fixes #271

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Consensus
  - Consolidated number of validators in tests in a single constant: `numValidators`
  - Fixed typo in `make test_consensus_concurrent_tests` so that we can run the correct test matrix
  - Using `GetBus()` instead of `bus` wherever possible
- P2P
  - mempool cap is now configurable via P2PConfig. Tests implement the mock accordingly.
  - Introduced the concept of a `addrbookProvider` that abstracts the fetching and the mapping from `Actor` to `NetworkPeer`
  - Temporary hack to allow access to the `addrBook` to the debug client (will be removed in an upcoming PR already in the works for issues [#203](https://github.com/pokt-network/pocket/issues/203) and [#331](https://github.com/pokt-network/pocket/issues/331))
  - Transport related functions are now in the `transport` package
  - Updated tests to source the `addrBook` from the `addrbookProvider` and therefore `Persistence`
  - Updated Raintree network constructur with dependency injection
  - Updated stdNetwork constructur with dependency injection
  - Improved documentation for the `peersManager
  - Raintree mempool cannot grow unbounded anymore. It's now bounded by a constant limit and when new nonces are inserted the oldest ones are removed.
Olshansk marked this conversation as resolved.
Show resolved
  - Raintree is now capable of fetching the address book for a previous height and to instantiate an ephemeral `peersManager` with it.
- Persistence
  - Moved Actor related getters from `genesis.go` to `actor.go`
  - Added `GetAllStakedActors()` that returns all Actors
- Shared
  - Added `GetMaxMempoolCount`

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


consensus/leader_election/module.go
```diff
@@ -60,5 +60,14 @@ func (m *leaderElectionModule) ElectNextLeader(message *typesCons.HotstuffMessag
 
 func (m *leaderElectionModule) electNextLeaderDeterministicRoundRobin(message *typesCons.HotstuffMessage) typesCons.NodeId {
 	value := int64(message.Height) + int64(message.Round) + int64(message.Step) - 1
-	return typesCons.NodeId(value%int64(len(m.GetBus().GetConsensusModule().ValidatorMap())) + 1)
+
+	ctx, err := m.GetBus().GetPersistenceModule().NewReadContext(int64(message.Height))
+	if err != nil {
+		log.Fatal(err)
+	}
+	vals, err := ctx.GetAllValidators(int64(message.Height))
```
Olshansk, line 68: Shouldn't this be part of `issue/270-p2p-integration-with-utility-for-addrbook-lookups`? I'm a little confused with the two different branches are dependant on each other.

p2p/module.go
```diff
@@ -31,6 +32,49 @@ type p2pModule struct {
 	network typesP2P.Network
 }
 
+// HandleEvent implements modules.P2PModule
+func (m *p2pModule) HandleEvent(message *anypb.Any) error {
```
Olshansk, line 36: I'm getting a bit worried that `p2p/module.go` is getting very big.

Do you think it would make it more readable to create a `p2p/event_handler.go` and `p2p/network_handler.go`?

p2p/module.go
```diff
@@ -31,6 +32,49 @@ type p2pModule struct {
...
+		m.broadcastAddressBookAtHeight(event.CurrentHeight)
+
+	default:
+		return typesP2P.ErrUnknownEventType(message.MessageName())
+	}
+	return nil
+}
+
+// broadcastAddressBookAtHeight broadcasts on the event bus the address book at the given height
+func (m *p2pModule) broadcastAddressBookAtHeight(height uint64) error {
```
Olshansk, line 57: **Starting a discussion here.**

I'm thinking whether we actually need to broadcast this as an event or not.

The persistence layer (i.e. staked validators) are going to tell us what the validator set is. The address book is going to tell us how to reach them. What if `p2pModule` exposes a `GetAddressBook` function that is called synchronously. That way, consensus can compare the URL found through peer / churn when it needs to do a `send` during the HotPOKT lifecycle.

p2p/raintree/network.go
```diff
@@ -179,6 +206,12 @@ func (n *rainTreeNetwork) HandleNetworkData(data []byte) ([]byte, error) {
 	}
 
 	n.mempool[rainTreeMsg.Nonce] = struct{}{}
+	n.mempool_keys = append(n.mempool_keys, rainTreeMsg.Nonce)
+	if len(n.mempool_keys) > mempoolMaxNonces {
```
Olshansk, line 210: What are your thoughts on potentially generalizing and reusing the mempool from `utility/types/mempool.go`?

It would require:

1. Me merging my PR
2. Moving mempool into general
3. Generalizing some names (e.g. Tx to a generic)
4. Modifying the appropriate call sites

Screenshot 2022-12-06 at 5 21 39 PM

Definitely another ticket but wanted to ask you first.

I think the terms `nonceSet` and `nonceList` would be more semantic here.

shared/messaging/proto/addressbook_at_height.proto
```diff
@@ -0,0 +1,16 @@
...
+option go_package = "github.com/pokt-network/pocket/shared/messaging";
+
+message AddressBookAtHeight {
+  repeated NetworkPeer network_peers = 1;
+  uint64 height = 2;
+}
+
+message NetworkPeer {
+  bytes public_key = 1;
+  bytes address = 2;
```
Olshansk, line 14: Can we change this into a string.

The fact that we're not consistent is legacy...

shared/node.go
```diff
@@ -159,6 +159,18 @@ func (node *Node) handleEvent(message *messaging.PocketEnvelope) error {
 	switch contentType {
 	case messaging.NodeStartedEventType:
 		log.Println("[NOOP] Received NodeStartedEvent")
+	case messaging.BeforeHeightChangedEventType:
+		log.Println("Received BeforeNewHeightEvent")
+		node.GetBus().GetP2PModule().HandleEvent(message.Content)
+	case messaging.AddressBookAtHeightEventType:
+		log.Printf("Received AddressBookAtHeightEvent: %v\n", message.Content)
+		// DISCUSS (deblasis): here we'd delegate a module to handle the event.
+		// We need an open PersistenceRWContext where we can update the validator list (sounds like churn management).
```
Olshansk, line 168: There is some confusion here.

The validator list is part of the world state (i.e. it is part of the persistence layer). When you apply a block, you should be able to say "Give me validators at height N from the SQL engine".

The address book is a list of all actors (staked actors, full nodes, light clients, etc...) that the current node can connect to. For example, the `serviceURL` of a validator can be matched with the address book so we know their active IP.

Olshansk (changes requested):
@deblasis Left a few comments, and the integration of consensus with persistence is great, but I think I may have led you down the wrong path.

- When we're doing something like consensus, the proposer does a direct send (address forwarded to P2P and P2P resolves it).

- When we're doing a structured gossip (where every staked actor has the same view of all the addresses on the network), the P2P module retrieves that list from persistence, and is then responsible for resolving addresses to IPs. When an address doesn't resolve is when we need to start looking at our redundancy/cleanup/adjust/resent layers

- This is making me think that in the case of full nodes, there might also be a a need for random gossip (increase data availability, but still need to mull on it).

----

Overall, given the comments, do you see a need for the broadcast events introduced here?

consensus/consensus_tests/utils_test.go
```diff
@@ -304,17 +304,22 @@ func basePersistenceMock(t *testing.T, _ modules.EventsChannel) *modulesMock.Moc
...
 
 	persistenceMock.EXPECT().Start().Return(nil).AnyTimes()
 	persistenceMock.EXPECT().SetBus(gomock.Any()).Return().AnyTimes()
-	persistenceMock.EXPECT().NewReadContext(int64(-1)).Return(persistenceContextMock, nil).AnyTimes()
+	persistenceMock.EXPECT().NewReadContext(gomock.Any()).Return(persistenceReadContextMock, nil).AnyTimes()
 	persistenceMock.EXPECT().ReleaseWriteContext().Return(nil).AnyTimes()
 
 	// The persistence context should usually be accessed via the utility module within the context
 	// of the consensus module. This one is only used when loading the initial consensus module
 	// state; hence the `-1` expectation in the call above.
```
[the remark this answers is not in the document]
deblasis: I am accessing the context for example here:

https://github.com/pokt-network/pocket/pull/374/files/dbb24d5918291dc28c8921bc2045de15180bc2e9#diff-9a2c4261d87cba56442982ea893473c0b02675123b3c6d09815ac15babef2cbdR64

and therefore we have Height that's != -1 as well. 

If I shouldn't be able to do that, we should probably look into ways for enforcing this "access constraint" in a more semantic way. 

I'll look into changing this so that we access the context via utility.

Stay tuned

consensus/consensus_tests/utils_test.go
```diff
@@ -304,17 +304,22 @@ func basePersistenceMock(t *testing.T, _ modules.EventsChannel) *modulesMock.Moc
...
 	persistenceContextMock.EXPECT().Close().Return(nil).AnyTimes()
-	persistenceContextMock.EXPECT().GetLatestBlockHeight().Return(uint64(0), nil).AnyTimes()
+	persistenceReadContextMock.EXPECT().GetLatestBlockHeight().Return(uint64(0), nil).AnyTimes()
+
+	// here we are mocking 4 actors since we are considering the 4 nodes in the LocalNet genesis
```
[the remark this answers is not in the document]
deblasis: Totally! Implemented suggestion 1 in https://github.com/pokt-network/pocket/pull/374/commits/09aa018a911b87a3259f2921d7ee90e77bff6a63

p2p/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.0.9] - 2022-12-04
+
+- Raintree mempool cannot grow unbounded anymore. It's now bounded by a constant limit and when new nonces are inserted the oldest ones are removed.
```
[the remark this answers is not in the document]
deblasis: 1. No it's not but it can be. Stay tuned

2. One of my goals in life is to make complex things very easy for myself and potentially anyone interested :)

No, jokes apart, you are right but it's not LIFO but FIFO

If the cap/limit on the size of the set is 2:

| T | What's up?| State |
| -- | -- | -- |
| 0 | put(first) | [ ] |
| 1 | put(second) | [ first ] | 
| 2 | put(third) | [ first, second ] | 
| 3 | 🅰️ where's **FIRST**? I saw it going **IN** at T 0 <br > 🅱️ dude, **FIRST** is **OUT**, it had to make space for the new guy, "three" | [ second, third ] |

p2p/raintree/network.go
deblasis: I totally agree, we should generalize the concept and apply it wherever needed. 

Adding a TODO (lemme know if I should create the issue as well or if this should be considered as part of the discussion in #278).

I am just renaming to `nonceSet` and `nonceList` for now.

consensus/pacemaker.go
```diff
@@ -262,3 +267,13 @@ func (p *paceMaker) getStepTimeout(round uint64) timePkg.Duration {
 	baseTimeout := timePkg.Duration(int64(timePkg.Millisecond) * int64(p.pacemakerCfg.GetTimeoutMsec()))
 	return baseTimeout
 }
+
+func (p *paceMaker) onBeforeNewHeight() {
+	event, _ := messaging.PackMessage(&messaging.BeforeHeightChangedEvent{CurrentHeight: p.consensusMod.CurrentHeight()})
```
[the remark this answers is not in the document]
deblasis: https://github.com/pokt-network/pocket/pull/374/commits/f77334595922520d36c641db8cee354659111b25

shared/node.go
deblasis: You basically answered my questions/clarified my doubts with this.

Since the P2P addressBook is just going to live in memory, the `peersManager` should suffice and therefore I believe that some of the events I introduced can be revisited/removed. I'll update and re-request review accordingly.

p2p/module.go
deblasis: I understand it now and correct me if I am wrong.

There's no need to broadcast the addressBook from P2P to other modules at every height. 
It would be necessary if we wanted to do something with it for some reason (I can think of observability but at the moment I can't think about anything else). Obviously it should not be part of the world state because, inherently, the addressBook is a "perspective" of the network from the current peer.

consensus/leader_election/module.go
deblasis: You are right, reverted.

The branches are dependent because from my understanding there were overlaps.

My confusion stems also from this https://github.com/pokt-network/pocket/pull/374#discussion_r1042044513 

What I mean is: If we have to access `GetAllValidators` from the context within `Utility`, shouldn't we simply integrate `P2P` with `Utility` (#270)?

Lemme know if this clarifies your confusion about my confusion :)

deblasis:
> Overall, given the comments, do you see a need for the broadcast events introduced here?

After having digested your comments and thought about it I think we should:
- Propagate the changes to the world state in terms of validators at height.
  - I envisage something like: `repeated Actor added_actors`  and `repeated Actor removed_actors` in the `HeightChangedEvent`
  
  This way P2P will react to changes and won't pull the validators set from persistence. Technically it won't rely on persistence directly anymore but on these messages received via the bus and handled in `handleMessage`. In terms of modularity and testability it should be an improvement.

  If this sounds like premature optimization and/or overcomplicating things, say no more. I can remove the events part and leave the `peerManagers` as is (with my changes to `getAddrBookLenght`).
  
On the latter, I have one observation:

Currently, we would never find ourselves in a situation where `getAddrBookLength` is called for a `height` that's lower than current because we only call this method in `getTargetsAtLevel` and we pass the `CurrentHeight` from `ConsensusModule` as an argument: 

image

Should we consider some tests to verify the use case: "we are propagating a message for a previous height"?

p2p/CHANGELOG.md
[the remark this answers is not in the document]
deblasis: 1. is now implemented in https://github.com/pokt-network/pocket/pull/374/commits/4ce80bc24b0cbad8a27df6d7468475626954defb

consensus/consensus_tests/utils_test.go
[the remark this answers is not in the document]
deblasis: This comment now doesn't apply to this PR anymore but it does in #378. Referencing it there

shared/messaging/proto/addressbook_at_height.proto
deblasis: This code has been deprecated already :) fail fast they say. It doesn't get faster than this IMHO

p2p/raintree/network.go
Olshansk: I created #388 to track this and assigned it to you as I had to fix one last unit test in #370.

p2p/module.go
Olshansk: > There's no need to broadcast the addressBook from P2P to other modules at every height.
It would be necessary if we wanted to do something with it for some reason

Yea, that's the only thing I'm imagining right now, but I'd prefer not to be broadcasting all this data until we are really using it.

If we were to use it for observability, what I would do is:

1. Create a function in the P2P interface called `GetAddressBook` that lets you retrieve it. It does not introduce new complexity, and can be multi-purposed. E.g. when we build a dashboard to view it. I bet I can asked ChatGPT to do this.

2. Create an event called "AddressBookChanged", that triggers us to poll.

> Obviously it should not be part of the world state because, inherently, the addressBook is a "perspective" of the network from the current peer.

Completely correct.

p2p/raintree/network.go
Olshansk: @deblasis #388 should be done separately but will really help with both this and utility.

p2p/types/proto/p2p_config.proto
```diff
@@ -8,9 +9,10 @@ message P2PConfig {
   uint32 consensus_port = 2;
   bool use_rain_tree = 3;
   bool is_empty_connection_type = 4; // TODO (Drewsky) switch back to enum
+  uint64 mempool_max_nonces = 5; // this is used to limit the number of nonces that can be stored in the mempool, after which a FIFO mechanism is used to remove the oldest nonces and make space for the new ones
```
Olshansk, line 12: max_mempool_count?

Olshansk (changes requested):
@deblasis Regarding the events, what are your thoughts on:

1. Exposing a function called `GetAddrBook` to the `p2p` module
2. Exposing an event called `AddrBookChanged`

I recognize that there will be a delay between when `GetAddrBook` is called and `AddrBookChanged` is sent, but it is sufficient for the observability purposes we need.

Olshansk (changes requested):
Lmk your thoughts on this: https://github.com/pokt-network/pocket/pull/374#pullrequestreview-1212334360

Marking as `RequestChanges` in the meantime (so it's easier to track). Lmk when I need to take another full look

deblasis:
> @deblasis Regarding the events, what are your thoughts on:
> 
> 1. Exposing a function called `GetAddrBook` to the `p2p` module

Where would this function be called currently? I would say that if it's not relevant to the scope of this issue we shouldn't do it here. Sounds like scope creep a little bit. No?

> 2. Exposing an event called `AddrBookChanged`

Similar comment to the one above. These are things that will be part of broader issues where we are not simply tending for tech debt but we are literally implementing features. For example, I envisage some form of observability of the p2p addressbook as long as we develop the raintree peer discovery and churn because in that case the testing harness itself could be listening to these events and make assertions based on them and/or the order in which these events happen.

p2p/raintree/network.go
deblasis: > @deblasis #388 should be done separately but will really help with both this and utility.

Are you saying that #388 is now a blocker for #271?

p2p/utils_test.go
```diff
@@ -165,13 +165,19 @@ func createMockGenesisState(t *testing.T, valKeys []cryptoPocket.PrivateKey) mod
...
+	telemetryMock *modulesMock.MockTelemetryModule,
+) *modulesMock.MockBus {
 	ctrl := gomock.NewController(t)
 	busMock := modulesMock.NewMockBus(ctrl)
 
 	busMock.EXPECT().PublishEventToBus(gomock.Any()).AnyTimes()
 	busMock.EXPECT().GetConsensusModule().Return(consensusMock).AnyTimes()
+	busMock.EXPECT().GetPersistenceModule().Return(persistenceMock).AnyTimes()
 	busMock.EXPECT().GetTelemetryModule().Return(telemetryMock).AnyTimes()
+	busMock.EXPECT().GetPersistenceModule().Return(persistenceMock).AnyTimes()
```
Olshansk, line 180: Remove. Duplicate two lines above

p2p/raintree/peers_manager_test.go
```diff
@@ -8,8 +8,10 @@ import (
 
 	"github.com/golang/mock/gomock"
 	"github.com/pokt-network/pocket/p2p/types"
+	"github.com/pokt-network/pocket/runtime/defaults"
 	"github.com/pokt-network/pocket/shared/crypto"
 	cryptoPocket "github.com/pokt-network/pocket/shared/crypto"
+	mock_modules "github.com/pokt-network/pocket/shared/modules/mocks"
```
Olshansk, line 14: Reuse `modulesMock` below

p2p/raintree/network.go
```diff
@@ -21,24 +21,53 @@ var _ modules.IntegratableModule = &rainTreeNetwork{}
...
 
-	selfAddr cryptoPocket.Address
+	selfAddr         cryptoPocket.Address
+	addrBookProvider typesP2P.AddrBookProvider
 
 	peersManager *peersManager
 
 	// TODO (#278): What should we use for de-duping messages within P2P?
-	mempool map[uint64]struct{} // TODO (#278) replace map implementation (can grow unbounded)
+	// TODO (team): we should generalize and use the FIFOMempool (in utility/types/mempool.go at the time of writing) in here as well for this. Same concept
```
Olshansk, line 30:
```suggestion
	// TODO(#388): generalize to use the shared `FIFOMempool` type (in `utility/types/mempool.go` at the time of writing) in here as well for this.
```

p2p/module.go
Olshansk: Resolving since its out of scope.

p2p/raintree/network.go
Olshansk: Nope. We created the ticket and I'm leaving a suggestion to include it in the comment in the code.

p2p/module.go
```diff
@@ -96,13 +120,31 @@ func (m *p2pModule) Start() error {
...
-	addrBook, err := ValidatorMapToAddrBook(m.p2pCfg, m.bus.GetConsensusModule().ValidatorMap())
+	currentHeight := m.bus.GetConsensusModule().CurrentHeight()
+	var (
+		addrBook typesP2P.AddrBook
+		err      error
+	)
+
+	if m.GetBus().GetPersistenceModule() == nil {
+		// we are getting called by the client and we use the "legacy behaviour"
+		// TODO (team): improve this.
```
Olshansk, line 131: I think we've kind of deprecated `team` in exchange for a name or a ticket. It's team by default. We also have an `IMPROVE` tag in the `Makefile`.

Can you change this to `IMPROVE: <explanation of why we can't delete the legacy behaviour>`

p2p/module.go
```diff
@@ -96,13 +120,31 @@ func (m *p2pModule) Start() error {
...
+	}
 	if err != nil {
 		return err
 	}
 
 	if m.p2pCfg.GetUseRainTree() {
-		m.network = raintree.NewRainTreeNetwork(m.address, addrBook)
+		if m.GetBus().GetPersistenceModule() == nil {
+			// we are getting called by the client and we use the "legacy behaviour"
+			// TODO (team): improve this.
```
Olshansk, line 143:
```suggestion
			// TECHDEBT: Deprecate the legacy behaviour altogether
```

p2p/raintree/network_test.go
```diff
@@ -3,21 +3,28 @@ package raintree
 import (
 	"testing"
 
+	"github.com/golang/mock/gomock"
 	"github.com/pokt-network/pocket/p2p/types"
 	typesP2P "github.com/pokt-network/pocket/p2p/types"
+	"github.com/pokt-network/pocket/runtime/defaults"
 	cryptoPocket "github.com/pokt-network/pocket/shared/crypto"
+	mock_modules "github.com/pokt-network/pocket/shared/modules/mocks"
```
Olshansk, line 11: I looked at https://go.dev/doc/effective_go on best guidelines for named important but it doesn't give guidance:

Screenshot 2022-12-12 at 5 40 31 PM

I think we should either:
1. Refer to this as `mocks` (the default)
2. Use `modulesMock` or `mockModules` in cameCase to be consistent with other parts of the code

p2p/module.go
```diff
@@ -127,6 +169,26 @@ func (m *p2pModule) Start() error {
 	return nil
 }
 
+func (m *p2pModule) getAddrBookPerHeight(height uint64) (typesP2P.AddrBook, error) {
```
Olshansk, line 172: Can you rename this to `getStakedAddrBookAtHeight`.

The address book P2P has will be greater (i.e. non-staked nodes), but this what we're going to be using for RainTree (i.e. structured gossip).

p2p/module.go
```diff
@@ -127,6 +169,26 @@ func (m *p2pModule) Start() error {
...
+	persistenceReadContext, err := m.bus.GetPersistenceModule().NewReadContext(int64(height))
+	if err != nil {
+		return nil, err
+	}
+	vals, err := persistenceReadContext.GetAllValidators(int64(height))
```
Olshansk, line 177: Either add a TODO to get other staked actors too (fisherman, service nodes, etc...) or incorporate it here.

p2p/module.go
```diff
@@ -96,13 +120,31 @@ func (m *p2pModule) Start() error {
...
 	}
 
 	if m.p2pCfg.GetUseRainTree() {
-		m.network = raintree.NewRainTreeNetwork(m.address, addrBook)
+		if m.GetBus().GetPersistenceModule() == nil {
+			// we are getting called by the client and we use the "legacy behaviour"
+			// TODO (team): improve this.
+			m.network = raintree.NewRainTreeNetwork(m.address, addrBook, m.p2pCfg)
+		} else {
+			m.network = raintree.NewRainTreeNetworkWithAddrBookProvider(m.address, m.getAddrBookPerHeight, currentHeight, m.p2pCfg)
```
Olshansk, line 146: If the `network` structur had access to the bus, wouldn't it be able to implement this internally so we can avoid passing around the function?

I think it'd be simpler given that `getAddrBookPerHeight` doesn't really have any dependencies on `getAddrBookPerHeight` other than bus access.

p2p/module.go
```diff
@@ -31,6 +32,29 @@ type p2pModule struct {
...
+		msg, err := codec.GetCodec().FromAny(message)
+		if err != nil {
+			return err
+		}
+		_, ok := msg.(*messaging.BeforeHeightChangedEvent)
+		if !ok {
+			return fmt.Errorf("failed to cast message to BeforeHeightChangedEvent")
+		}
+
+		// DISCUSS (https://github.com/pokt-network/pocket/pull/374#issuecomment-1341350786): decide if we want a pull or push model for integrating with persistence
```
Olshansk, line 48: Consolidating the discussions here:
- https://github.com/pokt-network/pocket/pull/374#issuecomment-1341350786
- https://github.com/pokt-network/pocket/pull/374#issuecomment-1347532500

The main reason I was suggesting those alternate events and functions is because `getAddrBookPerHeight` solves the problem of integrating with persistence fulfilling what this ticket intended to do.

I don't currently see how we're using these two events (in this PR) so am thinking we should just remove altogether, but wanted to leverage the work already done.

p2p/raintree/peers_manager.go
```diff
@@ -30,7 +30,16 @@ type peersManager struct {
...
+func NewPeersManagerWithAddrBookProvider(selfAddr cryptoPocket.Address, addrBookProvider typesP2P.AddrBookProvider, height uint64) (*peersManager, error) {
+	addrBook, err := addrBookProvider(height)
+	if err != nil {
+		return nil, err
+	}
+
+	return newPeersManager(selfAddr, addrBook, false)
+}
+
+func newPeersManager(selfAddr cryptoPocket.Address, addrBook typesP2P.AddrBook, handleEvents bool) (*peersManager, error) {
```
Olshansk, line 42: As I'm looking through the code, I think we'll end up having two types of address book:

- `stakedAddrBook`: map[stakedActorAddress] -> ActorStruct (from persistence)
- `addrBook`: map[address] -> IP address (peer discovery & churn)

It's out of scope for this PR but we should do it in a follow up. This way, we can have a peerManage that is static and one that is not.

Olshansk, line 42: Thoughts on renaming `handleEvents` to `isDynamic` or `isStatic`?

Olshansk (changes requested):

p2p/raintree/network_test.go
deblasis: https://github.com/pokt-network/pocket/pull/374/commits/30e2cfbda618799eb7f35f5f698181c6e33db014

p2p/module.go
deblasis: Maybe addressed in https://github.com/pokt-network/pocket/pull/374/commits/36d7129cfc1c0199ddb1931438c16b6480dda88e

p2p/module.go
deblasis: I have refactored things a bit, PTAL and let me know what you think.

The idea is that I am abstracting the `addrBook` sourcing from persistence into its own module so that it's easier to change that specific logic in a centralized place.

In a way this is going to connect with what you mention [here](https://github.com/pokt-network/pocket/pull/374#discussion_r1046585011) 

I am taking "baby steps" to share my WIP but definitely that work should be done separately.

p2p/module.go
deblasis: It makes total sense. However my code was there to convey a train of thought and to clarify my doubts, it was a poor-man's design spec in a way.

Now that things are clearer the P2P event handling can be removed, what I'd like to keep, if you agree, is just the events `BeforeNewHeightEvent` and `HeightChangedEvent` since they will likely be used as "hooks" later on. For now they are simply NOOP.

deblasis:
@Olshansk, thanks to your feedback as always, PTAL and let me know your thoughts.

I didn't update the changelogs because I have refactored things quite a bit and I would like to hear from you first.

p2p/addrbook_provider/getter.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"log"
+
+	typesP2P "github.com/pokt-network/pocket/p2p/types"
+	"github.com/pokt-network/pocket/shared/modules"
+)
+
+// GetAddrBook is a helper function that returns the addrBook depending on the availability of the persistence module
+//
+// this is a temporary solution simply used to centralize the logic that is going to be refactored in #331 and #203
```
Olshansk, line 12:
```suggestion
// HACK: This is a temporary solution simply used to centralize the logic that is going to be refactored in #331 and #203
```

p2p/addrbook_provider/getter.go
```diff
@@ -0,0 +1,28 @@
...
+func GetAddrBook(bus modules.Bus, addrBookProvider typesP2P.AddrBookProvider) typesP2P.AddrBook {
+	var (
+		addrBook typesP2P.AddrBook
+		err      error
+	)
+	if bus.GetPersistenceModule() == nil {
+		// TODO (#203): refactor this.
```
Olshansk, line 19:
```suggestion
		// TODO(#203): refactor or remote this once `ValidatorMap` logic is encapsulated.
```

p2p/addrbook_provider/getter.go
```diff
@@ -0,0 +1,28 @@
...
+	"log"
+
+	typesP2P "github.com/pokt-network/pocket/p2p/types"
+	"github.com/pokt-network/pocket/shared/modules"
+)
+
+// GetAddrBook is a helper function that returns the addrBook depending on the availability of the persistence module
+//
+// this is a temporary solution simply used to centralize the logic that is going to be refactored in #331 and #203
+func GetAddrBook(bus modules.Bus, addrBookProvider typesP2P.AddrBookProvider) typesP2P.AddrBook {
```
Olshansk, line 13:
```suggestion
func GetAddrBook(bus modules.Bus, addrBookProvider typesP2P.AddrBookProvider) (addrBook, typesP2P.AddrBook, err error) {
```

p2p/addrbook_provider/getter.go
```diff
@@ -0,0 +1,28 @@
...
+	var (
```
Olshansk, line 14: If you name the return values, you can avoid needing to declare these here

p2p/addrbook_provider/getter.go
```diff
@@ -0,0 +1,28 @@
...
+		err      error
+	)
+	if bus.GetPersistenceModule() == nil {
+		// TODO (#203): refactor this.
+		addrBook, err = addrBookProvider.ValidatorMapToAddrBook(bus.GetConsensusModule().ValidatorMap())
+	} else {
+		addrBook, err = addrBookProvider.GetStakedAddrBookAtHeight(bus.GetConsensusModule().CurrentHeight())
+	}
+	if err != nil {
+		log.Fatalf("[ERROR] Error getting addrBook: %v", err)
```
Olshansk, line 25: Return the error and let the caller handle it.

runtime/defaults/defaults.go
```diff
@@ -31,4 +31,5 @@ var (
 	DefaultRpcPort             = defaultRPCPort
 	DefaultRpcTimeout          = uint64(defaultRPCTimeout)
 	DefaultRemoteCLIURL        = fmt.Sprintf("http://%s:%s", defaultRPCHost, defaultRPCPort)
+	DefaultP2PMempoolMaxNonces = uint64(1e6)
```
Olshansk, line 34: DefaultP2PMempoolMaxNonceCount?

p2p/module.go
Olshansk: To quote you previous message: 

> Where would this function be called currently? I would say that if it's not relevant to the scope of this issue we shouldn't do it here. Sounds like scope creep a little bit. No?

I'm personally not a huge fan because these events were out of scope and are not being used, though I can see where your head is at.

If I were to introduce a new event, I'd probably create a `NewViewEvent(height, round)` and/or `NewHeight(height)` because I can only speculate how we'd use `BeforeNewHeightEvent` right now and the cost to add it is low.

I don't want this to continue being a blocker, but not a fan. Let's do this:
1. If you feel strongly that `BeforeNewHeightEvent` and `HeightChangedEvent` events will be useful, let's keep it as is.
2. If you're not sure, let's do one of the following:
2.1. Remove them altogether
2.2 Just add a `NewViewEvent` (per my suggestions)

p2p/module.go
Olshansk: 👍 Going to resolve this and review the refactored code.

consensus/pacemaker.go
```diff
@@ -262,3 +267,19 @@ func (p *paceMaker) getStepTimeout(round uint64) timePkg.Duration {
 	baseTimeout := timePkg.Duration(int64(timePkg.Millisecond) * int64(p.pacemakerCfg.GetTimeoutMsec()))
 	return baseTimeout
 }
+
+func (p *paceMaker) onBeforeNewHeight() {
```
Olshansk, line 271: Linking to our discussion here: https://github.com/pokt-network/pocket/pull/374#discussion_r1046583124

If we do decide to keep it, we should really leave a comment explaining "Not used, but can be used as a hook", etc..

p2p/types/addr_book.go
```diff
@@ -1,9 +1,20 @@
...
 
 // AddrBookMap maps p2p addresses to their respective NetworkPeer.
 //
 // Since maps cannot be sorted arbitrarily in Go, to achieve sorting, we need to rely on `addrList` which is a slice of addresses/strings and therefore we can sort it the way we want.
 type AddrBookMap map[string]*NetworkPeer
+
+// AddrBookProvider is an interface that provides AddrBook accessors
+type AddrBookProvider interface {
+	GetStakedAddrBookAtHeight(height uint64) (AddrBook, error)
+	ValidatorMapToAddrBook(validators map[string]modules.Actor) (AddrBook, error)
```
Olshansk, line 18: Add a TODO to remove all the `Validator` specific terminology from the interface.

persistence/genesis.go
```diff
@@ -321,3 +321,30 @@ func (p PostgresContext) GetAllFishermen(height int64) (f []modules.Actor, err e
 	}
 	return
 }
+
+// IMPROVE: This is a proof of concept. Ideally we should have a single query that returns all actors.
+func (p PostgresContext) GetAllStakedActors(height int64) (actors []modules.Actor, err error) {
```
Olshansk, line 326: I don't think `genesis.go` is the right place for this. Can you creator a new `actor.go` file?

persistence/genesis.go
```diff
@@ -321,3 +321,30 @@ func (p PostgresContext) GetAllFishermen(height int64) (f []modules.Actor, err e
...
+	actors = make([]modules.Actor, 0)
```
Olshansk, line 327: if we don't initialize this, does it break?

p2p/raintree/peers_manager.go
```diff
@@ -30,7 +30,20 @@ type peersManager struct {
...
+		return nil, err
+	}
+
+	return newPeersManager(selfAddr, addrBook, false)
+}
+
+// newPeersManager creates a new peersManager instance, it is in charge of handling operations on peers (like adding/removing them) within an AddrBook
+// it also takes care of keeping the AddrBook sorted and indexed for fast access
+//
+// If `isDynamic` is true, the peersManager will not handle addressBook changes, it will only be used for querying the AddrBook
```
Olshansk, line 45:
```suggestion
// If `isDynamic` is false, the peersManager will not handle addressBook changes, it will only be used for querying the AddrBook
```

Olshansk (changes requested):

Olshansk:
@deblasis Sorry for hom many back & forths this one is taking, but it's looking better with every iteration, and thank you for bearing!

deblasis:
> @deblasis Sorry for hom many back & forths this one is taking, but it's looking better with every iteration, and thank you for bearing!

It's all good @Olshansk ! It's all part of the process. I should be the one apologizing if you think about it. 

I'll address your comments and re-request review later today. 🙏

p2p/raintree/peers_manager.go
deblasis: Don't trust AI! 🙃

p2p/addrbook_provider/persistence.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (pabp *persistenceAddrBookProvider) GetBus() modules.Bus {
+	return pabp.bus
+}
+
+func (pabp *persistenceAddrBookProvider) SetBus(bus modules.Bus) {
+	pabp.bus = bus
+}
+
+func (pabp *persistenceAddrBookProvider) GetStakedAddrBookAtHeight(height uint64) (typesP2P.AddrBook, error) {
```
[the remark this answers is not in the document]
deblasis: for consistency, that's being handled in the #378 

https://github.com/pokt-network/pocket/pull/378/files#diff-7a8d5551d442e1630313d4aaa3a8fb51812ee580d00efb62c4b12c385c27f6d4R12-R13

that PR builds on top of this one.

persistence/genesis.go
deblasis: I agree with you but all the similar counterparts are here as well... I'll move all of them then.

On that note, I was about to start tackling #149 

```
// TODO(pocket/issues/149): All of the functions below following a structure similar to `GetAll<Actor>`
//	can easily be refactored and condensed into a single function using a generic type or a common
//      interface.
``` 
but then I remembered my promises from earlier today in planning. I might come back at it for vengeance later...

persistence/genesis.go
```diff
@@ -321,3 +321,30 @@ func (p PostgresContext) GetAllFishermen(height int64) (f []modules.Actor, err e
 	}
 	return
 }
+
```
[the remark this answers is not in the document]
deblasis: Unit test for a function marked as proof of concept? Mhhh I can, yes. Should I? Not sure. Probably not.

I am curious how would you test this. If I had to test it I'd implement configurable mocks within the test, driven by the table test to cover some scenarios and edge cases and I'd compare the expectation with the returned slice.

persistence/genesis.go
deblasis: Oh, well, good to know! Thanks @Olshansk 🙏

https://go.dev/play/p/GjBg0QuZEbB

image
image

Nice

persistence/genesis.go
```diff
@@ -321,3 +321,30 @@ func (p PostgresContext) GetAllFishermen(height int64) (f []modules.Actor, err e
...
+// IMPROVE: This is a proof of concept. Ideally we should have a single query that returns all actors.
+func (p PostgresContext) GetAllStakedActors(height int64) (actors []modules.Actor, err error) {
+	actors = make([]modules.Actor, 0)
+	var allActors []modules.Actor
```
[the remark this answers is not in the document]
deblasis: Still not solving the underlying problem but it looks more elegant, sure!

p2p/types/addr_book.go
deblasis: Not necessary, already handled in #378 

https://github.com/pokt-network/pocket/pull/378/files#diff-d4569455b8c8737b4fc4fd9fd0be440b504ea90e873af6c25ce01524c9957711R18-R19

persistence/genesis.go
```diff
@@ -205,119 +205,3 @@ func (p PostgresContext) GetAllPools(height int64) (accs []modules.Account, err
...
-		var actor *types.Actor
-		actor, height, err = p.getActorFromRow(rows)
-		if err != nil {
-			return
-		}
-		actors = append(actors, actor)
-	}
-	rows.Close()
-	for _, actor := range actors {
-		actor, err = p.getChainsForActor(ctx, tx, types.ApplicationActor, actor, height)
```
deblasis, line 258: I spotted this bug, this should be `ValidatorActor`, fixed in while I moved it to `actor.go`

deblasis:
I have removed the events, addressed comments and it's ready for another pass. Fewer things to look at I guess. We are getting there.

When you are happy I'll update the `CHANGELOGs`

persistence/genesis.go
Olshansk: Can you update that comment (since #149) was already resolved and just remove the ticket number for a plain TODO?

persistence/genesis.go
Olshansk: 200w

persistence/genesis.go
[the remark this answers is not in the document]
Olshansk: Yea, it's just about reduce copy-pasta

persistence/genesis.go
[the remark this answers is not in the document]
Olshansk: Fair enough, ignore!

p2p/raintree/utils_test.go
```diff
@@ -0,0 +1,25 @@
+package raintree
+
+import (
+	"github.com/golang/mock/gomock"
+	typesP2P "github.com/pokt-network/pocket/p2p/types"
+	mocksP2P "github.com/pokt-network/pocket/p2p/types/mocks"
+	mockModules "github.com/pokt-network/pocket/shared/modules/mocks"
+)
+
+func mockDummyBus(ctrl *gomock.Controller) *mockModules.MockBus {
```
Olshansk, line 10: Can we just call this `mockBus`?

persistence/genesis.go
deblasis: I have created a new issue and referenced that so that it doesn't get lost in the `void`

p2p/addrbook_provider/persistence.go
```diff
@@ -0,0 +1,108 @@
...
+	persistenceReadContext, err := pabp.GetBus().GetPersistenceModule().NewReadContext(int64(height))
+	if err != nil {
+		return nil, err
+	}
+	stakedActors, err := persistenceReadContext.GetAllStakedActors(int64(height))
+	if err != nil {
+		return nil, err
+	}
+	// TODO(#203): refactor `ValidatorMap``
```
Olshansk, line 60:
```suggestion
	// TODO(#203): refactor `ValidatorMap`
```

Olshansk (changes requested):
LGTM, let's just update the CHANGELOG and we're good to go

deblasis:
@Olshansk CHANGELOGs updated

runtime/docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.1] - 2022-12-14
+
+- Added `DefaultP2PMempoolMaxNonceCount`
```
Olshansk, line 12: Optional: Rename this to `DefaultP2PMempoolMaxCount` instead `DefaultP2PMempoolMaxNonceCount` to match the interface name

Olshansk (approved):

runtime/docs/CHANGELOG.md
deblasis: image

I see your `DefaultP2PMempoolMaxCount` and I am going all-in with `DefaultP2PMaxMempoolCount` since the interface method is called `GetMaxMempoolCount`

MichaelCeraMollysGameGIF

Olshansk: monkey-call

Issue #361, opened 2022-11-27
[Persistence] First State Hash Implementation - Followup Investigations


Olshansk:
## Objective

Identify missing gaps and define follow-up work to the first implementation of the state hash in #285.

## Origin Document

PR #285 originally set out to resolve #284, and it did so for the most part. However, the scope of the task is big and it is closely linked to the entire lifecycle of the Pocket Node causing the PR to grow in size and a long time to merge in.

In an effort to close to the loop and have smaller future iterations, this task is used to track open questions from the PR's self & peer review.

### Items of Investigation

1. Could / should we split all `prepareAndApply` functions into `prepare` and `apply` so the business logic can be reused in a more functional way between the proposer and others? Note that this spans persistence, utility and consensus.
2. Could `getTxsHash` be removed altogether?
3. Investigate whether we could/should remove `quorumCert` from proposal block.
4. Investigate whether the `proposalBlock` state should be maintained in the `persistenceContext` or the `utilityContext`.
5. Consolidate `README`s and `CHANGELOG`s that exist in the same module but more than one copy exists
6. Research whether we even need `transactionsHash` in `persistence/proto/block_persistence.proto`.
7. Genesis - research whether the geneis block requires a non-nil quorumCert and proposerAddr?
8. Clarify & consolidate (via code cleanup and documentation) the difference of Blockhash, AppHash and StateHash
 
## Goals

- Determining safety/security risks in the current state hash implementation
- Determining techdebt (i.e. lack of clarity in code flow) in the current implementation of the state hash
- Defining work that needs to be done in the short-term, mid-term and long-term as follow up to the state hash

## Deliverable

- [ ] Minor changes/updates to the documentation; structuring, formatting, organization, etc
- [ ] An explanation of which items need to be tended to and which do not (with backing research or justification)
- [ ] A list of prioritized issues to be tended to in subsequent PRs

## Non-goals / Non-deliverables

- Major (business logic) documentation changes
- Any code changes

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)

## Testing Methodology

- **All tests**: `make test_all`
[...]

Issue #283, opened 2022-10-04
[Consensus] Re-evaluate utility context management


Olshansk:
## Objective

Figure out how the utility context within consensus should be managed, refreshed and updated to avoid panics and race conditions.

## Origin Document

[This thread](https://github.com/pokt-network/pocket/pull/250#discussion_r982505262) In #250:

<img width="735" alt="Screen Shot 2022-10-04 at 4 46 43 PM" src="https://user-images.githubusercontent.com/1892194/193950461-0a0f5b02-ebef-4539-82e5-d9a86e34e0d4.png">

## Goals

- Avoid panics that can be a result of access to a `nil` `utilityContext`
- Avoid race conditions in the consensus lifecycle that can lead to issues in the use of `utilityContext`
- Update diagrams/documentation so it is abundantly clear how the consensus module should manage its instance of the `utilityContext`

## Deliverable

- [ ] A PR updating how the `utilityContext` in the consensus module is managed
- [ ] Updated unit tests validating these changes
- [ ] Add/update the relevant sequence diagram if necessary [mermaid](https://mermaid-js.github.io/mermaid/)
- [ ] Update documentation in the relevant READMes so this is clear
- [ ] Consider adding the following function per Gohkan's comment [here](https://github.com/pokt-network/pocket/pull/391#discussion_r1049469665): `func (u *UtilityContext) GetPersistanceReadContext() { return u.Context.PersistenceRWContext.PersistanceReadContext }`

## Non-goals / Non-deliverables

- Change any interfaces in the `ConsensusModule` or `UtilityModule`

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references

## Testing Methodology

- **All tests**: `make develop_test`
- **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @Olshansk 
**Co-Owners**: @andrewnguyen22

Olshansk:
@andrewnguyen22 I want to discuss this in a protocol hour before someone takes it on.

Olshansk:
Superseeded by @deblasis's "utilityUnitOfWork"

PR #391, opened 2022-12-11
[Persistence][Utility] Proposal block metadata maintained in `utilityContext` rather than `postgresContext`

Olshansk:
## Description

Maintain the proposed block in the `utilityContext` rather than the `postgresContext.`

This is a cleanup with some tech debt to help with the separation of concerns w.r.t contexts:
- Utility - has a semantic understanding of what is in the proposed block and whether it is valid
- Persistence - Responsible for computing & committing state integrity, but is agnostic to the actual state change of the transactions being proposed/applied.

This is a stepping stone to help with:
- Persistence rollback - changes to the SQL DB and Trees (i.e. KV stores) needs to be atomic and then either reverted or committed to disk
- Consensus state sync - when a node is catching up, it will also go through the `Propose` -> `Apply` -> `Commit` lifecycle, but it won't be dependant on the latency of consensus since it happened in the past so the `proposer` and `commitQC`  are available immediately

## Issue

Fixes part of #361:

Screenshot 2022-12-10 at 4 27 06 PM

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

**PersistenceModule Changes:**
- Remove `SetProposalBlock` and local vars to keep proposal state
- Add `proposerAddr` to the `Commit` function
- Move the `PostgresContext` struct to `context.db`

**Utility Module Changes:**
- Introduce `SetProposalBlock` and local vars to keep proposal state
- Maintaining proposal block state (proposer, hash, transactions) in local context

**Other Changes:**
- Partially start consolidating `appHash` and `stateHash` terminology
- Partially start consolidating all `hash` related vars to `string` instead of `[]byte`

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


consensus/consensus_tests/utils_test.go
```diff
@@ -364,17 +364,17 @@ func baseUtilityContextMock(t *testing.T) *modulesMock.MockUtilityContext {
...
 		CreateAndApplyProposalBlock(gomock.Any(), maxTxBytes).
-		Return(appHash, make([][]byte, 0), nil).
+		Return(stateHash, make([][]byte, 0), nil).
 		AnyTimes()
 	utilityContextMock.EXPECT().
 		ApplyBlock().
-		Return(appHash, nil).
+		Return(stateHash, nil).
 		AnyTimes()
+	utilityContextMock.EXPECT().SetProposalBlock(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
```
gokutheengineer, line 377: @Olshansk This causes tests to fail to compile, since struct `MockUtilityContext` doesn't implement function `SetProposalBlock`.

gokutheengineer (changes requested):

consensus/consensus_tests/utils_test.go
Olshansk: Did this happen to you locally?

Note how the CI checks passed:

Screenshot 2022-12-13 at 4 44 43 PM

Here's also a screenshot of it passing on my local machine:

Screenshot 2022-12-13 at 4 45 11 PM

And in my `utility_module_mock.go`:

Screenshot 2022-12-13 at 4 46 40 PM


When we change the interfaces (and in turns the mocks on which the interfaces are built), we need to run `make mockgen`, so they are recompiled. I like to run `make develop_test` whenever I want to do a clean end-to-end.

Lmk if that makes sense

consensus/types/proto/block.proto
```diff
@@ -14,7 +14,7 @@ message BlockHeader {
   uint32 numTxs = 5;
   int64 totalTxs = 6; // Total = total in the entire chain Num = total in block (Tendermint legacy)
   string lastBlockHash = 7;
-  bytes proposerAddress = 8;
+  bytes proposerAddress = 8; // TODO: Change this to a string
```
deblasis, line 17: Not actionable, only a "stupid question".

Just for my understanding: we are going to prefer strings to byte arrays for anything that's addresses and hashes, correct?

Is it for simplicity or there are other reasons?

persistence/test/block_test.go
```diff
@@ -10,16 +10,16 @@ func TestGetBlockStateHash(t *testing.T) {
 	db := NewTestPostgresContext(t, 0)
 
 	// Cannot get prev hash at height 0
-	appHash, err := db.GetBlockHash(0)
+	stateHash, err := db.GetBlockHash(0)
```
deblasis, line 13: I am not sure if it's intended but as basically a newbie in this part of the code, I find it "odd" even if it makes sense to see

```go
stateHash, err := db.GetBlockHash()
```
shouldn't this be 

```go
blockHash, err := db.GetBlockHash()
```
?

persistence/genesis.go
```diff
@@ -145,17 +145,14 @@ func (m *persistenceModule) populateGenesisState(state modules.PersistenceGenesi
 	}
 
 	// Updates all the merkle trees
-	appHash, err := rwContext.ComputeAppHash()
+	_, err = rwContext.ComputeStateHash()
```
deblasis, line 148: I would probably at least log the stateHash for debugging purposes.

Any thoughts against that?

deblasis (approved):
Solid work, just a couple of comments/questions

consensus/types/proto/block.proto
Olshansk: > Is it for simplicity or there are other reasons?

1. Simplicity - they're small, not big data blobs
2. Readability - Things like `address` or `stateHash` are "human readable" in the sense that we use & reference it (e.g. printing)
3. Performance (micro optimization) - lots of unnecessary encoding & decoding throughout the codebase

persistence/test/block_test.go
Olshansk: **tl;dr Yes**

Yea, the `appHash` and the `stateHash` are the same thing which was an inconsistency.

I still have to do a bit of research whether we are going to do anything different with `blockHash`. For example, the `quorumCert` and `proposer` are included in the block but not part of the state hash, but TBD.

Olshansk: Potentially addressed here: [`66ed7fc` (#391)](https://github.com/pokt-network/pocket/pull/391/commits/66ed7fc307771c9a57e1c31e21ee285984b471bf)

persistence/genesis.go
Olshansk: Potentially addressed here: [`39da0ed` (#391)](https://github.com/pokt-network/pocket/pull/391/commits/39da0ed1ab5e3dc8ac1101d2901fe8b752e5b7a2)

persistence/test/block_test.go
deblasis: Awesome, thank you! 🙏

gokutheengineer:
As a general comment, and probably not in the scope of this issue, I think we need to divide UtilityContext interface to smaller interfaces, and only leave common functions that are to be used by both block proposers and block verifiers together. This will not change how utility operates, however it can maybe provide better code clarity. 

For example, `CreateAndApplyProposalBlock` is to be used by the block proposers, as mentioned in the comment:

``` go
	//Reaps the mempool for transactions to be proposed in a new block, and applies them to this
	//context; intended to be used by the block proposer.
	CreateAndApplyProposalBlock(proposer []byte, maxTransactionBytes int) (stateHash string, transactions [][]byte, err error)
```

Having a separate interface implemented by UtilityContext, such as for Proposer:

```go
type UtilityProposerContext interface {
       CreateAndApplyProposalBlock(proposer []byte, maxTransactionBytes int) (stateHash string, transactions [][]byte, err error)
}
```

might make the roles of Proposer and Verifier much more clear. In later stages we can consider, implementing only subset of  interfaces interfaces for different roles, and enforce strict control.

Olshansk:
@gokutheengineer Completely agreed. Just a matter of prioritizing it. If you want to create a ticket or just add a TODO in the code, I'd 👍 it.

In #361, see the first action item:

Screenshot 2022-12-15 at 6 32 29 AM

I think this applies to the logic we have in utility, in persistence and in consensus too!

Olshansk:
@gokutheengineer Also, since you requested change sin the past, this PR is blocked until you re-review / approve.

persistence/block.go
```diff
@@ -36,43 +35,28 @@ func (p PostgresContext) GetLatestBlockHeight() (latestHeight uint64, err error)
 	return
 }
 
-// OPTIMIZE: evaluate if it's  faster to get this from the blockstore (or cache) than the SQL engine
-func (p PostgresContext) GetBlockHash(height int64) ([]byte, error) {
+func (p PostgresContext) GetBlockHash(height int64) (string, error) {
```
gokutheengineer, line 38: Similar to @deblasis's previous comment, out of curiosity, what is the main reason to move to string from bytes?

persistence/test/block_test.go
gokutheengineer: agreed, as my previous comment.

shared/modules/utility_module.go
```diff
@@ -26,17 +26,24 @@ type UtilityModule interface {
 type UtilityContext interface {
 	// Block operations
 
+	SetProposalBlock(blockHash string, proposerAddr []byte, transactions [][]byte) error
```
gokutheengineer, line 29: Shouldn't we add a comment stating this should be only used by the proposer? Or is there a reason for validators to run this?

gokutheengineer:
Added few more comments on minor things. Please have a look, otherwise LGTM!

gokutheengineer (approved):

consensus/hotstuff_leader.go
```diff
@@ -341,15 +342,13 @@ func (m *consensusModule) prepareAndApplyBlock(qc *typesCons.QuorumCertificate)
...
 	if err != nil {
 		return nil, err
 	}
 
-	persistenceContext := m.utilityContext.GetPersistenceContext()
-
-	// CONSOLIDATE: Last/Prev & AppHash/StateHash/BlockHash
-	prevAppHash, err := persistenceContext.GetBlockHash(int64(m.Height) - 1)
+	// IMPROVE: This data can be read via an ephemeral read context - no need to use the utility's persistence context
+	prevStateHash, err := m.utilityContext.GetPersistenceContext().GetBlockHash(int64(m.Height) - 1)
```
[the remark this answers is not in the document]
Olshansk: Adding this as a sub-bullet in https://github.com/pokt-network/pocket/issues/283#issue-1396979181 and updated the TODO in the code as well.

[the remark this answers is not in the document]
Olshansk: I believe you're looking at an older sha of the PR, already done.

shared/modules/utility_module.go
Olshansk: Adding the following comment:

```
	// This function is intended to be called by any type of node during state transitions. For example,
	// both block proposers and verifiers will use it to create a context (before finalizing it) during consensus,
	// and all verifiers will call it during state sync.
```

persistence/block.go
Olshansk: Same answer as here: https://github.com/pokt-network/pocket/pull/391#discussion_r1049075566

1. Simplicity - they're small, not big data blobs
2. Readability - Things like address or stateHash are "human readable" in the sense that we use & reference it (e.g. printing)
3. Performance (micro optimization) - lots of unnecessary encoding & decoding throughout the codebase

Olshansk:
@gokutheengineer Even though I merged it in, please `Resolve` the comments or let me know if there's anything additional I can clarify.

PR #397, opened 2022-12-14
[Tooling] CHANGELOG and Version Bump PreCommit Hook

### LINKED ISSUE #93: [Tooling] CHANGELOG and Version Bump PreCommit Hook
### Description

The v1 codebase is required to document all important changes in a `CHANGELOG.md` at the root of the repository and module specific `CHANGELOG.md` if the changes impacted the module in-question. In addition to that, the version should be bumped accordingly.

This issue is a perfect starter task for community contributions.

## Origin Document

This issue was created as a starter task for community contribution, as suggested per in the following [conversation](https://github.com/pokt-network/pocket/pull/80#discussion_r898622625)

## Goals

- [ ] Add pre-commit logic to validate if CHANGELOGs and the version were updated accordingly before commit
- [ ] Allow this to be disable locally
- [ ] Configure CI to always run this validation before merge of PRs

## Non-Goals

- [ ] Adding CI/CD to the V1 Codebase
- [ ] Automating the population of the CHANGELOG based on git-diff results

## Maintainers
Creator: @derrandz

h5law:
## Description

This PR contains the pre-commit logic to check and validate whether the relevant `CHANGELOG.md` files have (1) been included in the commit and (2) contain up-to-date versioning. This is in response to [Issue #93](https://github.com/pokt-network/pocket/issues/93).

Git Hook:
The client-side logic is contained in the bash script `.githooks/pre-commit` for it to actually perform this verification check prior to a commit it must be copied into `.git/hooks` and be executable.
```sh
cp .githooks/pre-commit .git/hooks
chmod +x .git/hooks/pre-commit
```
As it is a git hook, it can be disabled locally by running `git commit` with the `--no-verify` or `-n` flag:
```sh
git commit -n ...
```
The script will detect if a `CHANGELOG.md` is present and if it is it will check whether (1) the latest version in the changelog is greater than the one before it and (2) that the date alongside the latest version is the current date. If any of these checks fail (no changelog, invalid version, not current date) then the script will block the commit.

To verify the changelog's version is correct it converts the string `[0.0.0.8]` to `0008` and then strips the leading zeros and compares the number against the next version found if it is less than or equal to the one previous it sees this as an error. This is done by extracting the versions and dates from the `CHANGELOG.md` file using `sed`. It then compares the date of the newest version against today's date and if they aren't equal and it sees this as an error.

Currently, the script uses the directories `.githooks` `.github` `docs` `bin` as ignored directories and does not check for a `CHANGELOG.md` file in these subdirs.

Dependencies:
This script requires `bash` to be installed and requires `sed`, `head` and `date` commands - these should be present on all Linux/OSX/BSD systems. The script uses sed Basic Regular Expressions that are portable across GNU sed and FreeBSD sed (used by OSX as well) and also uses `date` formatting that is consistent across Linux/OSX/BSD versions of the command. Currently, there is no support for Windows when run locally.

GitHub Workflow:
The GitHub workflow will run the `pre-commit` script however it will populate the edited files by instead passing them as arguments to the script. This cannot be disabled and will run on all PRs when they are opened, reopened and synchronised. The jobs are configured to run on `ubuntu-latest` so there are no issues with dependencies as all the required binaries are present on ubuntu installs.

The files in question are found in:
```
.github
└── [ 112] workflows
      └── [ 1KB] changelog-verify.yml
.githooks
└── [ 4KB] pre-commit
docs
└── [ 60] development
      └── [ 6KB] README.md
```

## Issue

Fixes #93 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Add `pre-commit` git hook to validate changelogs presence in commits & correct versioning
- Add GitHub workflow to run the `pre-commit` logic on all PRs [opened, reopened, synchronised]
- Add a guide to activate `pre-commit` hook to `docs/development/README.md`
- Updated `utility/doc/CHANGELOG.md`

## Testing

Videos:
[changelog-verify-test-video.webm](https://user-images.githubusercontent.com/53987565/207724785-71f2bcc1-f3a5-401e-b543-17efaadaa6f2.webm)
[github-workflow-test-video1.webm](https://user-images.githubusercontent.com/53987565/207724787-b7c7c303-ab22-4689-a13c-341ef1cd7901.webm)
[github-workflow-test-video2.webm](https://user-images.githubusercontent.com/53987565/207724793-ed4c2b38-5c57-407d-97c0-98999ba11d4f.webm)

## Required Checklist

- [x] I have tested my changes using the available tooling --> I have tested locally, and using GitHub PRs for the workflow

### If Applicable Checklist


.github/workflows/changelog-verify.yml
```diff
@@ -0,0 +1,42 @@
+# This runs the pre-commit hook logic to check whether the relevent
+# CHANGELOG.md files have been updated for both the root directory
+# and/or modules edited if they have been changed.
+
+name: Validate Changelogs
+
+on:
+  pull_request:
+    types: [opened, reopened, edited, synchronize]
```
Olshansk, line 9: Maybe remove `edited` to avoid rerunning it "EVERY" time?

.githooks/pre-commit
```diff
@@ -0,0 +1,166 @@
...
+# Pre-commit hook to check if CHANGELOG.md has been included
+# when the project root or relevent module has been edited.
+# If it has been included the latest version is compared
+# against the previous version by stripping all leading zeros
+# and comparing them. If it is greater it goes on to check
+# that the latest change's date is the current date today.
+# If any of these tests fail the script exits and cancels the
+# commit.
+# To bypass this validation add the `--no-verify` flag:
+# 		`git commit --no-verify ...`
```
Olshansk, line 12:
```suggestion
# Pre-commit hook to check if CHANGELOG.md has been included
# when the project root or relevant module has been edited.
#
# If it has been included, the latest version is compared
# against the previous version by stripping all leading zeros
# and comparing them. If it is greater it goes on to check
# that the latest change's date is the current date (i.e. today).
# If any of these tests fail the script exits and cancels the
# commit.
#
# To bypass this validation add the `--no-verify` or `-n` flag:
# 		`git commit --no-verify ...`
```

.githooks/pre-commit
```diff
@@ -0,0 +1,166 @@
...
+
+# Get names of staged files being committed. If being run as
+# a git hook take the result of git diff. If not recieve
+# the edited files from the CLI arguments provided
```
Olshansk, line 16:
```suggestion
# Get names of staged files being committed. If being run as
# a git hook, take the result of git diff. If not, recieve
# the edited files from the CLI arguments provided.
```

.githooks/pre-commit
```diff
@@ -0,0 +1,166 @@
...
+if [[ "${#}" -eq 0 ]]
+then
+    STAGED_FILES=($(git diff --cached --name-only))
```
Olshansk, line 19:
```suggestion
    STAGED_FILES=($(git diff --cached --name-only)) # Local git hook
```

.githooks/pre-commit
```diff
@@ -0,0 +1,166 @@
...
+else
+    STAGED_FILES=("${@}") # Github workflow passes arguements
+fi
+
+if [[ "${#STAGED_FILES[@]}" == 0 ]]
+then
+	echo "No files added"
```
Olshansk, line 26:
```suggestion
	echo "No files modified"
```

.githooks/pre-commit
```diff
@@ -0,0 +1,166 @@
...
+		fi
+	fi
+done
+
+# Check if the root directory has been changed and if the
+# root CHANGELOG.md file was included in the commit
+if [[ "${ROOT_EDITED}" == 1 && "${ROOT_CHANGELOG_INCLUDED}" == 0 ]]
+then
+	printf 'Root changelog not included\n\n' >&2
+    CHANGELOG_ERRORS_FOUND=1
```
Olshansk, line 83:
```suggestion
	printf 'Root changelog not included\n\n' >&2
        CHANGELOG_ERRORS_FOUND=1
```

.githooks/pre-commit
```diff
@@ -0,0 +1,166 @@
...
+do
+    if [[ "${MODULES_MISSING_CHANGELOG[*]}" =~ "${module}" ]]
+    then
+        continue
+    fi
+	# sed using BRE expression for portability
+	head2versions=($(cat "${module}/CHANGELOG.md" | sed -n 's/^.*\(\[[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\]\) - \([0-9][0-9]*-[0-9][0-9]*-[0-9][0-9]*\).*$/\1 \2/p' | head -n2))
+	# Get latest and previous versions and strip any leading 0s
+	# and check the latest version is greater than the previous
+	latestversion=$(($(echo ${head2versions[0]:1:-1} | sed 's/\.//g' | sed 's/^0*//')+0))
```
Olshansk, line 135: Why not snake case? Ditto elsewhere

.github/workflows/changelog-verify.yml
```diff
@@ -0,0 +1,42 @@
...
+  validate:
+    runs-on: ubuntu-latest
+      # require the first job to have ran
+    needs: changedfiles
+      # only run there are changed files
+    if: ${{needs.changedfiles.outputs.all}}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Verify changelogs
+        run: bash ./.githooks/pre-commit ${{needs.changedfiles.outputs.all}}
```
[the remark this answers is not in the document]
h5law: From my understanding: on the client side, the `exit 1` prevents the commit unless the flag `--no-verify` is used and this is bypassed; for the GitHub workflow the `exit 1` will fail the test but it can still be merged if you wish. To change this I think you would have to specify that this test completes successfully in the repository settings. If this is the desired outcome I could change the workflow to assign the `do not merge` label to prevent merging if the test fails. 

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule

h5law:
I have addressed the comments you gave @Olshansk you didn't mention it so I think it will be fine but if you want me to stop checking for changes in the root directory (no real code there + no changelog) and any others let me know I can quickly do this if needed.

docs/development/README.md
```diff
@@ -65,6 +65,12 @@ $ git clone git@github.com:pokt-network/pocket.git && cd pocket
 $ make develop_start
 ```
 
+Activate changelog pre-commit hook
```
Olshansk, line 68:
```suggestion
Optionally activate changelog pre-commit hook
```

docs/development/README.md
```diff
@@ -65,6 +65,12 @@ $ git clone git@github.com:pokt-network/pocket.git && cd pocket
...
+```bash
+cp .githooks/pre-commit .git/hooks/pre-commit
+chmod +x .git/hooks/pre-commit
+```
+
```
Olshansk, line 73:
```suggestion
_Please note that the Github workflow will still prevent this from merging unless the CHANGELOG is updated._ 

```

.github/workflows/changelog-verify.yml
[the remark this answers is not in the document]
Olshansk: We have a `do-not-merge` workflow here: https://github.com/pokt-network/pocket/blob/main/.github/workflows/do-not-merge.yml

The idea is that if we apply the label, it prevents the PR from being merged to master.

The reason I ask is because our usual workflow is:
1. Make a bunch of commits locally
2. Push up for review
3. Get feedback
4. Repeat steps 1-3 a few times
5. Update changelog
6. Squash & Merge the github PR into main

I don't think the script will be used as a precommit, but will be very much more valuable as a workflow.

Can you confirm that I'm not missing anything?

Olshansk (changes requested):
A couple more comments, a question and then good to go

h5law:
@Olshansk I have addressed the issues you brought up in the documentation file and added a third job to run after the validation step in the workflow. If this job sees that the validation step failed, it will automatically add the 'do not merge' label, which _should_ trigger the pre-existing do-not-merge workflow to prevent any merging until the label is removed. I don't think automatically removing the label if the test passes would be a good idea as it may be applied for some other reason besides a CHANGELOG so I left it at simply applying the label.

Olshansk:
> it will automatically add the 'do not merge' label

Do we really need to do this? I figured that the existing `CHANGELOG` workflow would throw an exit 1, which will make the job fail and prevent merging as is. My questions was meant as more of a confirmation to validate my understanding.

If that's the case, I think its simpler to remove the workflow of adding the label, and then merge it in!

h5law:
> > it will automatically add the 'do not merge' label
> 
> Do we really need to do this? I figured that the existing `CHANGELOG` workflow would throw an exit 1, which will make the job fail and prevent merging as is. My questions was meant as more of a confirmation to validate my understanding.
> 
> If that's the case, I think its simpler to remove the workflow of adding the label, and then merge it in!

I misunderstood the question sorry. I have removed the `do not merge` auto label job from the workflow. The `pre-commit` script that the workflow runs will exit with code `1` whenever there are any issues. It will log _all_ the issues it finds not only the first one. This will cause the test to fail.

From my understanding whether or not a failed test prevents the PR being merged is decided in the repository settings. However I am not 100% on this, I am getting this from the documentation I linked in a previous message. I believe the `exit 1` essentially does the same as the `do not merge` workflow that is already present so I am sure that you are right in thinking that adding the label is unnecessary.

I hope this is all alright 😃

Olshansk:
>  I have removed the do not merge auto label job from the workflow

Appreciate you for bearing on the back-and-forth :)

> From my understanding whether or not a failed test prevents the PR being merged is decided in the repository settings.

That is correct, and is the case for us.

> However I am not 100% on this, I am getting this from the documentation I linked in a previous message

This isn't "mission critical" and I have high-confidence in the changes after the review. 

If there's a small update/change we need to make after merging, would you be able to help us out?

> I hope this is all alright 😃

🤝

h5law:
> If there's a small update/change we need to make after merging, would you be able to help us out?

Yeah of course, I hope I can get more involved anyway

h5law:
To address those final commits - I realised the `CHANGELOG.md` file detection was wrong as the CHANGELOGs are stored in different sub directories (root of module, doc, docs) and so to address this when a `CHANGELOG.md` is detected in the path of a file the script will now add this path to a list and loop through that instead of trying to "build" the `CHANGELOG.md` path itself.

I also removed the dependency on the date command (past the basic usage that is portable across linux/osx/bsd) and am now comparing the dates from the changelog as a string against today's date.

Finally I have updated the relevant changelog in `utility/doc/CHANGELOG.md` as Jessica requested. Sorry for the large number of commits 😅

utility/doc/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.0.14] - 2022-12-20
```
Olshansk, line 10: We don't actually have a CHANGELOG  for infra/tooling.

Putting this under `build` seems wrong (irrelevant), but creating a CHANGELOG under `root` also seems wrong (it's just tooling).

Since I don't have a good answer for it for now. Seems that your script checks for a CHANGELOG under root, which we don't have yet. DO you want to add it?

.githooks/pre-commit
```diff
@@ -0,0 +1,146 @@
...
+if [[ "${ROOT_EDITED}" == 1 && "${ROOT_CHANGELOG_INCLUDED}" == 0 ]]
+then
+    printf 'Root changelog not included\n\n' >&2
+    CHANGELOG_ERRORS_FOUND=1
+fi
+
+# Check that for each module edited the CHANGELOG.md file
+# was included in the commit
+MODULES_MISSING_CHANGELOG=()
+CHANGELOG_FILES=() # Store included changelog locations
```
Olshansk, line 76:
```suggestion
CHANGELOG_FILES=() # Full paths to CHANGELOG.md files
```

.githooks/pre-commit
```diff
@@ -0,0 +1,146 @@
...
+fi
+
+if [[ "${#STAGED_FILES[@]}" == 0 ]]
+then
+    echo "No files modified"
+    exit 0
+fi
+
+ROOT_EDITED=0
+ROOT_CHANGELOG_INCLUDED=0
```
Olshansk, line 33: Do we even have a root changelog in the current repo?

.githooks/pre-commit
```diff
@@ -0,0 +1,146 @@
...
+MODULES_EDITED=()
```
Olshansk, line 34: Is this correct?

```suggestion
MODULES_EDITED=() # Modules are considered top-level directories directly under root
```

.githooks/pre-commit
```diff
@@ -0,0 +1,146 @@
...
+CHANGELOG_FILES=() # Store included changelog locations
+if [[ "${#MODULES_EDITED[@]}" != 0 ]]
+then
+    # Loop through edited modules checking for CHANGELOG.md
+    for module in "${MODULES_EDITED[@]}"
+    do
+        MODULE_CHANGELOG_INCLUDED=0
+        for file in "${STAGED_FILES[@]}"
+        do
+            if [[ "${file}" =~ "${module}"/* ]]
```
Olshansk, line 85: Is this correct?

```suggestion
            # If any files under /root/module/* were edited, we need to make sure that /root/module/CHANGELOG was updated
            if [[ "${file}" =~ "${module}"/* ]]
```

Olshansk (changes requested):
> To address those final commits - I realised the CHANGELOG.md file detection was wrong as the CHANGELOGs are stored in different sub directories (root of module, doc, docs) and so to address this when a CHANGELOG.md is detected in the path of a file the script will now add this path to a list and loop through that instead of trying to "build" the CHANGELOG.md path itself.

Could you update the comment in the pre-commit if you think its relevant to add more details?

> I also removed the dependency on the date command (past the basic usage that is portable across linux/osx/bsd) and am now comparing the dates from the changelog as a string against today's date.

Nice

> Finally I have updated the relevant changelog in utility/doc/CHANGELOG.md as Jessica requested.

Left a comment about this.

> Sorry for the large number of commits 😅

No worries assuming your okay with the back-and-forth! Let's get it right :)

Either way, what we do is a "squash & merge" where the git message is the PR description, so commit as many times as you want while working in progress.

.githooks/pre-commit
h5law: Yes this is correct `=~` basically means do a RegEx comparison of the two strings. So this checks only the files for the module currently being checked, and is looking for the `CHANGELOG.md` where ever it may be in the module

.githooks/pre-commit
h5law: Yes, this is the definition I am using for module. I have added the comment to clarify.

.githooks/pre-commit
h5law: There is no root `CHANGELOG.md`

utility/doc/CHANGELOG.md
h5law: This was a question I posed in the original PR, I don't see the need to create a CHANGELOG simply for this one piece of tooling. I have removed the root directory check as there is nothing there. I think it is probably better _in my opinion_ to remove this check as there isn't really a need for it.

If you want me to remove the changes I made to `utility/doc/CHANGELOG.md` I can but I don't see the point in making a new root level `CHANGELOG.md` as it probably will not see any updates besides this - there is no top-level code. All the other modules have their own as well so let me know what you think about this. I can if that's what you are looking for.

h5law: Adding the root changelog would mean that for any update to a module that requires `go.mod` or `go.sum` to change then both the module and root `CHANGELOG.md` files would need to be included - which would be an pointless repetition. We could get around this by having the script ignore all files in the root directory - as they don't require a changelog really. But this is the same as not checking it at all...

.githooks/pre-commit
Olshansk: Appreciate the explanation!

utility/doc/CHANGELOG.md
Olshansk: > I don't see the need to create a CHANGELOG simply for this one piece of tooling

Agreed

> . I think it is probably better in my opinion to remove this check as there isn't really a need for it.

Since I'm not 100% sure at the moment, but I can see us adding a root changelog soon (for main releases), can you  add a `ROOT_CHANGELOG_ENABLE=0 ` with a comment explaining that it should be enabled once a root CHANGELOG is added. We get to keep the code and document why it is off at the moment.

Olshansk (changes requested):
Left one comment here:  https://github.com/pokt-network/pocket/pull/397/files#r1053724399

Wdyt of this compromise (best of both words w/ optionality) and then we'd be good to merge?

utility/doc/CHANGELOG.md
h5law: > Scan you add a `ROOT_CHANGELOG_ENABLE=0 ` with a comment explaining that it should be enabled once a root CHANGELOG is added. We get to keep the code and document why it is off at the moment.

So instead of this I have the script check the root level _ONLY_ for a `CHANGELOG.md` file. This avoids having to do the full root level check/ignore certain files but allows for the root-level `CHANGELOG.md` - _only when included_ to be validated.

This does mean there can never be an error situation for not including it in a commit. However it means that there doesn't need to be any code changes when the CHANGELOG gets implemented.

I would be happy to implement a checking system when the CHANGELOG is made and the circumstances defining when and when not to include it in a commit are clearly defined. But for now if its included it will be validated (date+version) like the rest but if not there will be no errors.

h5law:
> Left one comment here: https://github.com/pokt-network/pocket/pull/397/files#r1053724399
> 
> Wdyt of this compromise (best of both words w/ optionality) and then we'd be good to merge?

I have implemented something similar I explained the details in the reply to the comment above - but simply instead of having a hard "toggle" which would require an update when the CHANGELOG is implemented at root level the script now just checks for the presence of a root-level changelog and if its there it will validate it. No errors if its not there.

I have also removed the changes made to the `utility/doc/CHANGELOG.md` file.

Olshansk (approved):
Thank you @h5law and appreciate your collaboration on the back-and-forth. I learnt a few things along the way so lmk if you'd be interested in picking up more tasks in the future :)

As I mentioned earlier, we usually do a "Squash & Merge" + Put the PR description as the body so I will go ahead and do so!

Issue #313, opened 2022-10-19
[Utility] Design & implement a better approach to `GenericParam` /  `ActorSpecificParam`


Olshansk:
## Objective

Design, consolidate & refactor how parameters unique to different actors are managed throughout the codebase.

## Origin Document

At the time of writing this issue, `persistence/test/setup_test.go` has the following code snippet:

<img width="770" alt="Screen Shot 2022-10-19 at 3 23 07 PM" src="https://user-images.githubusercontent.com/1892194/196815934-cb0449a3-8c4e-47e1-96c9-ecf09cac1c82.png">

In #206, we consolidated some instances in the codebase where `GenericParam` and `ActorSpecificParam` are conflated, but the current definition for an actor does not make it clear what the "generic param" is:

```proto
message Actor {
  ActorType actor_type = 1;
  string address = 2;
  string public_key = 3;
  repeated string chains = 4;
  string generic_param = 5;
  string staked_amount = 6;
  int64 paused_height = 7;
  int64 unstaking_height = 8;
  string output = 9;
}
```

For example, it could be a `ServiceURI` in the context of a `ServiceNode` or the `MaxRelays` count in the context of an `App`. Furthermore, it does not permit us to have more than one "actor specific parameter." This need to be cleared up and improved.

## Goals

- [ ] Avoid confusing new contributors to the code base of what "generic param" or "actor specific param" may be and make it explicit
- [ ] Make the code self-explanatory and clear with naming conventions
- [ ] Enable giving actors the option to have more than one "actor specific parameter"
- [ ] Updating both protobufs & structs where appropriate

## Deliverable

- [ ] A design and README explaining how this task should be tackled
- [ ] A presentation and agreement from the team on how this should be approached
- [ ] A PR updating structs, protobufs and related dependencies to the new "actor specific parameter" schemas

## Non-goals / Non-deliverables

- Changing the parameter values associated with actors to actors
- Adding new functionality
- Refactoring unrelated parts of the codebase

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)

## Testing Methodology

- **All tests**: `make develop_all`
[...]

Issue #295, opened 2022-10-06
[TECHDEBT][Shared] Address legacy Run golangci-lint errors


jessicadaugherty:
## Objective

Improve unit tests by addressing legacy linting errors 

## Origin Document

Raised by review of https://github.com/pokt-network/pocket/pull/235#issuecomment-1270639788

![Screen_Shot_2022-10-06_at_10 50 28_AM](https://user-images.githubusercontent.com/55156619/194425205-473f6575-4eeb-49de-a761-f23e5bbcecf6.png)

## Goals

- Block PRs that break unit tests
- Block PRs that break linting errors
- Display code changes that do not follow style guidelines 

## Deliverables

- [ ] A PR that tends to all of the legacy linting errors in the repo
- [ ] A PR that updates the linting github workflow to block (rather than simply display a warning) on golang and repo linting errors

## Non-goals / Non-deliverables

- Changing any business logic in the codebase

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)

---

**Creator**: @jessicadaugherty
**Co-Owners**: @Olshansk

Olshansk:
@jessicadaugherty Can we include this in iteration 4? It's a lower priority, but it blocks us from enabling the autolinters

h5law:
@jessicadaugherty I'll do this (I don't know when iteration 4 is but can have it done this week). I find these errors annoying anyway so it'll be nice to get rid of em!

jessicadaugherty:
Amazing, thanks @h5law! I assigned this out to you on DeWork. LOL iteration 4 was in like October or something so this was originally a priority backlog item that kept getting refactored. Agreed it would improve contributor QoL! Appreciate it.

> Just to confirm - fix the errors from the golanglint-ci test throughout the codebase and make the workflow blocking on failure?

Exactly! I'd take a look at the linked PR in "Origin Document" for more context on where we stumbled on this need. And, of course, lmk if you have any questions!

okdas:
Hey @h5law, thank you for looking into this!

Here is a related thread: https://github.com/golangci/golangci-lint-action/issues/23

[...]

PR #403, opened 2022-12-15
[Core] Consolidate all node configurations into a new configs package - (Issue #396)

### LINKED ISSUE #334: [TECHDEBT] implement ValidateConfig and ValidateGenesis in all modules
## Objective

This issue tends to track the implementation of the validations that are currently marked as **TODOs** without a reference to any actual task.

Specifically, all modules that implement `ConfigurableModule` and/or `GenesisDependentModule` have a `ValidateConfig(Config) error` and/or a `ValidateGenesis(GenesisState) error` method.

An example of a valid implementation that could apply to all modules is:

```go
func (*persistenceModule) ValidateConfig(cfg modules.Config) error {
	persistenceConfig:= cfg.GetPersistenceConfig()
	if persistenceConfig == (*types.PersistenceConfig)(nil) {
		return fmt.Errorf("persistenceConfig is nil")
	}
        // additional validations
	return nil
}
```

Here, we are simply verifying that the configuration for the module is set, but we could also validate the individual values such as `PostgresUrl` (for which we expect a PostgreSQL connection string) by replacing `// additional validations` in the example above with something similar to:

```go
if !strings.HasPrefix(persistenceConfig.GetPostgresUrl(),"postgres://") {
    return fmt.Errorf("persistenceConfig.postgres_url must be a valid PostgreSQL connection string")
}
```

The other obvious ones are port numbers (which naturally have lower and upper bounds), paths that must be valid, etc.

Clearly, it's also possible that a particular configuration currently doesn't have any validation rule. It might or might not be defined in the future.

Designing and defining how we address (or not) existing values and future-proofing configs we will add in the future is part of the task.

## Origin Document

https://github.com/pokt-network/pocket/pull/176/files#r1012343547

## Goals

- Defining a pattern that all modules will follow for `Config` and `Genesis` validation
- Future-proofing validation for config values we have not created yet
- Implement validations wherever needed for the modules that implement `ConfigurableModule` and/or `GenesisDependentModule`

## Deliverable

- [ ] Validation functions for `Config` and `Genesis` in every respective module
- [ ] Removal of the TODO items
- [ ] Documenting the pattern for how this should be implemented

- Change the existing logic in the respective modules

## General issue deliverables

## [Required] Testing Methodology

- **New tests**: Add new unit tests that can be executed via `make test_configs` to validate all non-genesis configs
- **New tests**: Add new unit tests that can be executed via `make test_genesis` to validate all genesis configurations

---

**Creator**: @deblasis 
**Co-Owners**: @Olshansk

### LINKED ISSUE #396: [Core] Consolidate all node configurations into a new `configs` package
## Objective

Simplify the codebase and consolidate configurations.

## Origin Document

A few different mechanisms for configuration management were discussed in #234 and a compromise was reached. However, there are slight challenges to how new developers approach the codebase (e.g. understanding where the configs are), verification and validation.

## Goals

- Design & Identify patterns of how we should manage configs in the codebase
- Create a single package encapsulating all the logic related to configs
- Remove the onus of independent modules to design their own configs
- Enforce any module substitutes to conform to the configs in the shared package
- Improve readability & understandability of all the configs across the codebase

## Deliverable

- [ ] A PR that consolidates all the configs into a single package under the root director
- [ ] Refactoring & organize all existing configs into the package created above
- [ ] An updated README explaining how configs are managed & maintained in the codebase

- Introducing new validation logic for existing configs
- Updating any existing configs

## General issue deliverables

## Testing Methodology

---

**Creator**: @Olshansk 
**Co-Owners**: @deblasis

### LINKED ISSUE #414: Define a better approach to generating deterministic keys for testing purposes

deblasis:
## Description

The main goal of this PR is to centralize configurations into a single `configs` package.
The feedback received was that the previous implementation was probably overly abstracted. Mockability is still preserved thanks to the fact that we are ultimately just dealing with structs without any logic attached to them.

It also tends to address:

```
// IMPROVE(#361): Design a better way to generate deterministic keys for testing.
```
https://github.com/pokt-network/pocket/compare/issue/396-consolidate-config?expand=1#diff-ee06bf613911836fdf48424397b7ec556c2e1a8ddba5adecca4c1782aea24087L30

## Issue

- Fixes #396
- Fixes #334
- Fixes #414

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [x] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Centralized configs into the `configs` package
- Introduced a keygenerator that can work deterministically or randomly at will
- Centralize genesis as well, from two structs with overlapping states into a single one
- Simplified shared `core` types (Actor, Account, etc) under `coreTypes`
- Added missing `max_mempool_count` in config (it was causing P2P instabilities in LocalNet)

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


deblasis:
@Olshansk I would appreciate a quick scan on this, just to make sure it's getting where it's supposed to be. 

Please point out blatant gaps (apart from the fact that I need to centralize genesis as well but I am waiting for this initial feedback on the design before proceeding)

🙏

Olshansk (commented):
tl;dr I ❤️  where this is going

persistence/genesis.go
```diff
@@ -44,97 +45,72 @@ func (m *persistenceModule) populateGenesisState(state modules.PersistenceGenesi
...
-			log.Fatalf("an error occurred inserting a fisherman in the genesis state: %s", err.Error())
-		}
-		if err = addValueToPool(types.PoolNames_FishermanStakePool.String(), act.GetStakedAmount()); err != nil {
-			log.Fatalf("an error occurred inserting staked tokens into %s pool: %s", types.PoolNames_FishermanStakePool.String(), err.Error())
-		}
+
+	stakedActorsInsertConfigs := []struct {
+		Name     string
+		Getter   func() []*coreTypes.Actor
+		InsertFn func(address []byte, publicKey []byte, output []byte, paused bool, status int32, serviceURL string, stakedTokens string, chains []string, pausedHeight int64, unstakingHeight int64) error
```
Olshansk, line 59:
```suggestion
		InsertFn func(address, publicKey, output []byte, paused bool, status int32, serviceURL, stakedTokens string, chains []string, pausedHeight, unstakingHeight int64) error
```

persistence/genesis.go
```diff
@@ -44,97 +45,72 @@ func (m *persistenceModule) populateGenesisState(state modules.PersistenceGenesi
...
-		}
-		err = rwContext.InsertFisherman(addrBz, pubKeyBz, outputBz, false, StakedStatus, act.GetGenericParam(), act.GetStakedAmount(), act.GetChains(), act.GetPausedHeight(), act.GetUnstakingHeight())
-		if err != nil {
-			log.Fatalf("an error occurred inserting a fisherman in the genesis state: %s", err.Error())
-		}
-		if err = addValueToPool(types.PoolNames_FishermanStakePool.String(), act.GetStakedAmount()); err != nil {
-			log.Fatalf("an error occurred inserting staked tokens into %s pool: %s", types.PoolNames_FishermanStakePool.String(), err.Error())
-		}
+
+	stakedActorsInsertConfigs := []struct {
```
Olshansk, line 56: 💯

persistence/genesis.go
```diff
@@ -44,97 +45,72 @@ func (m *persistenceModule) populateGenesisState(state modules.PersistenceGenesi
...
+		Pool     coreTypes.PoolNames
+	}{
+		{
+			Name:     "app",
+			Getter:   state.GetApplications,
+			InsertFn: rwContext.InsertApp,
+			Pool:     coreTypes.PoolNames_POOL_NAMES_APP_STAKE_POOL,
+		},
+		{
+			Name:     "node",
```
Olshansk, line 69:
```suggestion
			Name:     "serviceNode",
```

TOL(Thinking Out Loud): I'm beginning to think we should rename these actors to `Servicers`

runtime/test_artifacts/keygenerator/keygen.go
```diff
@@ -0,0 +1,66 @@
...
+	"bytes"
+	"encoding/binary"
+	"math/rand"
+	"sync"
+
+	"github.com/pokt-network/pocket/shared/crypto"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+var keygenInitOnce sync.Once
```
Olshansk, line 13: Optional NIT: we can use factoring here.

Aside: Big fan of this singleton pattern.

runtime/genesis.go
```diff
@@ -4,43 +4,17 @@ import (
...
-
-func (g *runtimeGenesis) GetPersistenceGenesisState() modules.PersistenceGenesisState {
-	return g.PersistenceGenesisState
-}
-func (g *runtimeGenesis) GetConsensusGenesisState() modules.ConsensusGenesisState {
-	return g.ConsensusGenesisState
-}
-
-func parseGenesisJSON(genesisPath string) (g *runtimeGenesis, err error) {
+func parseGenesisJSON(genesisPath string) (g *genesis.GenesisState, err error) {
```
Olshansk, line 10: NIT: Can you rename this to `loadGenesis` and add a comment that `genesisPath` expects a JSON. 

`parseGenesisJSON(path)` feels incorrect.

persistence/test/setup_test.go
```diff
@@ -80,17 +83,19 @@ func NewTestPostgresContext(t testing.TB, height int64) *persistence.PostgresCon
...
-	defer os.Unsetenv(test_artifacts.PrivateKeySeedEnv)
-
-	cfg := runtime.NewConfig(&runtime.BaseConfig{}, runtime.WithPersistenceConfig(&types.PersistenceConfig{
-		PostgresUrl:    databaseUrl,
-		NodeSchema:     testSchema,
-		BlockStorePath: "",
-		TxIndexerPath:  "",
-		TreesStoreDir:  "",
-	}))
+	teardownDeterministicKeygen := keygenerator.GetInstance().SetSeed(42)
```
Olshansk, line 86: This interface looks very clean IMO, so do we even need #414? Did you have something else in mind?

shared/core/types/proto/core_actor.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package core;
+
+option go_package = "github.com/pokt-network/pocket/shared/core/types";
+
+enum ActorType {
+  ACTOR_TYPE_UNSPECIFIED = 0;
+  ACTOR_TYPE_APP = 1;
+  ACTOR_TYPE_NODE = 2;
```
Olshansk, line 10:
```suggestion
  ACTOR_TYPE_NODE = 2; // TODO: Consider rename `node` or `serviceNode` to `servicers` throughout the codebase.
```

runtime/test_artifacts/keygenerator/keygen.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	privateKey = privateKeyBz.String()
+	publicKey = privateKeyBz.PublicKey().String()
+	address = privateKeyBz.PublicKey().Address().String()
+
+	return
+}
+
+func getPrivateKeySeed() (privateKeySeed int) {
```
Olshansk, line 62: Why did you expose this as a separate function rather than embedding the logic inside of `reset`? Could simplify the logic a bit.

```go
func GetInstance() *keyGenerator {
	if keygen == nil {
		keygenInitOnce.Do(func() {
			keygen = &keyGenerator{}
			keygen.reset()
		})
	}
	return keygen
}

// ...

func (k *keyGenerator) reset() {
	rand.Seed(timestamppb.Now().Seconds)
	k.privateKeySeed = rand.Int()
}
```

persistence/genesis.go
```diff
@@ -44,97 +45,72 @@ func (m *persistenceModule) populateGenesisState(state modules.PersistenceGenesi
...
+			}
+			outputBz, err := hex.DecodeString(act.GetOutput())
+			if err != nil {
+				log.Fatalf("an error occurred converting output to bytes %s", act.GetOutput())
+			}
+			err = saic.InsertFn(addrBz, pubKeyBz, outputBz, false, StakedStatus, act.GetGenericParam(), act.GetStakedAmount(), act.GetChains(), act.GetPausedHeight(), act.GetUnstakingHeight())
+			if err != nil {
+				log.Fatalf("an error occurred inserting an %s in the genesis state: %s", saic.Name, err.Error())
+			}
+			if err = addValueToPool(coreTypes.PoolNames_POOL_NAMES_APP_STAKE_POOL.String(), act.GetStakedAmount()); err != nil {
```
Olshansk, line 108:
```suggestion
			if err = addValueToPool(saic.Pool.String(), act.GetStakedAmount()); err != nil {
```

persistence/genesis.go
```diff
@@ -44,97 +45,72 @@ func (m *persistenceModule) populateGenesisState(state modules.PersistenceGenesi
...
+				log.Fatalf("an error occurred inserting staked tokens into %s pool: %s", coreTypes.PoolNames_POOL_NAMES_APP_STAKE_POOL, err.Error())
```
Olshansk, line 109:
```suggestion
				log.Fatalf("an error occurred inserting staked tokens into %s pool: %s", coreTypes.Pool, err.Error())
```

runtime/manager_test.go
```diff
@@ -0,0 +1,365 @@
...
+)
+
+func TestNewManagerFromReaders(t *testing.T) {
+	type args struct {
+		configReader  io.Reader
+		genesisReader io.Reader
+		options       []func(*Manager)
+	}
+
+	buildConfigBytes, err := os.ReadFile("../build/config/config1.json")
```
Olshansk, line 26: NIT: Please move this path (and the one below) into constants at the top of the file with `Default` prefix

runtime/test_artifacts/generator.go
```diff
@@ -2,149 +2,112 @@ package test_artifacts
...
-	)
+		},
+	}
 }
 
-func NewPools() (pools []modules.Account) { // TODO (Team) in the real testing suite, we need to populate the pool amounts dependent on the actors
-	for _, name := range typesPers.PoolNames_name {
-		if name == typesPers.PoolNames_FeeCollector.String() {
-			pools = append(pools, &typesPers.Account{
+func NewPools() (pools []*coreTypes.Account) { // TODO (Team) in the real testing suite, we need to populate the pool amounts dependent on the actors
```
Olshansk, line 86:
```suggestion
 // REFACTOR: Test artifact generator should reflect the sum of the initial account values to populate the initial pool values
func NewPools() (pools []*coreTypes.Account) {
```

Olshansk (changes requested):
This PR is a 🕎 miracle!

Left a few minor comments throughout, but should be good to go right after. This is really going to help us start streamlining our work but also make the codebase much more understandable for anyone joining the codebase. Ty @deblasis!

shared/core/types/proto/core_account.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
deblasis: Totally! You made me also realise that the refactoring was incomplete. 🙏

persistence/genesis.go
deblasis: I would say yes but if we want to enforce a style I would say let's be consistent and scout the whole codebase for similar instances and rectify accordingly otherwise it becomes a mix and match.

Another thing we could do is adding newlines when the signatures are particularly long. I think you did that somewhere in the codebase and I like it since it allows the reader to avoid horizontal scrolling.

Let me know how you want to go about your suggestion and, more generally, about styling guidelines, perhaps this could be related to linting? (#295)

persistence/genesis.go
deblasis: I agree, serviceNode is an improvement here since node could mean anything. If we want to rename them to Servicers I guess the only thing I would do is make sure that we are consistent in the codebase, documentation and perhaps even in Notion. Thoughts?

persistence/test/setup_test.go
deblasis: I have created #414 only to track the individual todo-item in #361 (I have used the feature "convert to issue") so that when this PR is merged, the todo item is automatically marked as completed and there are also breadcrumbs to here basically.

runtime/test_artifacts/keygenerator/keygen.go
deblasis: Fair point

persistence/genesis.go
deblasis: I am not sure I understand this change

persistence/module.go
```diff
@@ -8,25 +8,24 @@ import (
...
 )
 
 // TODO: convert address and public key to string not bytes in all account and actor functions
 // TODO: remove address parameter from all pool operations
 type persistenceModule struct {
 	bus          modules.Bus
-	config       modules.PersistenceConfig
-	genesisState modules.PersistenceGenesisState
+	config       *configs.PersistenceConfig
+	genesisState *genesis.GenesisState
```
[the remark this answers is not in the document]
deblasis: Fair point. However, I guess we need to think about pros and cons. 
At a glance a con could be the fact that in tests we would have to mock the bus every time. (of course it's a one-off but still). 
So basically we are enforcing the bus as dependency. Makes sense if we think about how the whole thing works but what are the benefits in not having pointers to the config and genesis inside the modules?

deblasis:
@Olshansk:
> This PR is a 🕎 miracle!
> 
> Left a few minor comments throughout, but should be good to go right after. This is really going to help us start streamlining our work but also make the codebase much more understandable for anyone joining the codebase. Ty @deblasis!

LOL, too kind! I have addressed the majority of the comments, left some open questions (that might lead to creating separate issues) and I have to update the CHANGELOGs but I guess it's ready for another pass :)

utility/types/proto/stake_status.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package utility;
+
+option go_package = "github.com/pokt-network/pocket/utility/types";
+
+// DISCUSS: *Design Decision* deprecating StakeStatus
+enum StakeStatus {
```
Olshansk, line 7:
```suggestion
REFACTOR(#258): Rename the protobuf types to follow best practices
enum StakeStatus {
```

Olshansk (changes requested):
It's a pretty big PR so hard to catch small errors. Can you run the [iteration 3 demo](https://github.com/pokt-network/pocket/blob/main/docs/demos/iteration_3_end_to_end_tx.md) to see if everything checks out and still works?

Also, at some point (not in this PR) we should update the main READMEs around our code structure, but I think we have some more cleanup to do first.

deblasis:
@Olshansk 
> It's a pretty big PR so hard to catch small errors. Can you run the [iteration 3 demo](https://github.com/pokt-network/pocket/blob/main/docs/demos/iteration_3_end_to_end_tx.md) to see if everything checks out and still works?

100%

> Also, at some point (not in this PR) we should update the main READMEs around our code structure, but I think we have some more cleanup to do first.

Agreed

deblasis:
> @Olshansk
> 
> > It's a pretty big PR so hard to catch small errors. Can you run the [iteration 3 demo](https://github.com/pokt-network/pocket/blob/main/docs/demos/iteration_3_end_to_end_tx.md) to see if everything checks out and still works?
> 
> 100%

I ran the iteration demo and also updated it 

https://github.com/pokt-network/pocket/pull/403/files#diff-79ce7fc016a2250751d0d0f7a5621957e9feb4f4971e497ff49ad50d3f547899R100-R144

It looks like it working. The only weird thing / **stupid question** that I have noticed is that if I send 1000 POKT I see that the DAO takes 9000 POKT commission?

```bash
go run app/client/*.go --path_to_private_key_file=./val1.json Account Send 6f66574e1f50f0ef72dff748c3f11b9e0e89d32a 67eb3f0a50ae459fecf666be0e93176e92441317 1000
```
image

image

If the answer is simply: inflation we are all good, just wanted to make sure I get it. 🙂


> 
> > Also, at some point (not in this PR) we should update the main READMEs around our code structure, but I think we have some more cleanup to do first.
> 
> Agreed

I have updated the CHANGELOG. If you don't flag other major changes I'll update the readmes

runtime/configs/config.go
```diff
@@ -12,3 +14,64 @@ type Config struct {
...
+	}
+
+	for _, option := range options {
+		option(cfg)
+	}
+
+	return cfg
+}
+
+func WithPK(pk string) func(*Config) {
```
Olshansk, line 65:
```suggestion
// There is an option to configure module-specific keys in order to enable different "identities"
// for different purposes (i.e. validation, P2P, servicing, etc...).
func WithPK(pk string) func(*Config) {
```

docs/demos/iteration_3_end_to_end_tx.md
```diff
@@ -96,12 +97,32 @@ Show all the commands available in the CLI:
 go run app/client/*.go
 ```
 
+#### Accounts setup
+
+Since our Keybase is under development, currently we have to manually inject the private keys of the accounts we want to use in the CLI.
```
Olshansk, line 102:
```suggestion
Since our Keybase is under development, we have to manually inject the private keys of the accounts we want to use in the CLI.
```

docs/demos/iteration_3_end_to_end_tx.md
```diff
@@ -96,12 +97,32 @@ Show all the commands available in the CLI:
...
+
+For the following steps you'll need to use the accounts of the first two validators, therefore you have some options:
```
Olshansk, line 104:
```suggestion
For the following steps, you'll need to use the accounts of the first two validators in the hard-coded development genesis file. Therefore you have some options:
```

Olshansk (changes requested):

runtime/manager_test.go
```diff
@@ -91,275 +244,46 @@ func TestNewManagerFromReaders(t *testing.T) {
...
+						UseRainTree:           true,
+						IsEmptyConnectionType: false,
+						MaxMempoolCount:       defaults.DefaultP2PMaxMempoolCount,
 					},
 				},
+				expectedGenesis,
 				clock.New(),
 			},
+			assertion: func(tt require.TestingT, want, got interface{}, _ ...interface{}) {
+				require.Equal(tt, want.(*Manager).config.P2P, got.(*Manager).config.P2P)
```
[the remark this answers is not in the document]
deblasis: I selected a configuration value I was interested in and used it as a "probe" to prove the concept and that the behaviour is as expected.

Do you think is necessary to add the same test for all the other modules as well?
Perhaps I should name the test case more generically like "an unset configuration value should fallback to its default"

shared/core/types/pool_names.go
```diff
@@ -0,0 +1,18 @@
+package types
```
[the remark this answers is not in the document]
deblasis: Done, for consistency, I renamed the enum as well

persistence/docs/CHANGELOG.md
```diff
@@ -5,19 +5,91 @@ All notable changes to this module will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-TODO: consolidate `persistence/docs/CHANGELOG` and `persistence/CHANGELOG.md`
-
 ## [Unreleased]
 
-## [0.0.0.10] - 2022-12-21
+## [0.0.0.17] - 2022-12-28
+
+- Added missing `ActorType` in `GetAllXXXX()` functions
```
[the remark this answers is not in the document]
deblasis: It was never there to begin with, it was a bug that I am attempting to fix.

`getActor` returns the actor with no notion of its type. 

https://github.com/pokt-network/pocket/blob/main/persistence/shared_sql.go#L85-L110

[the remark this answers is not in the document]
deblasis: I think I can genericize it more not that we have `coreTypes` 😉, I can leverage the `actorSchema`.

Apart from that, I am confident that I am not breaking anything, the type was unspecified before 🙂

PTAL: https://github.com/pokt-network/pocket/pull/403/commits/d7a521eb816c33d810f1e29f848028727512a149

p2p/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.0.14] - 2022-12-28
+
+- `ActorsToAddrBook` now skips actors that are not validators since they don't have a serviceUrl generic parameter
```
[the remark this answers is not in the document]
deblasis: Currently, we use the serviceUrl to connect to peers and to send messages to them, I believe that the `serviceUrl` is likely gonna be replaced by something like a LibP2P Address

https://docs.libp2p.io/concepts/fundamentals/addressing/

Eg: 
`/ip4/7.7.7.7/tcp/4242/p2p/QmRelay/p2p-circuit/p2p/QmRelayedPeer`

but that's going to be dynamic and transient and I don't see how it would be part of the state.

That said, I am not sure we should change the genesis state at all to handle addressing. 
It's the addressing that has to change instead.

Also, what is this concept of generic parameter? Where can I learn more about it?

shared/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.0.8] - 2022-12-28
+
+- Added `PoolNames.FriendlyName` method
```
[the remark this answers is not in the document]
deblasis: Is it? Nice then! It sounded "friendly" enough and I honestly didn't know how else to name this.

shared/core/types/pools.go
```diff
@@ -0,0 +1,18 @@
+package types
+
+var poolFriendlyNames map[Pools]string
+
+func init() {
+	poolFriendlyNames = map[Pools]string{
+		Pools_POOLS_UNSPECIFIED:        "Unspecified",
+		Pools_POOLS_DAO:                "DAO",
+		Pools_POOLS_FEE_COLLECTOR:      "FeeCollector",
```
Olshansk, line 9:
```suggestion
		Pools_POOLS_FEE_COLLECTOR:      "FeeCollectorPool",
```

shared/core/types/pools.go
```diff
@@ -0,0 +1,18 @@
+package types
+
+var poolFriendlyNames map[Pools]string
+
+func init() {
+	poolFriendlyNames = map[Pools]string{
+		Pools_POOLS_UNSPECIFIED:        "Unspecified",
+		Pools_POOLS_DAO:                "DAO",
```
Olshansk, line 8:
```suggestion
		Pools_POOLS_DAO:                "DAOPool",
```

persistence/genesis.go
Olshansk:
```
I would say yes but if we want to enforce a style I would say let's be consistent and scout the whole codebase for similar instances and rectify accordingly otherwise it becomes a mix and match.
```

Agreed. I was trying to be consistent with line 83 of this file (just local consistency).

I believe we can enforce not specifying types (when they're the same) using the automatic linters @okdas set-up since these are standard golang suggestions. They're currently off because we need to do a one-time cleanup.

You are correct that #295 encompasses this.

```
Another thing we could do is adding newlines when the signatures are particularly long. I think you did that somewhere in the codebase and I like it since it allows the reader to avoid horizontal scrolling.
```

I do like it but think it's overkill here because the function is anonymous.

```
Let me know how you want to go about your suggestion and, more generally, about styling guidelines, perhaps this could be related to linting? (https://github.com/pokt-network/pocket/issues/295)
```

Let's just be consistent between this and line 83 and #295 will encompass the work of enforcing repo-wide guidelines.

persistence/shared_sql.go
```diff
@@ -92,7 +92,7 @@ func (p *PostgresContext) getActor(actorSchema types.ProtocolActorSchema, addres
 	if err != nil {
 		return
 	}
-
+	actor.ActorType = actorSchema.GetActorType()
```
Olshansk, line 95: `getActor` uses `getActorFromRow` under the hood

BUT

`GetAllXXX` uses `getActorFromRow` directly

So this won't actually fully work.

I'm okay with leaving it as is because we have a ton of scope creep, but could you add a TODO to fix it (in a follow up PR). I also think this is a really good opportunity to do some TDD to be future-proof.

persistence/docs/CHANGELOG.md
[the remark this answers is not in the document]
Olshansk: Left a comment in `persistence/shared_sql.go`

p2p/CHANGELOG.md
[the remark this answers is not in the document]
Olshansk: Regarding libp2p - makes sense when we get there.

Regarding the val filtering - see the comment I left in `p2p/addrbook_provider/persistence.go`

-----

`Also, what is this concept of generic parameter? Where can I learn more about it?`

This is just a poor design decision that was the result of a compromise during the shared module refactor.

We (currently) have 4 protocol actor types:
- Fisherman
- Validator
- Servicer
- Application

There are things they all share (e.g. an address) and there are things that are specific:
- Servicers need a "serviceURL" to service nodes
- Fisherman need a "serviceURL" to enable QoS
- Applications need a 'maxRelaysCount' which is only specific to them

The two potential future paths we take are:
1. One Protobuf per actor
2. One `BaseActor` protobuf that's embedded in each ProtocolActor type

See #313 for the origin document.

persistence/module.go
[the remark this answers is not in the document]
Olshansk: Ditto - fair point.

Given what you said, I don't love either approach but prefer what we have for now over the alternative.

Olshansk (changes requested):

p2p/addrbook_provider/persistence.go
```diff
@@ -69,9 +71,13 @@ func (pabp *persistenceAddrBookProvider) GetStakedAddrBookAtHeight(height uint64
 	return addrBook, nil
 }
 
-func (pabp *persistenceAddrBookProvider) ActorsToAddrBook(actors map[string]modules.Actor) (typesP2P.AddrBook, error) {
+func (pabp *persistenceAddrBookProvider) ActorsToAddrBook(actors map[string]coreTypes.Actor) (typesP2P.AddrBook, error) {
 	book := make(typesP2P.AddrBook, 0)
 	for _, v := range actors {
+		// only add validator actors since they are the only ones having a service url in their generic param at the moment
+		if v.ActorType != coreTypes.ActorType_ACTOR_TYPE_VAL {
```
Olshansk, line 78: Related to one of my other comments, fisherman and serviceNodes also have serviceURLs in their parameter, and can (and should) also participate in structured broadcast.

Should they not be included as well?

Screenshot 2023-01-03 at 11 49 05 AM

Olshansk (changes requested):

shared/core/types/pools.go
deblasis: Are you sure you want to change this? The current values reflect what's currently on `main`, also, I could argue that they are entities in the table `pool` so the suffix is redundant.

shared/core/types/pools.go
deblasis: https://github.com/pokt-network/pocket/pull/403#discussion_r1060922509

persistence/shared_sql.go
deblasis: I coudn't resist... https://github.com/pokt-network/pocket/pull/403/commits/fb42810947fe706561077ad191ee6a81246dab43 and https://github.com/pokt-network/pocket/pull/403/commits/6ff3c3ad2b446443df150931103c1b2b60df562b

p2p/addrbook_provider/persistence.go
deblasis: TL;DR: The way I see it, **NO**... 

Like you [said](https://github.com/pokt-network/pocket/pull/403#discussion_r1060764619), it looks like these `serviceUrl`s have a different meaning depending on the actor type and therefore using them to connect to a node for consensus related messaging/structured broadcast is a... shortcut.

Nodes are listening on the `ConsensuPort` which by default is 8080 and it seems that `serviceUrl` has been used to have an easy way to address them.

I guess this is because we don't have -yet- a full-fledged P2P module that owns the addressbook and can be relied upon for routing and messaging (that will come with peer discovery and churn).

We won't address messages via `serviceUrl` but rather by `address` or `nodeId`, then P2P would figure out how to reach that peer.

In `LocalNet` we spin up 4 validator nodes and the debug commands are all related to consensus.
This explains why previously we had 👇👈 while now we have 👉👇
image

If we consider other actors I can only imagine that since we have overlaps (eg: `node1.consensus:8080` in `genesis.json` is a Validator, a ServiceNode and also a Fisherman) we would have duplicate messages at a minimum.

I wish Brian was here to pick his brain on this :) the ways I see it, structured gossip happens under the hood in the P2P module, when you say "broadcast to all validators" you don't know/care if the message is also handled and retransmitted by a Fisherman in order to reach a "nearby" Validator. That's implementation detail of the P2P algorithm.

I would say that for now we have to filter. Happy to discuss further even offline. 

Just let me know how to proceed here.

p2p/CHANGELOG.md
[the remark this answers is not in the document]
deblasis: Thanks for the wholesome explanation and context🙏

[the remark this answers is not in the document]
Olshansk: 🙏

val1.json
```diff
@@ -0,0 +1 @@
+"6fd0bc54cc2dd205eaf226eebdb0451629b321f11d279013ce6fdd5a33059256b2eda2232ffb2750bf761141f70f75a03a025f65b2b2b417c7f8b3c9ca91e8e4"
```
Olshansk, line 1: Intentional?

persistence/test/actor_test.go
```diff
@@ -0,0 +1,38 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/pokt-network/pocket/shared/core/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetAllStakedActors(t *testing.T) {
```
Olshansk, line 10: 🙌

p2p/addrbook_provider/persistence.go
Olshansk: Solid explanation. I'm convinced.

Excited to see what our next steps are going to be here :)

Olshansk (changes requested):
One tiny comment about the addition of `val1.json` and `val2.json` to the root dir.

1. Add it to `.gitignore`
2. Update `docs/demos/iteration_3_end_to_end_tx.md` to move it into `/tmp` and reference that directly.

val1.json
deblasis: nope. Genuine mistake

Olshansk (approved):
LFG!

🚢 🚢🚢

p2p/addrbook_provider/persistence.go
deblasis: Added https://github.com/pokt-network/pocket/issues/426 PTAL 🙏

PR #407, opened 2022-12-18
[TECHDEBT][PERSISTENCE] Add: connection pooling to persistence layer db connection

Prajjawalk:
## Description

The current database connection is established via pgx.Connect() method which is not concurrency safe, so updated the database connection to connect via pgxpool module which establishes connection pooling. Updated configs to apply more connection configurations: max-connections, min-connections, max-connection lifetime, max-connection idle-time and health check intervals.  

## Issue

Fixes #NA. It was a TPDP w/o an issue

## Type of change

- [ ] Bug fix
- [X] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Upgrade `github.com/jackc/pgx/v4` -> `github.com/jackc/pgx/v5`
- Modified `persistence/db.go` -> `connectToDatabase()` function to use pgxpool for connecting to db
- Modified `persistence_config.proto`, `config1.json`, `config2.json`, `config3.json` and `config4.json` to include connection pool configurations

## Testing

- [X] `make develop_test`
- [X] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md) w/ all of the steps outlined in the `README`

## Required Checklist

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have tested my changes using the available tooling
- [X] I have updated the corresponding CHANGELOG

### If Applicable Checklist


persistence/db.go
```diff
@@ -66,14 +69,26 @@ func (pg *PostgresContext) ResetContext() error {
 }
 
 // TECHDEBT: Implement proper connection pooling
```
Olshansk, line 71: Is it safe to remove this TECHDEBT line item?

shared/modules/types.go
```diff
@@ -48,6 +48,11 @@ type PersistenceConfig interface {
 	GetBlockStorePath() string
 	GetTxIndexerPath() string
 	GetTreesStoreDir() string
+	GetMaxConns() int32
```
Olshansk, line 51:
```suggestion
	GetMaxConnsCount() uint32
```

shared/modules/types.go
```diff
@@ -48,6 +48,11 @@ type PersistenceConfig interface {
...
+	GetMinConns() int32
```
Olshansk, line 52: Why do we need a min?

shared/modules/types.go
```diff
@@ -48,6 +48,11 @@ type PersistenceConfig interface {
...
+	GetMaxConnLifetime() uint64
```
Olshansk, line 53: For these three, can you add a postfix explaining the time parameter? The code speaks for itself then

E.g.

`s/GetMaxConnLifetime/GetMaxConnSec`
`s/GetMaxIdleTime/GetMaxIdleSec`

persistence/module.go
```diff
@@ -183,7 +183,7 @@ func (m *persistenceModule) NewRWContext(height int64) (modules.PersistenceRWCon
 }
 
 func (m *persistenceModule) NewReadContext(height int64) (modules.PersistenceReadContext, error) {
-	conn, err := connectToDatabase(m.config.GetPostgresUrl(), m.config.GetNodeSchema())
+	conn, err := connectToDatabase(m.config, m.config.GetNodeSchema())
```
Olshansk, line 186: If we're going to be passing in the whole config, then there's no need for a second parameter.

Olshansk (changes requested):
Thank you for your contribution! I'm really curious how you found this specific task to work on?

Left a few comments throughout, primarily around the naming so the measurement units are self-explanatory.

persistence/db.go
Prajjawalk: Hi @Olshansk, Thanks for reviewing. I was just going through the code and found this Techdebt item which was not resolved so worked on that. I will go through the review and will make changes accordingly, after that we can remove this line.

shared/modules/types.go
Prajjawalk: Min is by default 0 but added it in case we may need it in the future.

shared/modules/types.go
Prajjawalk: Yep sure, will do

shared/modules/types.go
```diff
@@ -48,6 +48,11 @@ type PersistenceConfig interface {
 	GetBlockStorePath() string
 	GetTxIndexerPath() string
 	GetTreesStoreDir() string
+	GetMaxConnsCount() int32
+	GetMinConnsCount() int32
```
Olshansk, line 52: Thoughts on making this unsigned?

EDIT: My guess is that you made them signed to correspond with https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool?

shared/modules/types.go
```diff
@@ -48,6 +48,11 @@ type PersistenceConfig interface {
...
+	GetMaxConnLifetimeHour() uint64
+	GetMaxConnIdleTimeMinute() uint64
+	GetHealthCheckPeriodMinute() uint64
```
Olshansk, line 55: I looked into: https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool

And found that it uses found time.Duration: https://pkg.go.dev/time#Duration

It seems that we can easily parse things like `3s` or `1h` in golang: https://www.geeksforgeeks.org/time-parseduration-function-in-golang-with-examples/

Thoughts on removing the `Hour`, `Minute`, etc.., but making these strings so it could be configured to seconds, minutes, hours, etc in the config level.

This will also be similar to how tendermint designed things: github.com/pokt-network/pocket

Screenshot 2022-12-21 at 12 24 35 PM

Olshansk (changes requested):
Left another couple comments. Let me know what you think.

Make sure to also use `LocalNet` to test and update the `CHANGELOG`.

shared/modules/types.go
Prajjawalk: Yep, I had to make the connection count type int32 to correspond to pgxpool connection count type

shared/modules/types.go
Prajjawalk: Makes sense, will make changes accordingly

persistence/db.go
Olshansk: 💯

persistence/db.go
```diff
@@ -65,31 +68,51 @@ func (pg *PostgresContext) ResetContext() error {
...
-func connectToDatabase(postgresUrl string, schema string) (*pgx.Conn, error) {
+func connectToDatabase(cfg modules.PersistenceConfig) (*pgx.Conn, error) {
 	ctx := context.TODO()
 
-	conn, err := pgx.Connect(context.Background(), postgresUrl)
+	config, err := pgxpool.ParseConfig(cfg.GetPostgresUrl())
+	if err != nil {
+		return nil, fmt.Errorf("unable to create database config: %v", err)
+	}
+	MaxConnLifetime, err := time.ParseDuration(cfg.GetMaxConnLifetime())
```
Olshansk, line 78: Local var should start with a lowercase letter. Ditto below

persistence/proto/persistence_config.proto
```diff
@@ -9,4 +9,9 @@ message PersistenceConfig {
   string block_store_path = 3;
   string tx_indexer_path = 4;
   string trees_store_dir = 5;
+  int32 max_conns_count = 6;
```
Olshansk, line 12: Consider adding a comment above the proto saying "See https://pkg.go.dev/time#ParseDuration for expected string time parameters"

persistence/db.go
```diff
@@ -65,31 +68,51 @@ func (pg *PostgresContext) ResetContext() error {
...
+	if err == nil {
+		config.MaxConnLifetime = MaxConnLifetime
+	}
+	MaxConnIdleTime, err := time.ParseDuration(cfg.GetMaxConnIdleTime())
```
Olshansk, line 82: Let's return an error if parsing fails/

I understand the logic of "a quite success" of using the defaults, but it could lead to a bug where someone incorrectly configures a time parameter and spends hours debugging why it wasn't applied.

Olshansk (changes requested):
Left a couple of comments.

Also, can you rebase/merge this on top of #403 (or wait until it is merged to main)? We are doing an overhaul of how the configs are managed and I'm expecting some merge conflicts.

Apologize for the extra work!

persistence/db.go
Prajjawalk: Yep, shall I raise PR on top of "396-consolidate-config" branch?

Olshansk: Yup

Olshansk:
@Prajjawalk Could you close #424, merge this one with main (resolving any conflicts) and reupload it for review after testing?

I believe it should be good to merge then.

persistence/proto/persistence_config.proto
Olshansk: @Prajjawalk Wdyt?

persistence/db.go
```diff
@@ -65,31 +68,57 @@ func (pg *PostgresContext) ResetContext() error {
...
-	conn, err := pgx.Connect(context.Background(), postgresUrl)
+	config, err := pgxpool.ParseConfig(cfg.GetPostgresUrl())
+	if err != nil {
+		return nil, fmt.Errorf("unable to create database config: %v", err)
+	}
+	maxConnLifetime, err := time.ParseDuration(cfg.GetMaxConnLifetime())
+	if err == nil {
+		config.MaxConnLifetime = maxConnLifetime
+	} else {
+		log.Printf("unable to set max connection lifetime: %v", err)
```
Olshansk, line 82: @Prajjawalk Do you think just logging it (rather than returning an error) is more appropriate?

I was thinking that if we return `nil, err`, it'll inform the operator their configs are misconfigured and fix it. Not a lot of people look at logs :)

runtime/configs/proto/persistence_config.proto
```diff
@@ -10,4 +10,9 @@ message PersistenceConfig {
   string block_store_path = 3;
   string tx_indexer_path = 4;
   string trees_store_dir = 5;
+  int32 max_conns_count = 6;
+  int32 min_conns_count = 7;
+  string max_conn_lifetime = 8;
```
Olshansk, line 15:
```suggestion
  string max_conn_lifetime = 8; # See pkg.go.dev/time#ParseDuration for reference
```

Olshansk (changes requested):

persistence/db.go
Prajjawalk: If we return an error instead of logging then it may break the execution flow throwing a runtime error, I thought if configs are misconfigured then the error may not be that critical to break the flow as we can resort to default while informing user through logs, what do you think @Olshansk ?

persistence/db.go
```diff
@@ -65,31 +68,57 @@ func (pg *PostgresContext) ResetContext() error {
...
 	if err != nil {
 		return nil, fmt.Errorf("unable to connect to database: %v", err)
 	}
 
+	conn, _ := pool.Acquire(ctx)
+
 	// Creating and setting a new schema so we can running multiple nodes on one postgres instance. See
 	// more details at https://github.com/go-pg/pg/issues/351.
-	if _, err = conn.Exec(ctx, fmt.Sprintf("%s %s %s", CreateSchema, IfNotExists, schema)); err != nil {
+	if _, err = conn.Exec(ctx, fmt.Sprintf("%s %s %s", CreateSchema, IfNotExists, cfg.GetNodeSchema())); err != nil {
```
Olshansk, line 108: Can you create a local var for `schema := cfg.GetNodeSchema()` and reuse it in the three places below?

persistence/db.go
Olshansk: I don't think there's a right answer here as I can definitely argue both sides of it. I'd say it's probably a function of what were the worse situation a developer has experienced in the past:

1. Changing configs and spending a lot of time figuring out why they weren't being applied.
2. Misconfiguring configs and having everything crash.

There's a case to be made for each :)

Olshansk (changes requested):
One last comment and then it's good to go. Thanks for bearing with the reviews and the awesome change :)

persistence/db.go
Prajjawalk: yep

Olshansk (approved):
LGTM! Thanks for the high quality work @Prajjawalk!

Can you rebase on top of master so we can :shipit:

Olshansk:
@Prajjawalk Could you please:

- [ ] See the failing CI issue and tend to it
- [ ] Confirm that you've tested this manually with Localnet to confirm everything works. 👍 ?

Prajjawalk:
> @Prajjawalk Could you please:
> 
> * [ ]  See the failing CI issue and tend to it
> * [ ]  Confirm that you've tested this manually with Localnet to confirm everything works. 👍 ?



> @Prajjawalk Could you please:
> 
> * [X]  See the failing CI issue and tend to it
> * [X]  Confirm that you've tested this manually with Localnet to confirm everything works. 👍 ?

Olshansk:
I fixed the dates one more time and also verified that "everything on my machine". Thank you for the contribution and sorry for the delay.

Let me know if you're interested in helping out more :)

PR #417, opened 2022-12-23
[Docs] Add LocalNet issues and solutions to FAQ

h5law:
## Description

Fix errors with the `Makefile` by removing the `--build` flag from the incorrect position in the `client_start` command, and add the `rebuild_client_start` command as well.

Add a description of an issue relating to starting the LocalNet with SELinux and its solution to `docs/development/FAQ.md`. This will be useful when `make compose_and_watch` throw errors, these have also been included in the file..

## Issue

No issue documentation update

## Type of change

Please mark the relevant option(s):

- [ ] New feature, functionality or library
- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

- Add issue and solution to FAQ for SELinux blocking the docker container from accessing local files to start the LocalNet
- Remove the `--build` flag from the `client_start` command in the `Makefile` and add the `rebuild_client_start` command with the `--build` flag in the correct position

## Testing

- [x] `make develop_test`
- [x] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md) w/ all of the steps outlined in the `README`

## Required Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have tested my changes using the available tooling
- [x] I have updated the corresponding CHANGELOG

### If Applicable Checklist
- [x] I have updated the corresponding README(s); local and/or global
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added, or updated, [mermaid.js](https://mermaid-js.github.io) diagrams in the corresponding README(s)
- [ ] I have added, or updated, documentation and [mermaid.js](https://mermaid-js.github.io) diagrams in `shared/docs/*` if I updated `shared/*`README(s)

PR #421, opened 2022-12-28
Fix target for `rebuild_client_start`

Olshansk:
Quick fix for #417 which was overriding the target for `client_start` rather than creating a new one for `rebuild_client_start`


Olshansk:
cc @h5law

PR #422, opened 2022-12-28
[TECHDEBT] Change the PersistenceRWContext interface to pass paramName when getting a parameter - (Issue #286)

### LINKED ISSUE #286: [TECHDEBT] Change the `PersistenceRWContext` interface to pass `paramName` when getting a parameter
## Objective

Deprecate 
- [ ] `GetBlocksPerSession(height int64) (int, error)`
and
- [ ] `GetServiceNodesPerSessionAt(height int64) (int, error)` 

for 

- [ ] `GetParameter(paramName string, height int64) (interface, error)`

## Origin Document

We genericized the param setters like:

`SetParam(paramName string, value interface{}) error` 

We need a similar pattern for getters

## Goals

- Lower code footprint,
- Remove placeholder constants
- Maintain a strict pattern

## Deliverable

- Code complete implementation
- Updated documentation

## General issue deliverables

---

**Creator**: @andrewnguyen22

h5law:
## Description

This PR is in response to issue #286. A general function to retrieve parameters from their `paramName` string will help to address the goals laid out in the issue. Namely a lower code footprint, the removal of placeholder constants and it allows for a pattern to be used from now on regarding retrieval of parameters - similar to the pattern for setting parameters.

The function uses the `ParameterNameTypeMap` built by the `init()` function in `persistence/gov.go` to find the correct return type from the parameter name string. The map is built by:
 - Initialising an empty `Params` struct as defined in `persistence/types/persistence_genesis.pb.go`
 - Loop through the fields of this stuct:
    - Extracting the parameter name from the json tag of each field
       - Match the json tag as it is in the format of `{paramName},omitempty` and the parameter name can easily be found 
    - Setting ParameterNameTypeMap[paramName] = stringified type of field in struct

Building the map with the `init()` function and storing it in memory allows for quick access each time `GetParameter` is called removing the need to parse the `Params` struct each time it is called.

The function's signature is `GetParameter(paramName string, height int64) (any, error)`

The function's logic is defined in the file `persistence/gov.go` and works as follows:
- Using the `paramName` argument find the parameter return type from the `ParameterNameTypeMap`   
- Using this stringified parameter type, call the proper getter function `getParamOrFlag[int | string | []byte]` and the values from this call are returned.

The old functions have been removed and replaced with the new generic function. This includes the unit test in `utility/test/gov.go` where the `GetBlocksPerSession` test now uses the `GetParameter` function - passing successfully. The `GetSetIntParam` and `GetSetStringParam` unit tests in `persistence/test/gov_test.go` have also been updated to use `GetParameter` also passing successfully.

```
ok  	github.com/pokt-network/pocket/persistence/test	20.742s
ok  	github.com/pokt-network/pocket/utility/test	17.468s
```

## Issue

Fixes #286 

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Deprecate old functions (`GetBlocksPerSession` and `GetServiceNodesPerSessionAt`) in favour of using the more general `GetParameter` in `shared/modules/persistence_module.go`
- Add `init()` function in `persistence/gov.go` to build `ParameterNameTypeMap` of parameter names and their return types
- Define the logic for the `GetParameter`  in `persistence/gov.go`
- Replace old function with `GetParameter` in `utility/gov.go`
- Change unit test for `GetBlocksPerSession` in `utility/test/gov_test.go` to use the new function
- Change unit tests in `persistence/test/gov_test.go` to use the `GetParameter()` function instead of the `GetIntParam`, and `GetStringParam` functions

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


h5law:
@Olshansk I cannot assign myself to this, I don't think I have permissions to do that sort of thing.

I am also aware that in this implementation there will never be an instance where `getParamOrFlag[[]byte]` is used as none of the functions in the `Param` struct have that type. I am happy to work on a different implementation, but this feels like a good starting point.

persistence/gov.go
```diff
@@ -35,6 +40,34 @@ func (p PostgresContext) InitParams() error {
 	return err
 }
 
+func (p PostgresContext) GetParameter(paramName string, height int64) (v any, err error) {
+	st := reflect.TypeOf(types.Params{}) // Extract type of parameter from the Params struct
+	var typ reflect.Type
```
Olshansk, line 45: Wdyt of initializing a local in memory map using the `init` function that keeps a mapping of the tags from `persistence_genesis.pb.go` to avoid looping every time GetParam is called.

persistence/gov.go
```diff
@@ -35,6 +40,34 @@ func (p PostgresContext) InitParams() error {
 	return err
 }
 
+func (p PostgresContext) GetParameter(paramName string, height int64) (v any, err error) {
```
Olshansk, line 43: This business logic is very proto-specific and how we configured the proto as well. Can you add a comment explaining what's happening? A condensed version of what you have in the github description should be good enough.

Olshansk (changes requested):
Reflections in business logic are usually eery, but this is pretty clean.

If we do this, we can simplify all of the gov params accessors in a follow-up task.

@h5law Would you be open to it?

persistence/gov.go
h5law: Added code explaining the `init()` function parsing the `Params{}` struct to build the in memory map and a shorter comment for the `GetParameter()` function now the logic is seperated

persistence/gov.go
h5law: @Olshansk Great idea speeds things up a lot! Implemented as well - removes the need for any RegEx check as well - 1 less potential error.

utility/test/gov_test.go
```diff
@@ -83,11 +83,23 @@ func TestUtilityContext_GetBaselineAppStakeRate(t *testing.T) {
...
+//		require.Equal(t, defaultParam, gotParam)
+//
+//		test_artifacts.CleanupTest(ctx)
+//	}
+func TestUtilityContext_GetParameter(t *testing.T) {
 	ctx := NewTestingUtilityContext(t, 0)
 	defaultParams := DefaultTestingParams(t)
 	defaultParam := int(defaultParams.GetBlocksPerSession())
-	gotParam, err := ctx.GetBlocksPerSession()
+	gotParam, err := ctx.GetParameter("blocks_per_session", 0)
```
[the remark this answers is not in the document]
h5law: Have uncommented the constants 👍🏼

h5law:
> Reflections in business logic are usually eery, but this is pretty clean.

I couldn't think of any other way. I found the `pokt` tags of these structs not very clear as to the actual return types. `BIGINT` and `SMALLINT` I expected would return either `int64` and `int32` or `big.Int` and `int` but most are both `int32` return types which is why I didn't end up exporting the `parseGovProto` function or the map it creates as I found simply using the return types of the functions this getter would ultimately be calling to be easier.

> If we do this, we can simplify all of the gov params accessors in a follow-up task.

@Olshansk I would be up for taking this on - I am about to go back to uni and have a couple exams but after those I would be happy to jump in and continue this work 😄 So in around 2 weeks time, I could start on it.

persistence/gov.go
```diff
@@ -19,12 +21,33 @@ const (
 // TODO (Team) BUG setting parameters twice on the same height causes issues. We need to move the schema away from 'end_height' and
 // more towards the height_constraint architecture
 
-func (p PostgresContext) GetBlocksPerSession(height int64) (int, error) {
-	return p.GetIntParam(BlocksPerSessionParamName, height)
-}
-
-func (p PostgresContext) GetServiceNodesPerSessionAt(height int64) (int, error) {
-	return p.GetIntParam(ServiceNodesPerSessionParamName, height)
+// Deprecate these functions in favour of the getter function:
```
Olshansk, line 24: Let's delete this 💩 !

Ditto in other places in the codebase

persistence/gov.go
```diff
@@ -19,12 +21,33 @@ const (
 // TODO (Team) BUG setting parameters twice on the same height causes issues. We need to move the schema away from 'end_height' and
```
Olshansk, line 21: We already this this, so can you remove this TODO while you're in here

persistence/gov.go
```diff
@@ -19,12 +21,33 @@ const (
...
+// func (p PostgresContext) GetServiceNodesPerSessionAt(height int64) (int, error) {
+// 	return p.GetIntParam(ServiceNodesPerSessionParamName, height)
+// }
+
+// Mapping of parameter names and their stringified type names
+var ParameterNameTypeMap = make(map[string]string)
+
+// Using the json tag of the fields in the Params struct to extract the name of the
+// parameters and build a mapping in memory of the types associated to each parameter
+// name according to the struct defined in: persistence/types/persistence_genesis.pb.go
```
Olshansk, line 40:
```suggestion
// name according to the struct generated from `persistence_genesis.proto`
```

persistence/gov.go
```diff
@@ -19,12 +21,33 @@ const (
...
+// func (p PostgresContext) GetBlocksPerSession(height int64) (int, error) {
+// 	return p.GetIntParam(BlocksPerSessionParamName, height)
+// }
+
+// func (p PostgresContext) GetServiceNodesPerSessionAt(height int64) (int, error) {
+// 	return p.GetIntParam(ServiceNodesPerSessionParamName, height)
+// }
+
+// Mapping of parameter names and their stringified type names
+var ParameterNameTypeMap = make(map[string]string)
```
Olshansk, line 36: lowercase

persistence/gov.go
```diff
@@ -39,6 +62,23 @@ func (p PostgresContext) InitGenesisParams(params *genesis.Params) error {
...
+// Match paramName against the ParameterNameTypeMap struct and call the proper
+// getter function getParamOrFlag[int | string | byte] and return its values
+func (p PostgresContext) GetParameter(paramName string, height int64) (v any, err error) {
+	paramType := ParameterNameTypeMap[paramName]
+	switch paramType {
+	case "int", "int32", "int64":
+		v, _, err = getParamOrFlag[int](p, types.ParamsTableName, paramName, height)
+	case "string":
+		v, _, err = getParamOrFlag[string](p, types.ParamsTableName, paramName, height)
+	case "[]uint8": // []byte
```
Olshansk, line 74: Nice comment

shared/modules/persistence_module.go
```diff
@@ -125,9 +125,9 @@ type PersistenceReadContext interface {
 
 	// CONSOLIDATE: BlockHash / AppHash / StateHash
 	// Block Queries
-	GetLatestBlockHeight() (uint64, error)         // Returns the height of the latest block in the persistence layer
-	GetBlockHash(height int64) (string, error)     // Returns the app hash corresponding to the height provided
-	GetBlocksPerSession(height int64) (int, error) // TECHDEBT(#286): Deprecate this method
+	GetLatestBlockHeight() (uint64, error)     // Returns the height of the latest block in the persistence layer
+	GetBlockHash(height int64) (string, error) // Returns the app hash corresponding to the height provided
+	//GetBlocksPerSession(height int64) (int, error) // TECHDEBT(#286): Deprecate this method
```
Olshansk, line 130: Delete. Ditto elsewhere

Olshansk (changes requested):
@h5law Can you create a new issue explaining that we should apply this business logic in other parts of the codebase and add a `TODO(#XXX): Replace all gov parameter accessors with GetParameter`. No rush though, exams come first :)

Also, I've added you as an external contributor to https://github.com/orgs/pokt-network/teams/external-contributors/members

spider-man-uncle-ben

h5law:
@Olshansk I have removed those old comments - I did make one change to the logic of `GetParameter()` though. I noticed that all the parameters ending with `_owner` are actually used as `[]bye` not strings and the rest of the string types are base 10 `*big.Int` types so I added this handling into the `init()` function. If this assumption is incorrect I can change it but this would be useful as we move towards fully generalising these parameter accessor. I could grok the types from the `pokt` tags so maybe this is better - however now instead of `[]uint8` never being called `string` will never be called - as it is now split into `*big.Int` and `[]uint8` depending on parameter names.

Before I create the issue can I just confirm the task:
 - Replace all uses of `GetIntParam` `GetStringParam` `GetBytesParam` with `GetParameter`, across `persistence`, `utility` and any other places they are used?
- For example in `utility/gov.go`

Change:
```go
func (u *UtilityContext) GetAppMinimumStake() (*big.Int, typesUtil.Error) {
	return u.getBigIntParam(typesUtil.AppMinimumStakeParamName)
}
...
func (u *UtilityContext) getBigIntParam(paramName string) (*big.Int, typesUtil.Error) {
	store, height, er := u.GetStoreAndHeight()
	if er != nil {
		return nil, er
	}
	value, err := store.GetStringParam(paramName, height)
	if err != nil {
		log.Printf("err: %v\n", err)
		return nil, typesUtil.ErrGetParam(paramName, err)
	}
	return typesUtil.StringToBigInt(value)
}
```
To
```go
func (u *UtilityContext) getBigIntParam(paramName string) (*big.Int, typesUtil.Error) {
	store, height, er := u.GetStoreAndHeight()
	if er != nil {
		return nil, er
	}
	value, err := store.GetParameter(paramName, height)
	if err != nil {
		log.Printf("err: %v\n", err)
		return nil, typesUtil.ErrGetParam(paramName, err)
	}
	return value
}
```
Or?
```go
func (u *UtilityContext) GetAppMinimumStake() (*big.Int, typesUtil.Error) {
        store, height, err := u.GetStoreAndHeight()
        if err != nil {
                return nil, err
        }
	stake, err := u.GetParameter(typesUtil.AppMinimumStakeParamName, height)
        if err != nil {
                return nil, err
        }
        return stake, nil
}
```

persistence/gov.go
```diff
@@ -16,15 +20,33 @@ const (
...
+		json := field.Tag.Get("json") // Match the json tag of field: json:"paramName,omitempty"
+		paramName := strings.Split(json, ",")[0]
+		typ := field.Type.Name() // Get string version of field's type
+		matchOwner := ownerRegEx.MatchString(paramName)
+		// Handle string types
+		if typ == "string" && matchOwner {
+			typ = "[]uint8" // All owners are string types but used as []byte slices
+		} else if typ == "string" {
+			typ = "*big.Int" // Rest of string types are used as *big.Int
+		}
```
Olshansk, line 47: I'd like to remove lines 42-47 altogether. It might be the case right now, but it means we can never have a non integer string parameter. 

Does anything fail if we remove this?

persistence/gov.go
```diff
@@ -39,6 +61,31 @@ func (p PostgresContext) InitGenesisParams(params *genesis.Params) error {
...
+		val, _, err := getParamOrFlag[string](p, types.ParamsTableName, paramName, height)
+		b := big.Int{}
+		v, ok := b.SetString(val, 10)
+		if !ok {
+			return nil, fmt.Errorf("error converting parameter string into base 10 *big.Int.")
+		}
+		return v, err
+	case "[]uint8": // []byte
+		v, _, err = getParamOrFlag[[]byte](p, types.ParamsTableName, paramName, height)
+	case "string":
```
Olshansk, line 81: With the change in `init`, we never reach this anymore.

utility/gov.go
```diff
@@ -34,16 +34,9 @@ func (u *UtilityContext) UpdateParam(paramName string, value interface{}) typesU
...
-		return 0, er
-	}
-	blocksPerSession, err := store.GetBlocksPerSession(height)
-	if err != nil {
-		return typesUtil.ZeroInt, typesUtil.ErrGetParam(typesUtil.BlocksPerSessionParamName, err)
-	}
-	return blocksPerSession, nil
+func (u *UtilityContext) GetParameter(paramName string, height int64) (any, error) {
+	store := u.Store()
+	return store.GetParameter(paramName, height)
```
Olshansk, line 39:
```suggestion
	return u.Store().GetParameter(paramName, height)
```

Olshansk (changes requested):
Need a bit more time to think through the followup issue, but wanted to post a couple questions/commnets I had here in the meantime.

persistence/gov.go
h5law: @Olshansk No nothing will fail, I have changed 1 unit test back to using `GetBytesParam` as removing the `_owner` check on the param name means again that there will not be any `[]byte` types, as none of the Params have this type. I will leave the `[]byte` case in the switch though so it doesn't need to be added at another time. I am imagining a follow up to this task would include sorting out this parameter type issue so better to have the 3 getter functions covered.

h5law:
@Olshansk What do you think about storing the `[]byte` type parameters (so far those ending in `_owner`) as `[]byte` instead of `strings`? This would 1. get rid of the encoding and decoding steps, and 2. simplify the issue of retrieving them in the intended type as they are to be used?

persistence/gov.go
Olshansk: `I am imagining a follow up to this task would include sorting out this parameter type issue so better to have the 3 getter functions covered.`

I think that's a good idea and I like the simplification right now.

Olshansk:
> @Olshansk What do you think about storing the `[]byte` type parameters (so far those ending in `_owner`) as `[]byte` instead of `strings`? This would solve the issue around retrieving the correct type of each parameter, for the way it is to be used. I realise this is quite a big change, as editing the `runtime/genesis/proto/genesis.proto` file will affect the `utility`, `shared`, `persistence` and `consesus` modules I believe

This might make parsing easier, but creating the protobufs (reading, viewing, editing) will create a much worse developer experience since we wouldn't be able to just use strings anywhere, so I'm not a big fan of the idea at the moment.

---

SIde note: the PR looks pretty good to me so could you merge with main and just re-run the tests when you have a chance.

h5law:
@Olshansk 

> This might make parsing easier, but creating the protobufs (reading, viewing, editing) will create a much worse developer experience since we wouldn't be able to just use strings anywhere, so I'm not a big fan of the idea at the moment.

Yeah I understand that, no worries will continue to think on it!

> SIde note: the PR looks pretty good to me so could you merge with main and just re-run the tests when you have a chance.
 On it now 🫡

Olshansk (approved):
Let's :shipit:

Olshansk:
@h5law I added you to https://github.com/orgs/pokt-network/teams/external-contributors. I'm curious to know if this allows you to merge this in yourself, given that you have approval from someone on the core team. If so, do a "squash & merge" and use the body of the GH description as the commit message.

Also, can you create the follow-up issue and add in the details we've already discussed here? We can narrow down the requirements and scope out the solution in that thread before continuing to the implementation.

PR #437, opened 2023-01-12
[Core] Externalize a shared Mempool interface & implementation - (Issue #388)

### LINKED ISSUE #388: [Core] Externalize a shared `Mempool` interface & implementation
## Objective

Make the `Mempool` interface and implementation a common high-level type that can be reused across different modules

## Origin Document

This discussion in #374: https://github.com/pokt-network/pocket/pull/374#discussion_r1041652264

Screenshot 2022-12-08 at 4 53 32 PM

## Goals

- Create a Mempool type that other modules can easily import & reuse
- Clean up the code so all the modules are using the same mempool implementation

## Deliverable

- [ ] Create a shared mempool object based on `utility/types/Mempool` into `shared/mempool`
- [ ] Import and use `shared/mempool` in the utility module
- [ ] Import and use `shared/mempool` in the p2p module
- [ ] Add a couple mempool specific tests
- [ ] In `utility/test/module_test.go`: `TODO(#388): Expose a `GetMempool` function in `utility_module` so we can remove this reflection.`

- Modify existing business logic

## General issue deliverables

## Testing Methodology

- [ ] **Mempool tests **: `make test_mempool`

---

**Creator**: @Olshansk 
**Co-Owners**: @deblasis

deblasis:
## Description

This PR Makes the `Mempool` interface and implementation a common high-level type that can be reused across different modules

## Issue

Fixes #388

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Created generic/configurable fifo mempool
- Implemented TxMempool variant WIP

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


deblasis:
Hey @Olshansk

I would really appreciate a quick glance at this one, **ONLY TWO FILES!** 
It's the first pass. Kinda like my design idea. If you like where this is going I'll do some hardening, implement also the P2P variant and of course add tests.

Au contraire, if this sucks just delete the branch and let me know :)

🙏

utility/types/mempool_new.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/pokt-network/pocket/shared/crypto"
+	"github.com/pokt-network/pocket/shared/mempool"
+)
+
+var _ Mempool = &TxFIFOMempool{}
+
+type TxFIFOMempool struct {
+	g          *mempool.GenericFIFOSet[string, string]
+	m          sync.Mutex
```
Olshansk, line 15: Mutexes are necessary but I like to avoid them whenever possible.

Provided that there is a mutex in the underlying FIFO Set, could we avoid one here? 

Could be a fun opportunity for concurrent fuzz testing

shared/mempool/generic_fifo_set.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		return fmt.Errorf("item %v already exists", item)
+	}
+
+	g.set[index] = struct{}{}
+	g.queue.PushBack(item)
+	g.onAdd(item, g)
+	if g.isOverflowing != nil && g.isOverflowing(g) {
+		//if g.queue.Len() > g.maxElements {
+		front := g.queue.Front()
```
Olshansk, line 40: Seems like this is somethins that should be the default "customIsOverflowing" but the logic shouldn't be executed if a custom overflow handler is provided

Olshansk (commented):
Fascinating. I like the direction and also see us just open-sourcing this as a standalone 
package.

Please proceed!

utility/types/mempool_new.go
```diff
@@ -0,0 +1,107 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/pokt-network/pocket/shared/crypto"
+	"github.com/pokt-network/pocket/shared/mempool"
+)
+
+var _ Mempool = &TxFIFOMempool{}
+
+type TxFIFOMempool struct {
```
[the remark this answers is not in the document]
deblasis: "Accept interfaces, return structs" - **Go**nfucios 🙂

utility/types/mempool_new.go
deblasis: I agree but in this case we have a state that it's split into the generic part and the custom part (where we track specific things like `txBytes` and `size`, therefore we need two mutexes.

Unless you see a cleverer way to handle this. I thought about injecting a map into the generic part but it sounded very hacky and needlessly complex

shared/mempool/generic_fifo_set.go
deblasis: It's already implemented in the constructor, I made it more explicit now `defaultIsOverflowing`.

utility/types/mempool_new.go
```diff
@@ -0,0 +1,107 @@
...
+
+// AddTransaction implements Mempool
+func (t *TxFIFOMempool) AddTransaction(tx []byte) Error {
+	if err := t.g.Push(string(tx)); err != nil {
+		return ErrDuplicateTransaction()
+	}
+	return nil
+}
+
+// Clear implements Mempool
```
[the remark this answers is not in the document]
deblasis: Do you mean the `Mempool` interface (that I renamed to `TXMempool`?

If so, do you envisage other modules where this specific implementation that deals with transactions will be used in the codebase? 

My mind is at: let's leave it here for now, close to it's implementation. When more structs implement it, move to `shared`.

What's really shared here is the underlying `GenericFIFOSet` IMHO

utility/types/mempool_new.go
```diff
@@ -0,0 +1,107 @@
...
+	g          *mempool.GenericFIFOSet[string, string]
+	m          sync.Mutex
+	size       uint32 // current number of transactions in the mempool
+	txBytes    uint64 // current sum of all transactions' sizes (in bytes)
+	maxTxBytes uint64 // maximum total size of all txs allowed in the mempool
+}
+
+// AddTransaction implements Mempool
+func (t *TxFIFOMempool) AddTransaction(tx []byte) Error {
+	if err := t.g.Push(string(tx)); err != nil {
```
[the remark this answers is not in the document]
deblasis: Fair point, I didn't like it either but I figured that we only need to implement `comparable` for the index part so now it should be all good.

deblasis:
The next batch, that will include the below todos, should be the last one and then I guess this is ready for a full-fledged review.

TODO:
- tests for `utility.mempool`
- tests for `GenericFIFOSet`

I have added tests for the simplest implementation to give you an idea of what I am doing.

deblasis:
Looking for a couple of smarter eyes on this, E2E tests in consensus are failing at the moment, I am sure it's something stupid but I have tunnel vision now

deblasis:
As discussed on discord, I identified and addressed the issue in https://github.com/pokt-network/pocket/pull/437/commits/c0f7a3996108e0883ae791e59d06249cd73822cc

From discord:

> it seems that my machine is slower than usual and the previous timeout value wasn't providing enough time for consensus to work its way till the end.

> A larger number in there doesn't change the logic of the tests because we use relative values to trigger timeouts, it just gives more real time for the messages to reach the eventsChannel and be checked down the line if that makes sense

@Olshansk , @gokutheengineer and myself concurred that there's some room for improvement wrt determinism, a followup issue/PR will make life and DevEx easier going forward 🚀

Olshansk:
Follow-up ticket on determinism can be found here: #462

utility/types/mempool_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/pokt-network/pocket/shared/crypto"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMempool(t *testing.T) {
+	type args struct {
+		maxTransactionBytes uint64
+		maxTransactions     uint32
+		initialElements     *[][]byte
```
Olshansk, line 14: s/initialElements/initialTxs?

utility/types/mempool_test.go
```diff
@@ -0,0 +1,55 @@
...
+		actions             *[]func(*fIFOMempool)
+	}
+	tests := []struct {
+		name      string
+		args      args
+		wantItems [][]byte
+	}{
+		//TODO: Add test cases.
```
Olshansk, line 22: I've found that I usually forget what I'd ideally want to test if I had time, and this sort of thing is great as starter tasks or bounty tickets for people getting involved from the community.

```suggestion
		// ADDTEST: Test1
		// ADDTEST: Test2
```

consensus/hotstuff_mempool.go
```diff
@@ -0,0 +1,141 @@
...
+	typesCons "github.com/pokt-network/pocket/consensus/types"
+	"github.com/pokt-network/pocket/shared/codec"
+	"github.com/pokt-network/pocket/shared/crypto"
+	mempool "github.com/pokt-network/pocket/shared/mempool/list"
+)
+
+// TODO: index HighQC incrementally
+// TODO: expose findHighQC
+// TODO: implement iterator
+// TODO: tests for list as well
```
Olshansk, line 20: Can you just add a bit more details here. I realize it was a note you made to yourself at some point, but pretty confusing to the reader.

consensus/hotstuff_mempool_test.go
```diff
@@ -0,0 +1,230 @@
...
+	typesCons "github.com/pokt-network/pocket/consensus/types"
+	"github.com/pokt-network/pocket/shared/core/types"
+	"github.com/stretchr/testify/require"
+)
+
+// hotstuffMessageFactory returns a hotstuff message with a block with a single transaction of the given size
+//
+// it is used in tests to make sure that the mempool is correctly limiting the size of the transactions it stores
+//
+// Important: the size of the transaction is not the size of the whole message on the wire, there's obviously some overhead in the protobuf
```
Olshansk, line 16:
```suggestion
// IMPORTANT: the size of the transaction is not the size of the whole message on the wire, there's obviously some overhead in the protobuf
```

consensus/hotstuff_mempool_test.go
```diff
@@ -0,0 +1,230 @@
...
+func hotstuffMessageFactory(size int) *typesCons.HotstuffMessage {
+	tx := make([]byte, size)
+	for i := range tx {
```
Olshansk, line 19: Could we do `rand.Read(tx)` here insetad?

utility/types/mempool.go
```diff
@@ -1,166 +1,98 @@
...
 }
 
-func (f *FIFOMempool) Size() uint32 {
-	f.l.RLock()
-	defer f.l.RUnlock()
-
-	return f.size
+func (t *fIFOMempool) PopTransaction() ([]byte, error) {
+	popTx, err := t.g.Pop()
+	return []byte(popTx), NewError(-1, err.Error()) // TODO: prettier
```
Olshansk, line 42:
```suggestion
	return []byte(popTx), NewError(-1, err.Error())
```

Olshansk (approved):
Reviewed everything except for the 3 `_test.go` files. The coverage looks great and I'll take a deeper look after most of the comments are addressed. 

Note that a left a handful of "non actionables" so feel free to just resolve those.

This is really awesome work!

Olshansk (changes requested):
Oops, meant to request changes 😅

consensus/hotstuff_mempool_test.go
deblasis: Good shout, we could but then we couldn't use the same factory method to make assertions on the expected messages in the tests below.

I thought that the simplest, most idiomatic Go way to deterministically create a message of len(n) is to just fill it with a well known sequence. 

I went for: **"All 1s"**

consensus/hotstuff_leader.go
```diff
@@ -51,7 +49,7 @@ func (handler *HotstuffLeaderMessageHandler) HandleNewRoundMessage(m *consensusM
 
 	// Likely to be `nil` if blockchain is progressing well.
 	// TECHDEBT: How do we properly validate `prepareQC` here?
-	highPrepareQC := m.findHighQC(m.messagePool[NewRound])
+	highPrepareQC := m.findHighQC(m.consensusMessagePool[NewRound].GetAll()) //TODO: improve this
```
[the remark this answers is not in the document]
deblasis: Lemme say it out loud, perhaps it's a new ticket straight away.

I was thinking to further optimize this by having an abstraction around/extension on the current logic so that every time a new message is added to the mempool, it's immediately checked if it's a highPrepareQC and in that case that information is "memoized".

This way, we don't have to scan all the messages in the mempool to get the HighQC, we would have it already.

👍new ticket with the above description
👎small enough to add scope here

Curious about your thoughts on this. Maybe I am optimizing prematurely

shared/mempool/generic_fifo_set.go
```diff
@@ -0,0 +1,181 @@
+package mempool
+
+import (
+	"container/list"
+	"fmt"
+	"sync"
+)
+
+type GenericFIFOSet[TIdx comparable, TData any] struct {
```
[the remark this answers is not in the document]
deblasis: Oh, thank you!🙏

When you think it's mature enough give me a name and a repo and I'll standalone it :)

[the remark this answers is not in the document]
deblasis: No problem

shared/mempool/generic_fifo_set.go
```diff
@@ -0,0 +1,181 @@
...
+	g.m.Lock()
+	defer g.m.Unlock()
+
+	if g.queue.Len() == 0 {
+		return any(nil).(TData), fmt.Errorf("empty set")
+	}
+
+	front := g.queue.Front()
+	item := front.Value.(TData)
+	delete(g.set, g.indexerFn(item))
```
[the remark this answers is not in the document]
deblasis: Fair point. Now my brain is already thinking about how to potentially handle this more gracefully and if it's appropriate. 
I guess at least printing a meaningful error message before panicing would be nice.

Definitely a 🧠📌
I might end up creating a PR when this is a separate package :P

shared/mempool/generic_fifo_set.go
```diff
@@ -0,0 +1,181 @@
...
+
+func WithOnCollision[TIdx comparable, TData any](fn func(TData, *GenericFIFOSet[TIdx, TData]) error) func(*GenericFIFOSet[TIdx, TData]) {
+	return func(g *GenericFIFOSet[TIdx, TData]) {
+		g.onCollision = fn
+	}
+}
+
+// private methods
+
+func defaultIsOverflowing[TIdx comparable, TData any](g *GenericFIFOSet[TIdx, TData]) bool {
```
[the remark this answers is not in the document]
deblasis: Please flag any misnaming or if any of them is unclear

https://github.com/pokt-network/pocket/pull/437/commits/b314c72ddbb3a41c6515bc72573b3914af152b18

shared/mempool/list/generic_fifo_list.go
```diff
@@ -0,0 +1,199 @@
...
+
+import (
+	"container/list"
+	"encoding/json"
+	"fmt"
+	"log"
+	"sync"
+)
+
+type GenericFIFOList[TData any] struct {
```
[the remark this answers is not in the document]
deblasis: They are two different animals.

A set has inherently unique elements, a list can have multiple identical elements

shared/mempool/mempool.go
```diff
@@ -0,0 +1,15 @@
+package mempool
```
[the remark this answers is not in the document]
deblasis: 💯

shared/mempool/mempool.go
```diff
@@ -0,0 +1,15 @@
...
+	Contains(hash string) bool
+	AddTransaction(tx []byte) error
+	RemoveTransaction(tx []byte) error
+
+	Clear()
+	Size() uint32 // Returns the number of transactions stored in the mempool
+	IsEmpty() bool
+	TxsBytes() uint64 // Returns the total sum of all transactions' sizes (in bytes) stored in the mempool
+	PopTransaction() (tx []byte, err error)
+}
```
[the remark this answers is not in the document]
deblasis: When it comes to naming things I recognize my limitations (second language, pretty shitty in general... I can give you examples in private of things I have named very badly in my career 😅) 
I almost blindly trust your judgement when it makes sense

utility/test/module_test.go
```diff
@@ -89,8 +88,8 @@ func NewTestingUtilityContext(t *testing.T, height int64) utility.UtilityContext
 	})
 
 	return utility.UtilityContext{
-		Height:  height,
-		Mempool: mempool,
+		Bus:    testUtilityMod.GetBus(), // TODO
```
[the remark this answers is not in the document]
deblasis: This is what I had in mind 

https://github.com/pokt-network/pocket/pull/437/commits/e9e86c3667dab0234bbde7f7d7f135d89f520850

so that bus can be unexported

p2p/raintree/network.go
```diff
@@ -28,13 +28,7 @@ type rainTreeNetwork struct {
...
 
 	peersManager *peersManager
-
-	// TODO (#278): What should we use for de-duping messages within P2P?
-	// TODO(#388): generalize to use the shared `FIFOMempool` type (in `utility/types/mempool.go` at the time of writing) in here as well for this.
-	nonceSet         map[uint64]struct{}
-	nonceList        []uint64
-	mempoolMaxNonces uint64
-	nonceSetMutex    sync.Mutex
+	nonceDeduper *mempool.GenericFIFOSet[uint64, uint64]
```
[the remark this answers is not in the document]
deblasis: because it's not doing anything "special" just using the basic behaviour offered by the genericFIFOSet. 

In the other cases there are some nuances that I captured creating a specific type.

I am not sure we should add a comment to be honest... Should we really make the reader aware of the fact that there are other implementations of this type that instead are wrapped in a specific type because they have some extra specific behaviour here even if it's completely unrelated with raintree and deduping?

consensus/hotstuff_mempool.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return hotstuffFifoMempool
+}
+
+func (mp *hotstuffFifoMempool) Push(msg *typesCons.HotstuffMessage) error {
+	return mp.g.Push(msg)
+}
+
+func (mp *hotstuffFifoMempool) Clear() {
+	mp.g.Clear()
```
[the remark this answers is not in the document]
deblasis: We are clearing the underlying GenericFIFOList first and then the state that is hotstuffFIFOMempool specific. The order doesn't really matter. 
If you are referring to the Clear not being after the Lock that's why ☝️

Does this make sense?

consensus/hotstuff_mempool.go
deblasis: 💯 these are going to disappear soon

deblasis:
Yo @Olshansk! I think this is ready for another pass. PTAL 🙏

consensus/hotstuff_leader.go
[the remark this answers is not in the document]
Olshansk: Let's rename `DISCUSS` to `CONSIDERATION(M5)`. 

This approach is basically how Hotstuff specs out the default design and you're providing a good implementation optimization. Low priority enough for now to not need a ticket.

shared/mempool/generic_fifo_set.go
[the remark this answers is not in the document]
Olshansk: Easy: fifo-go!

shared/mempool/mempool.go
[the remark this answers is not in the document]
Olshansk: >  I can give you examples in private of things I have named very badly in my career

Looking forward to it :)

> I almost blindly trust your judgement when it makes sense

Likewise

consensus/hotstuff_mempool_test.go
Olshansk: Sounds > 💯

shared/mempool/tx_mempool.go
```diff
@@ -0,0 +1,15 @@
+package mempool
+
+// type GenericFIFOList = mempool_list.GenericFIFOList[any] // TODO: fix this aliasing
```
Olshansk, line 3: Why do we need/want to create these types?

shared/mempool/generic_fifo_set.go
```diff
@@ -68,16 +60,17 @@ func (g *GenericFIFOSet[TIdx, TData]) Push(item TData) error {
 		front := g.queue.Front()
 		delete(g.set, g.indexerFn(front.Value.(TData)))
 		g.queue.Remove(front)
+		g.onRemove(item, g)
```
Olshansk, line 63: I did not catch this...

utility/types/tx_fifo_mempool.go
```diff
@@ -0,0 +1,108 @@
...
+func (t *txFIFOMempool) AddTx(tx []byte) error {
+	if err := t.g.Push(tx); err != nil {
+		return ErrDuplicateTransaction()
+	}
+	return nil
+}
+
+// Clear clears the mempool
+func (t *txFIFOMempool) Clear() {
+	t.g.Clear()
```
Olshansk, line 30: Same as before. My inutition just makes me think:

1. Lock outer structure
2. Clear inner structure
   2.1. Capture inner lock
   2.2. Unlock inner lock
3. Clear outer strcture
4. Unlock outer structure

utility/types/tx_fifo_mempool_test.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/pokt-network/pocket/shared/codec"
+	"github.com/pokt-network/pocket/shared/crypto"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTxFIFOMempool(t *testing.T) {
+	type args struct {
+		maxTxBytes uint64
+		maxTxs     uint32
+		initialTxs *[][]byte
```
Olshansk, line 16: 1. Remove pointer?
1.1 Why are we doing a pointer to a slice? When slices are passed around, it's already just a reference to the first element.
1.2 If we do this, we should do it for `actions` too

2. If we don't do this, why is this a pointer but `wantItems` is not?

The diagram here helped me think through it: https://medium.com/swlh/golang-tips-why-pointers-to-slices-are-useful-and-how-ignoring-them-can-lead-to-tricky-bugs-cac90f72e77b

Screenshot 2023-02-02 at 5 18 58 PM

utility/types/tx_fifo_mempool_test.go
```diff
@@ -0,0 +1,298 @@
...
+		})
+	}
+}
+
+// messageSendFactory returns a MessageSend message marshalled in []byte format
+//
+// it is used in tests to make sure that the mempool is correctly limiting the size of the transactions it stores
+//
+// IMPORTANT: the size of the transaction is not the size of the whole message on the wire, there's obviously some overhead in the protobuf
+func messageSendFactory(size int) []byte {
```
Olshansk, line 285: s/size/amountByteSize

utility/types/tx_fifo_mempool_test.go
```diff
@@ -0,0 +1,298 @@
...
+	for i := range amountBz {
+		amountBz[i] = 1
+	}
+
+	message := &MessageSend{
+		FromAddress: []byte{},
+		ToAddress:   []byte{},
+		Amount:      string(amountBz),
+	}
+	messageBz, _ := codec.GetCodec().Marshal(message)
```
Olshansk, line 296: I think passing in `t` and running a `require.Error` here is fair game

Olshansk (changes requested):
One last round of small comments and we should be able to merge it in tomorrow

shared/mempool/tx_mempool.go
deblasis: I was trying to be fancy. Basically `GenericFIFOList` resides under the `list` folder 
image
and when importing we need to do this:
image

I wanted to alias it so that it was coming from the root package but with generics it's not possible it seems, I haven't spent more than 10secs on it.

shared/mempool/generic_fifo_set.go
deblasis: whenever there's an overflow, we need to compensate for the onAdd action we did earlier because effectively we are deleting an element from the data structures

utility/types/tx_fifo_mempool_test.go
deblasis: I normally use pointers in tests like these when I feel like I might not want to configure some cases with `InitialTxs`

Basically to provide that flexibility to the tester.

image

if nil we just skip that.

utility/types/tx_fifo_mempool_test.go
deblasis: I think it's a bit overkill and would add a "distracting" element to the factory signature. Ultimately a panic would be telling and regardless if we can't marshal an empty message our codebase would look like this anyway:

image

I hear you but I think it's clean enough and the corner we are cutting by not having a `require.NoError()` is negligeable.

If I didn't convince you with my hell picture I'll just update the code :)

utility/types/tx_fifo_mempool.go
deblasis: My 🚂 of 🤔 is:
1. Do what the generic implementation does, 
2. Do what's special to this implementation

Basically this:  https://github.com/pokt-network/pocket/pull/437/commits/b875b37f4a85499120adfccd8b3f99faea654972

shared/mempool/tx_mempool.go
deblasis: I have removed that comment in the meantime BTW

consensus/hotstuff_mempool.go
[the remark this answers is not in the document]
deblasis: Is [this](https://github.com/pokt-network/pocket/pull/437/commits/711d3df4117f340c167f401c40efc19c6212ca79) any better? (I explained in https://github.com/pokt-network/pocket/pull/437#discussion_r1095913547
my reasoning

deblasis:
@Olshansk I think this is almost there

CaffBuongiornoGIF


 a couple of open points that might have a quick turnaround anyway but yeah, I like this. 
Thanks for your inputs 🙏

utility/types/tx_fifo_mempool_test.go
Olshansk: You convinced me with your hell picture. Will put it out by resolving this conversation 🚒

utility/types/tx_fifo_mempool_test.go
Olshansk: I see the use case and raise you this mini demo:

Screenshot 2023-02-03 at 10 52 16 AM

If we just don't set a slice, it's nil by default.

I would request one of following:
1. Document (in the struct next to the field) why it it's a pointer
2. Not use a pointer at all

Source code for reference:
```go
package main

import "fmt"

func main() {
	type args struct {
		ptrSlice *[]byte
		slice    []byte
	}

	a := &args{
		ptrSlice: &[]byte{},
		slice:    []byte{},
	}

	b := &args{}

	fmt.Println("a.ptrSlice:", a.ptrSlice == nil)
	fmt.Println("a.slice   :", a.slice == nil)

	fmt.Println("b.ptrSlice:", b.ptrSlice == nil)
	fmt.Println("b.slice   :", b.slice == nil)
}
```

utility/types/tx_fifo_mempool.go
Olshansk: Won't push back against this since it doesn't matter.

---

Going to share my personal 🚄  of 🤔 thought:

1. In polymorphism (when a subclass inherits from a superclass), we usually do:

```
super.constructor()
self.constructioner()
```

2. In this case, where the generic implementation is an internal field, I'm thinking:

```
self.clearFields()
attributes.clearFields()
```

Olshansk:
@deblasis Just followed up on one comment and nothing else: https://github.com/pokt-network/pocket/pull/437#discussion_r1095250157

utility/types/tx_fifo_mempool_test.go
deblasis: 💯 but I meant it to be used like this:

image

utility/types/tx_fifo_mempool.go
deblasis: Well, it's not a rule, you can call the super before or after depending on how the subclass needs to behave.

If for example `super.constructor()` prints something that is dependent on some internal state (accessible and modifiable by the `self`, it makes sense for the `self.constructioner()` to happen before.

Olshansk (approved):
@deblasis Can you just add a short comment on why we have pointers for the `actions` and `initialTxs` (w.r.t. nil comparison)? Otherwise, let's 🚢

Issue #239, opened 2022-09-22
[Automation] CI checks for cross-module dependencies


Olshansk:
## Objective

Use custom CI linter checks to avoid cross-module dependencies.

## Origin Document

Even though the Pocket Network V1 implementation is a monolith that will shares artifacts, helpers, interfaces, structures, etc, across different modules, we want to avoid explicit dependency of one module on another and facilitate that communication via the application-specific bus.

In #235, the following conversation came up:

<img width="1419" alt="Screen Shot 2022-09-22 at 3 48 57 PM" src="https://user-images.githubusercontent.com/1892194/191866709-d97febae-96a2-4747-8bfc-979ae8d9ba33.png">

While we don't want to enable these linters right now, we want to enable them in the future.

Some examples include:
```bash
grep --exclude-dir='./persistence' -r "github.com/pokt-network/pocket/persistence"
grep --exclude-dir='./consensus' -r "github.com/pokt-network/pocket/consensus"
...
```

## Goals

- Enable automated ways of raising warnings on or blocking explicit cross-module (i.e. not via the application-specific bus)

## Deliverable

- [ ] Infrastructure that enables general-purpose linters based on the output of a bash command
- [ ] Documentation on how to add additional lintesr to the above

## Non-goals / Non-deliverables

- Enable these linters in mainline today

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)
- [ ] Update any relevant global documentation & references

---

**Creator**: @olshansky 
**Co-Owners**: @okdas

Olshansk:
Adding a few more to the list as our custom linter checks:

- Using `any` instead of `interface{}` -> needs to be changes
- Check if `#XXX` issue closed when trying to merge in a PR -> needs to be deleted
- Avoid using `proto` directly outside the `codec` package in favour of `coded.GetCodec()`

Olshansk:
Closing this out as unplanned for the foreseeable future. Ticket can be reopened if/when needed.

PR #453, opened 2023-01-20
[Persistence] Add logic for `updateParamsTree` and `updateFlagsTree` functions for the Sparse Merkle Tree

h5law:
## Description

This PR is in response to the comments left in the `peristence/state.go` file:

image

The functions work by utilising the new `getParamsUpdated(height int64)` and `getFlagsUpdated(height int64)` functions that return a slice of `[]*coreTypes.Param` and `[]*coreTypes.Flag` respectively. The function runs a query against the PostgreSQL DB to get all the parameter's/flag's at the given height, their names, values, and heights (+enabled status for flags) are then loaded from the returned rows into individual `Param` or `Flag` protobufs and these are then returned.

image

The `updateParamsTree` and `updateFlagsTree` functions then iterate over each protobuf in this slice and for each one runs the SMT `Update` function using the result of `crypto.SHA3Hash([]byte(proto.String()))` function (where `proto` is either `param` or `flag` depending on the function) as the key, and the marshalled protobuf as a `[]byte` as the value.

The corresponding hashes have been updated in the `TestStateHash_DeterministicStateWhenUpdatingAppStake` test, following the comment at the start of the function.

## Issue

Fixes N/A

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Create `Param` and `Flag` protobufs
- Create the `getParamsOrFlagsAtHeightQuery()` and `getParamsUpdated()` and `getFlagsUpdated()` functions to return the a slice of the respective protobufs for the given height
- Define the logic in the `updateParamsTree()` and `updateFlagsTree()` functions to utilise these functions and update their respective Merkle Trees
- Update the `TestStateHash_DeterministicStateWhenUpdatingAppStake` test hardcoded hashes to use new ones as the business logic has changed for calculating these
- Replace all uses of `proto.Marshal` with `codec.GetCodec().Marshal()` in `persistence/state.go`

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


persistence/account_shared_sql.go
```diff
@@ -99,3 +101,17 @@ func (p *PostgresContext) insertAccount(accountType types.ProtocolAccountSchema,
 	}
 	return nil
 }
+
+func (p *PostgresContext) getParamsOrFlagsAtHeightQuery(tableName string, height int64, descending bool) string {
```
Olshansk, line 105: I think `persistence/gov.go` would be the appropriate place for this function as well.

persistence/state.go
```diff
@@ -354,11 +354,39 @@ func (p *PostgresContext) updateTransactionsTree() error {
 }
 
 func (p *PostgresContext) updateParamsTree() error {
-	// TODO(#361): Create a core starter task to implement this
+	paramsOrFlags, err := p.getParamsOrFlags(types.ParamsTableName, p.Height, false)
```
Olshansk, line 357: Let's avoid having `state.go` have knowledge of the table names and.

1. We want to avoid having `state.go` have knowledge of implementation details such as the table name. 

2. Change these calls to `getParamsUpdate(height)` and `getFlagsUpadted(height)`.

3. Since the tree maintains a state, the order in which the results are returned should not affect the root hash, so we shouldn't need to pass that variable here.

persistence/types/gov.go
```diff
@@ -42,6 +43,36 @@ const (
 		)`
 )
 
+type ParamOrFlag struct {
```
Olshansk, line 46: The solution you have works, but I don't think it's great that the structure and serialization of what determines the state hash is embedded this deep in the code and is golang specific.

Fwiw, I don't think our current code structure is good at explaining this, but the idea is that we're using protobufs (a universal serialization codec) to maintain what the state hash is. This will enable other implementations (in other languages) to implement versions of the Pocket protocol as well.

1. See this diagram in the persistence spec for more details: https://github.com/pokt-network/pocket-network-protocol/tree/main/persistence#521-state-hash-computation

2. For now, can you create a new protobuf under `shared/core/types/proto` called `params.go`, and have a `message Param {...}` and `message Flag {...}`, so the "Hash" implementation you have below will be a proto marshalling of the object.

Olshansk (changes requested):
Definitely heading in the right direction and most of the logic is in place. Left a couple of major regarding next steps.

Let me know if you need any more clarifying details

persistence/types/gov.go
h5law: I think I have implemented this correctly but I've not used Protobufs before so my implementation could definitely be improved upon - I am using `[]byte(param.String())` and `proto.Marshal(param)` as the key and values in the tree now same for flag

h5law:
@Olshansk The local tests are passing - specifically the one with hard-coded hashes. However the Github test is not passing. There must be something wrong with how I am updating the merkle trees, am looking into this now

persistence/state.go
```diff
@@ -354,17 +354,18 @@ func (p *PostgresContext) updateTransactionsTree() error {
...
+	params, err := p.getParamsUpdated(p.Height)
 	if err != nil {
 		return err
 	}
 
-	for _, pf := range paramsOrFlags {
-		pfHash, err := pf.Hash()
+	for _, param := range params {
+		paramHash := crypto.SHA3Hash([]byte(param.String()))
+		paramBz, err := proto.Marshal(param)
```
Olshansk, line 364: Instead of `proto.Marshal`, look into calling `codec.GetCodec.Marshal(param)` from our `codec` library

persistence/state.go
```diff
@@ -354,17 +354,18 @@ func (p *PostgresContext) updateTransactionsTree() error {
...
-	paramsOrFlags, err := p.getParamsOrFlags(types.ParamsTableName, p.Height, false)
+	params, err := p.getParamsUpdated(p.Height)
 	if err != nil {
 		return err
 	}
 
-	for _, pf := range paramsOrFlags {
-		pfHash, err := pf.Hash()
+	for _, param := range params {
+		paramHash := crypto.SHA3Hash([]byte(param.String()))
```
Olshansk, line 363: s/paramHash/paramKey

This is the key used to index the state of that param.

`Merkle Trees` -> A data structure that can provide a succient argument (the root) of our entire state
`Key-Value Store` -> A data store engine backing the Merkle Tree. The path to every leaf is the key, and the value is the data.

Ditto for flag

persistence/gov.go
```diff
@@ -179,34 +180,62 @@ func getParamOrFlag[T int | string | []byte](p PostgresContext, tableName, param
...
+		if err != nil {
+			return nil, err
+		}
+		flag.Height = height
+		flagSlice = append(flagSlice, flag)
+	}
+	return flagSlice, nil
+}
+
+func (p *PostgresContext) getParamsOrFlagsAtHeightQuery(tableName string, height int64) string {
```
Olshansk, line 233: s/getParamsOrFlagsAtHeightQuery/getParamsOrFlagsUpdateAtHeightQuery

There's a difference between:
1. Getting all the param values at a certain height (i.e. all of them - the state)
2. Getting all the param values updated at a certain height (i.e. the delta)

persistence/state.go
h5law: After implementing this, locally tests are passing however the github action fails - I change the hashes to the ones the github workflow says it is - then the local tests fail saying the hashes should be the previous ones...

Looking into this now

h5law: @Olshansk Also in `persistence/state.go` should this also be using the codec marshaller? As all the other update tree functions are using `proto.Marshal` directly

h5law: > After implementing this, locally tests are passing however the github action fails - I change the hashes to the ones the github workflow says it is - then the local tests fail saying the hashes should be the previous ones...

Maybe the workflow was wrong? Passing now... still same hashes not changed them. My best guess was the local environment the workflow span up was dodgy?

Olshansk: > @Olshansk Also in persistence/state.go should this also be using the codec marshaller? As all the other update tree functions are using proto.Marshal directly

Yes! Do you mind changing all of them? Eventually we'll block the use of `proto.` directly outside of the codec library: https://github.com/pokt-network/pocket/issues/239#issuecomment-1405417782

> Maybe the workflow was wrong 

I'm not really sure what cause it to be flaky. Is it passing everywhere now?

Olshansk (approved):
Do you mind replacing the other `proto.Marshal` to `codec.GetCodec().Marhsal()`?

After that, feel free to merge!

persistence/state.go
h5law: @Olshansk yeah passing now locally and on GH - updated to use codec where proto was used directly doing changelogs then will merge 🫡

PR #457, opened 2023-01-23
[Tooling] changelog date verification: compare against last relevant commit

### LINKED ISSUE #456: [Tooling] Changelog verification date comparison
## Objective

The current (39302d6, at time of writing) [changelog verification job](https://github.com/pokt-network/pocket/blob/main/.github/workflows/changelog-verify.yml) compares the date of the most recent version in the changelog against the current date. This causes CI runs to fail if a PR sits in review for long enough such that when the next commit triggers CI, it's the next day.

## Origin Document

This behavior was introduced in #397.

## Goals

- Improve developer experience by eliminating a common case of CI failures.

## Deliverable

- The CI job should compare against the date of the most recent commit in the PR branch which affected the relevant files, as opposed to the current date. 
- The existing pre-commit hook should remain functionally the same.

## Testing Methodology

- [x] Manual testing via `git` CLI.

### Objectives:

- Ensure pre-commit hook does not permit a commit without a version in the respective changelog with today's date (i.e. still works the same)
- Ensure GitHub actions passes on a branch where the changelog version **is not** the current date but does match the date of the most recent commit which modified the respective files.

### Methodology

1. Ensure pre-commit hook date verification still works:
    ```
    # Checkout new branch off of main for testing
    git checkout main
    git pull --ff-only
    git checkout -b test/changelog-verification
  
    # Create and stage new file to trigger changelog verification
    touch ./runtime/deleteme
    git add ./runtime/deleteme
    
    # Ensure pre-commit hook is setup
    # (see https://github.com/pokt-network/pocket/blob/main/docs/development/README.md#prepare-local-environment)
    
    # Expect changelog verification to fail on pre-commit
    git commit -m "should not see me in git log"
    ```
    ```
    Missing changelog in module: runtime/
    
    Changelog verification failed. See error messages for more detail.
    ```
1. Update respective CHANGELOG with a past date:
    ```
    # In runtime/docs/CHANGELOG.md
    ...
    [0.0.0.9] - 2023-01-22
    ...
    ```
    ```
    # Stage respective CHANGELOG changes
    git add runtime/docs/CHANGELOG.md
    ```
1. Ensure pre-commit hook date prevents committing with past date:
    ```
    # Expect changelog verification to fail on pre-commit
    git commit -m "should not see me in git log"
    ```
    ```
    Latest date in runtime/docs/CHANGELOG.md is incorrect.
    Latest: 2023-01-22, Current: 2023-01-23
    
    Changelog verification failed. See error messages for more detail.
    ```
1. Construct a commit and changelog entry with a past date:
    ```
    # Commit, ignore pre-commit hook, and set date(s) to the past
    GIT_COMMITTER_DATE="2023-01-22 12:00" git commit -n -m "past commit" --date="2023.01.22 12:00"
    ```
1. Push _(recommend to your own fork)_.
    ```
    git push <remote> HEAD
    ```
1. Open a PR and wait for the changelog verification job to run, it should pass.
---

**Creator**: @bryanchriswhite

bryanchriswhite:
## Description

Duplicates and modifies the existing pre-commit hook to create a pre-push hook. This pre-push hook compares the CHANGELOG dates against the date of the most recent commit which affects the respective CHANGELOG's module.

## Issue

Fixes #456

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [x] Other: developer experience

## List of changes

- Derive pre-push hook from existing pre-commit which compares CHANGELOG version dates against the date of the last commit which affects the respective module, instead of the current date.
- Swap new pre-push hook out for pre-commit in changelog verification GitHub actions job.

## Testing

- [ ] `make develop_test`

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,140 @@
+#!/usr/bin/env bash
+
+# An example hook script to verify what is about to be pushed.  Called by "git
+# push" after it has checked the remote status, but before anything has been
+# pushed.  If this script exits with a non-zero status nothing will be pushed.
+#
+# This hook is called with the following parameters:
+#
+# $1 -- Name of the remote to which the push is being done
+# $2 -- URL to which the push is being done
+#
+# If pushing without using a named remote those arguments will be equal.
+#
+# Information about the commits which are being pushed is supplied as lines to
+# the standard input in the form:
+#
+#   <local ref> <local oid> <remote ref> <remote oid>
+#
+# This sample shows how to prevent push of commits where the log message starts
+# with "WIP" (work in progress).
+
+remote="$1"
+url="$2"
+
+BRANCH_CHANGED_FILES=($(git diff --name-only main..HEAD))
+
+# Initialise arrays and script wide variables
+IGNORE_DIRS=(".github" ".githooks" "docs" "bin")
+MODULES_EDITED=() # Modules are considered top-level directories directly under root
+MODULES_MISSING_CHANGELOG=()
+CHANGELOG_FILES=() # Full paths to CHANGELOG.md files
+CHANGELOG_ERRORS_FOUND=0
+
+# Loop through staged files
+for file in "${BRANCH_CHANGED_FILES[@]}"
+do
+    # Check the files contained in modules/top-level subdirs only
+    if [[ "${file}" =~ /+ ]]
+    then
+        module="$(echo "${file}" | sed 's/\/.*//')"
+        # Only add to modules list if not already present
+        if [[ ! "${MODULES_EDITED[*]}" =~ ${module} ]]
+        then
+            # Skip ignored modules if present
+            [[ ! "${IGNORE_DIRS[*]}" =~ ${module} ]] && MODULES_EDITED+=("${module}")
+        fi
+    # If not a subdir but root level CHANGELOG.md
+    elif [[ "${file}" == 'CHANGELOG.md' ]]
+    then
+        CHANGELOG_FILES+=("${file}") # Validate root changelog
+    fi
+done
+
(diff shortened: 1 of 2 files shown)


.githooks/pre-push
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
```
Olshansk, line 1: I reviewed the GH issue and only the comments at the top of this file, but not the code in place.

Doing a diff of `diff .githooks/pre-push .githooks/pre-commit`, seems like a lot of logic is shared, so I was wondering what your thoughts are on extracting the common logic/helpers into a separate bash file? The goal is to  have it potentially be reusable in our scripts/hooks in the future.

I realize this is a fair amount of extra work, so don't feel strongly about it.

Lmk what you think

Olshansk (changes requested):
One non-blocking comment. Lmk if you think it'll take too much time or overkill.

.githooks/pre-push
bryanchriswhite: Haha - indeed! I actually had to restrain myself from attempting a refactor as I was unsure about the portability of something like `declare -g ...` which was the first thing I would've reached for in a first pass. To avoid that, I think one would have to think more comprehensively about how the script could be broken down into more functional parts that are compatible with bash's calling conventions, and no doubt, consume some more reference material. As it's currently written, each loop modify multiple arrays declared in a common scope.

Olshansk: ❤️ this reasoning

Olshansk (approved):

PR #459, opened 2023-01-23
[Client Tooling] Keybase (issue #455)

### LINKED ISSUE #455: [Client Tooling] Keybase
## Objective

Miniature Keybase MVP functionality that allows accounts to interact with keys (e.g. import, create, export, etc...) 

## Origin Document

[V1 Roadmap: M1 Proof of Stake](https://github.com/pokt-network/pocket/milestone/7)

In order to achieve the M1 milestone - a mini-keybase implementation is needed to enable dynamic accounts and actors.

This keybase may be used for the following tasks:

- Signing blocks
- Signing transactions
- Signing arbitrary messages
- Signing multi-sig messages

### Research Resources

If you are unfamiliar with keybases you can gain high level context by reviewing these API specifications:
- https://keybase.io/docs/api/1.0
- https://keybase.io/docs/api/1.0/call/key/add

Key management exists as part of Pocket V0. It is recommended to install the CLI to see what commands exist as part of key management today. It's also recommended to review the current keybase implementation in the codebase.
- [Installing the Pocket-Core CLI](https://github.com/pokt-network/pocket-core/tree/staging/doc/specs/cli)
- [keys.go](https://github.com/pokt-network/pocket-core/blob/staging/crypto/keys.go)

As V0 is a ~2 yr old project, it is recommended to compare existing functionality against the latest industry standards to look for opportunities to modernize or enhance how keys are managed in V1
- [Ignite](https://github.com/ignite/cli/search?q=keybase) is using [Cosmos](https://github.com/cosmos/cosmos-sdk/tree/main/client/keys)

It's recommend that the assignee borrows what works from the above resources and integrates it with Pocket-specific utility.

### Technical Requirements

- Please use /shared/crypto/ed25519 as the PK Library
- Please use a filesystem.db for implementation, either Badger.db or Level.db

## Goals

- Evaluate existing V0 Keybase functionality using the CLI as well as industry approaches documented above
- Determine what can be imported into V1 and where Pocket-specific Utility needs to be applied
- Determine which filesystem.db should be used from options documented above
- Build a functional mini-keybase that can create, delete, import, export, and list and show actor keys 
- Document the functionality and key architecture of the mini-keybase

## Deliverables

- [ ] https://github.com/pokt-network/pocket/issues/150 
- [ ] https://github.com/pokt-network/pocket/issues/204
- [ ] https://github.com/pokt-network/pocket/issues/205

- Integrating Keybase with the CLI or other parts of the codebase 
- Implementing multisig 

---

h5law:
## Description

This is an implementation of a Keybase for the pocket V1 client.

It uses `BadgerDB` as the local database for the persistent storage of keys. The current `Keybase` interface has the following methods

image

These are covered with the following tests

image

The tests use an in-memory database created with `NewKeybaseInMemory()` whereas when exposed to the user the CLI should call `NewKeybase()` instead with a default path/user-supplied path. The CLI integration has not been implemented as part of this PR however.

### How it works

The keybase heavily relies upon `shared/crypto/ed25519.go` and the `PublicKey` and `PrivateKey` interfaces it exposes.

The keys are created using the methods from this library. Keys are then encrypted and armoured in the same fashion as in V0 and stored in the DB as a `KeyPair` struct (encoded to a `[]byte`).

image

This struct contains the public key and the encrypted JSON encoded private key string. This struct has the following methods:
 - `GetAddressBytes() []byte` -> returns `[]byte` of the public key address
 - `GetAddressString() string` -> returns the hex `string` of the public key address
 - `Unarmour(passphrase string) (crypto.PrivateKey, error)` -> returns the unencrypted unarmoured private key if passphrase is correct
 - `ExportString(passphrase string) (string, error)` -> returns raw private key string is passphrase is correct
 - `ExportJSON(passphrase string) (string, error)` -> returns json armoured private key string if the passphrase is correct

Keys are stored in the database in the following manner:
- The keys for the BadgerDB key-value storage are the `[]byte` value returned by `KeyPair.GetAddressBytes()` aka the `[]byte` address of the public key
- The values are the `[]byte` encoded `KeyPair` structs of the key

Key encryption/decryption works with or without a passphrase (when no passphrase is given `""` must still be passed to the function as the passphrase argument)

For ease of use the hex `string` returned by `KeyPair.GetAddressString()` is used to access the keys in storage.

Importing and exporting keys in either JSON string format or the private key hex string is fully interoperable between V1 and V0

Signing and Verification of messages works - and is covered with a Tx in the tests. I would need to look into the use cases for this more (for example multisig) as the current implementation is very rudimentary and in theory should work for producing a signature on any `[]byte` message but I would like to look into this more.

## Issue

Fixes #455 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

- Implement a keybase with BadgerDB local DB
- Add functionality to Import/Export keys, Get/List keys, Create new keys, Delete Keys, Sign Messages, Verify Signatures
- Add `make test_app` entry point in Makefile
- Add unit tests to cover Keybase use cases
- Add documentation for Keybase

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


h5law:
Thanks to @jessicadaugherty for giving me a hint to what might be the issue - HEX ENCODING!. V0's `ed25519` library has 2 extra functions for the `PrivateKey` type `RawBytes()` and `RawString()` I previously ignored these as they are not in the V1 library of the same name but they were the key to this issue.

Import/Export JSON is fully interoperable between V0 and V1 now: 

When encrypting the private key now convert the hex encoded private key string to `[]byte`

image

This is instead of directly getting the result of `PrivateKey.Bytes()` to match how V0 does the same task - although this extra conversion step does nothing for security it changes the length of the cipher text in the JSON encoded string.

Now when decrypting we must do another conversion/decoding step:

image

And the private key is then created from `bz`. When originally implementing this I thought this was weird and instead used the more direct approach (in my opinion) however this made the encrypted cipher text much shorter (despite representing the same values). This is now fully interoperable with V0 and is covered in a unit test with an account generated by `wallet.pokt.network` (manually testing the reverse works by `t.Log(jsonStr)` in the `ExportJSON` test works too)

app/client/keybase/keystore.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+	return &badgerKeybase{db: db}, nil
+}
+
+// Close the DB
+func (keybase *badgerKeybase) Stop() error {
+	return keybase.db.Close()
+}
+
+// Crate a new key and store it in the DB by encoding the KeyPair struct into a []byte
```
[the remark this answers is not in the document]
Olshansk: Similar typo in a couple of other places. Ditto elsewhere

app/client/keybase/keys.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func init() {
+	gob.Register(poktCrypto.Ed25519PublicKey{})
+	gob.Register(ed25519.PublicKey{})
+	gob.Register(KeyPair{})
+}
+
+// KeyPair struct stores the public key and the passphrase encrypted private key
+type KeyPair struct {
```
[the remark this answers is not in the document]
Olshansk: Also, could you create an interface for this so its easy to see the functions it exposes (for other devs in the future). I don't think we'll have alternative implementations, but mostly as a form of documentaiton

app/client/keybase/keystore.go
```diff
@@ -0,0 +1,391 @@
...
+			return err
+		}
+
+		// Use key address as key in DB
+		key := keyPair.GetAddressBytes()
+		if bytes.Compare(key, addrBz) != 0 {
+			return fmt.Errorf("Key address does not match previous address.")
+		}
+		// Encode entire KeyPair struct into []byte for value
+		bz := new(bytes.Buffer)
```
Olshansk, line 331: s/bz/keypairBz

These small nits go a long way when it comes to spending less time of other devs.

Just leaving one comment to avoid being pedantic, but use your best judgment elsewhere

app/client/keybase/keystore.go
```diff
@@ -0,0 +1,391 @@
...
+
+		return nil
+	})
+
+	return err
+}
+
+// Crate a new KeyPair from the private key hex string and store it in the DB by encoding the KeyPair struct into a []byte
+// Using the PublicKey.Address() return value as the key for storage
+func (keybase *badgerKeybase) ImportFromString(privStr, passphrase string) error {
```
Olshansk, line 119: s/privStr/privKeyHex

app/client/keybase/keystore.go
```diff
@@ -0,0 +1,391 @@
...
+	err := keybase.db.Update(func(tx *badger.Txn) error {
+		keyPair, err := CreateNewKeyFromString(privStr, passphrase)
+		if err != nil {
+			return err
+		}
+
+		// Use key address as key in DB
+		key := keyPair.GetAddressBytes()
```
Olshansk, line 127: s/key/addrKey

app/client/keybase/keys.go
```diff
@@ -0,0 +1,326 @@
...
+var (
+	// Errors
+	ErrorWrongPassphrase = errors.New("Can't decrypt private key: wrong passphrase")
+)
+
+func ErrorAddrNotFound(addr string) error {
+	return fmt.Errorf("No key found with address: %s", addr)
+}
+
+func init() {
```
Olshansk, line 42: I don't think you need these in the `keys.go` file

app/client/keybase/keybase_test.go
```diff
@@ -0,0 +1,467 @@
...
+	require.NoError(t, err)
+
+	privKey, err := db.GetPrivKey(testAddr, testPassphrase)
+	require.NoError(t, err)
+	require.Equal(t, privKey.Address().String(), testAddr)
+	require.Equal(t, privKey.String(), testPrivString)
+}
+
+func initDB() (Keybase, error) {
+	db, err := NewKeybaseInMemory("")
```
Olshansk, line 450: Shouldn't need the param here

app/client/keybase/keybase_test.go
```diff
@@ -0,0 +1,467 @@
...
+	err = db.ImportFromJSON(jsonStr, testPassphrase)
+	require.NoError(t, err)
+
+	privKey, err := db.GetPrivKey(testAddr, testPassphrase)
+	require.NoError(t, err)
+	require.Equal(t, privKey.Address().String(), testAddr)
+	require.Equal(t, privKey.String(), testPrivString)
+}
+
+func initDB() (Keybase, error) {
```
Olshansk, line 449: Consider passing in t *testing.T and running `require.NoError(t, err)` inside of the helper

app/client/keybase/keybase_test.go
```diff
@@ -0,0 +1,467 @@
...
+	db, err := NewKeybaseInMemory("")
+	if err != nil {
+		return nil, err
+	}
+	return db, nil
+}
+
+func createTestKey() (crypto.PrivateKey, error) {
+	return crypto.GeneratePrivateKey()
```
Olshansk, line 458: Optional: It's not a must here but look into `keygenerator.GetInstance()`. @deblasis put it together to create test key generation deterministic which can be very important for tests. 

Less so for these, but something to keep in mind.

app/client/keybase/keys.go
```diff
@@ -0,0 +1,326 @@
...
+	saltBytes = randBytes(16)
+
+	// Derive key for encryption, see: https://pkg.go.dev/golang.org/x/crypto/scrypt#Key
+	key, err := scrypt.Key([]byte(passphrase), saltBytes, n, r, p, klen)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Encrypt using AES
+	privKeyBytes := privKey.String()
```
Olshansk, line 207: Saw your comment here: https://github.com/pokt-network/pocket/pull/459#issuecomment-1402780661

Let's just make the var names appropriate.

Olshansk (changes requested):
This is really impressive work, in terms of quality, speed and impact.

Regarding your comment that fixes the V0 <-> V1 conversion. Tbh, I'm not a fan of all the data transformations, and I'm sure there's some redundancy since we ported some of the crypto library logic in v0.

If you see opportunities to clean it up in the future, just go for it.

giphy (1)

I will follow up in #455 regarding some next steps on this & other tickets.

app/client/keybase/keystore.go
```diff
@@ -0,0 +1,391 @@
...
+type Keybase interface {
+	// Close the DB connection
+	Stop() error
+
+	// Create new keypair entry in DB
+	Create(passphrase string) error
+	// Insert a new keypair from the private key hex string provided into the DB
+	ImportFromString(privStr, passphrase string) error
+	// Insert a new keypair from the JSON string of the encrypted private key into the DB
+	ImportFromJSON(jsonStr, passphrase string) error // TODO: Figure out why V0 keys are too long to import
```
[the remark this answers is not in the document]
h5law: I have addressed this and fixed the import issue the `testJSONString` + `testJSON...` constants are created with V0 and pass tests now

app/client/keybase/keybase_test.go
```diff
@@ -0,0 +1,463 @@
...
+	testPrivString = "045e8380086abc6f6e941d6fe47ca93b86723bc246ec8c4beee411b410028675ed78c49592f836f7a4d47d4fb6a0e6b19f07aebc201d005f6b2c6afe389086e9"
+	testPubString  = "ed78c49592f836f7a4d47d4fb6a0e6b19f07aebc201d005f6b2c6afe389086e9"
+	testAddr       = "26e16ccab7a898400022476332e2972b8199f2f9"
+	testJSON       = `{"kdf":"scrypt","salt":"AD165E014A42709A7EBF03594867C970","secparam":"12","hint":"","ciphertext":"BJ49gRklmHmnZP9cMBfExmpbQ7aGxyVgq/6ItDqbY5gFoses2+g3+aaHdgFv2PIWIAEojfqX3k1A6tBWF47Yw3OJ+vaTzje8sN/roKFJop4="}`
+	testPassphrase = "Testing@Testing123"
+
+	// Other
+	testNewPassphrase = "321gnitsetgnitset"
+	testTx            = "79fca587bbcfd5da86d73e1d849769017b1c91cc8177dec0fc0e3e0d345f2b35"
+	//testJSONString    = `{"kdf":"scrypt","salt":"6abd9619621dc8c2a62860c05dea6df3","secparam":"12","hint":"pocket wallet","ciphertext":"SZHQyzlYHke9g/7u/x1pCMPpA7wdeJSDjUN+9Oe5W0ep+y5enf2r2qVUnEgsZiQLin+t/kd5zS5DtgIS1IQwDrU3CR0ulbc7B5qy16oKkmE6lSCYO5MoLT1ixyenqNINvlmCwN2yaGpZ0LgDumI2pVAI1acjwXdNydG+Ph2m4RsoMwbEhftXN5LNekzjgbuV"}`
```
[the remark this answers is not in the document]
h5law: See comment on the ImportFromJSON method of interface - solved

app/client/keybase/keystore.go
```diff
@@ -0,0 +1,391 @@
+package keybase
```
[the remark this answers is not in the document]
h5law: Will write this up today 🫡

app/client/keybase/keybase_test.go
h5law: Have changed key generation to use this in the tests

h5law:
@Olshansk is it worth creating a `docs/` folder in the keybase directory to hold a `README.md` and a new `CHANGELOG.md` file as well? I am expecting the keybase will continue to be continued to develop on so a changelog may be relevant?

h5law:
@Olshansk I figured out that the passphrase issue was around `base64.StdEncoding` and have changed it to `base64.RawStdEncoding` now a key can be decrypted when `""` is used as the passphrase - this removes the need for a `defaultPassphrase` constant

h5law:
The [README](https://github.com/pokt-network/pocket/blob/keybase_implementation/app/client/keybase/README.md) is done. I would like some feedback on how its laid out etc, Ive never really wrote documentation for other people before so I tried to give as much detail as possible - but not a line by line code review as they can just read the code if they want the specifics

app/client/keybase/README.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+
+## Encryption and Armouring
+
+Whenever a new key is created or imported it is encrypted using the passphrase provided (this can be `""` for no passphrase).
+
+The packages used for the encryption and armouring of a key are: `golang.org/x/crypto/scrypt` (key generation), `crypto/aes`, `crypto/cipher` (AES256-GCM encryption cipher) as well as `encoding/hex`, `encoding/base64` and `encoding/json` for the armouring of the encrypted key.
+
+1. First the OS's randomness is used to generate a `[]byte` of length `16` which is used as the salt for generating the key.
```
Olshansk, line 171: It feels like there's too much detail here.

The details, libraries, functions, etc made me want to go and read through the code rather than the documentation.

The following isn't complete but I wanted to put together something simple as a starting point that would give a general idea of what how the implementation works:

Since this is just pseudocode, note that we have the freedom to play around where/when we add a certain amount of detail.

```mermaid
flowchart LR
    subgraph C[core lib]
        A["rand([16]byte)"]
    end
    subgraph S[scrypt lib]
        B["key(salt, pass, ...)"]
    end
    subgraph AES-GCM
        direction TB
        D["Cipher(key)"]
        E["GCM(block)"]
        F["Seal(plaintext, nonce)"]
        D--Block-->E
        E--Nonce-->F
    end
    C--Salt-->S
    S--Key-->AES-GCM
```

Screenshot 2023-01-27 at 8 07 06 PM

Olshansk (changes requested):
@h5law I didn't finish a re-review of the entire PR, but just wanted to share what I had as a starting point.

The library and implementation look to be in good shape but I think we can improve on the docs a bit.

Re default password - 💯 
Re docs & changelog - Let's do it

h5law:
@Olshansk I have left a reply to the `tx.Delete()` comment you made - personally I think it is better to stick to using the `Update()` transaction wrapper provided by `BadgerDB` as 1) It fits the pattern 2) it is less lines of code (it handles commit for us). I can change this if you really want the explicit `tx.Delete()` outside of a wrapper but this is why I didn't change it first time.

shared/crypto/README.md
```diff
@@ -0,0 +1,96 @@
...
+    C--Salt-->S
+    S--Key-->AES-GCM
+    AES-GCM--encryptedPrivateKey-->Armour
+    C--Salt-->Armour
+    kdf --> Armour
+    hint --> Armour
+    Armour--Marshal-->Return(encryptedArmouredPrivateKey)
+```
+
+The process above is reversed when unarmouring and decrypting a key in the keybase:
```
Olshansk, line 75: Could you look at the diagram I updated above and see if there's an opportunity to update the decryption one as well?

Olshansk (changes requested):
@h5law Almost over the finish line.

Left a couple small comments but also took the liberty to make some updates to the README while reading through it.

Take a look and fill free to update anything I missed or misrepresented.

Should be good to go afterwards.

app/client/keybase/keystore.go
```diff
@@ -335,7 +334,9 @@ func (keybase *badgerKeybase) Delete(address, passphrase string) error {
 	}
 
 	err = keybase.db.Update(func(tx *badger.Txn) error {
-		tx.Delete(addrBz)
+		if err := tx.Delete(addrBz); err != nil {
+			return err
+		}
 		return nil
```
Olshansk, line 340:
```suggestion
		return tx.Delete(addrBz)
```

app/client/keybase/keystore.go
```diff
@@ -25,9 +25,8 @@ type badgerKeybase struct {
 }
 
 // Creates/Opens the DB at the specified path
-// DISCUSS: Should this create the directory if it doesn't exist?
 func NewKeybase(path string) (Keybase, error) {
-	pathExists, err := dirExists(path)
+	pathExists, err := dirExists(path) // Creates path if it doesn't exist
```
Olshansk, line 29: Make sure to add this comment at the top of `dirExists` as well

Olshansk (approved):
One small comment but otherwise LGTM 🔥

PR #465, opened 2023-01-25
 [TECHDEBT] Update usages of `PersistenceRWContext` gov parameter accessors to use the more general `GetParameter()` function (Issue #436)

### LINKED ISSUE #436: [TECHDEBT] Update usages of `PersistenceRWContext` gov parameter accessors to use the more general `GetParameter()` function
## Objective

Update the different instances across `Persistence` and `Utility`, and anywhere else the `PersistenceRWContext` interface is used, to use the general `GetParameter(paramName string, height int64) (any, error)` function to access gov parameters.
 
This will include replacing the usages of `GetStringParam()`, `GetIntParam()` and `GetBytesParam()` and replacing them all with `GetParameter()` where this is appropriate. This is with the aim of reducing the code footprint across the modules affected.

Another part of this issue involves updating the logic of `GetParameter()` to automatically distinguish between `string` and `[]bytes` types given the `paramName` string. This must be done in a way that does not restrict the project from expanding the parameters without breaking the logic of this function.

## Origin Document

Since the merge of [PR#422](https://github.com/pokt-network/pocket/pull/422) the new generalised getter function `GetParameter()` has been implemented and has deprecated `GetBlocksPerSession()` and `GetServiceNodesPerSessionAt()`. However, all the other parameter accessor functions are not using this new generalised function yet.

## Goals

- Replace all gov parameter accessor functions using `GetStringParam()`, `GetIntParam()` or `GetBytesParam()` to use `GetParameter()`
- Update unit tests in `persistence/test/gov_test.go` and `utility/test/gov_test.go` to use the `GetParameter()` function
- Solve the issue around when to call the `[]bytes` variation of `GetParameter()` depending on the parameter name string

## Deliverable

- [ ] Update gov parameter accessors to use the `GetParameter()` function
- [ ] Update unit tests in `persistence/test/gov_test.go` and `utility/test/gov_test.go` to use the `GetParameter()` function where appropriate
- [ ] Update the logic of `GetParameter()` to detect `[]byte` types and distinguish them from `string` types automatically

- [ ] Restrict the expansion of the gov parameters by adding specific type detection rules to `GetParameter()` that would inhibit parameter additions that don't follow the _current_ patterns.

## General issue deliverables

## Testing Methodology

- [ ] **Task specific tests or benchmarks**: `make ...`
- [ ] **New tests or benchmarks**: `make ...`

---

**Creator**: @h5law

h5law:
## Description

This PR is a follow up from the changes merged in PR #422. It fully deprecates the `GetStringParam`, `GetIntParam` and `GetBytesParam` functions in favour of the more general `GetParameter` function. This is an aim at reducing the code footprint.

As interface functions cannot have type parameters and the return type of the parameters themselves cannot easily be determined from their name - or `pokt` tag/function return type. The `getByteArrayParam` helper function in `utility/gov.go` has been updated and a helper function `StringToBytes` has been created to make this task easier.

All usages in the codebase of the old functions are now using `GetParameter` and all unit tests are passing successfully.

## Issue

Fixes #436 

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Fully deprecate `GetIntParam`, `GetStringParam` and `GetBytesParam` in favour of `GetParameter`
- Update all usages and helper functions to use the `GetParameter` function where appropriate
- Add `StringToBytes` helper function and corresponding error message and error code

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


utility/types/util.go
```diff
@@ -40,3 +41,11 @@ func BigIntLessThan(a, b *big.Int) bool {
 	}
 	return false
 }
+
+func StringToBytes(s string) ([]byte, Error) {
```
Olshansk, line 45: s/StringToBytes/HexStringToBytes

utility/gov.go
```diff
@@ -559,46 +559,46 @@ func (u *UtilityContext) getBigIntParam(paramName string) (*big.Int, typesUtil.E
...
-	value, err := store.GetIntParam(paramName, height)
+	value, err := store.GetParameter(paramName, height)
 	if err != nil {
 		return typesUtil.ZeroInt, typesUtil.ErrGetParam(paramName, err)
 	}
-	return value, nil
+	return value.(int), nil
 }
 
 func (u *UtilityContext) getInt64Param(paramName string) (int64, typesUtil.Error) {
```
Olshansk, line 582: In these helpers `getInt64Params`, etc. We should validate if the type casting you're doing (e.g. `.(int)`) is valid and return an error if `!ok`.

Ref: https://stackoverflow.com/a/11580545/768439

utility/actor.go
```diff
@@ -149,11 +149,11 @@ func (u *UtilityContext) GetMaxPausedBlocks(actorType coreTypes.ActorType) (maxP
 		return 0, typesUtil.ErrUnknownActorType(actorType.String())
 	}
 
-	var er error
```
Olshansk, line 152: This is the only major issue I'm seeing in this PR.

I realize that part of the motivation was to remove `GetIntParam` and the other 2 functions, but now the onus of type casting is on the caller of `GetParamter`, which doesn't validate/check the cast at all.

What if we change `GetParameter` to accept a generic `[int | string | []byte]` and let the called specify it explicitly, so an error is returned if the cast fails?

Olshansk (changes requested):

utility/test/gov_test.go
```diff
@@ -87,7 +87,7 @@ func TestUtilityContext_GetBlocksPerSession(t *testing.T) {
 	ctx := NewTestingUtilityContext(t, 0)
 	defaultParams := DefaultTestingParams(t)
 	defaultParam := int(defaultParams.GetBlocksPerSession())
-	gotParam, err := ctx.GetParameter(typesUtil.BlocksPerSessionParamName, 0)
+	gotParam, err := ctx.GetBlocksPerSession()
```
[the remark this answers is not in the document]
h5law: I thought it was weird that all the other parameters had their own wrapper around a `GetParameter()` call and this was the only one without it - I will change it back but I thought it would be better to fit the pattern and just make the function a wrapper like the other parameters

utility/actor.go
h5law: @Olshansk I tried this but interface functions cannot have type parameters - this I thought would be the best work around, next to having a type casted wrapper for `GetParameter()` but this seems like a step backwards to what there was before.

Unless we have something like 
```go
func GetParam[T int | string | []byte](paramName string, height int64) (T, error)
```

utility/test/gov_test.go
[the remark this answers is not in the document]
Olshansk: Sgtm. Let's keep it

utility/actor.go
Olshansk: I don't have an immediate answer to this. It's one of the reasons I needed to think about #436 and search through the codebase to make sure no better/duplicate solutions exist.

I think you're getting pretty familiar with the layout, so I have two follow-up questions:

1. Does this duplicate work with  `func getParamOrFlag[T int | string | []byte]` in `gov.go`? Is there an opportunity to have logic consolidation there?

2. Do you think it makes sense to have the caller of `GetParam` be forced to type cast and check for errors?

Again, this isn't a loaded question, I just haven't had time to sit and go through all the flows to see what the cleanest path is.

h5law: So the `GetParam` function is probably a bad idea - I have just push a commit that removes the need for type casting by relying on the same helper functions in the utility context (now with added type casting checks) these are wrappers around `GetParameter()` returning solid types

h5law: @Olshansk regarding point 1 this would have probably just been a duplication of this function which is already wrapped by `GetParameter()` it would be better to avoid type casting and type parameter usages by the end user. However I do believe that whenever someone is using the parameters they will know the type they aim to be using - so the `utility/gov.go` helper functions can easily wrap the `GetParameter()` call handle any errors around type casting and shoot out the type they need at the current height

h5law: Quick thought: Change the `pokt:"val_type=..."` tags in the Protobuf to reflect the type the parameter actually is. This allows for the same method of encoding the data (`*big.Int`=`string`, `[]byte`=`string`, etc.). Currently the `pokt:"val_type=BIGINT"` refers to both `int64` and `*big.Int` (`persistence/types/gov.go`)

image

This would also mean the constants above would be changed to reflect the type in the switch statement - not altering any encoding but helping the parsing - I may be wrong here but I don't see a benefit of having a `*big.Int` val type be string and also a `[]byte` val type be a string and an `int64` val type be a `*big.Int`?

However this means the `GetParameter()` function could leverage `govParamMetadataMap` map built in `persistence/types/gov.go` (instead of using its own reflection of the functions return values) and accurately return the correct type of the parameter - then we would simply need the wrappers to do typecasting + error checking on this.

Whenever we return a generic and want to pass this into another function there must be a type-casting step, unless we allow the receiving functions to accept generics which is probably a bad idea. So with a generic function we should at least have it handle the type conversions (all parameters are stored as strings and the `GetParameter()` function only actually handles the `int` and `string` types due to its inability to accurately detect types) and any errors that may come from this

Olshansk: @h5law

**Side context**: I'm in the process of refactoring the utility module (#476) and revisiting (in the future) the entire `string/byte/BigInt` conversions is something I'd like to do. It'll require both thought and discussion, so I don't want that to be a blocker/distraction right now though.

The PR, as it stands right now, since to be a slight code health improvement.

Regarding your idea - I think it's a potential approach but have lost some context, and just don't see it as an immediate need/priority. What if we just create a ticket for the M5 milestone and revisit it then?

---

If you think this is something that should be done now though, to help/ease development, happy to discuss further.

Olshansk: Yea, I'd prefer we close it out and leave it as a reference in the ticket you created.

I think it requires a bit more thought on what we're trying to achieve here, which will become clearer once utility matures a bit.

PR #472, opened 2023-01-28
[Persistence] Add deterministic `Hash` functions to `Param` and `Flag` protobufs

h5law:
## Description

This PR is in response to and issue where the local `make test_all` command will pass and the GitHub worklow test will also pass, but at times will fail on the `TestStateHash_DeterministicStateWhenUpdatingAppStake` test.

image

The `String()` function generated by a protobuf to output its contents in a `string` format is not deterministic. As such the fields can appear in different orders. This would mean that the resulting hashing of these `string` values can differ if run numerous times.

When this function is used to update the `param` and `flags` trees their is a possibility that the hashes will be different to those hard-coded in the test case.

To fix this the marshalled protobuf's bytes are now being hashed using `crypto.SHA3Hash()` and used as the keys in the tree when updating. As these marshalled protobufs are using the `deterministic: true` option this will result in a deterministic hashing for the keys being used.

## Issue

Fixes N/A

## Type of change

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Use `crypto.SHA3Hash(protobufBz)` for the keys in the `Param` and `Flag` tree update methods
- Update hard-coded hashes in the `TestStateHash_DeterministicStateWhenUpdatingAppStake` test
- Use lowercase names for fields in the `Param` and `Flag` message definitions to fit convention

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


shared/core/types/proto/param.proto
```diff
@@ -5,14 +5,14 @@ package core;
 option go_package = "github.com/pokt-network/pocket/shared/core/types";
 
 message Param {
-    string Name = 1;
-    string Value = 2;
-    int64  Height = 3;
+    string name = 1;
```
Olshansk, line 8: Why did you make these lowercase?

shared/core/types/param.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+import (
+	"fmt"
+	"github.com/pokt-network/pocket/shared/crypto"
+)
+
+func (p Param) Hash() []byte {
```
Olshansk, line 8: In `codec.go`, the marshal function looks like so (note the `Deterministic: true`):

```go
func (p *ProtoCodec) Marshal(message proto.Message) ([]byte, error) {
	bz, err := proto.MarshalOptions{Deterministic: true}.Marshal(message)
	if err != nil {
		return nil, err
	}
	return bz, nil
}
```

Rather than "rolling our own" serialization for the key, what if the hash is just the sha of the serialized proto?

```go
paramBz, err := codec.GetCodec().Marshal(param)
paramKey := crypto.SHA3Hash(paramBz)
```

This should still be deterministic and result in less proprietary business logic.

Olshansk (changes requested):

shared/core/types/proto/param.proto
h5law: This was to match the pattern in the other protobuf files and the protobuf documentation - `words_like_this` are turned into `WordsLikeThis` when the go files are generated

shared/core/types/param.go
h5law: @Olshansk 🤦🏼‍♂️ This is much cleaner ahah

shared/core/types/proto/param.proto
Olshansk: You are correct!

protoc uppercases the vars when generating the Golang structs

Olshansk (approved):

PR #481, opened 2023-01-31
[Tooling] Add example of The CLI is meant to be an user but also a ma…

Olshansk:
## Description

Made the `p1` debug binary from localhost.

## Issue

NA

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Added `make build` to `make develop_start`
- Created a genesis file to be used from the host machine (not docker container)
- Added `localhost` and `workdir` to the `p1 debug` command so it can work not from within the docker container
- Updated the README with a bash helper other devs can use

## Testing

- [x] `make develop_test`

https://user-images.githubusercontent.com/1892194/215901991-076734e5-bc94-4755-9f2a-3d1f3c1e4aef.mov

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/app/client/cli/debug.go
+++ b/app/client/cli/debug.go
@@ -48,10 +48,17 @@ var (
 	// Its purpose is to allow the CLI to "discover" the nodes in the network. Since currently we don't have churn and we run nodes only in LocalNet, we can rely on the genesis state.
 	// HACK(#416): This is a temporary solution that guarantees backward compatibility while we implement peer discovery
 	validators []*coreTypes.Actor
+
+	// While the `p1` binary is in development, the debug commands require a config and a debug genesis file to operate.
+	// These currently live inside of the pocket repo so a workdir needs to be specified if `p1` is used from a directory
+	// other than the root of the pocket repo.
+	workdir string = ""
 )
 
 func init() {
-	rootCmd.AddCommand(NewDebugCommand())
+	debugCmd := NewDebugCommand()
+	debugCmd.Flags().StringVar(&workdir, "workdir", "", "workdir where the pocket repo is located, relative to which config & genesis files can be loaded")
+	rootCmd.AddCommand(debugCmd)
 }
 
 func NewDebugCommand() *cobra.Command {
@@ -61,7 +68,7 @@ func NewDebugCommand() *cobra.Command {
 		Args:  cobra.ExactArgs(0),
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			var err error
-			runtimeMgr := runtime.NewManagerFromFiles(defaultConfigPath, defaultGenesisPath, runtime.WithClientDebugMode(), runtime.WithRandomPK())
+			runtimeMgr := runtime.NewManagerFromFiles(workdir+defaultConfigPath, workdir+defaultGenesisPath, runtime.WithClientDebugMode(), runtime.WithRandomPK())
 
 			// HACK(#416): this is a temporary solution that guarantees backward compatibility while we implement peer discovery.
 			validators = runtimeMgr.GetGenesis().Validators
--- a/app/client/cli/debug.go
+++ b/app/client/cli/debug.go
@@ -25,8 +25,11 @@ const (
 	PromptTogglePacemakerMode    string = "TogglePacemakerMode"
 	PromptShowLatestBlockInStore string = "ShowLatestBlockInStore"
 
-	defaultConfigPath  = "build/config/config1.json"
-	defaultGenesisPath = "build/config/genesis.json"
+	defaultConfigPath = "build/config/config1.json"
+	// HACK: Note that `genesis.json` is a copy-pasta of `genesis_debug.json` with the only
+	// difference being that `node{X}.consensus:8080` is replaced with `localhost:808{X}` because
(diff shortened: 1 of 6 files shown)


app/client/cli/debug.go
```diff
@@ -25,8 +25,14 @@ const (
 	PromptTogglePacemakerMode    string = "TogglePacemakerMode"
 	PromptShowLatestBlockInStore string = "ShowLatestBlockInStore"
 
-	defaultConfigPath  = "build/config/config1.json"
+	defaultConfigPath = "build/config/config1.json"
```
okdas, line 28: I've changed this in LocalNet PR - https://github.com/pokt-network/pocket/blob/1261079d7bfe62b747809ad52470c3720888423f/app/client/cli/debug.go#L41 - and we could make use of environment variables instead of passing arguments. I can make adjustments/fix conflicts after the other PR is merged. We can keep both - envars and arguments, in that case I can make changes in LocalNet PR after this is merged.

app/client/cli/debug.go
```diff
@@ -25,8 +25,14 @@ const (
...
+	// Genesis file to be used from within a docker container to connect to other containers
 	defaultGenesisPath = "build/config/genesis.json"
+	// HACK: Genesis file to be used from the host machine to connect to docker containers.
+	// Note that `genesis.json` is a copy-pasta of `genesis_localhost.json` with the only
+	// difference being that `node{X}.consensus:8080` is replaced with `localhost:808{X}` because
+	// container names cannot be resolved by Docker's DNS from the host.
+	hostGenesisPath = "build/config/genesis_localhost.json"
```
okdas, line 35: In new LocalNet we do not expose these ports. We can, even with scaling actors dynamically, but I first want to understand what is the benefit of executing this binary directly versus in a container (what `make client_connect` already does)?

okdas: Similarly, this can be switched to an environment variable—happy to help and adjust the `p1` bash function, as we are going to hit a few git conflicts here anyway.

https://github.com/pokt-network/pocket/blob/1261079d7bfe62b747809ad52470c3720888423f/app/client/cli/debug.go#L42

build/docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.0.7] - 2023-01-31
+
+- Added `genesis_localhost.json`, a copy of `genesis.json` to be used by the localhost instead of a debug docker container
```
okdas, line 12: I do not see a `genesis_localhost.json` in this PR though, is this intentional?

docs/development/README.md
```diff
@@ -99,6 +99,23 @@ The cli binary will be available at `bin/p1` and can be used instead of `go run
...
 
+2.1 [OPTIONAL] Add the binary to your `.rc`
+
+You can add the following function so you can run the `p1` from anywhere on your host:
+
+```bash
+function p1 {
+    pocket_workdir="/Users/olshansky/workspace/pocket/pocket/"
+    if [ "$1" = "debug" ]; then
+        ${pocket_workdir}/bin/p1 debug --localhost=true --workdir="${pocket_workdir}"
```
okdas, line 110: If we opt-in for utilizing environment variables here, here is how this part will look like (not tested though - this is an example):
```bash
        POCKET_WORKDIR = "${HOME}/workspace/pocket/pocket/"
        CONFIG_PATH = "${POCKET_WORKDIR}/build/config/config1.json"
        GENESIS_PATH = "${POCKET_WORKDIR}/build/config/genesis_localhost.json
        ${POCKET_WORKDIR}/bin/p1 debug
```

okdas (commented):
Looks good! I have a few questions in the other comments, if you don't mind.

Just to provide some additional context, I wanted to explain why using environment variables might be a better option.

By using environment variables to configure file paths, the code becomes more flexible and adaptable, as it isn't limited to a specific execution environment. Changing the configuration or controlling code execution is easier using environment variables or arguments, and there's no need to recompile the binary if you want to modify the config file.

For example, with the current setup, if you want to use a different config file, you'll need to change the code and recompile the binary. But if this were controlled by environment variables, you could simply use the same binary and pass in a new configuration.

I understand that your solution works well for this case, but from my experience, it's always great to have control over simple things like configuration through environment variables or CLI arguments, just like what we have for `pocket` binary.

build/docs/CHANGELOG.md
Olshansk: Added

app/client/cli/debug.go
Olshansk: The `p1` binary has a `debug` option that is fundamentally broken right now. This was an attempt at fixing it

Screenshot 2023-02-02 at 8 36 18 PM

okdas (approved):
@Olshansk LGTM! This is not going to work with new LocalNet, and I can fix it, but first I'll need to learn about the use-case more. I suspect one of new make targets in other PR covers us but I'm not sure. :) A great stupid only question!

Olshansk:
@okdas Added an additional TODO and comment in the codebase

PR #500, opened 2023-02-08
[P2P] Initial libp2p integration

bryanchriswhite:
## Issue

Fixes #<issue_number>

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Change #1
- Change #2
- ...

## Testing

- [ ] `make develop_test`

## Required Checklist

### If Applicable Checklist


p2p/errors/error.go
```diff
@@ -0,0 +1,19 @@
+package errors
+
+import "fmt"
+
+// TODO: consolidate error helpers.
```
bryanchriswhite, line 5:
```suggestion
```

p2p/libp2p/identity/crypto.go
```diff
@@ -0,0 +1,22 @@
+package identity
+
+import (
+	"encoding/hex"
+	"github.com/libp2p/go-libp2p/core/crypto"
+	poktCrypto "github.com/pokt-network/pocket/shared/crypto"
+)
+
+// TODO: move
```
bryanchriswhite, line 9:
```suggestion
```

p2p/errors/error.go
```diff
@@ -0,0 +1,18 @@
...
+import "fmt"
+
+type ErrFactory func(msg string, err error) error
+
+func NewErrFactory(preMsg string) ErrFactory {
+	return func(msg string, err error) error {
+		msgStr := ""
+		if msg != "" {
+			// NB: with msg - "LibP2P module error: <msg>: <wrapped error>"
+			//  without msg - "LibP2P module error: <wrapped error>"
```
bryanchriswhite, line 12:
```suggestion
			// NB: with msg - "<preMsg>: <msg>: <wrapped error>"
			//  without msg - "<preMsg>: <wrapped error>"
```

p2p/errors/error.go
```diff
@@ -0,0 +1,18 @@
...
+			msgStr = fmt.Sprintf(": %s", msg)
+		}
+		// TODO: gracefully handle case(s) where msg and err is emtpy.
```
bryanchriswhite, line 15:
```suggestion
		// TODO: gracefully handle case(s) where preMsg, msg, and/or err are emtpy.
```

p2p/errors/error.go
bryanchriswhite: This is irrelevant, logging conventions are new and the codebase has yet to be made consistent in its usage.

p2p/errors/error.go
bryanchriswhite: Same as above.

p2p/libp2p/module.go
```diff
@@ -0,0 +1,344 @@
...
+	// TODO(future): investigate any unnecessary
+	// key exposure / duplication in memory
+	secretKey, err := identity.NewLibP2PPrivateKey(mod.cfg.PrivateKey)
+	if err != nil {
+		// TODO: wrap error
+		return nil, err
+	}
+
+	mod.identity = libp2p.Identity(secretKey)
+	// TODO: where is the hostname specified?
```
bryanchriswhite, line 115: I think it would make sense to allow node operators to specify the interface on which to listen. Does such a thing already exists and I just missed it?

bryanchriswhite:
Will fix the `hugeParam` when I'm back at my desk later. I didn't see any reason for that to delay review.

p2p/libp2p/module.go
```diff
@@ -0,0 +1,338 @@
...
+
+func Create(bus modules.Bus) (modules.Module, error) {
+	return CreateWithProviders(
+		bus,
+		persistence.NewPersistenceAddrBookProvider(bus),
+		bus.GetConsensusModule(),
+	)
+}
+
+func CreateWithProviders(bus modules.Bus, addrBookProvider addrbook_provider.AddrBookProvider, currentHeightProvider providers.CurrentHeightProvider) (modules.Module, error) {
```
Olshansk, line 75: consider mutlitlining so its easier to read

p2p/libp2p/module.go
```diff
@@ -0,0 +1,338 @@
...
+	return new(libp2pModule).CreateWithProviders(bus, addrBookProvider, currentHeightProvider)
+}
+
+func (mod *libp2pModule) GetModuleName() string {
+	// TODO: double check if this should change.
```
Olshansk, line 80: Since it seems like you're implementing t as a completely different module, we should have a different name (until they're consolidated)

p2p/common/error.go
```diff
@@ -0,0 +1,18 @@
+package common
+
+import "fmt"
+
+type ErrFactory func(msg string, err error) error
+
+func NewErrFactory(preMsg string) ErrFactory {
```
Olshansk, line 7: This is okay, but not a fan of introducing new patterns in just one specific module.

Would need one of the following:
1. TODO explaining we should adopt this pattern everywhere
2. A GitHub ticket or Github discussion where we do this
3. Just going with our general pattern for now:

```go
func ErrUnknownEventType(msg any) error {
	return fmt.Errorf("unknown event type: %v", msg)
}
```

p2p/libp2p/identity/crypto.go
```diff
@@ -0,0 +1,24 @@
+package identity
+
+import (
+	"encoding/hex"
+
+	"github.com/libp2p/go-libp2p/core/crypto"
+	poktCrypto "github.com/pokt-network/pocket/shared/crypto"
+)
+
+// NewLibP2PPrivateKey converts a hex-encoded ed25519d key
```
Olshansk, line 10: Should this be under `p2p` or under `shared/crypto` in your opinion?

p2p/libp2p/protocol/protocol.go
```diff
@@ -0,0 +1,9 @@
+package protocol
+
+import "github.com/libp2p/go-libp2p/core/protocol"
+
+// PoktProtocolID is the libp2p protocol ID matching current version of the pokt protocol.
+var PoktProtocolID = protocol.ID("pokt/v1.0.0")
+
+// DefaultTopicStr is a "default" pubsub topic string for use when subscribing.
+var DefaultTopicStr = "pokt/default"
```
Olshansk, line 9: 1. Why not constant?
2. Does this need to be exported (i.e. capital)
3. Use factoring to declare all the vars with one `var` or `const` statement

p2p/transport/libp2p.go
```diff
@@ -0,0 +1,39 @@
+package transport
```
Olshansk, line 1: The transport package contained the different "connection types" we have (i.e. empty and TCP).

I'm struggling to understand the code organization, but am thinking that if we have a libP2P transport implementing the transport interface, it should also be a connection type.

p2p/transport/libp2p.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/libp2p/go-libp2p/core/network"
+
+	"github.com/pokt-network/pocket/p2p/types"
+)
+
+type libP2PTransport struct {
+	stream network.Stream
+}
+
+var _ types.Transport = &libP2PTransport{}
```
Olshansk, line 15:
```suggestion
var _ types.Transport = &libP2PTransport{}

type libP2PTransport struct {
	stream network.Stream
}
```

p2p/libp2p/module.go
```diff
@@ -0,0 +1,338 @@
...
+	return modules.P2PModuleName
+}
+
+func (mod *libp2pModule) Create(bus modules.Bus) (modules.Module, error) {
+	return Create(bus)
+}
+
+func (mod *libp2pModule) CreateWithProviders(bus modules.Bus, addrBookProvider addrbook_provider.AddrBookProvider, currentHeightProvider providers.CurrentHeightProvider) (modules.Module, error) {
+	*mod = libp2pModule{
+		log:                   new(modules.Logger),
```
Olshansk, line 90: Please see the pattern we follow for loggers: https://github.com/pokt-network/pocket/wiki/Logger-Readme#logger-initialization

1. Name
2. Initialization with namespace
3. If we go with a pointer, we should add a TODO or update the documentation for that to be the case everywhere

p2p/libp2p/identity/identity.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/pokt-network/pocket/p2p/types"
+	poktCrypto "github.com/pokt-network/pocket/shared/crypto"
+)
+
+var (
+	ErrIdentity = common.NewErrFactory("")
+)
+
+// PoktPeerFromStream builds a pokt peer from a libp2p stream.
+// (NOTE: excludes `ServiceURL`)
```
Olshansk, line 25: 1. What does this note mean?
2. When do you pick between NOTE vs NB?

p2p/libp2p/identity/identity.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	poktPubKey, err := poktCrypto.NewPublicKeyFromBytes(remotePubKeyBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return &types.NetworkPeer{
+		Dialer:    transport.NewLibP2PTransport(stream),
+		PublicKey: poktPubKey,
+		// NB: pokt analogue of libp2p peer.ID
```
Olshansk, line 40: Maybe call this `TECHDEBT`? In our `Makefile we have a bunch of things you can search for (free free to update)

Screenshot 2023-02-13 at 8 31 55 PM

p2p/libp2p/identity/identity.go
```diff
@@ -0,0 +1,125 @@
...
+		Address:    poktPubKey.Address(),
+		ServiceUrl: stream.Conn().RemoteMultiaddr().String(),
```
Olshansk, line 42: Does this return a problem url?

p2p/libp2p/identity/identity.go
```diff
@@ -0,0 +1,125 @@
...
+	}, nil
+}
+
+// PubKeyFromPoktPeer retrieves the stdlib compatible public key from a pocket peer.
```
Olshansk, line 46: stdlib or p2plib?

p2p/libp2p/identity/identity.go
```diff
@@ -0,0 +1,125 @@
...
+func PubKeyFromPoktPeer(poktPeer *types.NetworkPeer) (crypto.PubKey, error) {
+	pubKey, err := crypto.UnmarshalEd25519PublicKey(poktPeer.PublicKey.Bytes())
+	if err != nil {
+		return nil, ErrIdentity("unable to unmarshal peer ed25519 public key", err)
+	}
+
+	return pubKey, nil
+}
+
+// PeerAddrInfoFromPoktPeer builds a libp2p AddrInfo which maps to the passed pcket peer.
```
Olshansk, line 56:
```suggestion
// PeerAddrInfoFromPoktPeer builds a libp2p AddrInfo which maps to the passed pocket peer.
```

p2p/libp2p/module.go
```diff
@@ -0,0 +1,338 @@
...
+	host                  host.Host
+	pubsub                *pubsub.PubSub
+	topic                 *pubsub.Topic
+	subscription          *pubsub.Subscription
+	network               typesP2P.Network
+}
+
+// NewReadStreamDeadline returns a future deadline
+// based on the read stream timeout duration.
+func NewReadStreamDeadline() time.Time {
```
Olshansk, line 51: Doesn't need to be exposed outside of this package - please lowercase.

Ditto other functions

p2p/libp2p/module.go
```diff
@@ -0,0 +1,338 @@
...
+		log:                   new(modules.Logger),
+		addrBookProvider:      addrBookProvider,
+		currentHeightProvider: currentHeightProvider,
+	}
+
+	if err := bus.RegisterModule(mod); err != nil {
+		return nil, ErrModule("unable to register module", err)
+	}
+
+	log.Println("Creating libp2p-backed network module")
```
Olshansk, line 99: Use the module's logger

p2p/libp2p/identity/identity.go
```diff
@@ -0,0 +1,125 @@
...
+	// NB: hard-code a scheme for URL parsing to work.
+	peerUrl, err := url.Parse("scheme://" + serviceURL)
+	if err != nil {
+		return nil, ErrIdentity(fmt.Sprintf(
+			"unable to parse peer service URL: %s", serviceURL,
+		), err)
+	}
+
+	// TODO: parameterize transport.
+	peerTransportStr := "tcp"
```
Olshansk, line 103: See `p2p/transport.go` - we already have a constant for this

p2p/libp2p/identity/identity.go
```diff
@@ -0,0 +1,125 @@
...
+	// NB: test if service URL hostname is an IP address or an FQDN
+	// NB: hard-code a scheme for URL parsing to work.
+	peerUrl, err := url.Parse("scheme://" + serviceURL)
+	if err != nil {
+		return nil, ErrIdentity(fmt.Sprintf(
+			"unable to parse peer service URL: %s", serviceURL,
+		), err)
+	}
+
+	// TODO: parameterize transport.
```
Olshansk, line 102: Isn't that what `connection type` is?

p2p/libp2p/identity/identity.go
```diff
@@ -0,0 +1,125 @@
+package identity
```
Olshansk, line 1: I like where this is going but had to google what a multiaddr was. Were you thinking of adding tests for this in the future?

p2p/libp2p/module.go
Olshansk: Search for `serviceUrlFormat`: `serviceUrlFormat  = "node%d.consensus:8080"`

consensus_port is set to `8080` in configs, but I just realize it's a bug that we don't use it

p2p/stdnetwork/libp2p.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+// TODO: factor out args which are common to network
+// implementations to an options or config struct.
+func NewLibp2pNetwork(
+	bus modules.Bus,
+	addrBookProvider providers.AddrBookProvider,
+	currentHeightProvider providers.CurrentHeightProvider,
+	log *modules.Logger,
+	host_ host.Host,
```
Olshansk, line 42: What does the _ suffix imply?

p2p/stdnetwork/libp2p.go
```diff
@@ -0,0 +1,184 @@
...
+	return nil
+}
+
+// TODO: fix typo in interface (?)
+func (p2pNet *libp2pNetwork) RemovePeerToAddrBook(peer *types.NetworkPeer) error {
+	delete(p2pNet.addrBookMap, peer.Address.String())
+	return nil
+}
+
+func (p2pNet *libp2pNetwork) GetBus() modules.Bus {
```
Olshansk, line 174: We won't need this after Alesasndro's PR. Just FYI

p2p/stdnetwork/libp2p.go
```diff
@@ -0,0 +1,184 @@
+package stdnetwork
```
Olshansk, line 1: I don't think stdnetwork is the right place here. It was meant for types that implement the `typesP2P.Network` interface (i.e. RainTreeNetwork, StdNetwork, etc...)

A separate packaged `p2p/libp2pnetwork` is probably best

Olshansk (changes requested):
@bryanchriswhite I took a first stab at this but will need to dive in deeper into the "meat of things". Mainly just left some questions and comments. 

The biggest things to keep in mind are:

1. Did you have a plan for this PR? Given that there are no tests and no documentation, I was thinking we could agree on an "MVP" here and then split it into smaller pieces of work (e.g. split out identity into its own PR w/ tests and an explanation).

2. I'm curious if you looked at any of the patterns of other go libp2p users? https://github.com/libp2p/go-libp2p#notable-users I briefly dove into celestia and flow but not enough to provide any actionable feedback.

3. Please try to look at some of the patterns we have elsewhere in the codebase. I think changing them is okay, but would rather stay consistent and then create a piece of work to change it all at once.

4. Did this work on LocalNet end-to-end?

5. Given that you're using some of the `network` abstractions, and that the implementation in the `p2p module` exposed interface is simple, I feel like there's a way to reuse the existing p2p module with libp2p as a connection type, so I was wondering why we passed on that approach?

If you think this is easier to discuss over a call. Happy to do so. I'll dive in deeper tomorrow

p2p/libp2p/module.go
bryanchriswhite: I did notice the port in the consensus config, thanks! I was referring specifically to being able to specify a hostname (i.e. `<hostname>:<port>`), in case we intend to support DNS resolution in production. Somewhat orthogonally, in cases where a node operator has multiple network interfaces (excl. loopback for this discussion), I think it would be prudent to allow a node operator to choose exactly which to expose. I didn't see any such degree of freedom in what I read so far of the current startup / config process. Perhaps this is related to some of the post-genesis address book work that I've seen references to?

It's not clear to me how the `serviceUrlFormat` piece connects as I only see references to test code when searching, could you elaborate a bit more?

p2p/common/error.go
```diff
@@ -0,0 +1,18 @@
+package common
```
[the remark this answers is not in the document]
bryanchriswhite: This is that file, I had to move it to avoid an import cycle. Sorry for the confusion.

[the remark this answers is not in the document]
bryanchriswhite: Thanks, I didn't see that there!

p2p/common/error.go
bryanchriswhite: Hehe - my intent was not to introduce a new pattern necessarily. I'm actually not especially proud of this and was hoping to get some feedback about what conventional error handling should look like.

p2p/libp2p/identity/crypto.go
bryanchriswhite: I don't think it really belongs in the "identity" concern but I was reluctant to put things into `shared/crypto` as it wasn't clear at the time of writing what pieces of code would be ephemeral and/or P2P specific. I think as it's currently written it would make more sense to move it. Thanks for calling that out! :raised_hands:

p2p/libp2p/protocol/protocol.go
bryanchriswhite: 1. My thinking was that perhaps we would want to set these with an `LDFLAGS=-X ...` at compile-time, which I thought required them to be `var`s but 1) I could be mistaken about that (see: [go tool link --help](https://pkg.go.dev/cmd/link)) and 2) don't have a strong opinion about it at this point.
2. This is all there is to the `protocol` package currently so to use it anywhere else, it would have to be exported, right? The only place it's currently used is in the libp2p module, are you suggesting to move it there?
3. :+1: thanks for calling that out!

p2p/libp2p/module.go
bryanchriswhite: Nice catch, thanks! :raised_hands:

p2p/libp2p/module.go
```diff
@@ -0,0 +1,338 @@
...
+	if err := bus.RegisterModule(mod); err != nil {
+		// TODO: wrap error
+		return nil, ErrModule("registering module", err)
+	}
+
+	return mod, nil
+}
+
+func (mod *libp2pModule) Start() error {
+	*mod.log = logger.Global.CreateLoggerForModule("P2P")
```
bryanchriswhite, line 132:
```suggestion
	*mod.log = logger.Global.CreateLoggerForModule(mod.GetModuleName())
```

p2p/libp2p/module.go
bryanchriswhite: I did read that one but see now that I missed a couple of details. Thanks for surfacing this!

The reason for the change to a pointer is (as @deblasis [pointed out](https://github.com/pokt-network/pocket/pull/502#issuecomment-1424590635)) is that the linter identifies it as a `hugeParam` otherwise.

More than happy to add a TODO or just update the docs if you prefer.

p2p/libp2p/identity/identity.go
bryanchriswhite: 1. My thinking is that we have competing representations for peers, "pokt" and "libp2p". In order to conform to our current interfaces, we need a pokt peer (i.e. [`types.NetworkPeer`](https://github.com/pokt-network/pocket/blob/main/p2p/types/network_peer.go#L7)) in a context where we only have a libp2p stream to construct one from.
2. Apparently when I'm tired :sweat_smile: - `NB` is a habit I picked up for short comments, I don't have a strong opinion about it.

p2p/libp2p/identity/identity.go
bryanchriswhite: Thanks for sharing that, I was not previously aware! :star_struck: 

My intention here was to point out that category theoretical equality of the two concepts. 

I'm not sure I see how it's `TECHDEBT`, are you trying to call out volatility due to the forthcoming consolidation of identity?

p2p/libp2p/identity/identity.go
bryanchriswhite: Nice catch - it should indeed be `libp2p`, thanks! :raised_hands:

p2p/libp2p/identity/identity.go
bryanchriswhite: I went back and forth on adding tests. After refactoring and looking at what's there, I got the impression we would either end up testing code that's already well tested (and just being composed) or would require extraneous effort to test (e.g. DNS resolution). Given the pace of iteration in this area and our current standing it seemed to me like the time was better spent elsewhere but would be more than happy to put more thought into a test plan if you feel like it would be prudent at this stage already.

bryanchriswhite: I hope to address any confusion around previously foreign concepts like multiaddr in forthcoming documentation updates.

bryanchriswhite:
Thanks @Olshansk! :raised_hands: 

> @bryanchriswhite I took a first stab at this but will need to dive in deeper into the "meat of things". Mainly just left some questions and comments.
> 
> The biggest things to keep in mind are:
> 
> 1. Did you have a plan for this PR? Given that there are no tests and no documentation, I was thinking we could agree on an "MVP" here and then split it into smaller pieces of work (e.g. split out identity into its own PR w/ tests and an explanation).

My main motivation was to deliver on just getting it working; this is that spike that we discussed. 

I agree that it would make sense to separate the identity piece, as we've discussed before. I think it might also make sense to separate the broadcast messaging piece and I would like to understand how to test that, if the current procedure doesn't already. There are some additional considerations around the using pubsub that we should discuss / think through.

Regarding tests, I didn't see any analogous tests for any of the other interface implementations, except for `rainTreeNetwork`, in which case the P2P module's `AddPeerToAddrBook` and `RemovePeerToAddrBook` methods are exercised. This seems like a reasonable thing to test for in the new module as well in my opinion and will add it. On the other hand, a lot of the new code is lightly wrapping well tested code from libp2p and I've been reluctant to spend time testing those things (e.g. identity and crypto helpers). I can add some tests for those pieces as well but they would be more regression tests than anything else, in my opinion. 

> 2. I'm curious if you looked at any of the patterns of other go libp2p users? https://github.com/libp2p/go-libp2p#notable-users I briefly dove into celestia and flow but not enough to provide any actionable feedback.

I have not but looked at anything outside of libp2p examples but that's a good idea, I'll take a look.

> 3. Please try to look at some of the patterns we have elsewhere in the codebase. I think changing them is okay, but would rather stay consistent and then create a piece of work to change it all at once.

I'll take a pass at the wiki. :+1: 

> 4. Did this work on LocalNet end-to-end?

I followed the ["LocalNet" steps](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md#basic-docker-compose) (still using docker compose) and it seems to be working. I'm not sure whether or not this exercises the pubsub piece as my understanding is still limited as to what all is and isn't going on in that context.

> 5. Given that you're using some of the `network` abstractions, and that the implementation in the `p2p module` exposed interface is simple, I feel like there's a way to reuse the existing p2p module with libp2p as a connection type, so I was wondering why we passed on that approach?

We could definitely merge the two modules but I think it would either detract from readability. While they do have some common structure and members, they are conceptually pretty different. Libp2p encapsulates concerns which are spread across the p2p package/subpackages and these new interface implemenations are better thought of as adapters to competing models in the libp2p regime.

> If you think this is easier to discuss over a call. Happy to do so. I'll dive in deeper tomorrow

Happy to as well, would you like to have a read through my responses first and let me know if / when you think it would be most beneficial?

shared/node.go
```diff
@@ -31,14 +32,20 @@ func CreateNode(bus modules.Bus) (modules.Module, error) {
 }
 
 func (m *Node) Create(bus modules.Bus) (modules.Module, error) {
+	p2pConfig := bus.GetRuntimeMgr().GetConfig().P2P
+	p2pCreate := p2p.Create
+	if p2pConfig.UseLibP2P {
```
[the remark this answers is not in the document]
bryanchriswhite: We could do something like this in `p2p/module.go`'s `Create` function but I'm not sure how I feel about it yet. I worry it's a bit indirect and could be confusing for the reader:

```
func Create(bus modules.Bus) (modules.Module, error) {
	p2pConfig := bus.GetRuntimeMgr().GetConfig().P2P
	if p2pConfig.UseLibP2P {
		return new(p2pModule).Create(bus)
	}

	return libp2p.Create(bus)
}
```

(This would DRY up the part in `app/client/cli/debug.go` that you were commenting on as well.)

p2p/libp2p/module.go
bryanchriswhite: My thinking was that the p2p modules are used mutually exclusively so I was expecting you to say it should be the same. Where does this present?

p2p/stdnetwork/libp2p.go
bryanchriswhite: I moved it into `p2p/libp2p` and named the file `network.go`. Does that seem reasonable to you?

p2p/common/error.go
bryanchriswhite: I feel like it's worth noting though that it did prove quite useful during debugging to have wrapped those errors in a string which I could search for. :sweat_smile:

bryanchriswhite:
I pulled the started separating the new module from the original one by making a new top-level directory and moving things over. As I got deeper it started getting more difficult and I ran out of time for the moment. There are still several types being from the original modules which are being referenced by the new one that need to be cleaned up.

libp2p/identity/identity.go
```diff
@@ -0,0 +1,134 @@
+package identity
```
Olshansk, line 1: 1. I really like the location of this now.
2. My biggest question, as I'm reading this, are "what functions are available". Try running `make go_doc`!

libp2p/identity/identity.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return peer.AddrInfo{
+		ID: peerID,
+		Addrs: []multiaddr.Multiaddr{
+			peerMultiaddr,
+		},
+	}, nil
+}
+
+// TECHDEBT: this probably belongs somewhere else, it's more of a networking helper.
```
Olshansk, line 95: Can you add a comment for it regardless?

libp2p/identity/identity.go
```diff
@@ -0,0 +1,134 @@
...
+func PeerMultiAddrFromServiceURL(serviceURL string) (multiaddr.Multiaddr, error) {
+	// NB: hard-code a scheme for URL parsing to work.
```
Olshansk, line 97: Do we always expect `scheme` to come from LibP2P service URLs?

If so: Move to const and add a comment
If not: add answer below

p2p/libp2p/module.go
Olshansk: Discussed offline but let me know if we need to reiterate here.

1. The `p2pConfig` exposes a port (e.g. 8080).
2. This port is where the node listens for consensus and p2p-specific events
3. Implicitly, we append the hostname to the `8080` port (e.g. `node1.consensus`) so we can send messages to another node based on their validator number
4. The `serviceUrl` also mimics (not enforced) what `(3):(1)` is based on the points above
5. The `serviceUrl` is defined via `generic_param` in `genesis.json`

NOTE: This is not done by design; this just happened and should change.

@okdas Have you put any thought into how these configs should be done?

p2p/stdnetwork/libp2p.go
Olshansk: Yes

p2p/libp2p/module.go
Olshansk: Feel free to resolve comments that don't need a second look. Gotta trust

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+var (
+	// TODO: consider security exposure of and "safe minimum" for timeout.
+	// TODO: parameterize and expose via config.
+	// ReadStreamTimeout is the duration to wait for a read operation on a
+	// stream to complete, after which the stream is closed ("timed out").
+	readStreamTimeoutDuration = time.Second * 10
+	// ErrModule wraps errors which occur within the libp2pModule implementation.
+	ErrModule = typesP2P.NewErrFactory("LibP2P module error")
```
Olshansk, line 56: does this need to be exposed outside of this package (i.e. uppercased)?

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+func CreateWithProviders(
+	bus modules.Bus,
+	addrBookProvider addrbook_provider.AddrBookProvider,
+	currentHeightProvider providers.CurrentHeightProvider,
+) (modules.Module, error) {
+	return new(libp2pModule).CreateWithProviders(bus, addrBookProvider, currentHeightProvider)
+}
+
+func (mod *libp2pModule) GetModuleName() string {
+	// TODO: double check if this should change.
```
Olshansk, line 76:
```suggestion
```

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+	return modules.P2PModuleName
+}
+
+func (mod *libp2pModule) Create(bus modules.Bus) (modules.Module, error) {
+	return Create(bus)
+}
+
+func (mod *libp2pModule) CreateWithProviders(bus modules.Bus, addrBookProvider addrbook_provider.AddrBookProvider, currentHeightProvider providers.CurrentHeightProvider) (modules.Module, error) {
```
Olshansk, line 84: multiline the method signature so its easier to read please

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+	*mod = libp2pModule{
+		logger:                new(modules.Logger),
```
Olshansk, line 86: We don't initialize a `new(modules.Logger)` anywhere in the codebase so you can remove this. We're creating a new one in `Start` regardless. 

Let it crash if the dev doesn't do it

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+		addrBookProvider:      addrBookProvider,
+		currentHeightProvider: currentHeightProvider,
+	}
+
+	if err := bus.RegisterModule(mod); err != nil {
+		return nil, ErrModule("unable to register module", err)
+	}
+
+	mod.logger.Print("Creating libp2p-backed network module")
```
Olshansk, line 95: I think using a global logger here is fair game

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+
+	mod.cfg = bus.GetRuntimeMgr().GetConfig().P2P
```
Olshansk, line 97: This, and some others, can be moved directly into the struct initialisation

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// TODO: support RainTree network
```
Olshansk, line 99:
```suggestion
	// INCOMPLETE: support RainTree network
```

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+	if mod.cfg.UseRainTree {
+		return nil, ErrModule("raintree is not yet compatible with libp2p", nil)
+	}
+
+	// TODO(future): investigate any unnecessary
```
Olshansk, line 104: TECHDEBT or TODO(M5) is more appropriate

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+	// key exposure / duplication in memory
+	secretKey, err := poktCrypto.NewLibP2PPrivateKey(mod.cfg.PrivateKey)
+	if err != nil {
+		// TODO: wrap error
+		return nil, err
+	}
+
+	mod.identity = libp2p.Identity(secretKey)
+
+	switch mod.cfg.ConnectionType {
```
Olshansk, line 114: This fit in really well!

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+	if err := bus.RegisterModule(mod); err != nil {
+		// TODO: wrap error
+		return nil, ErrModule("registering module", err)
+	}
+
+	return mod, nil
+}
+
+func (mod *libp2pModule) Start() error {
+	*mod.logger = logger.Global.CreateLoggerForModule("P2P")
```
Olshansk, line 139: Use the module name instead of `P2P`. Searching for `CreateLoggerForModule` in the codebase brings up some examples

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// IMPROVE: receive context in interface methods?
+	ctx := context.Background()
+
+	// TECHDEBT: metrics integration.
```
Olshansk, line 144: This is really great!

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		return ErrModule("unable to create pubsub", err)
+	}
+
+	mod.topic, err = mod.pubsub.Join(protocol.DefaultTopicStr)
+	if err != nil {
+		return ErrModule("unable to join pubsub topic", err)
+	}
+
+	mod.subscription, err = mod.topic.Subscribe()
```
Olshansk, line 178: Topic is where we publish and subscription is where we read? If so, maybe adding a small comment in the struct would help

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// TECHDEBT: metrics integration.
+	var err error
+	opts := []libp2p.Option{
+		mod.identity,
+		mod.listenAddrs,
+		// TECHDEBT / INCOMPLETE: add transport security!
+	}
+
+	if !mod.GetBus().GetRuntimeMgr().GetConfig().ClientDebugMode {
```
Olshansk, line 152:
```suggestion
        // Disabling some options for the local debug client because ...
	if !mod.GetBus().GetRuntimeMgr().GetConfig().ClientDebugMode {
```

shared/node.go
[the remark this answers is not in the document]
Olshansk: I think the `UseLibP2P` should be part of the base config and not the P2PConfig since its a competing module.

Theoretically, you could have a `LibP2PConfig` which would be a mostly copy-pasta which is how, theoretically, competing modules could do it in the future. I think we can stick with the same config for now, but move this out.

runtime/configs/proto/p2p_config.proto
```diff
@@ -7,10 +7,12 @@ import "connection.proto";
...
-  uint32 consensus_port = 2;
-  bool use_rain_tree = 3;
-  conn.ConnectionType connection_type = 4;
-  uint64 max_mempool_count = 5; // this is used to limit the number of nonces that can be stored in the mempool, after which a FIFO mechanism is used to remove the oldest nonces and make space for the new ones
-  bool is_client_only = 6;
+  string private_key = 1; // hex encoded
+  string hostname = 2;
+  uint32 consensus_port = 3;
+  bool use_rain_tree = 4;
+  bool use_lib_p2p = 5;
```
Olshansk, line 14: Mentinoed elsewhere, but I think this belongs in base?

shared/crypto/libp2p.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"encoding/hex"
+
+	"github.com/libp2p/go-libp2p/core/crypto"
+)
+
+// NewLibP2PPrivateKey converts a hex-encoded ed25519d key
+// string into a libp2p compatible Private Key.
+func NewLibP2PPrivateKey(hexString string) (crypto.PrivKey, error) {
```
Olshansk, line 11: In the identity package, you use `PoktX` as a prefix, but I think we should continue using `LibP2P` where appropriate and avoid using `Pokt` or `Pocket` unless we need to. It's implicit.

shared/crypto/libp2p.go
```diff
@@ -0,0 +1,23 @@
...
+	keyBytes, err := hex.DecodeString(hexString)
```
Olshansk, line 12: NIT: We got into the pattern of using `keyBz` for bz

libp2p/types/libp2p.go
```diff
@@ -0,0 +1,33 @@
...
+	host.Host
+}
+
+//type Message interface {
+//	// TODO: what?
+//}
+
+type Subscription interface {
+	Topic() string
+	Next(ctx context.Context) (interface{}, error)
```
Olshansk, line 23: s/interface{}/any

Ditto elsewhere

libp2p/types/libp2p.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+//type SubOpt interface {
+//	// TODO: what?
+//}
+
+type Topic interface {
```
Olshansk, line 30: A small sentence above each interface would really help.

Topic: LibP2P listeners can subscribe on topics, given  a set of options and its name
Subscription: A stream from which messages can be read.

libp2p/network/network.go
```diff
@@ -0,0 +1,219 @@
...
+type libp2pNetwork struct {
+	logger      *modules.Logger
+	bus         modules.Bus
+	host        host.Host
+	topic       *pubsub.Topic
+	addrBookMap types.AddrBookMap
+}
+
+var (
+	ErrNetwork = types.NewErrFactory("LibP2P network error")
```
Olshansk, line 28: reduce scope if possible

libp2p/network/network.go
```diff
@@ -0,0 +1,219 @@
...
+	DefaultPeerTTL = 2 * Year
+)
+
+// TECHDEBT: factor out args which are common to network
+// implementations to an options or config struct.
+func NewLibp2pNetwork(
+	bus modules.Bus,
+	addrBookProvider providers.AddrBookProvider,
+	currentHeightProvider providers.CurrentHeightProvider,
+	log *modules.Logger,
```
Olshansk, line 40: logger

libp2p/network/network.go
```diff
@@ -0,0 +1,219 @@
...
+	}, nil
+}
+
+// NetworkBroadcast uses the configured pubsub router to broadcast data to peers.
+func (p2pNet *libp2pNetwork) NetworkBroadcast(data []byte) error {
+	// IMPROVE: receive context in interface methods?
+	ctx := context.Background()
+
+	// NB: Routed send using pubsub
+	if err := p2pNet.topic.Publish(ctx, data); err != nil {
```
Olshansk, line 89: It's so simple!

libp2p/network/network.go
```diff
@@ -0,0 +1,219 @@
...
+			"writing to stream (peer address: %s)", poktAddr,
+		), err)
+	}
+	defer func() {
+		if err := stream.Close(); err != nil {
+			p2pNet.logger.Error().Err(err)
+		}
+	}()
+
+	// TECHDEBT: check if there's a more conventional way to send
```
Olshansk, line 143: Hmmm.. So topics let us publish to everyone (based on the configuration) but 1:1 comms require a stream?

libp2p/network/network.go
```diff
@@ -0,0 +1,219 @@
...
+	// messages directly than to close and re-open per direct message.
+	// NB: close the stream so that peer receives EOF.
+	if err := stream.Close(); err != nil {
```
Olshansk, line 146: You're already closing the stream via defer. Why do it here too?

libp2p/network/network.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+// TECHDEBT: factor out args which are common to network
+// implementations to an options or config struct.
+func NewLibp2pNetwork(
+	bus modules.Bus,
+	addrBookProvider providers.AddrBookProvider,
+	currentHeightProvider providers.CurrentHeightProvider,
+	log *modules.Logger,
+	host_ host.Host,
```
Olshansk, line 41: I would rename the import to `hostLibP2P` and just use `host` directly as a varname

libp2p/network/network.go
```diff
@@ -0,0 +1,219 @@
...
+	delete(p2pNet.addrBookMap, poktPeer.Address.String())
+
+	peer, err := identity.PeerAddrInfoFromPoktPeer(poktPeer)
+	if err != nil {
+		return ErrNetwork(fmt.Sprintf(
+			"converting peer info, pokt address: %s", poktPeer.Address,
+		), err)
+	}
+
+	p2pNet.host.Peerstore().RemovePeer(peer.ID)
```
Olshansk, line 205: Is the peerstore a conn pool?

libp2p/module.go
```diff
@@ -0,0 +1,358 @@
...
+	return identity.PeerMultiAddrFromServiceURL(fmt.Sprintf(
+		"%s:%d", mod.cfg.Hostname, mod.cfg.ConsensusPort,
+	))
+}
+
+// newReadStreamDeadline returns a future deadline
+// based on the read stream timeout duration.
+func newReadStreamDeadline() time.Time {
+	return time.Now().Add(readStreamTimeoutDuration)
+}
```
Olshansk, line 358: THe code handling all the streams is actually very clear!!

Olshansk (changes requested):
Took another round and overall, it's looking MUCH better! 🥂 

The code around the handling of the streams and topics is pretty. I would just add a comment in the structs so there's some context as you start tending to the code.

- Haven't reviewed the tests yet
- Feel free to `Resolve conversation` for trivial changes and I'll go back and continue the thread of conversation for those where I haven't followed up yet

okdas:
> @okdas Have you put any thought into how these configs should be done?

@Olshansk I sure did think of that part!

It does make sense to me to expose the port and hostname as the configuration parameter. Both needs to be adjusted for different environments. I do have a couple of notes on these two configs:
1. Port. We default the consensus port to `8080`, which is a common port for HTTP. In my opinion, we should change it to something less popular to cause less friction to our users. `8080` is often used for something else, and many people still avoid containerized solutions (that allow to remap ports easily). I am not sure if this PR is a good place, but I think changing this as a part of libp2p introduction makes sense. I would chose any Unassigned port from [RFC6335](https://www.rfc-editor.org/rfc/rfc6335.html), possibly containing [69](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=103) or [420](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=133), because, you know, memes. 42069 seems to be Unassigned btw.
2. Hostname. The value can not only be a valid DNS name, but IPv4/v6 as well. This is a necessary config because in some network topologies (think NAT) the IP address the node should be reached at from the internet is not known to the process. I am not sure though what value should be the default… Easy, or temporary, solution could be just a hostname from the environment (`os.Hostname()`). I did see a very cool pattern from other blockchain nodes - they integrate other 3rd party services to identify which hostname to “advertise” to the network. A good example is how Avax handles this: https://docs.avax.network/nodes/maintain/avalanchego-config-flags#public-ip - but we certainly can do that later! I can create a ticket for this (I suppose this is likely a North Star type of goal).

Also, here is an example on how to apply the configuration from docker-compose LocalNet to the LocalNet on Kubernetes: https://github.com/pokt-network/pocket/commit/ff185d16a4c242a27539df794ddc907068fc518c - @bryanchriswhite feel free to cherry-pick or [merge this branch](https://github.com/pokt-network/pocket/compare/chore/libp2p...chore/libp2p-localnet-configs).

libp2p/module.go
bryanchriswhite: Fair point, sorry for making you re-iterate on this one so much. I think in the case of error variables, my default thinking was to export them so that they can be used in assertions in other packages tests. I think there's value in having our tests make assertions about specific errors.

shared/node.go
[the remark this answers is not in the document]
bryanchriswhite: > I think the `UseLibP2P` should be part of the base config and not the P2PConfig since its a competing module.

Yes, thank you once again for the right context at the right time! :stuck_out_tongue_closed_eyes: 

> Theoretically, you could have a `LibP2PConfig` which would be a mostly copy-pasta which is how, theoretically, competing modules could do it in the future. I think we can stick with the same config for now, but move this out.

I had a similar thought but wasn't ready to commit yet either. Good to know we're on the same page! I'll add a comment and a mental note to do so before review in the future to better streamline these discussion. :sweat_smile:

p2p/libp2p/module.go
bryanchriswhite: Thanks for calling this out! I'll include it in the code review guidelines.

shared/crypto/libp2p.go
bryanchriswhite: Cool, is this a popular convention?

shared/crypto/libp2p.go
bryanchriswhite: I understand where you're coming from. My thinking was that 1) things which are currently converting between pocket and libp2p the analogues will be removed as we sort out the future of identity (because we shouldn't need them anymore) and 2) the libp2p based objects will become (or be lightly wrapped by - in future iterations -) the new "implicitly pocket" versions. **I'm getting onboard with perspective though**, I think it's more readable in the interim and I trust that you have the bigger picture in mind. :wink:

p2p/libp2p/module.go
bryanchriswhite: To add some detail to my thinking here, I would expect that since these modules are both implementations of the `P2PModule` interface, that it wouldn't make sense to duplicate `shared/modules/p2p_module.go` as part of isolating the 2 p2p modules from one another (i.e. I would expect then to share that interface). It just so happens that `P2PModuleName` is defined in the same place and the way I read that was something like "this is the abstract label for that kind of module".

p2p/libp2p/protocol/protocol.go
bryanchriswhite: @Olshansk did you have any more input on this point?

app/client/cli/debug.go
```diff
@@ -227,3 +233,27 @@ func sendDebugMessage(debugMsg *messaging.DebugMessage) {
...
+		mod, err = libp2p.CreateWithProviders(bus, addrBookProvider, currentHeightProvider)
+	} else {
+		mod, err = p2p.CreateWithProviders(bus, addrBookProvider, currentHeightProvider)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return mod.(modules.P2PModule), nil
+}
```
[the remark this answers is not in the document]
bryanchriswhite: This doesn't quite work as written because both p2p modules' `CreateWithProviders` functions returns a `modules.Module` interface which can't be assigned to the return variable directly because the return type is `modules.P2PModule`. I also do have a strong opinion against "naked returns" (as I've heard them called) as I was taught to avoid this when I was learning go. In my opinion, naked returns detract from readability and can become troublesome when copy/pasting or moving code such that it can negatively impact developer experience.

libp2p/identity/identity.go
```diff
@@ -0,0 +1,134 @@
+package identity
+
+import (
```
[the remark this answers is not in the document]
bryanchriswhite: Thanks for the hand-holding :sweat_smile:

libp2p/identity/identity.go
```diff
@@ -0,0 +1,134 @@
...
+
+var (
+	ErrIdentity = types.NewErrFactory("")
+)
+
+// PoktPeerFromStream builds a pokt peer from a libp2p stream.
+// (NOTE: excludes `ServiceURL`)
+func PoktPeerFromStream(stream network.Stream) (*types.NetworkPeer, error) {
+	remotePubKeyBytes, err := stream.Conn().RemotePublicKey().Raw()
+	// NB: abort handling this stream.
```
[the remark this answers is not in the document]
bryanchriswhite: This comment doesn't belong here (copy/pasta). I think It's the responsibility of the receiver of this error to close the stream, not this helper function.

libp2p/identity/identity.go
bryanchriswhite: I'm not sure if I followed the question exactly. My comment might not have been phrased clearly. To elaborate: we need to append something to fill the "scheme" space in the URL in order for the URL parser to be happy. The only examples I've looked at are the addresses in the genesis, none of which includes a scheme. I considered that we could use scheme to signal the expected transport to remote peers as it's currently only exposed to each node locally via their config (connection type). I figured that we would consider switching to multiaddrs, which would eliminate the need for that discussion.

libp2p/module.go
bryanchriswhite: (I continued the conversation in the previous thread - https://github.com/pokt-network/pocket/pull/500#discussion_r1108253634)

libp2p/types/libp2p.go
bryanchriswhite: Nice, I wasn't aware that was thing - thanks! This wasn't supposed to be committed though, I don't think we actually need it.

p2p/common/error.go
bryanchriswhite: Okay, I ended up going for option 2. I've added a `DISCUSS` to the code and opened #519.

libp2p/module.go
bryanchriswhite: I restructured things a bit, let me know if that's not what you had in mind.

libp2p/types/libp2p.go
bryanchriswhite: I added comments to the place where we're using these constructs from libp2p. This was part of bit of a misadventure of mine that I unintentionally committed.

libp2p/network/network.go
bryanchriswhite: Indeed. `pubsub` is a higher-level concept built on top of the hosts' streams as far as I understand.

bryanchriswhite: We could surely conceive of a way to use pubsub for 1:1 if we wanted to but I think it would be over-engineered which could make debugging more difficult. Depending on the properties that we want a "1:1 message exchange" to have, we may have to do additional work in the implementation but maybe also at the protocol level to ensure them. For example, in the most basic implementation, anyone may be able to learn the topic used for a "1:1 exchange". On the other hand, if we had a good reason to consider that route, it may worth taking a quick look to see if libp2p supports some kind of access control around topics or something. It does include options for message signing and verification so it's not too much of a stretch to imagine.

libp2p/module.go
```diff
@@ -0,0 +1,389 @@
...
+  - 2. Simplify libp2p module implementation
+  - - Network interface can be simplified
+  - - Consider renaming as it functions more like a "router"
+  - (NB: could be replaced in future iterations with a "raintree pubsub router")
+  - - Transport likely reduces to nothing
+  - 3. Remove "legacy" P2P module & rename libp2p module directory (possibly object names as well)
+  - - P2PModule interface can be simplified
+  - - Clean up TECHDEBT introduced in debug CLI and node startup
+*/
+package libp2p
```
bryanchriswhite, line 20: **NOTE**

libp2p/network/network.go
bryanchriswhite: (same rationale as https://github.com/pokt-network/pocket/pull/500#discussion_r1108124963)

p2p/libp2p/protocol/protocol.go
Olshansk: 1. My opinion is that iff we ever up needing it to be an `LDFLAG`, we can convert it to a var then, and make the flag is being used. Otherwise, please add a comment that says "this is a var and not a const because we might need it to be a flag one day".

2. I'm suggesting that you s/DefaultTopicStr/defaultTopicStr so that the developer knows "this is only being used in the context of this package"

3. 👍

p2p/libp2p/module.go
Olshansk: Please add a TODO in the logger module and/or the logger README

p2p/libp2p/module.go
Olshansk: Fair enough, let's remove the `// TODO: double check if this should change.`

p2p/libp2p/identity/identity.go
Olshansk: `We could of course convert the multiaddr to a proper URL here, if you think that would be a better approach.`

I think this makes more sense

app/client/cli/debug.go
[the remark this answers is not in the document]
Olshansk: I'll take it

libp2p/module.go
Olshansk: Please add a `// Exported for testing purposes` comment please

libp2p/network/network.go
Olshansk: Please add a `// Exported for testing purposes` comment please

libp2p/network/network.go
Olshansk: I know you were looking at some other codebases, so I think I'd be interested in learning what other chains using libp2p are using.

libp2p/network/network.go
Olshansk: Noted. Let's definitely build an understanding around these abstractions if we're going to be heavily dependant on them.

Olshansk:
@okdas I appreciate the detailed explanation and response!

1. Makes complete sense and seems @bryanchriswhite chose to go with that approach.

> I would chose any Unassigned port from [RFC6335](https://www.rfc-editor.org/rfc/rfc6335.html), possibly containing [69](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=103) or [420](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=133), because, you know, memes. 42069 seems to be Unassigned btw.

I support `42069`.

2. Regarding the 3rd party discovery service. Yup, let's create a ticket and assign it to the M5 milestone!

Olshansk (changes requested):
I reviewed everything except for `libp2p/network/network_test.go`, and I think this is looking REALLY good!!

The migration plan looks reasonable, and I think we'll find a few more along the way, but I suggest that we start splitting it and merging it piece by piece (creating tickets where necessary). 

Again, really impressed by how its turning out 👏

app/client/cli/debug.go
```diff
@@ -271,3 +275,19 @@ func fetchAddressBook(cmd *cobra.Command) (types.AddrBook, error) {
...
 }
+
+func getP2PModule(runtimeMgr *runtime.Manager) (p2pModule modules.P2PModule, err error) {
+	bus := runtimeMgr.GetBus()
+
+	var mod modules.Module
+	if runtimeMgr.GetConfig().UseLibP2P {
+		mod, err = libp2p.Create(bus)
+	} else {
+		mod, err = p2p.Create(bus)
```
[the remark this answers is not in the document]
bryanchriswhite: Already done - @deblasis and I resolved the merge conflicts together. :wink:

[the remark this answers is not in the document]
bryanchriswhite: Already done - @deblasis and I resolved the merge conflicts together. :wink:

libp2p/module.go
```diff
@@ -0,0 +1,389 @@
...
+	appMsgData, err := mod.network.HandleNetworkData(data)
+	if err != nil {
+		mod.logger.Error().Err(err).Msg("handling network data")
+		return
+	}
+
+	// There was no error, but we don't need to forward this to the app-specific bus.
+	// For example, the message has already been handled by the application.
+	if appMsgData == nil {
+		return
```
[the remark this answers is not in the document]
bryanchriswhite: This comment was copy/pasted from the original p2p module. Based on what it says, unless there's some specific detail that needs to be surfaced to the operator in this case, I think logging anything here would be unnecessarily verbose. I think we should prefer to log when the operator needs to take action, especially in dynamic places like this (as opposed to node startup, for example).

libp2p/network/network.go
```diff
@@ -0,0 +1,212 @@
...
+			), err)
+		}
+		libp2pPeer, err := identity.Libp2pAddrInfoFromPeer(peer)
+		if err != nil {
+			return nil, ErrNetwork(fmt.Sprintf(
+				"converting peer info, pokt address: %s", peer.Address,
+			), err)
+		}
+
+		host.Peerstore().AddAddrs(libp2pPeer.ID, libp2pPeer.Addrs, DefaultPeerTTL)
```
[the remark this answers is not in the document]
bryanchriswhite: Nope, guaranteed to succed apparently: https://pkg.go.dev/github.com/libp2p/go-libp2p/core/peerstore@v0.25.1#AddrBook.AddAddrs

libp2p/network/network.go
```diff
@@ -0,0 +1,212 @@
...
+	host libp2pHost.Host,
+	topic *pubsub.Topic,
+) (types.Network, error) {
+	addrBook, err := addrBookProvider.GetStakedAddrBookAtHeight(currentHeightProvider.CurrentHeight())
+	if err != nil {
+		logger.Fatal().Err(err).Msg("getting staked address book")
+	}
+
+	addrBookMap := make(types.AddrBookMap)
+	for _, peer := range addrBook {
```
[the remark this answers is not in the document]
bryanchriswhite: This will likely be completely redone in the very next steps as I add interfaces around the address book and peer  types.

libp2p/network/network.go
```diff
@@ -0,0 +1,212 @@
...
+	selfPoktAddr, err := p2pNet.GetBus().GetP2PModule().GetAddress()
+	if err != nil {
+		return ErrNetwork(fmt.Sprintf(
+			"sending to poktAddr: %s", poktAddr,
+		), err)
+	}
+
+	// NB: don't send to self.
+	if selfPoktAddr.Equals(poktAddr) {
+		return nil
```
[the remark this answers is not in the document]
bryanchriswhite: I would apply the same rationale here as with https://github.com/pokt-network/pocket/pull/500#discussion_r1112808400

libp2p/module.go
```diff
@@ -0,0 +1,389 @@
...
+	if err != nil {
+		mod.logger.Error().Err(err).Msg("reading from stream")
+		if err := stream.Close(); err != nil {
+			mod.logger.Error().Err(err)
+		}
+		// NB: abort this goroutine
+		// TODO: signal this somewhere?
+		return
+	}
+	defer func() {
```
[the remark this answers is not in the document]
bryanchriswhite: Nice one, thanks!

runtime/configs/proto/p2p_config.proto
```diff
@@ -7,10 +7,11 @@ import "connection.proto";
...
 message P2PConfig {
-  string private_key = 1;
-  uint32 consensus_port = 2;
-  bool use_rain_tree = 3;
-  conn.ConnectionType connection_type = 4;
-  uint64 max_mempool_count = 5; // this is used to limit the number of nonces that can be stored in the mempool, after which a FIFO mechanism is used to remove the oldest nonces and make space for the new ones
-  bool is_client_only = 6;
+  string private_key = 1; // hex encoded
+  string hostname = 2;
+  uint32 consensus_port = 3;
```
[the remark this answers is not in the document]
bryanchriswhite: Agreed, I was reluctant to introduce too much at once but can do! :smile: 

Option 1 would be my preference

p2p/libp2p/protocol/protocol.go
bryanchriswhite: We can continue the discussion in the relevant follow-on PR but I just wanted to clarify on point 2  to make sure we weren't talking past each other. Unless these are moved to the package where they're being used, they must be exported, no? 

I think moving them would be a reasonable suggestion; although, `PoktProtocolID` is being used in `module` and `network`, as currently written, so I think it must be exported. I anticipate that use of these by multiple packages would likely continue in the work to follow. While I agree that we should prefer to export things as necessary, in this case, for `DefaultTopicStr` I think keeping it close to `PoktProtocolID` improves maintainability and readability than moving it would.

libp2p/network/network.go
```diff
@@ -0,0 +1,212 @@
...
+	host        libp2pHost.Host
+	topic       *pubsub.Topic
+	addrBookMap types.AddrBookMap
+}
+
+var (
+	ErrNetwork = types.NewErrFactory("libp2p network error")
+	Year       = time.Hour * 24 * 365
+	// TECHDEBT: consider more carefully and parameterize.
+	DefaultPeerTTL = 2 * Year
```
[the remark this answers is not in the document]
bryanchriswhite: Apologies for the repetition. :sweat_smile: 1. There's a lot going on in the conversation here and 2. I keep managing to revert some changes as I stash experiments, etc.

Thanks for explaining the workflow, that's really helpful to understand (and now I can add this to my heuristics)! :raised_hands:

libp2p/module.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	switch mod.cfg.ConnectionType {
+	case types.ConnectionType_TCPConnection:
+		addr, err := mod.getMultiaddr()
+		if err != nil {
+			return nil, ErrModule("parsing multiaddr from config", err)
+		}
+		mod.listenAddrs = libp2p.ListenAddrs(addr)
+		logger.Global.Printf("listenAddrs: %s", addr)
```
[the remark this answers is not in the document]
bryanchriswhite: I'm not sure I followed. It looks like you improved my global logger call but them I'm confused by the second part. Are you suggesting moving logger setup back to create (and presumably updating the convetion)?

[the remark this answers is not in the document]
bryanchriswhite: (I'm also noticing that this specific line wasn't intended to be committed - it's a debug log that snuck in)

bryanchriswhite:
Closing in preparation for splitting.

Issue #489, opened 2023-02-03
[Tooling] Update CLI to support key management


Olshansk:
## Objective

Allow the user to manage keys via the CLI.

## Origin Document

This is a follow-up ticket #484. Specifically, the following deliverable:

> Documentation on what the default keys in the keybase are (until the CLI is added to support this)

## Goals

- Allow users to manage keys (create, delete, import, export, etc...) via the CLI
- Support all the necessary functions defined by the `Keybase` interface in `app/client/keybase/keybase.go`  via the CLI

## Deliverable

- [ ] Update the CLI (seen in the screenshot below) to have a `p1 Keys` namespace
- [ ] The `p1 Keys` namespace should have options to create, import, export, delete, etc commands
- [ ] Add the ability to have a default key in the keybase

<img width="1840" alt="Screenshot 2023-02-03 at 12 02 10 PM" src="https://user-images.githubusercontent.com/1892194/216701558-67090be7-53c3-4294-846d-2a82f2a808db.png">

## Non-goals / Non-deliverables

- Changes in any of the core node business logic

## General issue deliverables

- [ ] Update the appropriate CHANGELOG(s)
- [ ] Update any relevant local/global README(s)
- [ ] Update relevant source code tree explanations
- [ ] Add or update any relevant or supporting [mermaid](https://mermaid-js.github.io/mermaid/) diagrams

## Testing Methodology

- [ ] **All tests**: `make test_all`
- [ ] **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @Olshansk 
**Co-Owners**: @jessicadaugherty

Issue #150, opened 2022-08-05
[Keybase] Miniature Keybase Implementation to Enable M1 - Create and Delete


andrewnguyen22:
# Objective
Create a miniature Keybase that can create, store, and delete keys

# Origin Document
In order to achieve the M1 milestone a mini-keybase implementation is needed to enable dynamic accounts and actors.

This keybase may be used for the following tasks:
- Signing blocks
- Signing transactions
- Signing arbitrary messages
- Signing multi-sig messages

# Goals / Deliverables
Mini ED-25519 Keybase: 
- [ ] Code complete implementation of:
   - [ ] **CREATE** and store keys
   - [ ] **DELETE** keys
- [ ] Document the key architecture 

# Non-Goals / Deliverables
- Integrating Keybase with the CLI or other parts of the codebase
- Implementing multisig

## General issue checklist
- [ ] Update the appropriate CHANGELOG
- [ ] Update the README
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)
- [ ] Update any relevant global documentation & references
- [ ] Document small issues / TODOs along the way

---

Creator: @andrewnguyen22

andrewnguyen22:
**NOTE**: Please use the `/shared/crypto/ed25519` as the P/K library

andrewnguyen22:
**Recommend** to use a filesystem db for the implementation

Currently we are between `Badger.db` and `Level.db`

Please use one of these two options

Olshansk:
tl;dr @adshmh Could you or someone on your team take this on?

@andrewnguyen22 did a detailed job at explaining what we need from a basic keybase solution. This is a precursor that can be worked on in parallel with the CLI being built by @deblasis: https://github.com/pokt-network/pocket/pull/169/files

The keybase doesn't need to be anything fancy, and I think it can actually be _copy-pastad_ or extracted from V0, tendermint or ignite CLI? For example, just using [main/client/keys](https://github.com/cosmos/cosmos-sdk/tree/main/client/keys) and taking out the pieces we need.

Olshansk:
@Jasonyou1995 When you have time to tackle things, one thing I would suggest is doing a bit of research & learning about what a keybase is and how it works. For example, see the summary of the research @deblasis did on the state of RPC in the crypto industry here: https://github.com/pokt-network/pocket/discussions/213

[...]

PR #501, opened 2023-02-08
 [Tooling] Integrate Keybase w/ CLI (Issue #484 )

### LINKED ISSUE #484: [Tooling] Integrate Keybase w/ CLI
## Objective

Integrate the Keybase with the CLI to use keys from the keybase.

## Origin Document

In #459, the first version of the keybase was introduced as a separate package in the V1 codebase.

In #169, the first version of the CLI was introduced to the V1 codebase.

## Goals

- When sending commands via the CLI (stake, send, unstake, etc...), it should use keys from from the Keybase

## Deliverables

- [ ] Identify all entrypoints in `app/client/cli` where the keybase should be integrated
- [ ] Do a global search for `#150` and integrate all these entrypoint with the keybase (a subset of the above)
- [ ] Documentation on what the default keys in the keybase are (until the CLI is added to support this)

-  Adding new commands to the CLI to manage (create/import/export) keys

---

**Co-owner**: @Olshansk

h5law:
## Description

This PR addresses the `#150` comments throughout the `app/client/cli` directory and integrates the keybase to be used in place of reading keys from files locally.

The new method has been added to the `debug/keystore.go` file `InitialiseDebugKeybase()` which reads the 999 validator private keys from the `build/localnet/manifests/private-keys.yaml` file and stores them in the keybase. This is only run when there are less than 999 keys in the keybase. To speed this up `BadgerDB`'s `WriteBatch()` function is being used - however due to the encryption/JSON encoding and then Gob encoding steps of adding private keys to the keybase this takes a few seconds when ran for the first time.

This PR does not add any endpoints to interact with the keybase from the CLI this is instead tracked in the issue #489 

## Issue

Fixes #484 

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Add logic to initialise keybase with 999 validator keys
- Update `app/client/cli` to use the keybase

## Testing

- [x] `make develop_test`

[localnet_with_integration.webm](https://user-images.githubusercontent.com/53987565/217543043-21c35c3e-8389-44a5-968a-d3afb6cb8dd9.webm)

## Required Checklist

### If Applicable Checklist


app/client/keybase/keystore.go
```diff
@@ -47,6 +49,85 @@ func NewKeybaseInMemory() (Keybase, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(curAddr) < 999 {
+		// Use writebatch to speed up bulk insert
+		wb := db.NewWriteBatch()
+		for _, privHexString := range config.StringData {
+			// Import the keys into the keybase with no passphrase or hint
+			keyPair, err := crypto.CreateNewKeyFromString(privHexString, "", "")
```
Olshansk, line 106: Comment that hint and pwd are empty because these are debug keys

app/client/keybase/keystore.go
```diff
@@ -47,6 +49,85 @@ func NewKeybaseInMemory() (Keybase, error) {
...
+
+	// Add the keys if the keybase contains less than 999
+	curAddr, _, err := kb.GetAll()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(curAddr) < 999 {
+		// Use writebatch to speed up bulk insert
+		wb := db.NewWriteBatch()
```
Olshansk, line 103: good use of wb

app/client/keybase/keystore.go
```diff
@@ -47,6 +49,85 @@ func NewKeybaseInMemory() (Keybase, error) {
...
+	ApiVersion string            `yaml:"apiVersion"`
+	Kind       string            `yaml:"kind"`
+	MetaData   map[string]string `yaml:"metadata"`
+	Type       string            `yaml:"type"`
+	StringData map[string]string `yaml:"stringData"`
+}
+
+// Creates/Opens the DB and initialises the keys from the YAML file
+// FOR DEV/LOCANET PURPOSES ONLY
+func InitialiseKeybase(path string) (Keybase, error) {
```
Olshansk, line 62: I'd prefer to refactor this a bit. We are doing two things here:

1. Initialising a debug keybase for DEV purposes
2. Retrieving a keybase for PROD purposes

My request would be to move `InitialiseDebugKeybase` as a helper inside one of the `init` functions so it's a one-time check. Simply refer to that keybase in the normal code flow.

Olshansk (changes requested):
@h5law Two major comments:

1. I think there's some opportunity to reduce the amount of code duplication
2. To make the code clearer for other devs (and make future changes simpler), we should deliniate dev & prod functionality.

I gave some suggestions/guidelines in the comments but nothing too specific, so lmk if I should provide additional details.

app/client/keybase/keystore.go
h5law: @Olshansk I have added a `debug` folder inside the keybase and am initialising the DB at `$HOME/.pocket/keys` for the debug DB so this can be done in the `init()` function otherwise there would be no way to use `init()` with a path that could change.

app/client/cli/gov.go
```diff
@@ -76,5 +92,8 @@ func govCommands() []*cobra.Command {
 			},
 		},
 	}
+	for _, cmd := range cmds {
```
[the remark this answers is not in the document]
h5law: @Olshansk this command seems tightly coupled to be specific to the actor case as the `actorCmdDef` option part of the `applySubcommandOptions` function which uses the `attachPwdFlagToSubcommands` relies on the actor type being defined - it might be better to refactor this logic to be generally applicable to arbitrary commands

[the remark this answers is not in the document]
h5law: I changed the `applySubcommandOptions` to recieve only the options part of the `actorCmdDef` type so this can be used with the result of `attachPwdFlagToSubcommands` directly

app/client/cli/account.go
```diff
@@ -2,15 +2,14 @@ package cli
...
-
+	"github.com/pokt-network/pocket/app/client/keybase/debug"
 	"github.com/pokt-network/pocket/shared/crypto"
 	"github.com/pokt-network/pocket/utility/types"
 	"github.com/spf13/cobra"
 )
 
 func init() {
 	accountCmd := NewAccountCommand()
-	accountCmd.Flags().StringVar(&pwd, "pwd", "", "passphrase used by the cmd, non empty usage bypass interactive prompt")
```
Olshansk, line 13: Can you look into `cli/actor.go` and follow the same pattern w.r.t attaching the password params: 
1. Use options to set a pattern in case we need to add others
2. Don't embed it deep inside of `accountCommands` so its easier to find

```go
func init() {
	rootCmd.AddCommand(NewActorCommands(attachPwdFlagToSubcommands())...)

	rawChainCleanupRegex = regexp.MustCompile(rawChainCleanupExpr)

	oneMillion = big.NewInt(1e6)
}
```

app/client/cli/account.go
```diff
@@ -34,22 +33,31 @@ func accountCommands() []*cobra.Command {
...
+				fromAddrHex := args[0]
+				pk, err := kb.GetPrivKey(fromAddrHex, pwd)
+				if err != nil {
+					return err
+				}
+				if err := kb.Stop(); err != nil {
+					return err
+				}
+
+				fromAddr := crypto.AddressFromString(args[0])
```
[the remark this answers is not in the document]
Olshansk: Just TOL: Not going to ask for it here cause I think it's overkill, but unpacking the args into vars near the top could also help with clarity. Only do this if you personally feel it adds value.

app/client/cli/actor.go
```diff
@@ -86,17 +88,26 @@ A node can update relayChainIDs, serviceURI, and raise the stake amount with thi
...
+			fromAddrHex := args[0]
+			pk, err := kb.GetPrivKey(fromAddrHex, pwd)
+			if err != nil {
+				return err
+			}
+			if err := kb.Stop(); err != nil {
+				return err
+			}
+
+			fromAddr := crypto.AddressFromString(args[0])
```
Olshansk, line 110:
```suggestion
			fromAddr := crypto.AddressFromString(fromAddrHex)
```

Olshansk: Ditto below

app/client/cli/gov.go
```diff
@@ -76,5 +87,6 @@ func govCommands() []*cobra.Command {
 			},
 		},
 	}
+	applySubcommandOptions(cmds, attachPwdFlagToSubcommands())
```
Olshansk, line 90: Can you move this into `init` similar to the pattern in `actor.go`

app/client/cli/gov.go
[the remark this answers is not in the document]
Olshansk: Change to `applySubcommandOptions` makes sense to me

app/client/keybase/debug/keystore.go
```diff
@@ -0,0 +1,134 @@
...
+	"runtime"
+)
+
+const (
+	validatorNumber    = 999
+	debugKeybaseSuffix = "/.pocket/keys"
+)
+
+var (
+	DebugKeybasePath string
```
Olshansk, line 22: s/DebugKeybasePath/debugKeybasePath
```suggestion
         // TODO: Allow users to override this value via `datadir` flag
	debugKeybasePath string
```

app/client/keybase/debug/keystore.go
```diff
@@ -0,0 +1,134 @@
...
+		log.Fatalf("[ERROR] Cannot initialise the keybase with the validator keys: %s", err.Error())
+	}
+}
+
+// Creates/Opens the DB at `$HOME/.pocket/keys`
+func NewDebugKeybase() (keybase.Keybase, error) {
+	return keybase.NewKeybase(DebugKeybasePath)
+}
+
+// Struct to store the private-keys yaml file
```
Olshansk, line 43:
```suggestion
// Struct to process the yaml file of pre-generated private-keys
```

app/client/keybase/debug/keystore.go
```diff
@@ -0,0 +1,134 @@
...
+type yamlConfig struct {
+	ApiVersion string            `yaml:"apiVersion"`
+	Kind       string            `yaml:"kind"`
+	MetaData   map[string]string `yaml:"metadata"`
+	Type       string            `yaml:"type"`
+	StringData map[string]string `yaml:"stringData"`
+}
+
+// Creates/Opens the DB and initialises the keys from the YAML file
```
Olshansk, line 52:
```suggestion
// Creates/Opens the DB and initialises the keys from the pre-generated YAML file of private keys
```

app/client/keybase/debug/keystore.go
```diff
@@ -0,0 +1,134 @@
...
+// FOR DEV/LOCANET PURPOSES ONLY
+func InitialiseDebugKeybase() error {
+	// Get private keys from manifest file
+	_, current, _, _ := runtime.Caller(0)
+	//nolint:gocritic // Use path to find private-keys yaml file from being called in any location in the repo
+	yamlFile := filepath.Join(current, "../../../../../build/localnet/manifests/private-keys.yaml")
```
Olshansk, line 58: Move to a const

app/client/cli/account.go
h5law: @Olshansk without adding excess code the `gov.go` and `account.go` files don't need I feel this is the best solution:

```go
func NewAccountCommand() *cobra.Command {
	cmd := &cobra.Command{
		Use:     "Account",
		Short:   "Account specific commands",
		Aliases: []string{"account"},
		Args:    cobra.ExactArgs(0),
	}

	cmds := accountCommands()
	applySubcommandOptions(cmds, attachPwdFlagToSubcommands())
	cmd.AddCommand(cmds...)

	return cmd
}
```

similar with the gov one.

The way the actor one works is quite different from the others as it uses the `actorCmdDef` struct which is unique to this CLI endpoint - the others don't have something like this and its definitely not needed (as the code is now) for them either so they shouldn't go in the `init()` function as it would require changes that I don't really think are of any benefit to the structure of the other endpoints. Happy to do it if you think this move is not enough though.

app/client/keybase/keystore.go
Olshansk: It's a step in the right direction, but was hoping to take it further.

At the moment, every function in our PROD CLI starts with a `NewDebugKeybase()` and I think we can make it really prod-ready in this PR.

```
		RunE: func(cmd *cobra.Command, args []string) error {
			// Open the debug keybase at $HOME/.pocket/keys
			kb, err := debug.NewDebugKeybase()
```

The `init` function gets called when we import something. What I'm thinking is:

1. Create a file that contains an anonymous import of the debug package
2. By importing it, we immediately load the debug DB if its empty (we can exclude this from production binaries)
3. Replace the `kb` assignment with a prod ready assignment

app/client/keybase/debug/keystore.go
```diff
@@ -0,0 +1,134 @@
...
+type yamlConfig struct {
+	ApiVersion string            `yaml:"apiVersion"`
+	Kind       string            `yaml:"kind"`
+	MetaData   map[string]string `yaml:"metadata"`
+	Type       string            `yaml:"type"`
+	StringData map[string]string `yaml:"stringData"`
+}
+
+// Creates/Opens the DB and initialises the keys from the YAML file
+// FOR DEV/LOCANET PURPOSES ONLY
```
Olshansk, line 53: This is in the path name and method name, can be removed now.

app/client/keybase/keybase.go
```diff
@@ -1,9 +1,15 @@
...
-import "github.com/pokt-network/pocket/shared/crypto"
+import (
+	"github.com/dgraph-io/badger/v3"
+	"github.com/pokt-network/pocket/shared/crypto"
+)
 
 // Keybase interface implements the CRUD operations for the keybase
 type Keybase interface {
+	// Debug
+	GetDB() *badger.DB
```
Olshansk, line 11:
```suggestion
        // Debug
	// TECHDEBT: Exposed only so we can leverage badger's batch writing features when populating the development keybase
	GetBadgerDB() *badger.DB
```

Olshansk (changes requested):
@h5law Would like to iterate on this a bit (see my comments) so we have a "prod ready" integration between the CLI and the keybase, and not need to revisit this in the future.

Lmk if I can provide any more details.

h5law:
@Olshansk The CLI endpoints are now using "prod ready" code - however the debug keybase will only initialise to `$HOME/.pocket/keys` and I am unsure of a way to make this changeable with `--data_dir` as it is ran from the `init()` function - unless the datadir is binded with viper and used in the debug package?

Regarding the anonymous import I think I have followed what you asked for - not sure what you meant about making an empty file so I just did an anonymous import of the debug package in the `cmd.go` file as this works how you want it to.

app/client/cli/account.go
Olshansk: Great explanation and thanks for putting so much thought into it. The current solution LGTM

app/client/keybase/debug/keystore.go
```diff
@@ -0,0 +1,131 @@
...
+	"path/filepath"
+	"runtime"
+
+	"github.com/pokt-network/pocket/app/client/keybase"
+	"github.com/pokt-network/pocket/shared/crypto"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	validatorNumber     = 999
```
Olshansk, line 18: Rename to `numValidators` or `validatorCount` and add a comment `// The number of pre-generated keys in privateKeysYamlFile`

app/client/cli/actor.go
```diff
@@ -86,18 +90,34 @@ A node can update relayChainIDs, serviceURI, and raise the stake amount with thi
...
+		Args: cobra.ExactArgs(4),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// TODO(#150): update when we have keybase
-			pk, err := readEd25519PrivateKeyFromFile(privateKeyFilePath)
+			// Unpack CLI arguments
+			fromAddrHex := args[0]
+			fromAddr := crypto.AddressFromString(args[0])
+			amount := args[1]
+
+			// Open the keybase at the specified directory
```
Olshansk, line 100: Not a debug anymore?

app/client/cli/cmd.go
```diff
@@ -2,21 +2,31 @@ package cli
 
 import (
 	"context"
+	"log"
+	"os"
 
+	_ "github.com/pokt-network/pocket/app/client/keybase/debug"
```
Olshansk, line 8:
```suggestion
        // NOTE: Imported for debug purposes in order to populate the keybase with the pre-generated keys
	_ "github.com/pokt-network/pocket/app/client/keybase/debug"
```

Olshansk (approved):
@h5law Leaving a couple of minor last-minute nits but approving so you can merge right after.

Thanks for bearing on the back & forth, but I honestly think that final outcome is really good and much closer to being prod ready.

app/client/cli/actor.go
h5law: @Olshansk no this now uses the prod keybase - however while we do the anonymous import the prod keybase at `$HOME/.pocket/keys` is being initialised with the 999 keys. Remove the import and it has no default keys and its using the normal prod ready DB

Olshansk: 💯

PR #491, opened 2023-02-06
KISS Peer discovery, KISS Bootstrapping, Nodes auto-stake/unstake via K8S cluster-manager, Node Finite state machine - Issues #416, #490, #429, #498, #499

deblasis:
## Description

This PR aims at implementing a KISS implementation of Peer discovery #416 

TL;DR: I believe that:
**This** + **State sync** = **DevNet** 🚀

I still need to add the endpoints for accessing the addressbooks but this PR in the meantime shows the auto-Staking/Unstaking functionality.

It works by dogfooding our own CLI which also works as a way for testing the functionality and the "feel" of the tooling we are building. 

### Areas of improvement

- Since we can scale up to 999 validators, I had to consider their accounts in genesis and load them up with POKT, that means that our genesis files are quite big, not an issue per se but I don't particularly love it.
- It would be nice to have ways to interact with the `cluster-manager` from the debug CLI, for example to turn it on/off dynamically and also to scale up and down without the need for updating the `localnet-config.yaml`
- 

## Issue

Fixes #416 
Fixes #490 
Fixes #429 
Fixes #498 
Fixes #499 

## Type of change

Please mark the relevant option(s):

- [x] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [x] Major breaking change
- [ ] Documentation
- [ ] Other 

### CLI

- Added `not_interactive` flag to allow for non-interactive `Stake` and `Unstake` transactions (dogfooding in `cluster-manager`)
- Updated CLI to use to source the address book and the current height from the RPC server leveraging the `rpcAddressBookProvider` and `rpcCurrentHeightProvider` respectively and the `bus` for dependency injection

## Node binary

- Updated to handle `bootstrap-nodes` flag that can be used to override the default bootstrap nodes

### Infra

- Updated genesis to include accounts for all the validators that we can use in LocalNet
- Updated `docker-compose` to name the deployment as `pocket-v1` instead of `deployments` (default is the containing folder name)
- Introduced the `cluster-manager` that is a standalone microservice in the K8S LocalNet that takes care of (for now) staking/unstaking automatically nodes that are added/removed from the deployment
- Updated manifests and K8S resources to reflect the new `cluster-manager` addition
- In K8S LocalNet, the `cli-client` now waits for the `v1-validator001` since its required for address book sourcing

### Consensus

- Modules embed `base_modules.IntegratableModule` and `base_modules.InterruptableModule` for DRYness
- Updated modules `Create` to accept generic options
[...]

PR #502, opened 2023-02-09
[P2P] Logger module registration in p2p and network bugfix

deblasis:
## Description

We paired on this and we identified and fixed some issues that prevented `stdnetwork` from working (when turned on setting `use_rain_tree: false`

image

## Issue

No issue because it was "trivial", maybe very small 👕 🙂

## Type of change

- [x] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Updated logging initialization and passing to the network component instead of using the global logger
- Fixed incorrect use of `bus.GetLoggerModule()` in `stdnetwork.go` since it's never initialized when running the debug CLI
- `loggerModule` now embeds `modules.Module` making it registerable

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


logger/module.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 type loggerModule struct {
+	modules.Module
```
Olshansk, line 14: We're no longer type checking that `var _ modules.Modules = &loggerModule{}`. Should we create a `RegistrableModule` for this use case?

p2p/stdnetwork/network.go
```diff
@@ -20,21 +20,21 @@ var (
 type network struct {
 	addrBookMap typesP2P.AddrBookMap
 
-	logger modules.Logger
+	logger zerolog.Logger
```
Olshansk, line 23: We have our `modules.Logger` abstraction so we don't need to expose `zerolog` anywhere outside of `logger`. I can't find `logger zerolog.Logger` anywhere else in the code. 

Can we avoid this

Olshansk (changes requested):
@deblasis I did a quick review and raised a couple questions.

I realize this fixed the bug in place, but it also broke several patterns through the codebase in terms of exposing `zerlog` and the use `Module` interface.

My question is:
1. Do you think this is a reasonable long-term solution we should leave as is?
2. Do you think we should merge this in and fix it in a follow-up commit?
3. Is there an issue with the abstractions/design we have that we should tend to here?

Note that I didn't provide a solution since it would take a little bit of time to think through so happy to do so if need be.

p2p/stdnetwork/network.go
deblasis: We noticed that and we were confused by the fact that VSCode showed this: 

image

Already fixed :)

logger/module.go
deblasis: I wasn't really familiar with the logging implementation, I am getting my feet wet now :) I now understand that we don't need to register it because it's globally available and then within the `Start()` of each model we branch out.

This embed is not necessary anymore.

For consistency I have added 

```go
var _ modules.Module = &loggerModule{}
```

deblasis:
> @deblasis I did a quick review and raised a couple questions.

🙏

> I realize this fixed the bug in place, but it also broke several patterns through the codebase in terms of exposing `zerlog` and the use `Module` interface.
> My question is:
> 
> 1. Do you think this is a reasonable long-term solution we should leave as is?
> 2. Do you think we should merge this in and fix it in a follow-up commit?

VSCode confused me and yeah it looked weird to both of us initially. I just wanted to unblock @bryanchriswhite and quickly.
I have pushed updates PTAL 

> 3. Is there an issue with the abstractions/design we have that we should tend to here?

Not sure yet, I need to take a deeper look, I have noticed that in some cases I wasn't seeing logs but it might be simply misconfiguration/mistake on my end in my PR #491

One blatant thing I have noticed so far is that within the modules we are using a value reference for `logger` and not a pointer.
I suppose it should be a pointer, even the linter suggests that passing that value around is a big one. 

```
logger is heavy (89 bytes); consider passing it
by pointer (gocritic)
```

For consistency I'd apply the change across the board. since it would 100% cause linting issues to anyone that tries to pass the logger inside a function or sub-module if that makes sense.


> Note that I didn't provide a solution since it would take a little bit of time to think through so happy to do so if need be.

p2p/stdnetwork/network.go
```diff
@@ -20,21 +20,22 @@ var (
 type network struct {
 	addrBookMap typesP2P.AddrBookMap
 
-	logger modules.Logger
+	logger *modules.Logger
 }
 
 func NewNetwork(bus modules.Bus, addrBookProvider providers.AddrBookProvider, currentHeightProvider providers.CurrentHeightProvider) (n typesP2P.Network) {
+	logger := logger.Global.CreateLoggerForModule("network")
```
Olshansk, line 27: Rename local var to `log` or `networkLogger` to avoid overshadowing the import.

Ditto elsewhere.

Screenshot 2023-02-09 at 10 29 29 AM

p2p/stdnetwork/network.go
```diff
@@ -20,21 +20,22 @@ var (
 type network struct {
 	addrBookMap typesP2P.AddrBookMap
 
-	logger modules.Logger
+	logger *modules.Logger
```
Olshansk, line 23: The documentation instructs to use the value (not a pointer) to the logger: https://github.com/pokt-network/pocket/wiki/Logger-Readme#module-logging

Since it's a struct (not an interface), I see where you're coming from, but we should probably update the docs & other instances (can be done in a followup commit) if we're going down this approach.

Wdyt?

Olshansk (changes requested):

p2p/stdnetwork/network.go
deblasis: Ok, I took a deeper look and here are my 2c:

## First cent 🪙

Thinking about it, I see little to no value in having an instance of a logger inside a struct. Please correct me if I am missing something. 

this:
```go
type network struct {
	addrBookMap typesP2P.AddrBookMap

	logger modules.Logger
}

func NewNetwork(bus modules.Bus, addrBookProvider providers.AddrBookProvider, currentHeightProvider providers.CurrentHeightProvider) (n typesP2P.Network) {
	networkLogger := logger.Global.CreateLoggerForModule("network")
	networkLogger.Info().Msg("Initializing stdnetwork")

	addrBook, err := addrBookProvider.GetStakedAddrBookAtHeight(currentHeightProvider.CurrentHeight())
	if err != nil {
		networkLogger.Fatal().Err(err).Msg("Error getting addrBook")
	}

	addrBookMap := make(typesP2P.AddrBookMap)
	for _, peer := range addrBook {
		addrBookMap[peer.Address.String()] = peer
	}
	return &network{
		logger:      networkLogger,
		addrBookMap: addrBookMap,
	}
}
```

is probably simpler if it becomes:

```go
var (
	_ typesP2P.Network           = &network{}
	_ modules.IntegratableModule = &network{}
	// 👇👇👇👇👇
	moduleLogger = logger.Global.CreateLoggerForModule("network")
	// 👆👆👆👆👆
)

type network struct {
	addrBookMap typesP2P.AddrBookMap
}

func NewNetwork(bus modules.Bus, addrBookProvider providers.AddrBookProvider, currentHeightProvider providers.CurrentHeightProvider) (n typesP2P.Network) {
	moduleLogger.Info().Msg("Initializing stdnetwork")

	addrBook, err := addrBookProvider.GetStakedAddrBookAtHeight(currentHeightProvider.CurrentHeight())
	if err != nil {
		moduleLogger.Fatal().Err(err).Msg("Error getting addrBook")
	}

	addrBookMap := make(typesP2P.AddrBookMap)
	for _, peer := range addrBook {
		addrBookMap[peer.Address.String()] = peer
	}
	return &network{
		addrBookMap: addrBookMap,
	}
}
```

we would have the `moduleLogger` (or different/better name) defined once inside a package and within that package we would have a zeroLog sublogger that can be used across the board, bypassing any issue in terms of value/reference passing in a "2 birds with one stone way".

Unless we have packages where we have multiple modules, I don't see any drawback with this approach.

As it stands, the `LoggerModule` is not used like the other modules since the module is used to configure something that is subsequently used using a global reference, not something that comes from the bus. 

For simplicity sake, I'd keep it like this (as in keep using logger.Global) but since we are using a global that's public and used across all modules, why not also use a sublogger that's global, **within** the package?

I hope this makes sense.

## Second cent 🪙
Also, I have noticed a `UpdateFields` method used in consensus 

```go
func (m *consensusModule) setLogPrefix(logPrefix string) {
	logger.Global.UpdateFields(map[string]any{
		"kind": logPrefix,
	})
	// INVESTIGATE: Do we need to create a new logger here?
	m.logger = logger.Global.CreateLoggerForModule("consensus")
}
```

It looks wrong, in here I would use the `moduleLogger` approach from above and then instead of calling `setLogPrefix` I would use the built-in zeroLog `With()`: https://github.com/rs/zerolog#sub-loggers-let-you-chain-loggers-with-additional-context


### Suggested plan

1) Merge this now that it doesn't break patterns but fixes an issue that was blocking @bryanchriswhite 
2) Create an issue from this comment
3) Quick discussion if needed
4) Take action / or not depending on 3)

p2p/module.go
```diff
@@ -116,13 +116,15 @@ func (m *p2pModule) GetModuleName() string {
...
-	logger.Global.Info().Msg("Starting network module")
+	m.logger = logger.Global.CreateLoggerForModule(m.GetModuleName())
+	m.logger.Info().Msg("Starting network module")
 
 	addrbookProvider := getAddrBookProvider(m)
 	currentHeightProvider := getCurrentHeightProvider(m)
 
 	cfg := m.GetBus().GetRuntimeMgr().GetConfig()
 
+	// TODO: pass down logger
```
Olshansk, line 127:
```suggestion
	// TODO: propogate `p2pModule.logger` to the underlying network
```

p2p/stdnetwork/network.go
Olshansk: > Merge this now that it doesn't break patterns but fixes an issue that was blocking @bryanchriswhite

🆗 

>  I would use the built-in zeroLog With()

👍

> Create an issue from this comment

✅ 

> For consistency I'd apply the change across the board. since it would 100% cause linting issues to anyone that tries to pass the logger inside a function or sub-module if that makes sense.

🧠 

> Take action / or not depending on 3)

You are essentially changing the `moduleLogger` to a `packageLogger` in the context of zerlog's `sublogger`. I see where you're coming from, but it still makes sense to be to have the logger be part of the module.

Example:
1. Things that come from an "instance" of one of our modules, I would want to be namespace (i.e. use the logger).
2. A package might have a "static" function, where we'd be forced to use the global logger rather than the namespace one.
3. If we want to instantiate a module and override it with a noop logger, that'd be possible. 

I don't feel really strongly about this so am okay going with either approach, but am still personally leaning towards the embedded logger approach.

Olshansk (approved):
Thanks for helping unblock Bryan!

Let's merge this in and continue the convo in https://github.com/pokt-network/pocket/pull/502#discussion_r1102167581. 

giphy

p2p/module.go
deblasis: This TODO is now redundant since the logger is created inside the network components.

@bryanchriswhite, in your upcoming PR, would you mind simply removing this TODO? 🙏

PR #510, opened 2023-02-14
[WIP] SLIP-0010 HD Child Key Generation

h5law:
## Description

This PR contains an implementation of [SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) which allows for deterministic child key generation from a master key. This is done by using the seed of the master key and creating a `hmac` hash using the `sha512` hashing function. The hash is split into 2 `[]byte` slices representing the `SecretKey` and the `ChainCode`.

This `hmac` key can then be used to deterministically generate keys within the BIP-44 path `m/44'/635'/%d'/`. `44` referring to the BIP-44 hardened key purpose of the path and `635'` is the coin type for Pocket Network. `%d` can be any number between `0-2147483647` inclusive. Allowing for a large number of child keys to be deterministically generated and retrieved by index from a single master key's seed.

_NOTE_ All the child keys are hardened ed25519 keys. They are then converted into an encrypted `KeyPair` interface following the same pattern already found in the keybase.

Some documentation on this process can be seen [here](https://github.com/pokt-network/pocket/blob/slip-0010-keys/shared/crypto/README.md#slip-0010-hd-child-key-generation)

## Issue

Fixes N/A

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

- Implement SLIP-0010
- Integrate child key generation with the keybase
- Add unit tests for child generation
- Update documentation

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/app/client/keybase/keybase.go
+++ b/app/client/keybase/keybase.go
@@ -14,6 +14,15 @@ type Keybase interface {
 	// Insert a new keypair from the JSON string of the encrypted private key into the DB
 	ImportFromJSON(jsonStr, passphrase string) error
 
+	// DISCUSSION: Is there a more appropriate way to handle the HD interactions?
+	// SLIPS-0010 Key Derivation
+	// Deterministically generate and return the child key
+	DeriveChildFromKey(masterAddrHex, passphrase string, childIndex int32) (crypto.KeyPair, error)
+	DeriveChildFromSeed(seed []byte, childIndex int32) (crypto.KeyPair, error)
+	// Store the derived child key in the keybase
+	StoreChildFromKey(masterAddrHex, masterPassphrase string, childIndex int32, childPassphrase, childHint string) error
+	StoreChildFromSeed(seed []byte, childIndex int32, childPassphrase, childHint string) error
+
 	// Accessors
 	Get(address string) (crypto.KeyPair, error)
 	GetPubKey(address string) (crypto.PublicKey, error)
--- a/app/client/keybase/keystore.go
+++ b/app/client/keybase/keystore.go
@@ -223,6 +223,136 @@ func (keybase *badgerKeybase) GetAll() (addresses []string, keyPairs []crypto.Ke
 	return addresses, keyPairs, nil
 }
 
+// Deterministically generate and return the ith child key from the masterAddrHex key stored in the keybase
+func (keybase *badgerKeybase) DeriveChildFromKey(masterAddrHex, passphrase string, childIndex int32) (crypto.KeyPair, error) {
+	privKey, err := keybase.GetPrivKey(masterAddrHex, passphrase)
+	if err != nil {
+		return nil, err
+	}
+	seed := privKey.Seed()
+	path := fmt.Sprintf(crypto.PoktAccountPathFormat, childIndex)
+	childKey, err := crypto.DeriveKeyFromPath(path, seed)
+	if err != nil {
+		return nil, err
+	}
+	return childKey, nil
+}
+
+// Deterministically generate and return the ith child from the seed provided
+func (keybase *badgerKeybase) DeriveChildFromSeed(seed []byte, childIndex int32) (crypto.KeyPair, error) {
+	path := fmt.Sprintf(crypto.PoktAccountPathFormat, childIndex)
+	childKey, err := crypto.DeriveKeyFromPath(path, seed)
+	if err != nil {
+		return nil, err
+	}
+	return childKey, nil
+}
+
+// Deterministically generate and store the ith child from the masterAddrHex key stored in the keybase
+func (keybase *badgerKeybase) StoreChildFromKey(masterAddrHex, masterPassphrase string, childIndex int32, childPassphrase, childHint string) error {
+	masterPrivKey, err := keybase.GetPrivKey(masterAddrHex, masterPassphrase)
+	if err != nil {
+		return err
+	}
+	seed := masterPrivKey.Seed()
+	path := fmt.Sprintf(crypto.PoktAccountPathFormat, childIndex)
+	childKey, err := crypto.DeriveKeyFromPath(path, seed)
+	if err != nil {
+		return err
+	}
+	// No need to re-encrypt with provided passphrase
+	if childPassphrase == "" && childHint == "" {
+		err = keybase.db.Update(func(tx *badger.Txn) error {
+			// Use key address as key in DB
+			addrKey := childKey.GetAddressBytes()
+
+			// Encode KeyPair into []byte for value
+			keypairBz, err := childKey.Marshal()
+			if err != nil {
+				return err
+			}
+
+			return tx.Set(addrKey, keypairBz)
+		})
+	} else {
+		// Re-encrypt child key with passphrase and hint
+		err = keybase.db.Update(func(tx *badger.Txn) error {
+			// Get the private key hex string from the child key
+			privKeyHex, err := childKey.ExportString("") // No passphrase by default
+			if err != nil {
+				return err
+			}
+
+			keyPair, err := crypto.CreateNewKeyFromString(privKeyHex, childPassphrase, childHint)
+			if err != nil {
+				return err
+			}
+
+			// Use key address as key in DB
+			addrKey := keyPair.GetAddressBytes()
+
+			// Encode KeyPair into []byte for value
+			keypairBz, err := keyPair.Marshal()
+			if err != nil {
+				return err
+			}
+
+			return tx.Set(addrKey, keypairBz)
+		})
+	}
+	return nil
+}
+
+// Deterministically generate and store the ith child from the masterAddrHex key stored in the keybase
+func (keybase *badgerKeybase) StoreChildFromSeed(seed []byte, childIndex int32, childPassphrase, childHint string) error {
+	path := fmt.Sprintf(crypto.PoktAccountPathFormat, childIndex)
+	childKey, err := crypto.DeriveKeyFromPath(path, seed)
+	if err != nil {
+		return err
+	}
+	// No need to re-encrypt with provided passphrase
+	if childPassphrase == "" && childHint == "" {
+		err = keybase.db.Update(func(tx *badger.Txn) error {
+			// Use key address as key in DB
+			addrKey := childKey.GetAddressBytes()
+
+			// Encode KeyPair into []byte for value
+			keypairBz, err := childKey.Marshal()
+			if err != nil {
+				return err
+			}
+
+			return tx.Set(addrKey, keypairBz)
+		})
+	} else {
+		// Re-encrypt child key with passphrase and hint
+		err = keybase.db.Update(func(tx *badger.Txn) error {
+			// Get the private key hex string from the child key
+			privKeyHex, err := childKey.ExportString("") // No passphrase by default
+			if err != nil {
+				return err
+			}
+
+			keyPair, err := crypto.CreateNewKeyFromString(privKeyHex, childPassphrase, childHint)
+			if err != nil {
+				return err
+			}
+
+			// Use key address as key in DB
+			addrKey := keyPair.GetAddressBytes()
+
+			// Encode KeyPair into []byte for value
+			keypairBz, err := keyPair.Marshal()
+			if err != nil {
+				return err
+			}
+
+			return tx.Set(addrKey, keypairBz)
+		})
+	}
+	return nil
+}
+
 // Check whether an address is currently stored in the DB
 func (keybase *badgerKeybase) Exists(address string) (bool, error) {
 	val, err := keybase.Get(address)
--- a/shared/crypto/README.md
+++ b/shared/crypto/README.md
@@ -3,7 +3,7 @@
 - [KeyPair Interface](#keypair-interface)
   - [KeyPair Code Structure](#keypair-code-structure)
 - [Encryption and Armouring](#encryption-and-armouring)
-- [Child Key Generation](#slips-0010-hd-child-key-generation)
+- [Child Key Generation](#slip-0010-hd-child-key-generation)
 
 _DOCUMENT: Note that this README is a WIP and does not exhaustively document all the current types in this package_
 
@@ -105,9 +105,9 @@ flowchart LR
     AES-GCM-->PrivateKey
 ```
 
-## SLIPS-0010 HD Child Key Generation
+## SLIP-0010 HD Child Key Generation
 
-[SLIPS-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) key generation from a master key or seed is supported through the file [slip.go](./slip.go)
+[SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) key generation from a master key or seed is supported through the file [slip.go](./slip.go)
 
 The keys are generated using the BIP-44 path `m/44'/635'/%d'` where `%d` is the index of the child key - this allows for the deterministic generation of up to `2147483647` hardened ed25519 child keys per master key.
 
--- a/shared/crypto/README.md
+++ b/shared/crypto/README.md
@@ -3,6 +3,7 @@
 - [KeyPair Interface](#keypair-interface)
   - [KeyPair Code Structure](#keypair-code-structure)
 - [Encryption and Armouring](#encryption-and-armouring)
+- [Child Key Generation](#slips-0010-hd-child-key-generation)
 
 _DOCUMENT: Note that this README is a WIP and does not exhaustively document all the current types in this package_
 
@@ -102,6 +103,54 @@ flowchart LR
     C--encryptedBytes-->AES-GCM
     S--Key-->AES-GCM
     AES-GCM-->PrivateKey
+```
+
+## SLIPS-0010 HD Child Key Generation
+
+[SLIPS-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) key generation from a master key or seed is supported through the file [slip.go](./slip.go)
+
+The keys are generated using the BIP-44 path `m/44'/635'/%d'` where `%d` is the index of the child key - this allows for the deterministic generation of up to `2147483647` hardened ed25519 child keys per master key.
+
+Master key derivation is done as follows:
+```mermaid
+flowchart LR
+    subgraph HMAC
+        direction TB
+        A["hmacNew(sha512, seedModifier)"]
+        B["hmacWrite(seed)"]
+        C["convertToBytes(hmac)"]
+        A-->B-->C
+    end
+    subgraph MASTER-KEY
+        direction TB
+        D["SecretKey: hmacBytes[:32]"]
+        E["ChainCode: hmacBytes[32:]"]
+        D --> E --Secret+Chaincode--> KEY
+    end
+    HMAC--hmacBytes-->MASTER-KEY
+```
+
+Child keys are derived from their parents as follows:
+```mermaid
+flowchart LR
+    subgraph HMAC-CHILD
+        direction TB
+        C["append(0x0, parent.SecretKey, bigEndian(index))"]
+        A["hmacNew(sha512, parent.Chaincode)"]
+        B["hmacWrite(data)"]
+        D["convertToBytes(hmac)"]
+        A--hmac-->B
+        C--data-->B
+        B-->D
+    end
+    subgraph CHILD-KEY
+        direction TB
+        F["secret: hmacBytes[:32]"]
+        G["chaincode: hmacBytes[32:]"]
+        F --> G --Secret+Chaincode--> KEY
+    end
+    Parent --> HMAC-CHILD
+    HMAC-CHILD --hmacBytes--> CHILD-KEY
 ```
 
 <!-- GITHUB_WIKI: shared/crypto/readme -->
--- a/shared/crypto/keypair.go
+++ b/shared/crypto/keypair.go
@@ -30,6 +30,9 @@ type KeyPair interface {
 	ExportString(passphrase string) (string, error)
 	ExportJSON(passphrase string) (string, error)
 
+	// Seed
+	GetSeed(passphrase string) ([]byte, error)
+
 	// Marshalling
 	Marshal() ([]byte, error)
 	Unmarshal([]byte) error
@@ -97,6 +100,15 @@ func (kp encKeyPair) ExportJSON(passphrase string) (string, error) {
 		return "", err
 	}
 	return kp.PrivKeyArmour, nil
+}
+
+// Return the seed of the key
+func (kp encKeyPair) GetSeed(passphrase string) ([]byte, error) {
+	privKey, err := unarmourDecryptPrivKey(kp.PrivKeyArmour, passphrase)
+	if err != nil {
+		return []byte{}, err
+	}
+	return privKey.Seed(), nil
 }
 
 // Marshal KeyPair into a []byte
--- a/shared/crypto/slip.go
+++ b/shared/crypto/slip.go
@@ -1,0 +1,157 @@
+package crypto
+
+import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/sha512"
+	"encoding/binary"
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+const (
+	// PoktAccountPrefix is the prefix used for HD key derivation where 635 is the SLIP-0044 coin type
+	PoktAccountPrefix = "m/44'/635'"
+	// PoktPrimaryAccountPath is the derivation path for the primary account
+	PoktPrimaryAccountPath = "m/44'/635'/0'"
+	// PoktAccountPathFormat is to be used with fmt.Sprintf to generate child keys
+	PoktAccountPathFormat = "m/44'/635'/%d"
+	// FirstHardenedKey is the index for the first hardened key for ed25519 keys
+	FirstHardenedKey = uint32(0x80000000)
+	MaxHardenedKey   = ^uint32(0)
+	// As defined in: https://github.com/satoshilabs/slips/blob/master/slip-0010.md#master-key-generation
+	seedModifier = "ed25519 seed"
+)
+
+var (
+	ErrInvalidPath  = fmt.Errorf("invalid BIP-44 derivation path")
+	ErrNoDerivation = fmt.Errorf("no derivation for an ed25519 key is possible")
+	pathRegex       = regexp.MustCompile(`m(\/[0-9]+')+$`)
+)
+
+type slipKey struct {
+	SecretKey []byte
+	ChainCode []byte
+}
+
+// Derives a key from a BIP-44 path and a seed only operating on hardened ed25519 keys
+func DeriveKeyFromPath(path string, seed []byte, passphrase, hint string) (KeyPair, error) {
+	segments, err := pathToSegments(path)
+	if err != nil {
+		return nil, err
+	}
+
+	key, err := NewMasterKey(seed)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, i32 := range segments {
+		// Force hardened keys
+		if i32 >= MaxHardenedKey-FirstHardenedKey {
+			return nil, fmt.Errorf("hardened key index too large, max: %d, got: %d", MaxHardenedKey-FirstHardenedKey, i32)
+		}
+		i := i32 + FirstHardenedKey
+
+		// Derive the correct child until final segment
+		key, err = key.DeriveChild(i)
+		if err != nil {
+			return nil, err
+		}
+	}
+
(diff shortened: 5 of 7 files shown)


shared/crypto/slip.go
```diff
@@ -0,0 +1,156 @@
...
+const (
+	// PoktAccountPathFormat used for HD key derivation where 635 is the SLIP-0044 coin type
+	// To be used with fmt.Sprintf to generate child keys
+	PoktAccountPathFormat = "m/44'/635'/%d'"
+	// firstHardenedKey is the index for the first hardened key for ed25519 keys
+	firstHardenedKey = uint32(0x80000000)
+	maxHardenedKey   = ^uint32(0)
+	MaxChildKeyIndex = maxHardenedKey - firstHardenedKey
+	// As defined in: https://github.com/satoshilabs/slips/blob/master/slip-0010.md#master-key-generation
+	seedModifier = "ed25519 seed"
```
Olshansk, line 23: s/seedModifier/hmacKey

shared/crypto/slip.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Derives a key from a BIP-44 path and a seed only operating on hardened ed25519 keys
+func DeriveKeyFromPath(path string, seed []byte) (KeyPair, error) {
+	segments, err := pathToSegments(path)
+	if err != nil {
+		return nil, err
+	}
+
+	key, err := newMasterKey(seed)
```
Olshansk, line 44: s/key/masterKey

NIT: DO this first (easier to follow the code)

shared/crypto/slip.go
```diff
@@ -0,0 +1,156 @@
...
+
+	for _, i32 := range segments {
+		// Force hardened keys
+		if i32 > MaxChildKeyIndex {
+			return nil, fmt.Errorf("hardened key index too large, max: %d, got: %d", MaxChildKeyIndex, i32)
+		}
+		i := i32 + firstHardenedKey
+
+		// Derive the correct child until final segment
+		key, err = key.deriveChild(i)
```
Olshansk, line 57: s/key/childKey

shared/crypto/slip.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, i32 := range segments {
+		// Force hardened keys
+		if i32 > MaxChildKeyIndex {
+			return nil, fmt.Errorf("hardened key index too large, max: %d, got: %d", MaxChildKeyIndex, i32)
+		}
+		i := i32 + firstHardenedKey
```
Olshansk, line 54: Can you point me to these 4 lines in the spec?

shared/crypto/slip.go
```diff
@@ -0,0 +1,156 @@
+package crypto
```
Olshansk, line 1: Can you add `slip_test.go` (positive & negative test cases) for https://github.com/satoshilabs/slips/blob/master/slip-0010.md to make sure it works.

Sleep need a bit more time to understand the spec and verify.

Olshansk (changes requested):
1. Code is pretty clean 👌

2. I learnt a bunch of new stuff, so ty

3. Usually people say to "never roll your own crypto" so what was the motivation here? Were there no go libraries supporting this or was it just for the sake of doing something fun & useful while also learning?

shared/crypto/slip.go
h5law: @Olshansk I think this should stay as `seedModifier` it is only the first part of the HMAC key and the seed provided (which in theory should be a mnemonic or something like this or just a key's seed) is the other part

shared/crypto/README.md
```diff
@@ -104,4 +105,52 @@ flowchart LR
     AES-GCM-->PrivateKey
 ```
 
+## SLIP-0010 HD Child Key Generation
+
+[SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) key generation from a master key or seed is supported through the file [slip.go](./slip.go)
+
+The keys are generated using the BIP-44 path `m/44'/635'/%d'` where `%d` is the index of the child key - this allows for the deterministic generation of up to `2147483647` hardened ed25519 child keys per master key.
```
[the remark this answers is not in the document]
h5law: @Olshansk I actually chose BIP-44 as the spec was using it 😅  However I have actually removed the BIP-44 path from the implementation as it is kind of overkill - the implementation works exactly the same now (hardened ed25519 keys only) but without the path - which was not even exposed to the user anyway

shared/crypto/slip.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+const (
+	// PoktAccountPathFormat used for HD key derivation where 635 is the SLIP-0044 coin type
+	// To be used with fmt.Sprintf to generate child keys
+	PoktAccountPathFormat = "m/44'/635'/%d'"
+	// firstHardenedKey is the index for the first hardened key for ed25519 keys
+	firstHardenedKey = uint32(0x80000000)
+	maxHardenedKey   = ^uint32(0)
+	MaxChildKeyIndex = maxHardenedKey - firstHardenedKey
```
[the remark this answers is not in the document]
h5law: These come from the BIP-32 spec which basically says normal keys are indexes 0-(2^31)-1 and hardened keys are 2^31 - (2^32)-1 (`^uint32(0)`)

h5law:
@Olshansk

> Usually people say to "never roll your own crypto" so what was the motivation here? Were there no go libraries supporting this or was it just for the sake of doing something fun & useful while also learning?

I had actually found a few libraries - they were all 3-5 years old without any updates. The only recent one is a ("demo" implementation by Iota](https://github.com/Wollac/iota-crypto-demo/tree/b10619eccb983ca50b26c72c4afa6534e5476092/pkg/slip10) - however because it was called a demo implementation this seemed like it maybe shouldn't be used.

However, after reading through all the different implementations I saw that this is not really any new cryptography but is instead just utilising it in a different manner. I was also excited to do it myself but my code is very similar to all the other implementations, the only real difference is linking it with our `KeyPair` interface - most other implementations are using separate key types for theirs but I decided to instead use a separate type and convert it to our existing `KeyPair` interface as this is simply an abstraction of an `ed25519` key anyway.

shared/crypto/README.md
```diff
@@ -104,4 +105,51 @@ flowchart LR
...
+        C["convertToBytes(hmac)"]
+        A-->B-->C
+    end
+    subgraph MASTER-KEY
+        direction TB
+        D["SecretKey: hmacBytes[:32]"]
+        E["ChainCode: hmacBytes[32:]"]
+        D --> E --Secret+Chaincode--> KEY
+    end
+    HMAC--hmacBytes-->MASTER-KEY
```
Olshansk, line 129:
```suggestion
flowchart LR
    subgraph HMAC
        direction TB
        A["hmac = hmacNew(sha512, seedModifier)"]
        B["hmac.Write(seed)"]
        C["convertToBytes(hmac)"]
        A-->B
        B--hmac-->C
    end
    subgraph MASTER-KEY
        direction LR
        D["SecretKey: hmacBytes[:32]"]
        E["ChainCode: hmacBytes[32:]"]
        D --> KEY
        E --> KEY
    end
    seed-->HMAC
    HMAC--hmacBytes-->MASTER-KEY
```

shared/crypto/README.md
```diff
@@ -104,4 +105,51 @@ flowchart LR
...
+        B-->D
+    end
+    subgraph CHILD-KEY
+        direction TB
+        F["secret: hmacBytes[:32]"]
+        G["chaincode: hmacBytes[32:]"]
+        F --> G --Secret+Chaincode--> KEY
+    end
+    Parent --> HMAC-CHILD
+    HMAC-CHILD --hmacBytes--> CHILD-KEY
```
Olshansk, line 152:
```suggestion
flowchart LR
    subgraph HCHILD["HMAC-CHILD"]
        direction TB
        C["append(0x0, parent.SecretKey, bigEndian(index))"]
        A["hmacNew(sha512, parent.Chaincode)"]
        B["hmac.Write(data)"]
        D["convertToBytes(hmac)"]
        A--hmac-->B
        C--data-->B
        B--hmac-->D
    end
    subgraph CKEY[CHILD-KEY]
        direction LR
        F["SecretKey: hmacBytes[:32]"]
        G["ChainCode: hmacBytes[32:]"]
        F --> KEY
        G --> KEY
    end
    Index-->HCHILD
    Parent-->HCHILD
    HCHILD--hmacBytes-->CKEY
```

shared/crypto/README.md
```diff
@@ -104,4 +105,51 @@ flowchart LR
...
+```
```
Olshansk, line 153: Solid diagrams. Just made a couple small suggestions

shared/crypto/slip.go
```diff
@@ -0,0 +1,164 @@
...
+
+const (
+	// PoktAccountPathFormat used for HD key derivation where 635 is the SLIP-0044 coin type
+	// To be used with fmt.Sprintf to generate child keys
+	// Ref: https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+	PoktAccountPathFormat = "m/44'/635'/%d'" // m/purpose'/coin_type'/account_idx'
+	// As defined in: https://github.com/satoshilabs/slips/blob/master/slip-0010.md#master-key-generation
+	seedModifier = "ed25519 seed"
+	// Hardened key values, as defined in: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys
+	firstHardenedKeyIndex = uint32(1 << 31) // 2^31
```
Olshansk, line 22:
```suggestion
	firstHardenedKeyIndex = uint32(1 << 31) // 2^31; [0, 2^31) are non-hardened keys which are not supported for ed25519
```

shared/crypto/slip.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, err
+	}
+
+	// Iterate over segments in path regenerating the child key until correct
+	for _, i32 := range segments {
+		// Force hardened keys
+		if i32 > maxChildKeyIndex {
+			return nil, fmt.Errorf("hardened key index too large, max: %d, got: %d", maxChildKeyIndex, i32)
+		}
+		i := i32 + firstHardenedKeyIndex
```
Olshansk, line 58:
```suggestion
		keyIdx := i32 + firstHardenedKeyIndex
```

app/client/keybase/slip_test.go
```diff
@@ -0,0 +1,308 @@
...
+				path: "m",
+				seed: seed,
+			},
+			wantPrivHex: "2b4be7f19ee27bbf30c667b642d5f4aa69fd169872f8fc3059c08ebae2eb19e7",
+			wantPubHex:  "00a4b2856bfec510abab89753fac1ac0e1112364e7d250545963f135f2a33188ed",
+
+			wantErr: false,
+		},
+		{
+			name: "Key(m/0')",
```
Olshansk, line 44: Can you update the names with what it's testing?

app/client/keybase/slip_test.go
```diff
@@ -0,0 +1,308 @@
...
+			wantErr:     true,
+		},
+	}
+	for _, tv := range tests {
+		t.Run(tv.name, func(t *testing.T) {
+			childKey, err := crypto.DeriveChild(tv.args.path, tv.args.seed)
+			if (err != nil) != tv.wantErr {
+				t.Errorf("DeriveChild() error = %v, wantErr %v", err, tv.wantErr)
+				return
+			}
```
Olshansk, line 114: I think this might do the same thing:

```suggestion
			if tv.wantErr {
				require.Error(t, err)
			}
```

app/client/keybase/slip_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+			// Slip-0010 keys are prefixed with "00" in the test vectors
+			pubHex := childKey.GetPublicKey().String()
+			require.Equal(t, "00"+pubHex, tv.wantPubHex)
+		})
+	}
+}
+
+// https://github.com/satoshilabs/slips/blob/master/slip-0010.md#test-vector-2-for-ed25519
+func TestSlip_DeriveChild_Vector2(t *testing.T) {
```
Olshansk, line 135: You can just consolidate these two tests into one and change the params.  That's what the structs & forloop are setup for.

app/client/keybase/slip_test.go
```diff
@@ -0,0 +1,308 @@
...
+				path: "m/0'/1'/2'/2'",
+				seed: seed,
+			},
+			wantPrivHex: "30d1dc7e5fc04c31219ab25a27ae00b50f6fd66622f6e9c913253d6511d1e662",
+			wantPubHex:  "008abae2d66361c879b900d204ad2cc4984fa2aa344dd7ddc46007329ac76c429c",
+
+			wantErr: false,
+		},
+		{
+			name: "Key(m/0'/1'/2'/2'/1000000000')",
```
Olshansk, line 88: Can you add at least a couple tests for `m/44'/635'/%d` to capture our use case?

Olshansk (changes requested):
Took a deeper dive and everything makes sense now. Appreciate all the added context you provided to help clarify things.

Should be good after this next set of changes

shared/crypto/keypair.go
```diff
@@ -111,6 +115,16 @@ func (kp encKeyPair) GetSeed(passphrase string) ([]byte, error) {
 	return privKey.Seed(), nil
 }
 
+// Derives a new child key using the key as the master
```
Olshansk, line 118: Just a NIT (but its outside the scope of this PR) that I learnt this this week: proper godoc comments can be found here: https://go.dev/blog/godoc

tl;dr `<FunctionName> <ExplanationWhatFunctionDoes>`

shared/crypto/keypair.go
```diff
@@ -111,6 +115,16 @@ func (kp encKeyPair) GetSeed(passphrase string) ([]byte, error) {
...
+func (kp encKeyPair) DeriveChild(passphrase string, index uint32) (KeyPair, error) {
+	seed, err := kp.GetSeed(passphrase)
+	if err != nil {
+		return nil, err
+	}
+	path := fmt.Sprintf(PoktAccountPathFormat, index)
+	return DeriveChild(path, seed)
```
Olshansk, line 125: Seeing this usage (and being confused by where it's defined) made me think that maybe we should move `slip.go` and `slip_test.go` into `shared/crypto/slip`? 

It might be premature, but I'm hoping to have all the packages we're building here to be importable (more modularly) by others in the future. Will definitely help with growth, adoption, etc

shared/crypto/README.md
```diff
@@ -113,43 +113,48 @@ The keys are generated using the BIP-44 path `m/44'/635'/%d'` where `%d` is the
 Master key derivation is done as follows:
 ```mermaid
 flowchart LR
-    subgraph HMAC
+    subgraph HCHILD["HMAC-CHILD"]
```
Olshansk, line 116: These diagrams are great btw!

app/client/keybase/slip_test.go
```diff
@@ -2,225 +2,213 @@ package keybase
...
-			wantErr: false,
+			wantErr:     false,
 		},
 		{
-			name: "Key(m/0')",
-			args: args{
-				path: "m/0'",
-				seed: seed,
-			},
+			name:        "TestVector1 Key derivation is deterministic for path `m/0'`",
```
Olshansk, line 42: For my personal understanding, since th epath is `m/purpose'/coint_type'/child_idx'`, what is this deriving?

app/client/keybase/slip_test.go
```diff
@@ -2,225 +2,213 @@ package keybase
...
+			name:        "TestVector2 Key derivation is deterministic for path `m/0'/2147483647'/1'/2147483646'/2'`",
+			path:        "m/0'/2147483647'/1'/2147483646'/2'",
+			seed:        testVector2SeedHex,
 			wantPrivHex: "551d333177df541ad876a60ea71f00447931c0a9da16f227c11ea080d7391b8d",
 			wantPubHex:  "0047150c75db263559a70d5778bf36abbab30fb061ad69f69ece61a72b0cfa4fc0",
-
-			wantErr: false,
+			wantErr:     false,
+		},
+		// Pocket specific test vectors
```
Olshansk, line 137: 🙏

app/client/keybase/keystore.go
```diff
@@ -283,40 +284,47 @@ func (keybase *badgerKeybase) StoreChildFromSeed(seed []byte, childIndex uint32,
...
 
-	// Re-encrypt child key with passphrase and hint
-	return keybase.db.Update(func(tx *badger.Txn) error {
-		// Get the private key hex string from the child key
-		privKeyHex, err := childKey.ExportString("") // No passphrase by default
-		if err != nil {
-			return err
-		}
+		return childKey, err
+	} else {
```
Olshansk, line 289: Why did you reintroduce the else statement?

Since we return in the `if`, its easier to read without it.

app/client/keybase/keybase.go
```diff
@@ -25,8 +25,8 @@ type Keybase interface {
 	DeriveChildFromKey(masterAddrHex, passphrase string, childIndex uint32) (crypto.KeyPair, error)
```
Olshansk, line 25: Seeing how you introduced the `store_child` flag. Do we even need both `DeriveChild` and `StoreChild` functions anymore?

I feel like just having `DeriveChild` with `store_child` defaulting to `true` is simple, descriptive and maintains optionality.

Olshansk (changes requested):

app/client/keybase/slip_test.go
h5law: @Olshansk So what we actually do to derive the key is first derive a master key `m/` then for each segment derive a child - for our use case this means we derive 3 keys before we actually get to the indexed derivation `m(key 1)/44'(key 2)/635'(key 3)/%d'(child key index)` this is the idea behind the of BIP-44. So we can change the purpose (`44'`) part of the path and with the same seed generate new accounts. For our use case this probably wont be the case but that is what we are actually doing. Many layers to the child 🧅

shared/crypto/keypair.go
h5law: This might be a good starter task actually, I will go through the keybase and slips stuff later today and get the comments in order 🤝

Olshansk: 🤝

app/client/cli/keys.go
```diff
@@ -725,25 +732,17 @@ func keysSlipCommands() []*cobra.Command {
...
+				kp, err := kb.DeriveChildFromKey(parentAddr, pwd, index, childPwd, childHint)
+				if err != nil {
+					return err
 				}
 
 				if err := kb.Stop(); err != nil {
 					return err
 				}
 
+				logger.Global.Info().Str("address", kp.GetAddressString()).Str("parent", parentAddr).Uint32("index", index).Bool("stored", storeChild).Msg("Child key derived")
```
Olshansk, line 744: Nice logline

app/client/cli/keys.go
```diff
@@ -98,6 +99,12 @@ func NewKeysCommand() *cobra.Command {
 	cmd.AddCommand(signTxCmds...)
 	cmd.AddCommand(slipCmds...)
 
+	// Bind the store_child flag
+	if err := viper.BindPFlag("storeChild", cmd.Flags().Lookup("store_child")); err != nil {
```
Olshansk, line 103: Seems inconsistent that we're using cobra flags for everything except for this one exception, which is only applicable to the `DeriveChild` function.

I think we should simply update the interface:

```go
// Before:
DeriveChildFromKey(parentAddr, pwd, index, childPwd, childHint)

// After: 
DeriveChildFromKey(parentAddr, pwd, index, childPwd, childHint, shouldStore)
```

Olshansk (changes requested):

app/client/keybase/slip_test.go
Olshansk: Got it. So hypothetically (we don't need to do it), an application that wants to maintain privacy by generating lots of different receiving addresses could cache m/44/635 and only update the index, right?

h5law: @Olshansk so If someone really wanted they could store the seed of the key generated from `m/44'/635'` and this is equivalent to the `SecretKey` of the `HMAC` key. Then they could derive the indexed child directly from this - however our current setup means this wouldn't work as we force the use of a `BIP-44` path when deriving children - this means if they saved `m/44'/635'` they couldn't use the path `193872'` to get that child as this is not valid if they used `m/44'/635'/193872'` as the path at this point they would not get the right child as they would re-generate the other "layers" of the child key. I hope this makes sense.

It would be possible to change this but I don't think the current use case requires this - maybe in a follow up PR or something

shared/crypto/slip/slip.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Initialise a master key from seed to start child generation
+	key, err := newMasterKey(seed)
+	if err != nil {
+		return nil, err
+	}
+
+	// Iterate over segments in path regenerating the child key until correct
+	for _, i32 := range segments {
+		// Force hardened keys
```
Olshansk, line 54: Does this also implicitly check that `purpose` and `coin_type` are < `maxChildKeyIndex`? If so, let's just update the comment to make it clear.

shared/crypto/slip/slip.go
```diff
@@ -0,0 +1,147 @@
...
+		if i32 > maxChildKeyIndex {
+			return nil, fmt.Errorf("hardened key index too large, max: %d, got: %d", maxChildKeyIndex, i32)
+		}
+		keyIdx := i32 + firstHardenedKeyIndex
+
+		// Derive the correct child until final segment
```
Olshansk, line 60: This makes sense now. Ty

shared/crypto/slip/slip_test.go
```diff
@@ -0,0 +1,228 @@
...
+			seed, err := hex.DecodeString(tv.seed)
+			require.NoError(t, err)
+			childKey, err := DeriveChild(tv.path, seed)
+			if tv.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+
+			if err != nil {
```
Olshansk, line 212: Do we need this if structure?

shared/crypto/slip/slip_test.go
```diff
@@ -0,0 +1,228 @@
...
+			}
+
+			// Slip-0010 private keys in test vector are only the seed of the full private key
+			// This is equivalent to the SecretKey of the HMAC key used to generate the ed25519 key
+			privSeed, err := childKey.GetSeed("")
+			require.NoError(t, err)
+			privHex := hex.EncodeToString(privSeed)
+			require.Equal(t, privHex, tv.wantPrivHex)
+
+			// Slip-0010 keys are prefixed with "00" in the test vectors
```
Olshansk, line 223: Why is this?

shared/crypto/slip/slip_test.go
```diff
@@ -0,0 +1,228 @@
...
+			if err != nil {
+				return
+			}
+
+			// Slip-0010 private keys in test vector are only the seed of the full private key
+			// This is equivalent to the SecretKey of the HMAC key used to generate the ed25519 key
+			privSeed, err := childKey.GetSeed("")
+			require.NoError(t, err)
+			privHex := hex.EncodeToString(privSeed)
+			require.Equal(t, privHex, tv.wantPrivHex)
```
Olshansk, line 221: Reorder expected & actual: `require.Equal(t, tv.wantPrivHex, privHex)`. It's important because it impacts the error message that's outputted if there's an issue.

https://pkg.go.dev/github.com/stretchr/testify/require#Equal

Olshansk (changes requested):

shared/crypto/slip/slip_test.go
h5law: @Olshansk So normally ed25519 public keys have no prefix - but for some reason the Test Vector ones in the SLIP spec are prefixed with `00` this means the same keys are derived but we just have to prefix them explicitly.

shared/crypto/slip/slip_test.go
h5law: Yeah this makes sure that if the error is present (for the tests when we want it) it doesn't try to do the rest of the test and throw null pointer exceptions because the other stuff isn't there.

app/client/keybase/keystore.go
```diff
@@ -242,7 +247,78 @@ func (keybase *badgerKeybase) GetAll() (addresses []string, keyPairs []crypto.Ke
...
+
+	if shouldStore {
+		// No need to re-encrypt with provided passphrase
+		if childPassphrase == "" && childHint == "" {
+			err = keybase.db.Update(func(tx *badger.Txn) error {
+				// Use key address as key in DB
+				addrKey := childKey.GetAddressBytes()
+
+				// Encode KeyPair into []byte for value
+				keypairBz, err := childKey.Marshal()
```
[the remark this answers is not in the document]
h5law: So what we return from `DeriveChild` is actually a `crypto.KeyPair` interface the same as we are using normally this has both the `crypto.PublicKey` interface and the `JSON` encoded, encrypted private key string making up the keypair. All `Marshal()` is doing here is the conversion to a `[]byte`

shared/crypto/slip/slip_test.go
```diff
@@ -0,0 +1,228 @@
...
+	tests := []struct {
+		name        string
+		path        string
+		seed        string
+		wantPrivHex string
+		wantPubHex  string
+
+		wantErr bool
+	}{
+		// https://github.com/satoshilabs/slips/blob/master/slip-0010.md#test-vector-1-for-ed25519
```
Olshansk, line 30:
```suggestion
		// https://github.com/satoshilabs/slips/blob/master/slip-0010.md#test-vector-1-for-ed25519
		// Note that ed25519 public keys normaly don't have a "00" prefix, but we are reflecting the 
		// test vectors from the spec which do
```

shared/crypto/slip/slip_test.go
Olshansk: Fair enough. Adding a suggestion to a comment

shared/crypto/slip/slip_test.go
```diff
@@ -0,0 +1,228 @@
...
+			require.NoError(t, err)
+			childKey, err := DeriveChild(tv.path, seed)
+			if tv.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+
+			if err != nil {
+				return
```
Olshansk, line 213:
```suggestion
				continue
```

shared/crypto/slip/slip_test.go
Olshansk: Adding a suggestion, should it be a `continue` then so we run the remainder of the tests?

app/client/keybase/keystore.go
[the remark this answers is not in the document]
Olshansk: 👌

shared/crypto/slip/slip_test.go
```diff
@@ -0,0 +1,228 @@
...
+			// Slip-0010 private keys in test vector are only the seed of the full private key
+			// This is equivalent to the SecretKey of the HMAC key used to generate the ed25519 key
+			privSeed, err := childKey.GetSeed("")
+			require.NoError(t, err)
+			privHex := hex.EncodeToString(privSeed)
+			require.Equal(t, tv.wantPrivHex, privHex)
+
+			// Slip-0010 keys are prefixed with "00" in the test vectors
+			pubHex := childKey.GetPublicKey().String()
+			require.Equal(t, "00"+pubHex, tv.wantPubHex)
```
Olshansk, line 225:
```suggestion
			require.Equal(t, tv.wantPubHex, "00"+pubHex)
```

Olshansk (approved):
PTAL at the last couple of minor comments I left, but otherwise LGTM. Appreciate you bearing through all the back & forth.

Would you be down to present this at a devlog?

h5law:
> Would you be down to present this at a devlog?

WiFi issues fixed ✅
New headphones ✅
You will not steal my ⚡ twice 😉

shared/crypto/slip/slip_test.go
h5law: @Olshansk I don't think `continue` would work as this is not in a for loop. From my understanding the `t.Run` will run all of the test cases in separate "loops" so the return is only blocking the error cases from failing the tests it shouldn't pass anyway. This return doesn't block the other tests from running

Olshansk:
> WiFi issues fixed ✅ New headphones ✅ You will not steal my ⚡ twice 😉

🤝 

**Side note**: after all the iterations, I think we might have the best golang slip-0010 keygen library out there

Issue #352, opened 2022-11-14
[Consensus] State Sync - Implement & expose interface to advertise and provide height + blocks


Olshansk:
# Objective

Implement the interface that exposes the maximum & minimum blocks a node can expose and share

# Origin Document

State sync block by block design: #125

# Goals

- A foundation for advertising/exposing data a single node has to offer to others during state sync
- Allow another node to request a block from the advertising node

# Deliverables

- [ ] Implementation of the interface (fully integrated) that exposes the data a node has available
- [ ] Tooling and/or tests that call these endpoints to retrieve the advertised information

# Non-goals
- Actual implementation of state sync

## General issue checklist
- [ ] Update the appropriate CHANGELOG
- [ ] Update the README
- [ ] If applicable, update the source code tree explanation
- [ ] If applicable, add or update a state, sequence or flowchart diagram using [mermaid](https://mermaid-js.github.io/mermaid/)
- [ ] Update any relevant global documentation & references
- [ ] Document small issues / TODOs along the way

---

Creators: @jessicadaugherty @olshansk

PR #528, opened 2023-02-20
[Consensus] Configuration entry point state sync

### LINKED ISSUE #351: [Consensus] State Sync - Configuration & Entrypoints for Node State
# Objective

Maintain the state (e.g. synching, synched, etc...) of the node for state sync & execution purposes. the consensus module will have a variable that maintains state. This will require a constant that indicates its state (synced, syncing, pacemaker). The consensus module should still control state sync but there should be no validation as part of this implementation. 

# Origin Document

State sync block by block design: #125

# Goals

- A foundation for the full state sync implementation
- Configurations & a rough "state machine" for all the states involved in state sync

# Deliverables

- [ ] Implementation of entrypoints into the different states required for state sync
   - [ ] Interfaces that show the lifecycle of syncing without implementing state sync fully
- [ ] At a minimum, enable "synching" and "synced" states
- [ ] Placeholders / TODOs for the real state sync implementation

# Non-goals
- Actual implementation of state sync

## General issue checklist
- [ ] Update the README

---

Creators: @jessicadaugherty @olshansk

gokutheengineer:
## Description

Defines entry points and transitions of the state machine for the following modes:

- Unsynched mode
- Sync Mode
- Synced Mode
- Pacemaker Mode

This PR aims to define state transitions between different modes. Currently, server mode is not represented by the node's FSM, but rather it is set through a config variable, and it can be turned on and off via exported `EnableServerMode` and `DisableServerMode` functions

## Issue

Fixes #351 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Added `fsm_handler.go` to the consensus module to trigger business logic in consensus module accordingly to the FSM events
- Added `FSMConsensusEvents` interface to the shared module
- Updated and added FSM states, events, state transitions

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


consensus/state_sync/module.go
```diff
@@ -34,9 +25,14 @@ type StateSyncModule interface {
...
 	IsServerModEnabled() bool
-	EnableServerMode()
+	EnableServerMode() error
+	DisableServerMode() error
 
 	SendStateSyncMessage(msg *typesCons.StateSyncMessage, nodeAddress cryptoPocket.Address, height uint64) error
+	IsOutOfSync() bool
+	AggregateMetadataResponses() error
+	GetAggregatedSyncMetadata() *typesCons.StateSyncMetadataResponse
+	Sync() error
```
Olshansk, line 35: I think this should be `StartSynching` that triggers it to start requesting blocks from others

consensus/state_sync/module.go
```diff
@@ -34,9 +25,14 @@ type StateSyncModule interface {
 	HandleGetBlockResponse(*typesCons.GetBlockResponse) error
 
 	IsServerModEnabled() bool
-	EnableServerMode()
+	EnableServerMode() error
+	DisableServerMode() error
 
 	SendStateSyncMessage(msg *typesCons.StateSyncMessage, nodeAddress cryptoPocket.Address, height uint64) error
+	IsOutOfSync() bool
```
Olshansk, line 32: s/IsOutOfSync/IsSynched

Note that it's the opposite (so you need to change logic) but it's clearer.

consensus/state_sync/module.go
```diff
@@ -34,9 +25,14 @@ type StateSyncModule interface {
...
+	AggregateMetadataResponses() error
```
Olshansk, line 33: I kind of imagined this to be an independent process that happens implicitly. Anytime you get a response from others, you auto-aggregate.

consensus/fsm_handler.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Currently, FSM never transition to this state and a non-validator node always stays in syncmode.
+// CONSIDER: when a non-validator sync is implemented, maybe there is a case that requires transitioning to this state
+func (m *consensusModule) HandleSynced(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Debug().Msg("FSM of non-validator node is in Synced mode")
+	return nil
+}
+
+// Pacemaker mode is when the validator is synced and it is waiting for a new block proposal to come in
```
Olshansk, line 96: Let's start using proper godoc. 

HandlePacamer handles an event ...

consensus/fsm_handler.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+
+	return nil
+}
+
+// Currently, FSM never transition to this state and a non-validator node always stays in syncmode.
+// CONSIDER: when a non-validator sync is implemented, maybe there is a case that requires transitioning to this state
+func (m *consensusModule) HandleSynced(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Debug().Msg("FSM of non-validator node is in Synced mode")
```
Olshansk, line 92: Can you add a TODO to check that this never happens if you're `IsValidator` is true?

consensus/fsm_handler.go
```diff
@@ -0,0 +1,102 @@
...
+	return nil
+}
+
+// Pacemaker mode is when the validator is synced and it is waiting for a new block proposal to come in
+func (m *consensusModule) HandlePacemaker(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Debug().Msg("FSM of validator node is synced and in Pacemaker mode. It will stay in this mode until it receives a new block proposal that has a higher height than the current block height")
+	// validator receives a new block proposal, and it understands that it doesn't have block and it transitions to unsycnhed state
```
Olshansk, line 99: Can you add a TODO to check that this never happens if you're `IsValidator` is false?

Olshansk (changes requested):
@gokutheengineer I took a first pass at this review. There's definitely some more work and thinking to do to design this properly but it's a great start. 

Hopefully the comments I left throughout provide enough direction to get going.

Something that I think might really help is drawing out (and adding it to the readme) things like:

1. I got a Block and applied it
2. If no on e in my aggregated metadata has a higher block -> send event that I'm synched
3. If someone still has a higher block height -> keep requesting blocks fro my peers
4. In the background, keep aggregating metadata at different rates.

I think drwaing out a state diagram can really help with this: https://mermaid.js.org/syntax/stateDiagram.html

Looking forward to the next round of review on this

consensus/fsm_handler.go
```diff
@@ -0,0 +1,102 @@
...
+
+	fsm_state := msg.NewState
+	m.logger.Debug().Fields(map[string]any{
+		"event":          msg.Event,
+		"previous_state": msg.PreviousState,
+		"new_state":      fsm_state,
+	}).Msg("Received state machine transition msg")
+
+	switch fsm_state {
+	case string(coreTypes.StateMachineState_P2P_Bootstrapped):
```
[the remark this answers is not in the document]
gokutheengineer: very good, thanks.

consensus/helpers.go
```diff
@@ -26,8 +26,9 @@ const (
 
 	ByzantineThreshold = float64(2) / float64(3)
 
-	HotstuffMessageContentType  = "consensus.HotstuffMessage"
-	StateSyncMessageContentType = "consensus.StateSyncMessage"
+	HotstuffMessageContentType      = "consensus.HotstuffMessage"
```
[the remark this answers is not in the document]
gokutheengineer: Makes a lot of sense to move them into common location, moved.

consensus/module.go
```diff
@@ -149,6 +150,10 @@ func (*consensusModule) Create(bus modules.Bus, options ...modules.ModuleOption)
 
 	consensusCfg := runtimeMgr.GetConfig().Consensus
 
+	if consensusCfg.ServerModeEnabled {
```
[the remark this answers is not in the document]
gokutheengineer: I agree with that, and in fact, I first did it like this. However, currently server mode is not managed through with the FSM. Because I faced a limitation with the FSM library when trying to transition from one state to two new states. So, if server mode is enabled, this must happen with the FSM:

After `StateMachineEvent_Start` :
- FSM must transition to `P2P_Bootstrapping` state (which continues as `P2P_Bootstrapped`).
- FSM must also transition to EnableServerMode state, which must be running in parallel to the all states. Therefore this state transition must happen mutually exclusive to the `P2P_Bootstrapping` state transition. However, there is an issue with transitioning to more than one states with the current lib.

I will be investigating this issue, and if I find a easy solution I will enable State_Machine events.

consensus/module.go
```diff
@@ -317,6 +332,19 @@ func (m *consensusModule) loadPersistedState() error {
...
-	m.stateSync.EnableServerMode()
+func (m *consensusModule) EnableServerMode() error {
+	return m.stateSync.EnableServerMode()
+}
+
+func (m *consensusModule) DisableServerMode() error {
+	return m.stateSync.DisableServerMode()
+}
+
+// TODO! Implement this, placeholder function
```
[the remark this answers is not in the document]
gokutheengineer: done using the `getValidatorsAtHeight(uint64)` method.

shared/modules/consensus_module.go
```diff
@@ -36,7 +37,8 @@ type ConsensusModule interface {
 	CurrentStep() uint64
 
 	// State Sync functions
-	EnableServerMode()
+	EnableServerMode() error
```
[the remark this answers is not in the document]
gokutheengineer: I totally agree, but see the previous comment. I will be checking FSM lib more throughly, and try to find a solution.

shared/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.0.35] - 2023-02-27
+
+- Add FSMConsensusEvents() interface to handle state transition events in consensus module
```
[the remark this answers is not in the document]
gokutheengineer: Yeah totally makes sense, there is no need to export these functions. Rather, I generated` HandleStateTransitionEvent(transitionMessageAny *anypb.Any) error` function that calls these functions inside the consensus module.

consensus/module.go
```diff
@@ -317,6 +332,19 @@ func (m *consensusModule) loadPersistedState() error {
 	return nil
 }
 
-func (m *consensusModule) EnableServerMode() {
-	m.stateSync.EnableServerMode()
+func (m *consensusModule) EnableServerMode() error {
```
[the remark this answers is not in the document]
gokutheengineer: That is right, they are not helpful removed. Thanks for pointing it out.

consensus/pacemaker/module.go
```diff
@@ -114,11 +115,24 @@ func (m *pacemaker) ShouldHandleMessage(msg *typesCons.HotstuffMessage) (bool, e
...
+	// If this case happens, there are two possibilities:
+	// 1. The node is behind and needs to catch up, so state sync should start
+	// 2. The leader is sending a malicious proposal, node needs to reject the proposal.
+	// For both cases, nodes needs to ask peers for the metadata, and:
+	// if the node is behind, he should start state sync, else node should reject the block proposal.
 	if msg.Height > currentHeight {
-		m.logger.Info().Msgf("⚠️ [WARN][DISCARDING] ⚠️ Node at height %d > message height %d", currentHeight, msg.Height)
-		return false, nil
+		m.logger.Info().Msgf("⚠️ [WARN] ⚠️ Node at height %d > message height %d", currentHeight, msg.Height)
+		if err := m.GetBus().GetStateMachineModule().SendEvent(coreTypes.StateMachineEvent_Consensus_IsCaughtUpNonValidator); err != nil {
```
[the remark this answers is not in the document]
gokutheengineer: that is correct, fixed along with the previous comment. Your previous comment is the state change here, thanks for pointing it out.

state_machine/fsm.go
```diff
@@ -48,13 +48,33 @@ func NewNodeFSM(callbacks *fsm.Callbacks, options ...func(*fsm.FSM)) *fsm.FSM {
...
+			{
+				Name: string(coreTypes.StateMachineEvent_Consensus_IsCaughtUpNonValidator),
 				Src: []string{
-					string(coreTypes.StateMachineState_P2P_Bootstrapped),
 					string(coreTypes.StateMachineState_Consensus_SyncMode),
 				},
 				Dst: string(coreTypes.StateMachineState_Consensus_Synced),
 			},
+			{
+				Name: string(coreTypes.StateMachineEvent_Consensus_IsUnsynched),
```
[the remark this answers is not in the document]
gokutheengineer: updated.

consensus/state_sync/module.go
gokutheengineer: That is right, I left it there to clarify the process for the next steps, removed with the new structure.

consensus/e2e_tests/utils_test.go
```diff
@@ -98,6 +98,8 @@ func CreateTestConsensusPocketNode(
 	bus modules.Bus,
```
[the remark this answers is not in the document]
gokutheengineer: I added the first version of the test, please let me know how it can be improved.

consensus/fsm_handler.go
gokutheengineer: I think this is not correct. Because, a validator transitions to this state after it synchs with the rest of the network. Validator stays in Pacemaker mode until it receives a block proposal that is higher than its current height, and it transitions to the unsynched state. 

Lmk what do you think.

consensus/pacemaker/module.go
```diff
@@ -114,11 +115,24 @@ func (m *pacemaker) ShouldHandleMessage(msg *typesCons.HotstuffMessage) (bool, e
...
+			typesCons.ErrSendingStateTransitionEvent(coreTypes.StateMachineEvent_Consensus_IsUnsynched)
+			return false, err
+		}
+
+		// at this point node is synched. Therefore, the block height should not be higher than node's current height
+		// if not node must reject the proposal since it must be a malicious proposal
+		if msg.Height > currentHeight {
```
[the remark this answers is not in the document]
gokutheengineer: Changed this logic, please see the new version.

consensus/fsm_handler.go
```diff
@@ -0,0 +1,102 @@
...
+
+// Sync mode is when the node (validator or non-valdiator) is syncing with the rest of the network
+func (m *consensusModule) HandleSync(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Debug().Msg("FSM is in Sync Mode, starting syncing")
+
+	m.stateSync.AggregateMetadataResponses()
+
+	// try sycing until node is synced
+	// CONSIDER: consider putting a limit on number of tries, or timeout
+	err := m.stateSync.Sync()
```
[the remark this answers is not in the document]
gokutheengineer: That is a great input, and I modified. There can be still some changes to make, so please have a look at the new version and let me know.

consensus/state_sync/helpers.go
```diff
@@ -6,6 +6,12 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
+// TODO! Implement, a placeholder function
+func (m *stateSync) BroadCastStateSyncMessage(stateSyncMsg *typesCons.StateSyncMessage, height uint64) error {
```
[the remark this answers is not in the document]
gokutheengineer: Thanks for the suggestion, I updated comments.

shared/node.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/pokt-network/pocket/state_machine"
 	"github.com/pokt-network/pocket/telemetry"
 	"github.com/pokt-network/pocket/utility"
+	"go.uber.org/multierr"
```
Olshansk, line 17: cc @bryanchriswhite 

@gokutheengineer is actually adding it right now. Let's add a TODO to remove this one we move to Go 1.2: https://stackoverflow.com/a/74645028/768439

shared/modules/consensus_module.go
```diff
@@ -74,6 +73,8 @@ type ConsensusPacemaker interface {
 type ConsensusStateSync interface {
 	GetNodeIdFromNodeAddress(string) (uint64, error)
 	GetNodeAddress() string
+	IsSynched() (bool, error)
+	GetValidatorsAtHeight(uint64) ([]*types.Actor, error)
```
Olshansk, line 77: Why are we exposing this? I couldn't find any usages outside of the `consensus` package.

General piece of feedback: I know it's really easy to add this in the shared interface, but always think twice if its necessary.

shared/modules/consensus_module.go
```diff
@@ -85,4 +86,6 @@ type ConsensusDebugModule interface {
 	SetStep(uint8)
 	SetBlock(*types.Block)
 	SetUtilityContext(UtilityContext)
+
+	SetAggregatedMetadata(uint64, uint64, string)
```
Olshansk, line 90: Add param names in the interface header so its easier to understand

shared/modules/consensus_module.go
```diff
@@ -74,6 +73,8 @@ type ConsensusPacemaker interface {
 type ConsensusStateSync interface {
 	GetNodeIdFromNodeAddress(string) (uint64, error)
 	GetNodeAddress() string
+	IsSynched() (bool, error)
+	GetValidatorsAtHeight(uint64) ([]*types.Actor, error)
 }
 
 type ConsensusDebugModule interface {
```
Olshansk, line 80: Can you add some comments on where/how the debug interface is intended to be used.

state_machine/docs/README.md
```diff
@@ -61,4 +61,8 @@ If you make any changes to it, you can re-generate it via:
 make generate_node_state_machine_diagram
 ```
 
+
+## State Machine Lifecycle
+Node FSM starts from `stopped` state, and transitions to first `starting` and then to the `bootstapping` state, which is handled by the P2P module. In `bootstrapping` state, the node fills its address book, and upon completing mode transitions to `unsynched` state and then to `sync mode` to start requesting block metadata and receive the blocks it is missing. Upon completing syncing its state in `sync mode`; if the node is a validator node transitions to `pacemaker` state and starts participating in the consensus state, else if the node is a non-validator, node transitions to the `synced` state. However, it is important to node that a non-validator node practically always stays in the `snyc mode` to keep getting blocks from the peers. The node transitions to the `unsynched` state if it stays behind on the network, i.e. in `pacemaker` mode if the node receives a block proposal which is higher than it's current height.
```
Olshansk, line 66: This is a lot of condense text that's really hard to read. I'd recommend two things:
1. User a mermaid flow chart (or link to the fsm one)
2. Use bullet points

For example:

1. **Node:** `stopped` -> `starting` -> `bootstrapping`
2. **P2P** `bootstrapping`: fill address book -> `unsynched` -> `sync mode`
3. **Consensus** `synching`: request & receiving blocks

I might be biased, but I think its easier to read. Also found some small typos. E.g. `snyc mode`

consensus/state_sync/module.go
```diff
@@ -12,47 +14,50 @@ const (
...
-
 type StateSyncModule interface {
 	modules.Module
 	StateSyncServerModule
+	DebugStateSync
 
-	// Handle a metadata response from a peer so this node can update its local view of the state
-	// sync metadata available from its peers
-	HandleStateSyncMetadataResponse(*typesCons.StateSyncMetadataResponse) error
+	IsServerModEnabled() bool
```
Olshansk, line 22: Adding some small comments around the grouped methods is simple but would help

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -236,12 +236,12 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
 	advanceTime(t, clockMock, paceMakerTimeout+10*time.Millisecond)
 }
 
-// TODO: Implement these tests and use them as a starting point for new ones. Consider using ChatGPT to help you out :)
-
 func TestPacemakerDifferentHeightsCatchup(t *testing.T) {
 	t.Skip()
 }
 
+// TODO: Implement these tests and use them as a starting point for new ones. Consider using ChatGPT to help you out :)
```
Olshansk, line 243: Revert

consensus/debugging.go
```diff
@@ -36,6 +36,14 @@ func (m *consensusModule) SetUtilityContext(utilityContext modules.UtilityContex
 	m.utilityContext = utilityContext
 }
 
+func (m *consensusModule) SetAggregatedMetadata(maxHeight, minHeight uint64, peerAddress string) {
```
Olshansk, line 39: Are we using this anywhere?

consensus/e2e_tests/utils_test.go
```diff
@@ -45,6 +46,14 @@ const (
 
 type IdToNodeMapping map[typesCons.NodeId]*shared.Node
 
+// Used to mock state sync
+type StateSyncMockHelper struct {
```
Olshansk, line 50: Does this need to be exposed outside of this package?

consensus/e2e_tests/utils_test.go
```diff
@@ -45,6 +46,14 @@ const (
...
+	maxheight uint64 // the max height of the node
+	minheight uint64 // the min height of the node
```
Olshansk, line 52: s/minheight/minHeight

consensus/helpers.go
```diff
@@ -83,7 +80,7 @@ func (m *consensusModule) getQuorumCertificate(height uint64, step typesCons.Hot
 		pss = append(pss, msg.GetPartialSignature())
 	}
 
-	validators, err := m.getValidatorsAtHeight(height)
+	validators, err := m.GetValidatorsAtHeight(height)
```
Olshansk, line 83: I don't think `GetValidatorsAtHeight` should be exposed by `Consensus`. Feels like persistence layer logic

consensus/helpers.go
```diff
@@ -317,3 +314,18 @@ func (m *consensusModule) getValidatorsAtHeight(height uint64) ([]*coreTypes.Act
 
 	return persistenceReadContext.GetAllValidators(int64(height))
 }
+
+func (m *consensusModule) IsValidator() (bool, error) {
```
Olshansk, line 318: This seems very inefficient, but also not the primary goal of the PR.

Can you add a TODO that would do one of two things:
1. Single query to the DB in the persistence module
2. Cache it in the `consensus` module

consensus/module.go
```diff
@@ -317,6 +318,22 @@ func (m *consensusModule) loadPersistedState() error {
 	return nil
 }
 
-func (m *consensusModule) EnableServerMode() {
-	m.stateSync.EnableServerMode()
+func (m *consensusModule) IsSynched() (bool, error) {
```
Olshansk, line 321: In the interface, add a comment what this does: `Compares the local persisted state against the aggregated metadata from peers. There is never a 100% guarantee that it's synched`.

In the codebase, just say `IsSynched implements the interface function for ...`. 

Bryan's PR in #558 follows a goo pattern for this.

consensus/module.go
```diff
@@ -317,6 +318,22 @@ func (m *consensusModule) loadPersistedState() error {
...
+	lastPersistedBlockHeight := m.GetBus().GetConsensusModule().CurrentHeight() - 1
+	persistenceContext, err := m.GetBus().GetPersistenceModule().NewReadContext(int64(lastPersistedBlockHeight))
+	if err != nil {
+		return false, err
+	}
+	defer persistenceContext.Close()
+
+	maxHeight, err := persistenceContext.GetMaximumBlockHeight()
```
Olshansk, line 329: s/maxHeight/maxPersistedHeight

consensus/module.go
```diff
@@ -317,6 +318,22 @@ func (m *consensusModule) loadPersistedState() error {
...
+	if err != nil {
+		return false, err
+	}
+
+	if maxHeight == m.stateSync.GetAggregatedSyncMetadata().MaxHeight {
```
Olshansk, line 334: Something like this is easier to read

```go
maxSeenHeight = m.stateSync.GetAggregatedSyncMetadata().MaxHeight

return maxSeenHeight == maxPersistedHeight, nil
```

consensus/state_sync/helpers.go
```diff
@@ -6,6 +6,25 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
+// TODO(#352): Implement this function, currently a placeholder.
+// Helper function for broadcasting state sync messages to the all peers known to the node
+// It is used for:
+//
+//		requesting for metadata, via the periodicMetaDataSynch() function
+//	 	requesting for blocks, via the StartSynching() function
+func (m *stateSync) broadCastStateSyncMessage(stateSyncMsg *typesCons.StateSyncMessage, height uint64) error {
```
Olshansk, line 15: broadCast is one word -> `broadcastStateSyncMessage`

consensus/pacemaker/module.go
```diff
@@ -114,11 +115,28 @@ func (m *pacemaker) ShouldHandleMessage(msg *typesCons.HotstuffMessage) (bool, e
...
-	// Current node is out of sync
+	// If this case happens, there are two possibilities:
+	// 1. The node is behind and needs to catch up, node must start syncing,
+	// 2. The leader is sending a malicious proposal.
+	// There, for both cases, node rejects the proposal, because:
+	// 1. If node is out of sync, node can't verify the block proposal, so rejects it. But node will eventually sync with the rest of the network and add the block.
+	// 2. If node is synched, node must reject the proposal because proposal is not valid.
 	if msg.Height > currentHeight {
-		m.logger.Info().Msgf("⚠️ [WARN][DISCARDING] ⚠️ Node at height %d > message height %d", currentHeight, msg.Height)
+		m.logger.Info().Msgf("⚠️ [WARN] ⚠️ Node at height %d > message height %d", currentHeight, msg.Height)
```
Olshansk, line 125: The comment `>` is incorrect

consensus/pacemaker/module.go
```diff
@@ -114,11 +115,28 @@ func (m *pacemaker) ShouldHandleMessage(msg *typesCons.HotstuffMessage) (bool, e
...
+		isSynched, err := m.GetBus().GetConsensusModule().IsSynched()
+		if err != nil {
+			return false, err
+		}
+
+		if !isSynched {
+			err = m.GetBus().GetStateMachineModule().SendEvent(coreTypes.StateMachineEvent_Consensus_IsUnsynched)
+			if err != nil {
+				return false, err
+			}
```
Olshansk, line 135:
```suggestion
			err = m.GetBus().GetStateMachineModule().SendEvent(coreTypes.StateMachineEvent_Consensus_IsUnsynched)
			return false, err
```

consensus/state_sync/helpers.go
```diff
@@ -6,6 +6,25 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
+// TODO(#352): Implement this function, currently a placeholder.
+// Helper function for broadcasting state sync messages to the all peers known to the node
+// It is used for:
+//
+//		requesting for metadata, via the periodicMetaDataSynch() function
```
Olshansk, line 13: 1. Metadata is one word -> `MetaData` -> `Metadata`
2. An easier read: `requests for metadata using `periodicMetadataSync`; ditto below

consensus/state_sync/helpers.go
```diff
@@ -6,6 +6,25 @@ import (
...
+//	 	requesting for blocks, via the StartSynching() function
+func (m *stateSync) broadCastStateSyncMessage(stateSyncMsg *typesCons.StateSyncMessage, height uint64) error {
+	m.logger.Info().Fields(
+		map[string]any{
+			"height": height,
+			"nodeId": m.GetBus().GetConsensusModule().GetNodeId(),
+		},
+	).Msg("📣 Broadcasting state sync message... 📣")
+
+	_ = stateSyncMsg
```
Olshansk, line 23: Add a TODO above that this is temporary

consensus/state_sync/module.go
```diff
@@ -12,47 +14,50 @@ const (
...
-	// and catch up to the global world state
+	HandleStateSyncMetadataResponse(*typesCons.StateSyncMetadataResponse) error
 	HandleGetBlockResponse(*typesCons.GetBlockResponse) error
 
-	IsServerModEnabled() bool
-	EnableServerMode()
-
 	SendStateSyncMessage(msg *typesCons.StateSyncMessage, nodeAddress cryptoPocket.Address, height uint64) error
+
+	GetSyncMetadataBuffer() []*typesCons.StateSyncMetadataResponse
```
Olshansk, line 31: Do we need to expose both functions? I was thinking that one could (potentially) be enough?

I'm still looking through the rest of the code though

consensus/state_sync/module.go
```diff
@@ -12,47 +14,50 @@ const (
...
+	GetAggregatedSyncMetadata() *typesCons.StateSyncMetadataResponse
```
Olshansk, line 32: Can you replace `SyncMetadata` with `StateSyncMetadata` so its explicit

consensus/fsm_handler.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/pokt-network/pocket/shared/codec"
+	coreTypes "github.com/pokt-network/pocket/shared/core/types"
+	"github.com/pokt-network/pocket/shared/messaging"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+func (m *consensusModule) HandleStateTransitionEvent(transitionMessageAny *anypb.Any) error {
+	m.m.Lock()
+	defer m.m.Unlock()
+	m.logger.Info().Msgf("Received a state transition message: %s", transitionMessageAny)
```
Olshansk, line 16: Given your other in flight PR, we can probably improve this, right?

consensus/fsm_handler.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	coreTypes "github.com/pokt-network/pocket/shared/core/types"
+	"github.com/pokt-network/pocket/shared/messaging"
+)
+
+func (m *consensusModule) handleStateMachineTransitionEvent(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Info().Msgf("I am starting to handle  handleStateMachineTransitionEvent: %s", msg)
+
+	fsm_state := msg.NewState
+	m.logger.Debug().Fields(map[string]any{
```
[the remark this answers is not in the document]
Olshansk: Add a TODO linking to the issue created

consensus/fsm_handler.go
```diff
@@ -0,0 +1,111 @@
...
+
+	case coreTypes.StateMachineState_Consensus_SyncMode:
+		return m.HandleSyncMode(msg)
+
+	case coreTypes.StateMachineState_Consensus_Synced:
+		return m.HandleSynced(msg)
+
+	case coreTypes.StateMachineState_Consensus_Pacemaker:
+		return m.HandlePacemaker(msg)
+	default:
```
Olshansk, line 61:
```suggestion

		default:
```

consensus/fsm_handler.go
```diff
@@ -0,0 +1,111 @@
...
+		m.logger.Warn().Msg("Consensus module not handling this event")
```
Olshansk, line 62: Which event? :)

consensus/state_sync/docs/State_Sync.md
```diff
@@ -0,0 +1,7 @@
+## State Sync Lifecycle
+
+Node starts syching with the rest of the network upon starting, as an external process via the `periodicMetaDataSynch()` function. Node keeps adding the metadata information it receives in the `syncMetadataBuffer` field of the stateSync struct via the `HandleStateSyncMetadataResponse()`.
```
Olshansk, line 3: periodicMetaDataSynch

-> periodic Metadata Sync

consensus/state_sync/docs/State_Sync.md
```diff
@@ -0,0 +1,7 @@
+## State Sync Lifecycle
```
Olshansk, line 1: We already have `consensus/doc/PROTOCOL_STATE_SYNC.md`.

Can you update that doc appropriate rather than creating a new one?

consensus/state_sync/docs/State_Sync.md
```diff
@@ -0,0 +1,7 @@
...
+
+Upon receiving a block, validator node checks the node's height, and if it's higher than it's current round, it checks if it is out of synch via `IsSynched()` function, which triggers aggregation of the metadata responses in the `syncMetadataBuffer` to the `aggregatedSyncMetadata`, compares with the current node height, and returns true if node is out of synch. In this case, node sends `StateMachineEvent_Consensus_IsUnsynched` event, which in turn, through FSM state transitions, triggers `StartSynching()` function. Or, if the node is in synch with its peers, then it rejects the block proposal. 
+
+`StartSynching()` function requests block one by one using the minimum and maximum height info in the `aggregatedSyncMetadata` field, using the `broadCastStateSyncMessage()` function. The node, receives the requested blocks from the peers and applies the "correct" block to the persistence via `HandleGetBlockResponse()` function which checks quorum certificate and block validity.
```
Olshansk, line 7: @gokutheengineer Can you try to take another stab at the documentation after you port it over to the proper state sync doc at consensus/doc/PROTOCOL_STATE_SYNC.md?

I realize writing all of this text is "easier" and much easier to understand for the you (the author) or someone who is in the weeds, but we want to make it really easy for anyone:
- Moe bullet points
- Less text
- More diagrams

consensus/fsm_handler.go
Olshansk: I think either works. Seems like you designed it so that a validator goes from synched -> pacemaker, right?

If so, feel free to update the comment appropriately.

consensus/fsm_handler.go
```diff
@@ -0,0 +1,111 @@
...
+func (m *consensusModule) HandleUnsynched(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Debug().Msg("FSM is in Unsyched state, as node is out of sync sending syncmode event to start syncing")
+
+	return m.GetBus().GetStateMachineModule().SendEvent(coreTypes.StateMachineEvent_Consensus_IsSyncing)
+}
+
+// HandleSyncMode handles FSM event Consensus_IsSyncing, and SyncMode is the destination state.
+// In Sync mode node (validator or non-validator) starts syncing with the rest of the network.
+func (m *consensusModule) HandleSyncMode(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Debug().Msg("FSM is in Sync Mode, starting syncing...")
```
Olshansk, line 90:
```suggestion
	m.logger.Debug().Msg("FSM is in Sync Mode, start syncing...")
```

consensus/fsm_handler.go
```diff
@@ -0,0 +1,111 @@
...
+// HandleSynced handles FSM event IsSynchedNonValidator, and Synced is the destination state.
+// Currently, FSM never transition to this state and a non-validator node always stays in syncmode.
+// CONSIDER: when a non-validator sync is implemented, maybe there is a case that requires transitioning to this state.
+// TODO: Add check that this never happens when IsValidator() is false, i.e. node is not validator.
+func (m *consensusModule) HandleSynced(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Debug().Msg("FSM of non-validator node is in Synced mode")
+	return nil
+}
+
+// HandlePacemaker handles FSM event IsSynchedValidator, and Pacemaker is the destination state.
```
Olshansk, line 104: `Synched` and `Sync` are correct. `Synch` is not.

Leaving one comment here (where there is no delta), but just FYI.

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,130 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
 	_, err = WaitForNetworkStateSyncEvents(t, clockMock, eventsChannel, errMsg, numExpectedMsgs, 250, false)
 	require.Error(t, err)
 }
+
+func TestStateSync_UnsynchedPeerSynchs_Success(t *testing.T) {
```
Olshansk, line 185: Even if you don't add more tests, think of other tests you want to implement and add headers for them. I can try to think of a few once you take another stab at it.

For example, having 4 nodes, all at different heights, and validating that they all catch up

consensus/e2e_tests/utils_test.go
```diff
@@ -45,6 +46,14 @@ const (
...
+}
+
+var stateSyncHelper StateSyncMockHelper
```
Olshansk, line 55: Global variables are usually a really big code smell. I know it's "easy" and gets it working, but it becomes really hard to maintain & understand really fast.

Can you think of a way to make it more functional? How we pass everything around through variables.

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,130 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+		consensusModImpl.MethodByName("SetUtilityContext").Call([]reflect.Value{reflect.ValueOf(utilityContext)})
+	}
+
+	// Debug message to start consensus by triggering first view change
+	for _, pocketNode := range pocketNodes {
+		TriggerNextView(t, pocketNode)
+	}
+	advanceTime(t, clockMock, 10*time.Millisecond)
+
+	// Assert that unsynched node has a separate view of the network than the rest of the nodes
```
Olshansk, line 260: Nice one

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,130 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+		P2PBroadcast(t, pocketNodes, message)
+	}
+	advanceTime(t, clockMock, 10*time.Millisecond)
+
+	// 2. Propose
+	numExpectedMsgs := numberOfValidators
+	_, err = WaitForNetworkConsensusEvents(t, clockMock, eventsChannel, consensus.Prepare, consensus.Propose, numExpectedMsgs, 250, true)
+	require.NoError(t, err)
+
+	for nodeId, pocketNode := range pocketNodes {
```
Olshansk, line 299: Nice

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,130 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+					Round:  uint8(testRound + 1),
+				},
+				nodeState)
+		}
+		require.Equal(t, false, nodeState.IsLeader)
+		require.Equal(t, typesCons.NodeId(0), nodeState.LeaderId)
+	}
+
+	// Mock the unsynched node's periodic metadata sync
+	MockPeriodicMetaDataSynch(testHeight, 1)
```
Olshansk, line 287: Try to avoid the global var. I'll think about this a bit more and can help out if you can't find a solution.

Olshansk (changes requested):
Great progress. Still a bit of work todo, but it's really coming together

shared/node.go
bryanchriswhite: 🙌

shared/modules/consensus_module.go
```diff
@@ -29,14 +29,13 @@ type ConsensusModule interface {
 	HandleMessage(*anypb.Any) error
 	// State Sync messages Handler
 	HandleStateSyncMessage(*anypb.Any) error
+	// FSM transition events handler
+	HandleStateTransitionEvent(transitionMessageAny *anypb.Any) error
```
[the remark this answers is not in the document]
gokutheengineer: Thanks for pointing this out, I will be doing this, appreciated the feedback.

shared/modules/consensus_module.go
gokutheengineer: Very good point, thanks. Removed.

consensus/debugging.go
gokutheengineer: I left this for the issue #352,  and for testing. I will remove it, and I will try to avoid adding it back, in parallel to your suggestion of not having more than necessary shared functions.

consensus/helpers.go
gokutheengineer: Agreed, removed, thanks.

consensus/fsm_handler.go
gokutheengineer: Correct, added the comment.

consensus/state_sync/module.go
gokutheengineer: I removed `GetSyncMetadataBuffer()` as issue #352 doesn't need it anymore, so that's right.

consensus/module.go
gokutheengineer: Thanks!

consensus/fsm_handler.go
Olshansk: Screenshot 2023-03-10 at 11 33 56 AM

consensus/e2e_tests/state_sync_test.go
gokutheengineer: I added few more test, I may update again.

consensus/e2e_tests/utils_test.go
```diff
@@ -494,13 +494,37 @@ func baseRpcMock(t *testing.T, _ modules.EventsChannel) *mockModules.MockRPCModu
...
+
+	stateMachineMock.EXPECT().SendEvent(gomock.Any()).DoAndReturn(func(event coreTypes.StateMachineEvent, args ...any) error {
+		switch coreTypes.StateMachineEvent(event) {
+		case coreTypes.StateMachineEvent_Consensus_IsUnsynched:
+			t.Logf("Node is unsynched")
+			return bus.GetStateMachineModule().SendEvent(coreTypes.StateMachineEvent_Consensus_IsSyncing)
+		case coreTypes.StateMachineEvent_Consensus_IsSyncing:
+			maxHeight := bus.GetConsensusModule().GetAggregatedStateSyncMetadataMaxHeight()
+			t.Logf("Node is syncing")
+			consensusModImpl.MethodByName("SetHeight").Call([]reflect.Value{reflect.ValueOf(maxHeight)})
```
Olshansk, line 514: Ditto the same comments I left above about casting and reflections.

consensus/e2e_tests/utils_test.go
```diff
@@ -494,13 +494,37 @@ func baseRpcMock(t *testing.T, _ modules.EventsChannel) *mockModules.MockRPCModu
...
 
-func baseStateMachineMock(t *testing.T, _ modules.EventsChannel) *mockModules.MockStateMachineModule {
+func baseStateMachineMock(t *testing.T, _ modules.EventsChannel, bus modules.Bus) *mockModules.MockStateMachineModule {
 	ctrl := gomock.NewController(t)
 	stateMachineMock := mockModules.NewMockStateMachineModule(ctrl)
 	stateMachineMock.EXPECT().Start().Return(nil).AnyTimes()
 	stateMachineMock.EXPECT().SetBus(gomock.Any()).Return().AnyTimes()
 	stateMachineMock.EXPECT().GetModuleName().Return(modules.StateMachineModuleName).AnyTimes()
 
+	consensusModImpl := reflect.ValueOf(bus.GetConsensusModule())
```
Olshansk, line 504: Why don't we just send the event directly on the events channel which is already being passed into this function?

consensus/fsm_handler.go
```diff
@@ -0,0 +1,116 @@
...
+		return m.handleStateTransitionEvent(stateTransitionMessage)
+	default:
+		return typesCons.ErrUnknownStateSyncMessageType(transitionMessageAny.MessageName())
+	}
+}
+
+func (m *consensusModule) handleStateTransitionEvent(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Info().Msgf("Begin handling StateMachineTransitionEvent: %s", msg)
+
+	// TODO (#571): update with logger helper function
```
Olshansk, line 42: You already did this, so why not use the `TransitionEventToMap` helper?

shared/modules/consensus_module.go
```diff
@@ -74,8 +74,13 @@ type ConsensusPacemaker interface {
 type ConsensusStateSync interface {
 	GetNodeIdFromNodeAddress(string) (uint64, error)
 	GetNodeAddress() string
+
+	// Compares the persisted state with the aggregated state of the network. If the persisted state is behind the network state, i.e. that node is not synched, it will return false.
```
Olshansk, line 78: Use proper godoc comments 
```suggestion
	// IsSynched compares the persisted state with the aggregated state of the network. If the persisted state is behind the network state, i.e. that node is not synched, it will return false.
```

shared/modules/consensus_module.go
```diff
@@ -85,4 +90,7 @@ type ConsensusDebugModule interface {
 	SetStep(uint8)
 	SetBlock(*types.Block)
 	SetUtilityContext(UtilityContext)
+
+	SetAggregatedStateSyncMetadata(minHeight, maxHeight uint64, peerAddress string)
```
Olshansk, line 94: Should this be `Add` because it's only for a single peer?

consensus/fsm_handler.go
```diff
@@ -0,0 +1,102 @@
...
+		return m.HandlePacemaker(msg)
+	}
+
+	return nil
+}
+
+// Bootrstapped mode is when the node (validator or non-valdiator) is out of sync with the rest of the network
+// This mode is a transition mode from node bootstrappin to node being out-of-sync.
+func (m *consensusModule) HandleBootstrapped(msg *messaging.StateMachineTransitionEvent) error {
+	m.logger.Debug().Msg("FSM is in bootstrapped state, so it is out of sync, and transitions to unsynched mode")
```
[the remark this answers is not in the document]
Olshansk: @gokutheengineer You resolved this comment but never made the change.

Screenshot 2023-03-23 at 8 16 51 PM

Olshansk (changes requested):
@gokutheengineer I reviewed half of this, but it still needs some work:

1. Please do a full self review before I look at it again.
2. If you resolve a comment, please make the corresponding change or explain why you're not doing it if you do not make the corresponding change.

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,132 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+					Step:   testStep,
+					Round:  uint8(testRound + 1),
+				},
+				nodeState)
+		}
+		require.Equal(t, false, nodeState.IsLeader)
+		require.Equal(t, typesCons.NodeId(0), nodeState.LeaderId)
+	}
+
+	unsynchedNodeModImpl.MethodByName("SetAggregatedStateSyncMetadata").Call([]reflect.Value{reflect.ValueOf(uint64(1)), reflect.ValueOf(testHeight), reflect.ValueOf(string(consensusPK.Address()))})
```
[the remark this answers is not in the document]
gokutheengineer: Updated, thanks for pointing it out.

shared/modules/consensus_module.go
gokutheengineer: This is used to update peer's aggregated metadata in testing scenarios, to simulate periodic metadata synchronization. So I think in this sense `Set` makes more sense, wdyt?

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,132 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+}
+
+func TestStateSync_Unsynched4PeersSynchs_Success(t *testing.T) {
+	t.Skip()
+}
+
+// Test if unsynched peer can catch up with the rest of the network
+func TestStateSync_UnsynchedPeerCatchsUpConsensus_ConsensusProceedsSuccessfully(t *testing.T) {
+	t.Skip()
+}
```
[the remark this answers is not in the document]
gokutheengineer: Added this one. I added the tests I already implemented in #352, as TODOs for this PR. Also added couple of more which I think has a value, but also can be coded via the help of ChatGPT. I will be actively thinking about more test scenarios that can have high value.

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,132 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+			nodeState)
+	}
+}
+
+func TestStateSync_Unsynched4PeersSynchs_Success(t *testing.T) {
+	t.Skip()
+}
+
+// Test if unsynched peer can catch up with the rest of the network
+func TestStateSync_UnsynchedPeerCatchsUpConsensus_ConsensusProceedsSuccessfully(t *testing.T) {
```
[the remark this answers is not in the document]
gokutheengineer: You are right. I removed that one and added the tests I already implemented for #352.

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,132 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+		PrevStateHash:     "",
+		ProposerAddress:   consensusPK.Address(),
+		QuorumCertificate: nil,
+	}
+	block := &coreTypes.Block{
+		BlockHeader:  blockHeader,
+		Transactions: make([][]byte, 0),
+	}
+
+	leaderConsensusModImpl := GetConsensusModImpl(leader)
```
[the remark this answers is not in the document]
gokutheengineer: Not using reflections, removed that lines.

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,132 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+		} else {
+			assertNodeConsensusView(t, nodeId,
+				typesCons.ConsensusNodeState{
+					Height: testHeight,
+					Step:   testStep,
+					Round:  uint8(testRound + 1),
+				},
+				nodeState)
+		}
+		require.Equal(t, false, nodeState.IsLeader)
```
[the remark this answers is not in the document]
gokutheengineer: No, because currently the check is performed on the `NewRound` messages received from all peers. Nodes will set the leader once we broadcast NewRound responses to all peers, in the following lines. I added a check to the next `assertNodeConsensusView()` for the `leaderId`.

consensus/e2e_tests/state_sync_test.go
```diff
@@ -179,3 +181,132 @@ func TestStateSync_ServerGetBlock_FailNonExistingBlock(t *testing.T) {
...
+	buses := GenerateBuses(t, runtimeMgrs)
+
+	// Create & start test pocket nodes
+	eventsChannel := make(modules.EventsChannel, 100)
+	pocketNodes := CreateTestConsensusPocketNodes(t, buses, eventsChannel)
+	StartAllTestPocketNodes(t, pocketNodes)
+
+	// Prepare leader info
+	leaderId := typesCons.NodeId(3)
+	require.Equal(t, uint64(leaderId), testHeight%uint64(numberOfValidators)) // Uses our deterministic round robin leader election
```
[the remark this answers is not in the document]
gokutheengineer: Thanks, updated. I added a new debugging function to the Consensus module `GetLeaderElectionResult()`.  It calls the `ElectNextLeader` function by forming a hotstuff message using the given height, round, and step, and returns the node id that will be elected as the next leader. 

In the tests, leaderId is set by calling the `GetLeaderElectionResult()` function.

consensus/e2e_tests/state_sync_test.go
```diff
@@ -289,24 +286,43 @@ func TestStateSync_UnsynchedPeerSynchs_Success(t *testing.T) {
...
+		fmt.Printf("Node state, node id: %d, height: %d, round: %d, step: %d, leaderId: %d \n", nodeId, nodeState.Height, nodeState.Round, nodeState.Step, nodeState.LeaderId)
 		assertNodeConsensusView(t, nodeId,
 			typesCons.ConsensusNodeState{
 				Height: testHeight,
 				Step:   uint8(consensus.Prepare),
-				Round:  uint8(testRound + 1),
+				Round:  uint8(currentRound),
 			},
 			nodeState)
+		require.Equal(t, leaderId, nodeState.LeaderId)
```
Olshansk, line 300: The issue lies in `electNextLeaderDeterministicRoundRobin`. I don't know why, but if you add the following:

```go
	fmt.Println("OLSH DEBUG", value, numVals, int64(message.Height), int64(message.Round), int64(message.Step))
```

You'll see that `message.Height` is `2` instead of `3` when the failure happens. Try to look into how this happens and hopefully you'll uncover the source of the bug.

----

Also, I noticed that we don't lock in these two places. I'm always averse to locking, so I'm not saying we need to, but asking you to look into it in the context of state sync.

```bash
--- a/consensus/pacemaker_consensus.go
+++ b/consensus/pacemaker_consensus.go
@@ -61,6 +61,9 @@ func (m *consensusModule) IsLeaderSet() bool {
 }

 func (m *consensusModule) NewLeader(msg *anypb.Any) error {
+       m.m.Lock()
+       defer m.m.Unlock()
+
```

```bash
+++ b/consensus/debugging.go
@@ -47,6 +47,9 @@ func (m *consensusModule) GetAggregatedStateSyncMetadataMaxHeight() (maxHeight u
 }

 func (m *consensusModule) GetLeaderElectionResult(height, round uint64, step uint8) uint64 {
+       m.m.Lock()
+       defer m.m.Unlock()
+
```

Olshansk (changes requested):

consensus/e2e_tests/state_sync_test.go
Olshansk: > I think the best thing to do now is to remove this test from this PR, and rather have it in #352 where the block actually receives blocks and stores. There, we must ensure leader election is performed upon syncing up to the latest state. Wdyt?

I would disagree.

This PR is almost 3 weeks of work and `TestStateSync_UnsynchedPeerSynchs_Success` is the only unit test we added. Let me know if I'm incorrect, but if it's the case, removing it is not an option.

Whenever only 1 (or few) unit tests are written, and they're difficult to write, incorrect or are flaky, it's a signal that the underlying work needs to be improved.

Olshansk (approved):

Olshansk:
@gokutheengineer Please do the following before merging it in:

1. Create a ticket to track the "Wait for FSM" events test helper we discussed.
2. Review the changes I pushed for your verification
3. Make sure no changes from other PRs you're working on are included in this branch when you merge
4. Focus on downloading and applying blocks after the merge

PR #529, opened 2023-02-20
patch: localnet dev QOL improvements

0xBigBoss:
These changes were helpful for me to dev on an Ubuntu server.

- `docker-compose` is deprecated and will not be supported anymore after June 2023. This prepares for that by also checking for v2 `docker compose` cli plugin
- add example .env.example for docker compose to allow port overrides on crowded host machines
- clean up the `p1` alias by adding a bash command to update your `~/.bash_aliases`


build/deployments/docker-compose.yaml
```diff
@@ -42,10 +42,10 @@ services:
       - "9000"
       - "50832"
     ports:
-      - "7081:7081"
-      - "8081:8080"
-      - "9081:9080"
-      - "50832:50832"
+      - "${NODE1_PORT_7081:-0.0.0.0:7081}:7081"
+      - "${NODE1_PORT_8081:-0.0.0.0:8080}:8080"
+      - "${NODE1_PORT_9081:-0.0.0.0:9080}:9080"
```
Olshansk, line 47:
```suggestion
      - "${NODE1_PORT_8081:-0.0.0.0:8081}:8080"
      - "${NODE1_PORT_9081:-0.0.0.0:9081}:9080"
```

build/deployments/.env.example
```diff
@@ -0,0 +1,18 @@
+# docker compose stuff
+# override with your own values if the default host port mappings do not suit you
```
Olshansk, line 2:
```suggestion
# override with your own values if the default host port mappings in docker-compose do not suit you
```

Olshansk (approved):
Thank you! Left a couple of NITs but feel free to merge. The pattern we're following right now is to squash and merge the pattern where we copy-pasta the github PR description as the git commit message. 

I added you as a contributor so you can create branches in the core repo and merge things yourself. 

undefined - Imgur

build/deployments/docker-compose.yaml
0xBigBoss: great suggestion, was also going to point out, I didn't know colloquial names for these ports or else I would have used those. 

But I guess these var names are already long enough and a motivated reader would eventually figure it out.

0xBigBoss:
big thanks @Olshansk won't let you down. we gonna build some big stuff together 🤝 .

Issue #347, opened 2022-11-13
[P2P] Integrate LibP2P


jessicadaugherty:
# Objective
Use the results of https://github.com/pokt-network/pocket-network-protocol/issues/17 to integrate LibP2P into the p2p module.

# Origin Document(s)
* https://github.com/pokt-network/pocket-network-protocol/issues/17
* [LibP2P Guide](https://docs.libp2p.io/guides/)

# Goals

- Integrate LibP2P into Pocket Core

# Deliverables
- [x] MVP integration where LibP2P transport layer is used to send messages

# Non-Goals / Deliverables
- Integrating P2P deeper into transport layer
- Implementing peer churn or discovery
- Exhaustively enabling all the features we will need from LibP2P

## General issue checklist
- [ ] Update the appropriate CHANGELOG
- [ ] Update the README
- [ ] Update the source code tree explanation
- [ ] Update any relevant global documentation & references
- [ ] Document small issues / TODOs along the way

---

Creators: @jessicadaugherty @Olshansk

Olshansk:
Per our team discussion, I believe the scope of this can be summarized to, at a very high level, to bullet point 3 from [the research SPIKE](https://github.com/pokt-network/pocket-network-protocol/issues/17#issuecomment-1315936241):

<img width="957" alt="Screen Shot 2022-11-15 at 2 19 31 PM" src="https://user-images.githubusercontent.com/1892194/202037095-281429e8-9a46-4d98-b9f9-72d667cd334b.png">

jessicadaugherty:
Embed LibP2P host in Pocket host. Do not enable.

Olshansk:
Synched up @bryanchriswhite on this offline and wanted to capture the potential action items for a next step.

**Idea**: LibP2P could be treated as another "connection type" similar to what we implemented in #268.

**Next steps**:
1. Add `LibP2P` as a connection type (alongside TCP & noop)
2. Create a new ticket titled "Consolidate LibP2P & Pocket Interface"
3. "Hack" things together (in terms of data types) to replace the transport layer in P2P
4. Add comments similar to `CLEANUP(#XXX): Cleanup these interfaces` wherever we kind of hacked things together

**Other things discussed**:
1. Lack of connection pooling in the current version was not an intentional "thought-through" deicsion - we just got things working using the standard Golang libraries and I did not document it explicitly.

Olshansk:
[...]

PR #534, opened 2023-02-22
[Libp2p] Add libp2p module directories and helpers (part 1)

bryanchriswhite:
## Description

This is the first of a series of PRs split out from #500. Here we set up the new libp2p module directory structure and add crypto and identity / network helpers.

## Issue

#347 

## Type of change

Please mark the relevant option(s):

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- prepare pocket repo new libp2p module
- add pocket / libp2p identity helpers
- add url <--> multiaddr conversion helpers for use with libp2p (see: https://github.com/multiformats/go-multiaddr)
- add pokt --> libp2p crypto helpers

## Testing

- [x] `make develop_test`
- [x] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md) w/ all of the steps outlined in the `README`

## Required Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have tested my changes using the available tooling
- [x] I have updated the corresponding CHANGELOG

### If Applicable Checklist

- [ ] I have updated the corresponding README(s); local and/or global
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added, or updated, [mermaid.js](https://mermaid-js.github.io) diagrams in the corresponding README(s)
- [ ] I have added, or updated, documentation and [mermaid.js](https://mermaid-js.github.io) diagrams in `shared/docs/*` if I updated `shared/*`README(s)

PR #535, opened 2023-02-22
[P2P, Runtime] Update P2P & base config (part 2)

bryanchriswhite:
## Description

This is another of a series of PRs split out from https://github.com/pokt-network/pocket/pull/500. Here we update the base and P2P configs in preparation for the addition of the libp2p module.

## Issue

#347 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Replace `consensus_port` with `port` in P2P config.
- Update default P2P config `port` to from `8080` to `42069`.
- Add `use_libp2p` field to base config.
- Add `hostname` field to P2P config.
- Update state hash test after modifying genesis (updated port numbers).

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


runtime/manager_test.go
```diff
@@ -4255,7 +4258,8 @@ func TestNewManagerFromReaders(t *testing.T) {
 				config: &configs.Config{
 					P2P: &configs.P2PConfig{
 						PrivateKey:      "4ff3292ff14213149446f8208942b35439cb4b2c5e819f41fb612e880b5614bdd6cea8706f6ee6672c1e013e667ec8c46231e0e7abcf97ba35d89fceb8edae45",
-						ConsensusPort:   8080,
+						Hostname:        "node1.consensus",
+						Port:            8080,
```
Olshansk, line 4262: Should we change the tests too?

runtime/manager_test.go
```diff
@@ -4243,7 +4245,8 @@ func TestNewManagerFromReaders(t *testing.T) {
 			args: args{
 				configReader: strings.NewReader(string(`{
 					"p2p": {
-					  "consensus_port": 8080,
+					  "hostname": "node1.consensus",
+					  "port": 8080,
```
Olshansk, line 4249: Might be worth doing a global search for "8080" across the whole codebase.

Olshansk (changes requested):
Found one minor place where we still use "8080".

1. Can you do a global search for `8080`-`8084` and update all of them for consistency? E.g. see `build/deployments/.env.example` as well.

2. I know you tested it with the `docker-compose` localnet, but please double check that the k8s localnet works too. Bonus points: updating the github template to include it.

runtime/manager_test.go
bryanchriswhite: For sure, I did this but maybe things have changed in the meantime. Will double-check. Good call to look for the other port numbers.

.github/workflows/main.yml
```diff
@@ -55,7 +55,7 @@ jobs:
         # Not utilizing makefile target here to make use of pipefail bash feature.
         run: |
           set -euo pipefail
-          go test -p 1 -json ./... -covermode=count -coverprofile=coverage.out 2>&1 | tee test_results.json
+          GODEBUG=netdns=cgo go test -p 1 -json ./... -covermode=count -coverprofile=coverage.out 2>&1 | tee test_results.json
```
bryanchriswhite, line 58: This was intended for #534, my mistake

bryanchriswhite: Resolved - no pun intended (okay, maybe a little :wink:).

Olshansk:
> Still have to add `hostname` to the k8s P2P config but I'm not sure how to do that just yet and won't have time in this sitting. Perhaps we can tag-team?
> 
> (cc @okdas @Olshansk)

Similar to the configs you updated in `build/config/config1.json`, you would do the same in `build/localnet/manifests/configs.yaml` and some of the other manifests. It seems like you already renamed the `port` from `consensus_port` but didn't add the hostname there.

Regarding Localnet and Tiltfile, a good starting point would be to look at `build/localnet/templates/v1-validator-template.yaml.tpl` which is used as a template by `build/localnet/templates/v1-validator-template.sh`. Since the naming convention for the scaling k8s validators is `v1-validatorXXXX`, that will likely need to be the `hostname` (don't quote me on this) for the cluster to resolve each container separately.

Hopefully that's a good enough starting point. Seeing how I'm OOO for ETH Denver next week, hopefully @okdas can help out with some pointers as well.

runtime/manager_test.go
```diff
@@ -4209,7 +4210,8 @@ func TestNewManagerFromReaders(t *testing.T) {
 					},
 					P2P: &configs.P2PConfig{
 						PrivateKey:      "0ca1a40ddecdab4f5b04fa0bfed1d235beaa2b8082e7554425607516f0862075dfe357de55649e6d2ce889acf15eb77e94ab3c5756fe46d3c7538d37f27f115e",
-						ConsensusPort:   8080,
+						Hostname:        "node1.consensus",
+						Port:            defaults.DefaultP2PPort,
```
Olshansk, line 4214: Why do we use `defaults.DefaultP2PPort` in one place and hardcode `42069` everywhere else?

Olshansk (approved):
I just have one question (not a blocker but might involve a change). 

Also, let's have @okdas review the Tiltfile changes as well to make sure I didn't miss anything there.

build/localnet/templates/v1-validator-configs.yaml.tpl
```diff
@@ -1,13 +1,14 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: v1-validator-default-config
```
okdas, line 4: Instead of using template for this file, I would suggest to keep that file static, and only add an environment variable for that config parameter in `build/localnet/templates/v1-validator-template.yaml.tpl`.

build/localnet/templates/v1-validator-template.yaml.tpl
```diff
@@ -29,7 +29,7 @@ spec:
             - -config=/configs/config.json
             - -genesis=/genesis.json
           ports:
-            - containerPort: 8080
+            - containerPort: 42069
               name: consensus
             - containerPort: 50832
               name: rpc
```
okdas, line 35: Below this line there is a list of environment variables. Here is what I'm suggesting:
```
    - name: POCKET_P2P_HOSTNAME
       value: "v1-validator${VALIDATOR_NUMBER}"
```

That will override the configuration that we have in config.json. It works, because:
1. We follow The Twelve-Factor App methodology for [configuration files](https://12factor.net/config), and all our configuration parameters can be changed via environment variables which take priority over config.json.
2. We create a Service object below in the same file with this name.

okdas (commented):
Unless you feel strongly about this, I would suggest keeping just one template file for StatefulSet and a Service to avoid creating more templates. I've supplied an example on how an existing pattern can be utilized to achieve the same result.

build/localnet/templates/v1-validator-template.yaml.tpl
bryanchriswhite: Thanks for explaining that makes much more sense!

runtime/manager_test.go
bryanchriswhite: Good question. Except for two instances, all other usage is inside of string literals. It didn't seem prudent to rewrite the tests to interpolate `defaults.DefaultP2PPort` into what's historically been (many) simple string literals. Do you disagree?

Of those two exceptions, one has a corresopnding string literal representation of itself in the test case which it asserts against; see the [`unset MaxMempoolCount should fallback to default value` sub-test](https://github.com/pokt-network/pocket/blob/main/runtime/manager_test.go#L4242). I would argue that this  one should be left alone due the expected literal correspondence.

I've pushed a commit for the other instance: 743d5cb0.

Olshansk: Makes sense to me. Ty for the explanation

build/localnet/templates/v1-validator-template.yaml.tpl
```diff
@@ -63,6 +63,8 @@ spec:
               value: "postgres://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@$(POSTGRES_HOST):$(POSTGRES_PORT)/$(POSTGRES_DB)"
             - name: POCKET_PERSISTENCE_NODE_SCHEMA
               value: validator${VALIDATOR_NUMBER}
+            - name: POCKET_P2P_HOSTNAME
```
Olshansk, line 66: nice

Olshansk (approved):

okdas (approved):
This looks great, thank you!

Issue #525, opened 2023-02-17
[CLI] Improve Keybase private-keys sourcing when running in Kubernetes


deblasis:
## Objective

Currently, now that we merged #522, the CLI has a `Kubernetes API` dependency because the `private-keys.yaml` file (required by the keybase) is not available locally inside the `Pod` and therefore, in order to have the same init logic, we need to source it by fetching the `Secret` via the `Kubernetes APIs`.

## Origin Document

During the Protocol Hour today (17th Feb 2023) @okdas raised a good point regarding having Kubernetes dependencies inside the CLI at all. The CLI should stay unopinionated and be flexible in terms of sourcing dependencies like the `private-keys.yaml` 

## Goals

- Decouple debug keybase initialization from the core logic
- At a minimum allow to specify a path where to fetch the private-keys from
- Decouple the debug/development private-keys from k8s

## Deliverable

- [ ] Allow specifying a path from where the development/debug private keys should be initialized from
- [ ] Make debug key base initialization a separate process (e.g. different make target) independent of the CLI
		- [ ]  Updated CLI/Keybase code
- [ ] Update the Kubernetes manifests so that the private-keys file is accessible without requiring us to interact with the Kubernetes API at all
	- [ ] Updated Localnet manifests

## Non-goals / Non-deliverables

- Additional modifications to keybase extensions

## General issue deliverables

- [ ] Update the appropriate CHANGELOG(s)
- [ ] Update any relevant local/global README(s)
- [ ] Update relevant source code tree explanations
- [ ] Add or update any relevant or supporting [mermaid](https://mermaid-js.github.io/mermaid/) diagrams

## Testing Methodology

- [ ] **Task specific tests or benchmarks**: `make ...`
- [ ] **New tests or benchmarks**: `make ...`
- [ ] **All tests**: `make test_all`
- [ ] **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @deblasis 
**Co-Owners**: @okdas, @Olshansk 

```[tasklist]
### Tasks
- [ ] Add a draft title or issue reference here
```

jessicadaugherty:
@deblasis wdyt about linking with @h5law for this one if he's available?

jessicadaugherty:
Consult with @deblasis on implementation

0xBigBoss:
[...]

PR #537, opened 2023-02-23
[KEYBASE] Add Keybase Config and Vault Backend

### LINKED ISSUE #527: [Keybase] HashiCorp Vault support as a keybase backend
## Objective

Add a more node-operator-oriented, production-ready backend for the keybase.

## Origin Document

The implementation for the CLI & keybase are available at [app/client](https://github.com/pokt-network/pocket/tree/main/app/client) but only supports a filesystem-based key based backed by [BadgerDb](https://github.com/dgraph-io/badger). Example from `/Users/olshansky/workspace/pocket/pocket/app/client/cli/account.go`:

```go
	// Open the debug keybase at the specified path
	pocketDir := strings.TrimSuffix(dataDir, "/")
	keybasePath, err := filepath.Abs(pocketDir + keybaseSuffix)
	// ...
	kb, err := keybase.NewKeybase(keybasePath)
	// ...
```

PNI and other major node runners make use of [HashiCorp Vault](https://www.vaultproject.io/) which makes it a great 1st candidate for an alternate keybase backend.

Screenshot 2023-02-20 at 9 06 16 AM

## Goals

- Leverage the existing keybase utilities & interfaces for production uses cases
- Make TestNet keybase management easier for PNI and other node operators
- Add support for [HashiCorp Vault](https://www.vaultproject.io/) as a keybase backend

## Deliverable

- [ ] A PR that enables using Valut as a backend for keybase management
- [ ] A flag or configuration allowing the user to specify which backend they want to use via the CLI
- [ ] A document explaining how to setup HashiCorp Vault as a backend

- Adding new commands to the CLI
- Modifying the core node business logic

## General issue deliverables

- [ ] Add a README w/ relevant documentation & [mermaid](https://mermaid-js.github.io/mermaid/) diagrams on how to use this for keybase management

## Testing Methodology

- [ ] Instructions (e.g. a demo) with a video and list of commands showing the functionality in action
---

**Creator**: @Olshansk 
**Co-Owners**: @h5law @okdas @jessicadaugherty

0xBigBoss:
## Description

Introduces a keybase config into the runtime config and adds a more node-operator-oriented, production-ready backend for the keybase. https://app.dework.xyz/pokt-network/v1-protocol?taskId=04cb2ab8-8b7b-4c81-a407-857649a928b7

## Issue

Fixes #527

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

- Adds KeybaseConfig to runtime config
- Add Hashicorp Vault keybase implementation
- Fix a bug where build binary could not resolve private keys
- Update CLI commands to parse keybase flags
- Updates CLI docs
- Adds confirm passphrase new key creation, importing unarmored private keys, and updating passphrases

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// Create new keypair entry in Vault
+func (vk *VaultKeybase) Create(passphrase, hint string) error {
+	keyPair, err := crypto.CreateNewKey(passphrase, hint)
+	if err != nil {
+		return err
+	}
+
+	keyPairBytes, err := json.Marshal(keyPair)
```
h5law, line 34: @0xBigBoss Here the `KeyPair` struct actually has its own marshalling functions so you could do `keyPairBz, err := keyPair.Marshal()`

h5law:
@0xBigBoss Just to let you know now that PR #501 is merged the create and import functions now return `(crypto.KeyPair, error)` and there is no longer `Exists` this was just a pointless wrapper on `Get` anyway. Left a comment about marshalling but looks 🔥

app/client/keybase/keyvault.go
0xBigBoss: @h5law originally i implemented it with this function, but I opted for JSON since it may need to be consumed by other SDK clients. I'm not an expert in go serialization and how portable the format is. Compare the two results. wdyt?

```
8��
EncKeyPair��	PublicKey
PrivKeyArmour����=github.com/pokt-network/pocket/shared/crypto.Ed25519PublicKey
" �~^9�2��|M��N�	j܁o�k�TN��V��({"kdf":"scrypt","salt":"9d812b1fcc5709c60b35dd132190cd95","secparam":"12","hint":"hint","ciphertext":"YH5R2dN1deUNO6NBAO0hXC5y3L8WuEgX3ur6FF5mNRdmwtvdxlkJxk2OeIZ5+DRHPcE6G6Lr0VlbbIC57PfnPB59UGX8Y9NWCefGutSe7gsdvQJxSA09R/qYs/XoxHAi896hacP6EVLtE7rfez7ZXAohvB+Ps1AeFAf8yCVtUHzqhMUJEPEQLt48LVHAj447"}
```

vs with json format

```json
{"PublicKey":"122004d236918c9a92bd5f21e02c33b4e6451b6d078796e6a0d7ea61d0867d0f","PrivKeyArmour":"{\"kdf\":\"scrypt\",\"salt\":\"bcfdd0c8eb441d33c27ce26cbbae3431\",\"secparam\":\"12\",\"hint\":\"hint\",\"ciphertext\":\"CPcnOHljTqgD9M6KBENZnL0uTLL8Vl0YAPREjh33KN6w+keGD/P3yknSTHyyyqoecgpZTx8KoAkGY44i46bggPhRoTbGLeb4baa5iMEbv3libXYxyeTfQiWVzLLN1hW6+F752nVh1R5o7t4UeiKciZmoT2XbPlZ4FGV50xQUKvts+tyrqjQCdkkZa0gkfzxY\"}"}
```

h5law: @0xBigBoss `encoding/gob` is very specifc to golang - maybe for cross-language/SDK support it might be better to define a `keypair.proto` file? This would make the transmission a lot easier as the serialisation is universal across all languages. WDYT?

shared/crypto/keypair.go
```diff
@@ -4,13 +4,14 @@ import (
...
 	"encoding/gob"
+	"encoding/json"
 )
 
 // Encoding is used to serialise the data to store the KeyPairs in the database
 func init() {
 	gob.Register(Ed25519PublicKey{})
 	gob.Register(ed25519.PublicKey{})
-	gob.Register(encKeyPair{})
+	gob.Register(EncKeyPair{})
```
h5law, line 14: Here we should really reduce the scope and just use the `KeyPair` interface. I had the same issue when trying to create an empty one see the function `GetKeypair()` in `keypair.go` this can be used instead of `var kp crypto.KeyPair` when you need it

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+	if data == nil {
+		return nil, errors.New("key not found")
+	}
+
+	keyPairStr, ok := data.Data["key_pair"].(string)
+	if !ok {
+		return nil, errors.New("invalid key pair")
+	}
+
+	var keyPairStruct crypto.EncKeyPair
```
h5law, line 117:
```suggestion
	keyPairStruct := crypto.GetKeypair()
```var keyPairStruct crypto.EncKeyPair

File: shared/crypto/keypair.go [hunk 4]
```diff
@@ -35,54 +36,81 @@ type KeyPair interface {
...
 
-// encKeyPair struct stores the public key and the passphrase encrypted private key
-type encKeyPair struct {
+// EncKeyPair struct stores the public key and the passphrase encrypted private key
+type EncKeyPair struct {
 	PublicKey     PublicKey
 	PrivKeyArmour string
 }
 
+func (ekp *EncKeyPair) MarshalJSON() ([]byte, error) {
```
h5law, line 47: Just a NIT you should add these to the `KeyPair` interface, use `(kp *encKeyPair)` instead to match the other methods and move them down to be with the other marshalling functions (or replace them if you think this is better) 😅

app/client/keybase/keyvault.go
Olshansk: 1. Let's avoid exposing `gob` encoded bytes unless it lives in the scope of the project (i.e. looking at external services).

2. @h5law We can add a TODO in the keypair source code and implement it in the future. Definitely better for cross-compatibility.

3. The approach I'm thinking about in this use case is adding json tags to the keypair struct and adding a `MarshalJSON` function to the interface.

```go
type EncKeyPair struct {
	PublicKey     PublicKey  `json:"public_key"`
	PrivKeyArmour string    `json:"priv_key_armour"`
}
```

Usage would be:  `kpJson :=  kp.MarshalJSON()`

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/vault/api"
+	"github.com/pokt-network/pocket/shared/crypto"
+)
+
+// VaultKeybase implements the Keybase interface using HashiCorp Vault
+type VaultKeybase struct {
```
Olshansk, line 14: Let's lowercase it if we don't need to expose `VaultKeybase` outside of this package.

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
+package keybase
```
Olshansk, line 1: Can you create a package called `vault` at `app/client/keybase/vault`. If we ever need to add other backends in the future, it'll be more modular that way.

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+	if err != nil {
+		return err
+	}
+
+	keyPairBytes, err := json.Marshal(keyPair)
+	if err != nil {
+		return err
+	}
+
+	data := map[string]interface{}{
```
Olshansk, line 39: SInce you're using this in a few places in the same file, let's create a small local struct near the top.

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+		"key_pair": string(keyPairBytes),
+		"hint":     hint,
+	}
+
+	_, err = vk.client.KVv2(vk.mount).Put(context.TODO(), keyPair.GetAddressString(), data)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
Olshansk, line 49:
```suggestion
	return err
```

Ditto below

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// Get a keypair from Vault
+func (vk *VaultKeybase) Get(address string) (crypto.KeyPair, error) {
+	data, err := vk.client.KVv2(vk.mount).Get(context.TODO(), address)
+	if err != nil {
+		return nil, err
+	}
+
```
Olshansk, line 107:
```suggestion
```

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+func (vk *VaultKeybase) GetPubKey(address string) (crypto.PublicKey, error) {
+	keyPair, err := vk.Get(address)
+	if err != nil {
+		return nil, err
+	}
+
+	return keyPair.GetPublicKey(), nil
+}
+
+// Get a private key from Vault
```
Olshansk, line 136: NIT (here and elsewhere): let's use proper godoc. We're not doing it everywhere in the codebase yet but the team discussed that we'll put more effort into it.
```suggestion
// GetPrivKey gets a private key from the vault
```

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	if data == nil {
+		return nil, nil, nil
+	}
+
+	var addresses []string
+	var keyPairs []crypto.KeyPair
+
+	for _, key := range data.Data["keys"].([]interface{}) {
```
Olshansk, line 166: 1. s/interface{}/any as of go 1.18
2. Why do we even need a typecast here? I feel like skipping it would avoid the need for string type assertions below

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+		keyPairs = append(keyPairs, keyPair)
+	}
+
+	return addresses, keyPairs, nil
+}
+
+// Check if a key exists in Vault
+func (vk *VaultKeybase) Exists(address string) (bool, error) {
+	_, err := vk.Get(address)
+	if err != nil {
```
Olshansk, line 181:
```suggestion
	if _, err := vk.Get(address); err != nil {
```

app/client/keybase/keyvault_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	return client, vk, nil
+}
+
+func TestVaultKeybase(t *testing.T) {
+
+	ctx := context.Background()
+
+	// Create a new Vault container
+	req := testcontainers.ContainerRequest{
```
Olshansk, line 40: 💯

app/client/keybase/keyvault_test.go
```diff
@@ -0,0 +1,192 @@
...
+	// Set the root token for the client
+	client.SetToken("dev-only-token")
+
+	// Create a new VaultKeybase instance
+	vk := NewVaultKeybase(client, "secret")
+
+	return client, vk, nil
+}
+
+func TestVaultKeybase(t *testing.T) {
```
Olshansk, line 35: Consider using `TestMain` for the setup and teardown of the vault container: https://pkg.go.dev/testing.

We do something similar with our postgres container in `persistence/test/setup_test.go`

Olshansk (changes requested):
Did a quick pass through and left a few comments throughout. The actual usage of vault looks pretty on point to me thought.

Per @h5law's comments, let's update the interface rather than exposing the types directly. Left a potential suggestion there.

Will do a deeper dive once the existing comments are tended to.

app/client/keybase/keyvault.go
```diff
@@ -0,0 +1,296 @@
...
+// NewVaultKeybase returns a new instance of VaultKeybase
+func NewVaultKeybase(client *api.Client, mount string) *VaultKeybase {
+	return &VaultKeybase{
+		client: client,
+		mount:  mount,
+	}
+}
+
+// Create new keypair entry in Vault
+func (vk *VaultKeybase) Create(passphrase, hint string) error {
```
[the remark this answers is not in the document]
0xBigBoss: 🇷🇺🤣

0xBigBoss:
still todo: 
- test localnet
- add new methods to badgerdb implementation
- clean up test comments
- profit???

Olshansk:
> profit???

That part is always fun.

---

Make sure to re-request my review when it's ready for another look so I get notified.

build/debug.go
```diff
@@ -0,0 +1,14 @@
+package build
```
deblasis, line 1: I am not reviewing the whole PR, just focusing on this since I am coming from #525 

Nice! One thing that I would change if we go with this approach is that I'd probably like to embed conditionally, depending on the presence of a `debug` build flag perhaps that we can use also to disable completely the debug menu in the CLI as well so that any `debug` related stuff is stripped out from the production binary since debug is used for development purposes only.

How does it sound guys? @Olshansk @0xBigBoss

Olshansk: I left another comment about this elsewhere, but I think we should use the runtime manager for this (see above)

@deblasis wdyt?

app/client/keybase/keybase.go
```diff
@@ -40,3 +43,44 @@ type Keybase interface {
...
 }
+
+type KeybaseType int
+
+const (
+	KeybaseTypeFile KeybaseType = iota
+	KeybaseTypeVault
+)
+
+type KeybaseOptions struct {
```
Olshansk, line 54: Per my other comments, I'm beginning to think that this should be part of `config.json`, which configures our modules, rpc, etc.. since it's part of a node's config.

If you look at `config.proto` it should be a good start.

Also, even though I think this is the right approach, I'm not married to it and happy to hear counterarguments.

Olshansk (changes requested):
@0xBigBoss I left multiple comments about this throughout, but I'm beginning to think that we should have a `Keybase` config in `config.proto` and then use `viper` to override it with env vars in the `runtime/manager.go`.

@deblasis Wdyt of this design?

app/client/cli/keys.go
```diff
@@ -405,6 +374,10 @@ func keysImportCommands() []*cobra.Command {
 						return err
 					}
 				case "raw":
+					// it is unarmoured so we need to confirm the passphrase
```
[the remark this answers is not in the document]
0xBigBoss: I believe we should have the user confirm the passphrase anytime we are armoring a private key interactively.

app/client/cli/keys.go
```diff
@@ -292,6 +270,9 @@ func keysGetCommands() []*cobra.Command {
 			},
 		},
 	}
+
+	applySubcommandOptions(cmds, attachKeybaseFlagsToSubcommands())
```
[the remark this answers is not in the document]
0xBigBoss: i refactored it a bit to make it more clear which subcommands need options, I'm pretty sure this was needed since the keysGetCommand did not have any flags before this PR.

app/client/keybase/keybase.go
0xBigBoss: 💯 I was a bit confused finding that the CLI and node keybase logic were disconnected. Though, I think that should be tackled in a separate PR especially since there aren't any unit tests on the CLI commands.

shared/crypto/keypair.go
```diff
@@ -33,14 +33,42 @@ type KeyPair interface {
...
+	PublicKey     PublicKey `json:"public_key"`
+	PrivKeyArmour string    `json:"priv_key_armour"`
+}
+
+type simpleEncKeyPair struct {
+	PublicKey     string `json:"public_key"`
+	PrivKeyArmour string `json:"priv_key_armour"`
+}
+
+func (ekp *encKeyPair) MarshalJSON() ([]byte, error) {
```
h5law, line 53: Bumping - just a NIT but these should be moved down next to the other marshalling methods

build/debug.go
deblasis: @Olshansk I agree with you. Especially now with keybase and vault integration, we are totally relying on flags that in reality represent something that is kinda static and runtime/environment dependent.

Just using a random command as an example: 
image

We don't want to expect the user to input all the time the `--keybase` and other flags if they use vault.

We want to offer the option to override that but I think we should reuse the pattern where we have defaults->config->overrides via environment variables/flags thanks to viper (12 factor app).

app/client/cli/utils.go
```diff
@@ -220,6 +223,43 @@ func attachHintFlagToSubcommands() []cmdOption {
 	}}
 }
 
+func attachKeybaseFlagsToSubcommands() []cmdOption {
+	return []cmdOption{func(c *cobra.Command) {
+		c.Flags().StringVar(&keybaseTypeFromCLI, "keybase", "file", "keybase type used by the cmd, options are: file, vault")
+		c.Flags().StringVar(&keybaseVaultAddrFromCLI, "vault-addr", "", "Vault address used by the cmd. Defaults to https://127.0.0.1:8200 or VAULT_ADDR env var")
+		c.Flags().StringVar(&keybaseVaultTokenFromCLI, "vault-token", "", "Vault token used by the cmd. Defaults to VAULT_TOKEN env var")
```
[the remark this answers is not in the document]
deblasis: I wholeheartedly support this proposal :)

app/client/keybase/vault_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+func TestVaultKeybase(t *testing.T) {
+
+	if vk == nil {
+		t.Fatalf("vk is nil")
+	}
+
+	// Test Create
+	_, err := vk.Create("passphrase", "hint")
+	if err != nil {
```
h5law, line 110: Just a thought these could be simplified to something like
```suggestion
	require.NoError(t, err)
```

Would reduce all the if statments

app/client/keybase/vault.go
```diff
@@ -0,0 +1,317 @@
...
+
+	versions := make([]int, 0, len(versionsMeta))
+	for _, version := range versionsMeta {
+		versions = append(versions, version.Version)
+	}
+
+	return vk.client.KVv2(vk.mount).Destroy(context.TODO(), address, versions)
+}
+
+func (vk *vaultKeybase) MarshalJSON() ([]byte, error) {
```
h5law, line 296: Just curious what these 2 methods are for? Do we need to marshal the entire keybase?

app/client/keybase/vault.go
```diff
@@ -0,0 +1,317 @@
...
+			return nil, nil, err
+		}
+		keyPairs = append(keyPairs, keyPair)
+	}
+
+	return addresses, keyPairs, nil
+}
+
+// Exists checks if a key exists in vault
+func (vk *vaultKeybase) Exists(address string) (bool, error) {
```
h5law, line 192: `Exists` has been removed from the current Keybase interface

app/client/keybase/debug/keystore.go
```diff
@@ -128,28 +128,16 @@ func fetchValidatorPrivateKeysFromK8S() (map[string]string, error) {
...
-	}
 	var config struct {
 		ApiVersion string            `yaml:"apiVersion"`
 		Kind       string            `yaml:"kind"`
 		MetaData   map[string]string `yaml:"metadata"`
 		Type       string            `yaml:"type"`
 		StringData map[string]string `yaml:"stringData"`
 	}
-	if err := yaml.Unmarshal(yamlData, &config); err != nil {
+	if err := yaml.Unmarshal(build.PrivateKeysFile, &config); err != nil {
```
h5law, line 140: 🔥

app/client/cli/keys.go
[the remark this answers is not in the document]
h5law: @0xBigBoss Does the get command need any flags? Personally I think the `applySubcommandOptions` should be as the top to be clear which flags are being added to which subcommand

shared/crypto/keypair.go
```diff
@@ -33,14 +33,42 @@ type KeyPair interface {
...
 
 // encKeyPair struct stores the public key and the passphrase encrypted private key
 type encKeyPair struct {
-	PublicKey     PublicKey
-	PrivKeyArmour string
+	PublicKey     PublicKey `json:"public_key"`
+	PrivKeyArmour string    `json:"priv_key_armour"`
+}
+
+type simpleEncKeyPair struct {
```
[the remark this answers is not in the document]
0xBigBoss: i think we could, but encoding byte arrays into JSON strings seems a bit strange. Seems more familiar to use the hex encoding.

app/client/cli/keys.go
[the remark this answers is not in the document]
0xBigBoss: `keysGetCommand` needs to know which keybase to get the key from.

I disagree, the options are tightly coupled with the commands and should be as close as possible to their definition.

app/client/keybase/vault.go
0xBigBoss: nope, just left over methods. prob messed up a rebase. should be dropped.

app/client/cli/keys.go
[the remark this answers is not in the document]
h5law: @0xBigBoss Just to give some context I also thought this when first integrating the keybase with the CLI but @Olshansk pointed out to me that its more for the other devs to pick up quickly instead of having to dive deep into the code to find out which cmds have which flags

https://github.com/pokt-network/pocket/pull/501#discussion_r1105082303

app/client/cli/keys.go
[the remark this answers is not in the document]
h5law: But they already have to provide either the password as a flag or type it in any way? Maybe this is useful for creation but elsewhere its already being confirmed - this is just my opinion though. Whatever you think fits best for the use case

app/client/keybase/README.md
```diff
@@ -52,21 +52,32 @@ Keys can be created without a password by specifying an empty (`""`) passphrase.
...
-```bash
-app
-└── client
-    └── keybase
-          ├── README.md
-          ├── keybase.go
-          ├── keybase_test.go
-          └── keystore.go
+```sh
+$ tree app/client/keybase/
```
h5law, line 56: Just a NIT - I don't think we need this line the output is really what we are trying to show

app/client/keybase/vault.go
```diff
@@ -0,0 +1,346 @@
...
+	return nil
+}
+
+// Create new keypair entry in vault
+func (vk *vaultKeybase) Create(passphrase, hint string) (crypto.KeyPair, error) {
+	keyPair, err := crypto.CreateNewKey(passphrase, hint)
+	if err != nil {
+		return nil, err
+	}
+	err = writeVaultKeyPair(vk, keyPair.GetAddressString(), keyPair, hint)
```
h5law, line 81:
```suggestion
	if err := writeVaultKeyPair(vk, keyPair.GetAddressString(), keyPair, hint); err != nil {
	        return nil, err
	}
```

Same for the ones below

h5law (changes requested):
Looks good to me - left a couple of final comments about simplifying if statements when only receiving an error from a function. Otherwise looks 🔥  IMO be ready to merge soon, will wait to see if @Olshansk has anything to add first

app/client/cli/keys.go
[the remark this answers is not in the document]
0xBigBoss: idk, i can't think of one CLI tool that doesn't ask you to confirm a password input interactively. seems like a deviation from standard password input ux.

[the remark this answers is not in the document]
h5law: My only thought is that unless the `--pwd` flag is being used the `pwd  = readPassphrase(pwd)` will cause a password prompt by the user anyway. Having to do this back to back seems excessive outside of key creation. But this is really not a huge deal, if you think this matches the general UX from other CLI apps then I'll go with your way 🔥

app/client/keybase/vault_test.go
0xBigBoss: sry missed this earlier, great suggestion.

app/client/keybase/README.md
0xBigBoss: I compare NITs to vegetables, I hate vegetables 🤣 but i still eat em.

app/client/cli/keys.go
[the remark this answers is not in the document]
0xBigBoss: Exactly, this `readPassphrase` should always be followed by a confirm if it's a new passphrase (else we should just force one passphrase for the whole keybase). 

decrypting never requires a confirmation since the feedback is immediate if it's wrong.

Confirmation prompts should be required any time things are being encrypted interactively. Just my thoughts on this matter though, I guess i'm basing this totally on my own experience.

app/client/cli/keys.go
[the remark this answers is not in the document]
0xBigBoss: Yah i get that it makes it easier to read, but we should be leveraging go compiler and IDE "lookup references" since these flags have bled into other files.

Plus I think it makes testing even more painful. Testing cobra cli apps is already kind of a dumpster fire though it is getting better. So I'm gonna say it's better to wrap it in a function that sets up the commands with all required inputs.

 That way tests like this, https://github.com/pokt-network/pocket/pull/537/files#diff-f16936f439473efd656c1cf821fa092b9eeb465ee369b8267c9c470832b2e2f4, are easier to write in the future.

app/client/cli/utils.go
```diff
@@ -239,6 +249,43 @@ func attachChildPwdFlagToSubcommands() []cmdOption {
...
+	}}
+}
+
+func keybaseForCLI() (keybase.Keybase, error) {
+	keybaseType := keybase.KeybaseTypeFile
+	keybaseOptions := keybase.KeybaseOptions{}
+
+	if keybaseTypeFromCLI != "" {
+		switch keybaseTypeFromCLI {
+		case "file":
```
Olshansk, line 267: Move "file"  and "vault" into constants

app/client/cli/utils.go
```diff
@@ -239,6 +249,43 @@ func attachChildPwdFlagToSubcommands() []cmdOption {
...
+				return nil, err
+			}
+			keybaseOptions.KeybasePath = keybasePath
+		case "vault":
+			keybaseType = keybase.KeybaseTypeVault
+			keybaseOptions.VaultAddr = keybaseVaultAddrFromCLI
+			keybaseOptions.VaultToken = keybaseVaultTokenFromCLI
+			keybaseOptions.VaultMountPath = keybaseVaultMountPathFromCLI
+		default:
+			fmt.Printf("❌ Invalid keybase type: %s\n", keybaseTypeFromCLI)
```
Olshansk, line 282: `logger.Global.Fatal()` causes an `os.exit(1)`

shared/crypto/keypair.go
```diff
@@ -36,18 +36,25 @@ type KeyPair interface {
...
 // The encrypted private key is stored as a JSON string with the fields needed to decrypt the key
 type encKeyPair struct {
-	PublicKey     PublicKey
-	PrivKeyArmour string
+	PublicKey     PublicKey `json:"public_key"`
+	PrivKeyArmour string    `json:"priv_key_armour"`
 }
 
-// newKeyPair Generates a new KeyPair interface given the public key and armoured private key
+type simpleEncKeyPair struct {
```
Olshansk, line 52: I was looking around and found this: https://medium.com/picus-security-engineering/custom-json-marshaller-in-go-and-common-pitfalls-c43fa774db05

The tl;dr is that if we define a `func (pk *PublicKey) MarshalJSON() ([]byte, error) { ...}`, we'd be able to avoid the need of having `simpleEncKeyPair` altogether.

Cc @h5law  for visibility

Here's an example (https://go.dev/play/p/a5OojMD3atJ):

Screenshot 2023-03-07 at 4 16 11 PM

Olshansk:
Did another round of review (not the entire thing) and left some nits/comments throughout.

However, I wanted to consolidate these three threads:
- https://github.com/pokt-network/pocket/pull/537/files#r1119419089
- https://github.com/pokt-network/pocket/pull/537/files#r1119580562
- https://github.com/pokt-network/pocket/pull/537/files#r1119564427

I believe what we should do is:
1. Create `keybase_config.proto` in `runtime/configs/proto` - thereby removing the new structrs introduced
2. Update `runtime/configs/config.go` by adding in its place `KeybaseConfig`
3. Add `POCKET_` prefix to the env variables introduced so they can be overriden by `viper`
4. Split out the embedding on the yaml file into a separate PR (see https://github.com/pokt-network/pocket/issues/525#issuecomment-1459073082)
5. In a future PR, we can update the k8s manifests to include the keybase configs and test in dev (cc @okdas)

cc @deblasis for visibility

Olshansk (changes requested):
Did another round of review (not the entire thing) and left some nits/comments throughout.

However, I wanted to consolidate these three threads:
- https://github.com/pokt-network/pocket/pull/537/files#r1119419089
- https://github.com/pokt-network/pocket/pull/537/files#r1119580562
- https://github.com/pokt-network/pocket/pull/537/files#r1119564427

I believe what we should do is:
1. Create `keybase_config.proto` in `runtime/configs/proto` - thereby removing the new structrs introduced
2. Update `runtime/configs/config.go` by adding in its place `KeybaseConfig`
3. Add `POCKET_` prefix to the env variables introduced so they can be overriden by `viper`
4. Split out the embedding on the yaml file into a separate PR (see https://github.com/pokt-network/pocket/issues/525#issuecomment-1459073082)
5. In a future PR, we can update the k8s manifests to include the keybase configs and test in dev (cc @okdas)

cc @deblasis for visibility

app/client/cli/keys_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"bytes"
+	"io"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestKeysCreateCommands is a test for the keys create command.
```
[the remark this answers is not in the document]
0xBigBoss: Yes, that is my suggestion.

app/client/cli/keys_test.go
```diff
@@ -0,0 +1,48 @@
...
+// This is a simple, contrived example of how to test one command.
+func TestKeysCreateCommands(t *testing.T) {
+
+	// boilerplate that should be in a setup function for all tests
```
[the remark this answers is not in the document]
0xBigBoss: 💯 but this is my fault for including changes out of scope of the original issue. I don't want to distract from the initial goals too much.

app/client/cli/keys.go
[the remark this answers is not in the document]
0xBigBoss: Yup, that is what's happening here. Confirm passphrase for the newly imported raw private key string if running interactively.

app/client/cli/keys.go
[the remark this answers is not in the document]
0xBigBoss: definitely a lot of this feedback on improving the DX of the CLI app should be collected and tackled in a separate PR IMO.

app/client/cli/keys.go
```diff
@@ -137,10 +83,17 @@ func keysCreateCommands() []*cobra.Command {
 
 				logger.Global.Info().Str("address", kp.GetAddressString()).Msg("New Key Created")
 
+				cmd.Println("New Key Created")
```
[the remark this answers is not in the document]
0xBigBoss: I could, but it'd break the contrived cli test I introduced.

app/client/keybase/doc/vault.md
```diff
@@ -0,0 +1,156 @@
+# Keybase Vault
+
+The Keybase Vault stores key pairs, public key and armoured private keys, using the Hashicorp Vault KV secrets engine. The Keybase Vault is a wrapper around the Hashicorp Vault API.
+
+Keybase Vault requires a few pieces of information to be able to connect to it:
+
+- Address (e.g. VAULT_ADDR=http://127.0.0.1:8200/)
+- Token (e.g. VAULT_TOKEN=hvs.25YM7qJDN8S2EpFEA4SL0ciD)
+- Mount Path (e.g. /secret)
```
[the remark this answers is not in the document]
0xBigBoss: Sadly, I believe not. /secret is the default though for the kv2 store.

shared/crypto/keypair.go
0xBigBoss: I racked my head against this a bit, I don't think there is a way to encode the public key as a string and unmarshal as a crypto.PublicKey without a temporary struct. Even that article uses that technique to unmarshal.

Though this new approach is much cleaner and simplified the interfaces a lot.

h5law: @0xBigBoss Just to jump back in on this one - I thought that the previous comment I left on the function in `ed25519.go` that can generate a new public key from bytes would be the solution here. The marshalled `string` does not need to be human readable does it? If we just do `string(rawBz)`  then we can do `NewPublicKeyFromBytes([]byte(rawBzString))`?

I think this should work, the encoded string doesn't have any requirements on being human readable - nobody will be reading it anyway so there is not a need for it to "look nice".

app/client/cli/keys.go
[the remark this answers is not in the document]
Olshansk: Can you think of a better way to test it? I feel like we're introducing bad form for tests without needing to.

Lmk if you need my help to ideate some tests.

app/client/cli/keys.go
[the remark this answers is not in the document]
Olshansk: Per my review, let's update the comment.

Olshansk:
@0xBigBoss Lmk what you think of https://github.com/pokt-network/pocket/pull/537#pullrequestreview-1329373725 when you have a chance.

0xBigBoss:
@Olshansk I totally agree that those items to introduce the keybase into the runtime config is the way to go, but as my comment mentioned https://github.com/pokt-network/pocket/pull/537#discussion_r1120017341, I believe they should be moved to a separate task as I believe the original deliverables and goals have been met. I still want to address the style, non-functional, and out-of-scope feedback that was provided. Just trying to budget my time appropriately.

Happy to chat through this too, maybe I am missing something.

shared/crypto/keypair.go
0xBigBoss: @h5law But don't we run into the same issue that gob uses it's own encoding so it's not easily consumed by other SDKs?

Being able to easily read the values from vault has come in handy before at least from my experience.

Screenshot 2023-03-08 at 16 41 56

Olshansk:
> @Olshansk I totally agree that those items to introduce the keybase into the runtime config is the way to go, but as my comment mentioned [#537 (comment)](https://github.com/pokt-network/pocket/pull/537#discussion_r1120017341), I believe they should be moved to a separate task as I believe the original deliverables and goals have been met. I still want to address the style, non-functional, and out-of-scope feedback that was provided. Just trying to budget my time appropriately.

I realize that "it works", but because:
1. It's not super urgent
2. Doesn't block downstream work
3. We're trying to limit tech debt on main
4. We have a well-defined solution as to how it should be refactored and the other owners of this part of the codebase (@h5law and @deblasis) agree.

Are you thinking of just splitting #527 into two PRs? This one + a follow-up with the proper config?

app/client/cli/keys.go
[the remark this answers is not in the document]
0xBigBoss: np should be fixed now.

app/client/cli/keys_test.go
```diff
@@ -0,0 +1,48 @@
...
+	cmd.SetOut(buf)
+
+	// Use a buffer to simulate stdin
+	cmd.SetIn(input)
+
+	// Execute the command with args and flags
+	err = cmd.Execute()
+	require.NoError(t, err)
+
+	// Verify the output
```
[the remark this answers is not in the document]
0xBigBoss: Test is updated now to be a bit better.

shared/crypto/keypair.go
h5law: @0xBigBoss So `gob` encoding would be specific to golang, I'm pretty sure, so you are correct on that front but with JSON encoding to `[]byte` this would be consistent across all potential SDKs. I am not really too invested in either now as the idea of a Protobuf for the `KeyPair` would be the most portable option - serialising these would also not be human readable though. I think the human readable encoding should be left to exporting the key either in RAW or JSON format IMO.

The only thing is that introducing another struct to ease the encoding seems unnecessary to me, that's all really. If we can do the same without introducing the `simpleEncodedKey` then I think we should go with that option.

0xBigBoss: The TLDR on why we need a temp/anon struct is cuz of the implementation `UnmarshalJSON` of the public key. We are assuming it's using hex encoding. Go json uses base64 encoding for byte[]. We'd have to change this implementation (which I don't recommend, base16 is common and expected). 

@Olshansk @h5law 

https://github.com/pokt-network/pocket/blob/main/shared/crypto/ed25519.go#L198-L214

app/client/cli/keys.go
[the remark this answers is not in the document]
0xBigBoss: sry i missed this one, wdym?

0xBigBoss:
> Are you thinking of just splitting https://github.com/pokt-network/pocket/issues/527 into two PRs? This one + a follow-up with the proper config?

@Olshansk this seems more appropriate to me and if everyone is onboard with this, that'd be awesome. This PR page already takes a few seconds to load for me 🤣 

cc @h5law @deblasis

Olshansk:
> @Olshansk this seems more appropriate to me and if everyone is onboard with this, that'd be awesome. This PR page already takes a few seconds to load for me 🤣

I'd say that this PR is below the bar in terms of quality we aim to maintain the v1 codebase, but I also appreciate just iterating on main rather than aiming for a perfect solution.

Before we take another stab at a round of reviews, @0xBigBoss, can you do the following:
1. Create a follow GitHub issue outline what you intend to do in the follow-up PR (e.g. the structure of the configs, the deliverables, etc...)
2. Add `TODO(#XXX)` (once you have the ticket issue) in this PR showing where the changes will take place. 

This will help with:
1. Scoping out the follow-up work
2. Avoid losing track of the "hacks" implemented in this PR
3. Make it clear to new-comers in the codebase what work on main is a WIP
4. Prevent others from following the patterns introduced in this PR

0xBigBoss:
@Olshansk @h5law still need to do manual testing and documentation of the new capabilities of the CLI, but I have most of the code in place now.

runtime/defaults/defaults.go
```diff
@@ -2,10 +2,26 @@ package defaults
...
+	"os"
 
 	"github.com/pokt-network/pocket/runtime/configs/types"
 )
 
+func init() {
+	initDefaultRootDirectory()
+}
+
+func initDefaultRootDirectory() {
```
h5law, line 14: Is there a reason this couldnt go in the `init()` function directly?

runtime/configs/types/keybase_ext.go
```diff
@@ -0,0 +1,25 @@
+package types
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+func (k KeybaseType) MarshalJSON() ([]byte, error) {
+	return json.Marshal(KeybaseType_name[int32(k)])
```
h5law, line 10: For my own understanding can you explain `KeybaseType_name` and `KeybaseType_value`? I am unsure what they are and where they come from.

app/client/cli/keys.go
```diff
@@ -141,6 +87,13 @@ func keysCreateCommands() []*cobra.Command {
 			},
 		},
 	}
+
+	// Add --pwd and --hint flags
```
h5law, line 91: Comments make this a lot clearer 👍🏼

shared/crypto/keypair.go
```diff
@@ -135,6 +136,26 @@ func (kp *encKeyPair) Unmarshal(bz []byte) error {
 	return nil
 }
 
+// UnmarshalJSON Unmarshals a JSON string into an encKeyPair struct
+func (ekp *encKeyPair) UnmarshalJSON(data []byte) error {
+	type Alias encKeyPair
+	aux := &struct {
```
h5law, line 142: This temporary struct is very similar to what was there before but I think its outside the scope of this PR to add a Protobuf for the KeyPair.

runtime/defaults/defaults.go
0xBigBoss: nope, just a style thing

runtime/configs/types/keybase_ext.go
0xBigBoss: Sure, they are generated code for the enum type. This code exists so we can have the human-readable names in the config.json instead of ints.

```go
// keybase.pb.go

type KeybaseType int32

const (
	KeybaseType_FILE KeybaseType = 0
	// Hashicorp Vault
	KeybaseType_VAULT KeybaseType = 1
)

// Enum value maps for KeybaseType.
var (
	KeybaseType_name = map[int32]string{
		0: "FILE",
		1: "VAULT",
	}
	KeybaseType_value = map[string]int32{
		"FILE":  0,
		"VAULT": 1,
	}
)
```

shared/crypto/keypair.go
0xBigBoss: right, yeah I think we're at the limits of what go can do for us with json marshalling and custom types.

Olshansk:
@0xBigBoss Appreciate all the changes, but I want to make sure we document the next steps in terms of the configs w/ a ticket and `TODO(#XXX): ...` comments throughout this codebase.

This PR currently adds the most "anti-patterns" that we've been trying to avoid using the runtime manager, configs, etc... I understand the importance of iterating on main, but just want to make sure it's clear to anydev who is looking at the codebase.

I can take another look at the PR then.

0xBigBoss:
> Did another round of review (not the entire thing) and left some nits/comments throughout.
> 
> However, I wanted to consolidate these three threads:
> 
> * [#537 (files)](https://github.com/pokt-network/pocket/pull/537/files#r1119419089)
> * [#537 (files)](https://github.com/pokt-network/pocket/pull/537/files#r1119580562)
> * [#537 (files)](https://github.com/pokt-network/pocket/pull/537/files#r1119564427)
> 
> I believe what we should do is:
> 
> 1. Create `keybase_config.proto` in `runtime/configs/proto` - thereby removing the new structrs introduced
> 2. Update `runtime/configs/config.go` by adding in its place `KeybaseConfig`
> 3. Add `POCKET_` prefix to the env variables introduced so they can be overriden by `viper`
> 4. Split out the embedding on the yaml file into a separate PR (see [#525 (comment)](https://github.com/pokt-network/pocket/issues/525#issuecomment-1459073082))
> 5. In a future PR, we can update the k8s manifests to include the keybase configs and test in dev (cc @okdas)
> 
> cc @deblasis for visibility

Hey @Olshansk Sry it wasn't clear in my previous reply. I ended up adding 1-3 to this PR. 4 & 5 were addressed in #578. Unless there were other items I missed?

@deblasis you may want to take a peek at the runtime and configs now.

deblasis:
Hey @0xBigBoss I am almost done reviewing this but I had to pause to look into something else. Bear with me. I should be able to resume and complete it late(r) today. 🙏

shared/crypto/ed25519.go
```diff
@@ -195,6 +195,10 @@ func GenerateAddress() (Address, error) {
 	return pk.Address(), nil
 }
 
+func (pub Ed25519PublicKey) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
0xBigBoss: This would work if `encKeyPair` had public key as a pointer receiver type. https://github.com/0xBigBoss/pocket/blob/147fd72e1516d813e4bfd21c14c4504508b5fac8/shared/crypto/keypair.go#L46

```go
// encKeyPair struct stores the public key and the passphrase encrypted private key
// The encrypted private key is stored as a JSON string with the fields needed to decrypt the key
type encKeyPair struct {
	PublicKey     PublicKey `json:"public_key"`
	PrivKeyArmour string    `json:"priv_key_armour"`
}
```

runtime/configs/config.go
```diff
@@ -15,11 +27,62 @@ type Config struct {
...
+	viper.AutomaticEnv()
+
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok && cfgFile == "" {
+			// if file does not exist and there is no override, return default config
+			// DISCUSS: how should we handle this? Should we return an error? Should we create the default file?
+			return config
+		} else {
+			// Some other error occurred while reading the config file
+			log.Fatalf("[ERROR] failed to read config %s", err.Error())
```
Olshansk, line 62: I think this was added before we had a logger.

Can you change these to `logger.Global.Error(err).Fatalf("failed to read config")`. Ditto below

app/client/keybase/hashicorp/vault.go
```diff
@@ -0,0 +1,339 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return pubKey.Verify(msg, sig), nil
+}
+
+// Delete a keypair from vault
+func (vk *vaultKeybase) Delete(address, passphrase string) error {
+
```
Olshansk, line 307:
```suggestion
```

app/client/keybase/doc/vault.md
```diff
@@ -0,0 +1,225 @@
+# Keybase Vault
+
+The Keybase Vault stores key pairs, `(public_key, armoured_private_key)`, using the [Hashicorp Vault](https://www.vaultproject.io/) KV secrets engine. The Keybase Vault is a wrapper around the Hashicorp Vault API.
+
+Keybase Vault requires a few pieces of information to be able to connect to it:
+
+- Address (e.g. `VAULT_ADDR`=http://127.0.0.1:8200/)
```
Olshansk, line 7: Add POCKET prefix

app/client/keybase/hashicorp/vault.go
```diff
@@ -0,0 +1,339 @@
...
+		return err
+	}
+
+	_, err = vk.client.KVv2(vk.mount).Put(context.TODO(), address, map[string]interface{}{
+		"key_pair": string(dataBz),
+		"hint":     hint,
+	})
+
+	return err
+}
```
Olshansk, line 339: Building on top of the work @h5law did, I honestly think the fact that we have a library for this for easy use with vault is absolute 🔥

app/client/keybase/doc/vault.md
```diff
@@ -0,0 +1,225 @@
...
+- Token (e.g. `VAULT_TOKEN`=hvs.25YM7qJDN8S2EpFEA4SL0ciD)
+- Mount Path (e.g. /secret)
+
+You can configure the CLI to use a Hashicorp Vault keybase in three different ways, listed in order of precedence:
+
+1. configuration file
+2. environment variables prefixed with POCKET\_ (e.g., POCKET_KEYBASE_TYPE=vault) to override the configuration file
+3. command-line flags (highest precedence)
+
+## Simple Vault Demo
```
Olshansk, line 17: Note that as a contributor, you can create branches directly in the core repo.

Makes it easier for us to pull and test rather than clone a lot of different repos.

app/client/keybase/doc/vault.md
```diff
@@ -0,0 +1,225 @@
+# Keybase Vault
+
+The Keybase Vault stores key pairs, `(public_key, armoured_private_key)`, using the [Hashicorp Vault](https://www.vaultproject.io/) KV secrets engine. The Keybase Vault is a wrapper around the Hashicorp Vault API.
+
+Keybase Vault requires a few pieces of information to be able to connect to it:
+
+- Address (e.g. `VAULT_ADDR`=http://127.0.0.1:8200/)
+- Token (e.g. `VAULT_TOKEN`=hvs.25YM7qJDN8S2EpFEA4SL0ciD)
```
Olshansk, line 8: Making sure this is not a prod token

Olshansk (changes requested):
Just have two more files to look through, but otherwise, this is some really high-quality work. 🙇 

Thank you for bearing through all the comments!

Screenshot 2023-03-22 at 8 28 21 PM

app/client/cli/cmd.go
```diff
@@ -2,38 +2,46 @@ package cli
...
 	cliExecutableName = "client"
-	keybaseSuffix     = "/keys"
 )
 
 var (
 	remoteCLIURL   string
 	dataDir        string
+	configPath     string
 	nonInteractive bool
+	cfg            configs.Config
```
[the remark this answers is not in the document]
0xBigBoss: Good catch, I made it a pointer to keep the same return type as ParseConfig.

Not sure about it ever being null though, I guess that would require refactoring more of the runtime since I think it everywhere just assumes it's not null.

app/client/cli/utils.go
```diff
@@ -239,6 +249,45 @@ func attachChildPwdFlagToSubcommands() []cmdOption {
 	}}
 }
 
+func attachKeybaseFlagsToSubcommands() []cmdOption {
+	return []cmdOption{func(c *cobra.Command) {
+		c.Flags().StringVar(&kbTypeStrFromCLI, "keybase", "file", "keybase type used by the cmd, options are: file, vault")
+		c.Flags().StringVar(&keybaseVaultAddrFromCLI, "vault-addr", "", "Vault address used by the cmd. Defaults to https://127.0.0.1:8200 or VAULT_ADDR env var")
```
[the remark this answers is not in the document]
0xBigBoss: So it's both, with the `POCKET_VAULT_ADDR` higher precedence. There is a caveat for the `POCKET_` prefix though. 

For `POCKET_VAULT_ADDR` to work, you need to have the `keybase.vault_addr` set in the config.json which is what I found when testing manually. I believe if we use their [`SetDefault` mechanism](https://pkg.go.dev/github.com/spf13/viper#readme-establishing-defaults) mechanism it _could_ work but I haven't confirmed this.

Hashicorp clients have standard env vars (similar to AWS) that they look for when initializing an SDK. `VAULT_ADDR` and `VAULT_TOKEN` are some of the more prominent ones.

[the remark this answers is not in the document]
0xBigBoss: i think i found a workaround, stay tuned.

app/client/keybase/doc/vault.md
0xBigBoss: nope, but it looked like one.

app/client/keybase/doc/vault.md
0xBigBoss: oh my bad thought I pushed to the PNI one.

app/client/keybase/hashicorp/vault.go
```diff
@@ -0,0 +1,339 @@
...
+	data, err := vk.client.KVv2(vk.mount).Get(context.TODO(), address)
+	if err != nil {
+		return nil, err
+	}
+
+	if data == nil {
+		return nil, errors.New("key not found")
+	}
+
+	keyPairStr, ok := data.Data["key_pair"].(string)
```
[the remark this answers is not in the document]
0xBigBoss: Cuz of the way we write it to vault. see https://github.com/pokt-network/pocket/blob/04f246c4662d504ce26ce99f2441a8849c71cc61/app/client/keybase/hashicorp/vault.go#L338

This makes it readable as a json string.

app/client/keybase/hashicorp/vault.go
```diff
@@ -0,0 +1,339 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return privKey, nil
+}
+
+// GetAll get all keys at this path, NOTE: these may not all be keypairs so practice good hygiene
+func (vk *vaultKeybase) GetAll() ([]string, []crypto.KeyPair, error) {
+	data, err := vk.client.Logical().List(fmt.Sprintf("%s/metadata", vk.mount))
```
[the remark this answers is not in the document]
0xBigBoss: yup, will do.

runtime/configs/config.go
0xBigBoss: Need to move things around with our logger_config.proto first. This causes import cycles.

runtime/configs/proto/keybase_config.proto
```diff
@@ -0,0 +1,23 @@
...
+package configs;
+
+import "keybase.proto";
+
+option go_package = "github.com/pokt-network/pocket/runtime/configs";
+
+message KeybaseConfig {
+
+  // The type of keybase to use. Can be FILE or VAULT. Defaults to FILE.
+  keybase.KeybaseType type = 1;
```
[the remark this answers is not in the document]
0xBigBoss: This looks very clean, how would you tell the config which one to load though? Could marshalling handle all of that?

app/client/keybase/doc/vault.md
0xBigBoss: ahh there we go should all be sorted out now . Screenshot 2023-03-23 at 23 58 09

runtime/configs/proto/keybase_config.proto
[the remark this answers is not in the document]
Olshansk: It's the presence/type of `keybase_config` that determines it.

We do it with stateSync messages:

```proto
message StateSyncMessage {
    oneof message {
        StateSyncMetadataRequest metadata_req = 2;
        StateSyncMetadataResponse metadata_res = 3;
        GetBlockRequest get_block_req = 4;
        GetBlockResponse get_block_res = 5;
    }
}
```

```go
func (m *consensusModule) handleStateSyncMessage(stateSyncMessage *typesCons.StateSyncMessage) error {
	switch stateSyncMessage.Message.(type) {
	case *typesCons.StateSyncMessage_MetadataReq:
	   // ...
	   // ...
	case *typesCons.StateSyncMessage_GetBlockReq:
	   // ...
	case *typesCons.StateSyncMessage_GetBlockRes:
	   // ...
	default:
		return fmt.Errorf("unspecified state sync message type")
	}
}

File: app/client/cli/utils.go (hunk 49 above)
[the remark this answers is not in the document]
Comment (Olshansk) [reply app/client/cli/utils.go]: > Hashicorp clients have standard env vars (similar to AWS) that they look for when initializing an SDK. VAULT_ADDR and VAULT_TOKEN are some of the more prominent ones.

Makes sense, let's keep it.

> For POCKET_KEYBASE_VAULT_ADDR to work, you need to have the keybase.vault_addr set in the config.json which is what I found when testing manually. I believe if we use their [SetDefault mechanism](https://pkg.go.dev/github.com/spf13/viper#readme-establishing-defaults) it could work without having it set in the config.json but I haven't confirmed this.

@deblasis actually ran into it in the past, and here's his fix: https://github.com/pokt-network/pocket/compare/issue/cli-viper-environment-vars-fix

@deblasis Why did we never merge this in? I figure @0xBigBoss can include it here.

File: runtime/configs/config.go (hunk 42 above)
Comment (Olshansk) [reply runtime/configs/config.go]: Just leave a TODO so it's not a blocker for this PR.

File: app/client/cli/utils.go (hunk 49 above)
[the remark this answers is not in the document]
Comment (0xBigBoss) [reply app/client/cli/utils.go]: That was the first approach I tried last night, it didn't work well for the nested struct values from what I remember. The approach I took in this PR handles the nested config structs. You can pass map[string]any to SetDefault and Viper handles overriding nested struct keys automagically.

> Establishing Defaults
> A good configuration system will support default values. A default value is not required for a key, but it’s useful in the event that a key hasn't been set via config file, environment variable, remote configuration or flag.
> 
> Examples:
> 
> viper.SetDefault("ContentDir", "content")
> viper.SetDefault("LayoutDir", "layouts")
> viper.SetDefault("Taxonomies", map[string]string{"tag": "tags", "category": "categories"})

File: app/client/keybase/doc/vault.md [hunk 53]
```diff
@@ -0,0 +1,225 @@
...
+# {"level":"info","address":"new_addr_goes_here","parent":"addr_goes_here","index":0,"stored":true,"time":"2023-02-28T09:26:11-04:00","message":"Child key derived"}
+
+p1 Keys Delete --config config-with-vault.json addr_goes_here
+
+## TODO: add SignTx and VerifyTx
+
+p1 Keys SignTx --config config-with-vault.json
+p1 Keys VerifyTx --config config-with-vault.json
+
+```
```
Olshansk, line 225: Why don't I see any secrets even though I can export?

Screenshot 2023-03-24 at 4 00 02 PM



Screenshot 2023-03-24 at 3 58 50 PM

app/client/keybase/hashicorp/vault_test.go
```diff
@@ -0,0 +1,175 @@
...
+	// You can't defer this because os.Exit doesn't care for defer
+	if err := pool.Purge(resource); err != nil {
+		log.Fatalf("Could not purge resource: %s", err)
+	}
+
+	os.Exit(code)
+}
+
+func TestVaultKeybase(t *testing.T) {
+
```
Olshansk, line 99:
```suggestion
```

app/client/keybase/hashicorp/vault_test.go
```diff
@@ -0,0 +1,175 @@
...
+	// Test GetPrivKey
+	origPrivKey := privKey.String()
+	privKey, err = vk.GetPrivKey(privKey.Address().String(), "passphrase")
+	require.NoErrorf(t, err, "error getting private key: %s")
+	assert.Equal(t, origPrivKey, privKey.String())
+
+	// Test ImportFromJSON
+	keyPair, err := crypto.CreateNewKey("passphrase2", "hint")
+	require.NoErrorf(t, err, "error creating key pair: %s")
```
Olshansk, line 124: Since we're not providing any value to format `%s`, what ends up showing up?

Ditto below

app/client/keybase/hashicorp/vault_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Test ExportPrivJSON
+	_, err = vk.ExportPrivJSON(keyPair.GetAddressString(), "passphrase2")
+	require.NoErrorf(t, err, "error exporting private key: %s")
+
+	// Test UpdatePassphrase
+	err = vk.UpdatePassphrase(keyPair.GetAddressString(), "passphrase2", "new-passphrase", "hint")
+	require.NoErrorf(t, err, "error updating passphrase: %s")
+
+	// // Test Sign
```
Olshansk, line 160:
```suggestion
	// Test Sign
```

app/client/keybase/hashicorp/vault_test.go
```diff
@@ -0,0 +1,175 @@
...
+	msg := []byte("hello world")
+	sig, err := vk.Sign(keyPair.GetAddressString(), "new-passphrase", msg)
+	require.NoErrorf(t, err, "error signing message: %s")
+
+	// Test Verify
+	verified, err := vk.Verify(keyPair.GetAddressString(), msg, sig)
+	require.NoErrorf(t, err, "error verifying signature: %s")
+
+	if !verified {
+		require.NoErrorf(t, fmt.Errorf("signature not verified"), "signature not verified")
```
Olshansk, line 170: require.True(t, verified, ...)

Olshansk (changes requested):

app/client/keybase/doc/vault.md
0xBigBoss: ah this was a great catch, ty. fixed here https://github.com/pokt-network/pocket/pull/537/commits/cb796ee4cf74e4d713850c48e6be244e685a7199

app/client/keybase/hashicorp/vault_test.go
0xBigBoss: my bad, the format string is printed :sweat:

Messages:       error marshaling key pair: %s

runtime/configs/proto/keybase_config.proto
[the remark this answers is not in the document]
0xBigBoss: Ah this is very cool, my guess is the trouble would be how to get Viper and json to unmarshall correctly?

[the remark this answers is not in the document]
0xBigBoss: how would u avoid this situation? to set this correctly in viper it wouldn't it be like `POCKET_KEYBASE_KEYBASE_CONFIG_FILE_CONFIG_FILE_PATH` 🤣 

Screenshot 2023-03-24 at 23 58 33
Screenshot 2023-03-24 at 23 58 12

app/client/keybase/doc/vault.md
Olshansk: 1. I rebuilt the binary
2. Created a new key
3. Launched vault - still don't see anything
4. What do I need to do?

Screenshot 2023-03-27 at 1 51 12 PM

runtime/configs/proto/keybase_config.proto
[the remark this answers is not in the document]
Olshansk: That one I do not have an answer for, unfortunately... 

How about creating a GitHub issue link to this comment, providing the context (i.e. screenshots are fine), so we can refactor those in a follow-up config?

Olshansk:
@0xBigBoss

Only two more things:
1. I want to see the secrets showing up in my local vault; https://github.com/pokt-network/pocket/pull/537#discussion_r1149778562
2. Let's create a ticket to cleanup/refactor the configs in a follow-up PR for https://github.com/pokt-network/pocket/pull/537#discussion_r1145620918

app/client/keybase/doc/vault.md
0xBigBoss: hmmm, not sure @Olshansk 

- how did u attempt to connect to the vault this time? flags, config, or env variables?
- what happens when you run `p1 --keybase vault keys list` without configuring address or token or starting the vault? Do you see `{"level":"fatal","error":"Get \"https://127.0.0.1:8200/v1/secret/metadata?list=true\": dial tcp 127.0.0.1:8200: connect: connection refused","time":"2023-03-27T17:06:58-04:00","message":"failed to execute command"}`
- where is ur vault hosted? If docker, are the ports published?

runtime/configs/proto/keybase_config.proto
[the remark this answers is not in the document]
0xBigBoss: cool -> https://github.com/pokt-network/pocket/issues/626

app/client/keybase/doc/vault.md
Olshansk: https://user-images.githubusercontent.com/1892194/228078783-2fffcd3c-89a3-4b60-92a3-56e686d0bcad.mov

0xBigBoss: Apologies, since no config file was provided, it returned early and did not go through the normal viper unmarshal flow., fixed in https://github.com/pokt-network/pocket/pull/537/commits/0382fa25c85b64368b0185e7b970f651630f3b47

Olshansk: Screenshot 2023-03-28 at 1 45 23 PM
Screenshot 2023-03-28 at 1 45 19 PM

Olshansk:
@0xBigBoss I've run everything locally and finally see the secrets in my local vault. Aside from #626, which we can do after this is merged, is there anything else you are working on? I'm asking because I think it's ready to approve but saw that you haven't requested a re-review.

0xBigBoss:
@Olshansk its ready to go. My bad forgot to re-request on this one.

Olshansk (approved):
This is a big one, so please make sure to rerun tests (unit and localnet) and verify locally before merging it in. Otherwise, :shipit:

0xBigBoss:
@Olshansk ended up hitting a snag once I merged in latest main. Documented my findings here https://github.com/pokt-network/pocket/issues/631. Will wait until I can complete localnet testing. Stay tuned.

Olshansk:
> @Olshansk ended up hitting a snag once I merged in latest main. Documented my findings here #631. Will wait until I can complete localnet testing. Stay tuned.

Appreciate the patience and DD @0xBigBoss. The work @dylanlott is doing should soon automate this E2E to prevent regressions on main in the future. 

As long as you can spin up a LocalNet (using compose and k8s) and it makes progress on a few blocks (my PR is not merged in yet), you have a 👍 from me to merge this in.

PR #539, opened 2023-02-23
[Consensus] patch: consensus interface updates in test

bryanchriswhite:
## Description

Attempt to fix failing consensus test introduced in #531.

## Issue

image

```
--- FAIL: TestPacemakerCatchupSameStepDifferentRounds (0.01s)
panic: reflect: Call using *types.NodeId as type uint64 [recovered]
	panic: reflect: Call using *types.NodeId as type uint64

goroutine 34 [running]:
testing.tRunner.func1.2({0xb1a5a0, 0xc00037fa90})
	/usr/lib/go/src/testing/testing.go:1526 +0x24e
testing.tRunner.func1()
	/usr/lib/go/src/testing/testing.go:1529 +0x39f
panic({0xb1a5a0, 0xc00037fa90})
	/usr/lib/go/src/runtime/panic.go:884 +0x213
reflect.Value.call({0xc23580?, 0xc000142300?, 0x1c?}, {0xc2b3b3, 0x4}, {0xc0000ebee8, 0x1, 0xd44798?})
	/usr/lib/go/src/reflect/value.go:442 +0x1a9f
reflect.Value.Call({0xc23580?, 0xc000142300?, 0x0?}, {0xc0000ebee8?, 0x0?, 0x0?})
	/usr/lib/go/src/reflect/value.go:370 +0xbc
github.com/pokt-network/pocket/consensus/e2e_tests.TestPacemakerCatchupSameStepDifferentRounds(0x4087f9?)
	/home/bwhite/Projects/pokt/pocket/consensus/e2e_tests/pacemaker_test.go:185 +0x885
```

## Type of change

Please mark the relevant option(s):

- [ ] New feature, functionality or library
- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Cast leader ID to the new type before passing to `SetLeaderId`.

## Testing

- [ ] `make develop_test`
- [ ] [LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md) w/ all of the steps outlined in the `README`

## Required Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have tested my changes using the available tooling
- [ ] I have updated the corresponding CHANGELOG

### If Applicable Checklist

- [ ] I have updated the corresponding README(s); local and/or global
- [ ] I have added tests that prove my fix is effective or that my feature works
[...]

PR #541, opened 2023-02-23
[Logger] Replace logger values with pointers

bryanchriswhite:
## Description

Previously when setting up a new module according to the logger module readme, one encounters [`hugeParam` linter error](https://golangci-lint.run/usage/linters/#gocritic) about the module's `logger` field.

## Issue

image

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Update the logger module interface to use logger pointers instead of values
- Update logger value references with pointers
- Update README

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/logger/module.go
+++ b/logger/module.go
@@ -57,10 +57,7 @@ func Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, e
 	return new(loggerModule).Create(bus, options...)
 }
 
-// CreateLoggerForModule returns the logger with additional context (e.g. module name)
-// (see: https://github.com/rs/zerolog#sub-loggers-let-you-chain-loggers-with-additional-context)
-// NB: returns a pointer to mitigate `hugParam` linter error.
-// (see: https://golangci-lint.run/usage/linters/#gocritic)
+// CreateLoggerForModule implements the respective `modules.Logger` interface member.
 func (*loggerModule) CreateLoggerForModule(moduleName string) *modules.Logger {
 	logger := Global.Logger.With().Str("module", moduleName).Logger()
 	return &logger
--- a/logger/module.go
+++ b/logger/module.go
@@ -57,8 +57,13 @@ func Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, e
 	return new(loggerModule).Create(bus, options...)
 }
 
-func (*loggerModule) CreateLoggerForModule(moduleName string) modules.Logger {
-	return Global.Logger.With().Str("module", moduleName).Logger()
+// CreateLoggerForModule returns the logger with additional context (e.g. module name)
+// (see: https://github.com/rs/zerolog#sub-loggers-let-you-chain-loggers-with-additional-context)
+// NB: returns a pointer to mitigate `hugParam` linter error.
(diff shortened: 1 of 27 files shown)


logger/module.go
```diff
@@ -57,8 +57,13 @@ func Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, e
 	return new(loggerModule).Create(bus, options...)
 }
 
-func (*loggerModule) CreateLoggerForModule(moduleName string) modules.Logger {
-	return Global.Logger.With().Str("module", moduleName).Logger()
+// CreateLoggerForModule returns the logger with additional context (e.g. module name)
```
Olshansk, line 60: Not actionable.

Thanks for this. We definitely don't do a good job at using godoc at the moment, so do continue doing this whenever you find an opportunity.

persistence/genesis.go
```diff
@@ -13,8 +13,6 @@ import (
 
 // CONSIDERATION: Should this return an error and let the caller decide if it should log a fatal error?
 func (m *persistenceModule) populateGenesisState(state *genesis.GenesisState) {
-	m.logger.Info().Msg("Populating genesis state...")
```
Olshansk, line 16: Not useful in your opinion?

shared/modules/logger_module.go
```diff
@@ -15,7 +15,9 @@ type LoggerModule interface {
 	// TODO(#288): Embed `ObservableModule` inside of the `Module` interface so each module has its own context specific logger
 	ObservableModule
 
-	// `CreateLoggerForModule` returns the logger with additional context (e.g. module name)
-	// https://github.com/rs/zerolog#sub-loggers-let-you-chain-loggers-with-additional-context
-	CreateLoggerForModule(string) Logger
+	// CreateLoggerForModule returns the logger with additional context (e.g. module name)
```
Olshansk, line 18: Do we need the same comment both next to the interface and the implementation?

Don't have a good answer to this, just feels repetative.

Olshansk (approved):
Thanks @bryanchriswhite! Left a couple questions/comments, but also lgtm (looks good to merge) as is!

shared/modules/logger_module.go
bryanchriswhite: I was similarly uncertain and defaulted to verbosity.

bryanchriswhite: Wdyt about leaving the on the interface and the making the comment on the implementation something like this:
```
// CreateLoggerForModule implements the `modules.Logger` interface member.
```

bryanchriswhite:
## CI Failure

As of 21c9fe0, tests are failing in CI due to #539. This is the only failure and can be confirmed by clicking on the gear icon --> "View raw logs" of [the respective run](https://github.com/pokt-network/pocket/actions/runs/4260978062/jobs/7414758531). Then search the raw logs for "--- fail" to find thee test failure.

shared/modules/logger_module.go
Olshansk: I LIKE IT!

Throw it in the code review guideline ;)

PR #543, opened 2023-02-23
[Consensus] Fixed `TestPacemakerCatchupSameStepDifferentRounds` test

deblasis:
## Description

This PR removes the call to `SetLeaderId` via reflection and it seems that this fixes the test. 

Previously we were calling
```go
        consensusModImpl.MethodByName("SetLeaderId").Call([]reflect.Value{reflect.Zero(reflect.TypeOf(&leaderId))})
```

and the "Zero Value" of a pointer is `nil`:

https://go.dev/play/p/gml5ADTLNLZ

image

SetLeaderId` was:
```go
func (m *consensusModule) SetLeaderId(leaderId *typesCons.NodeId) {
	m.leaderId = leaderId
}
```

so we were essentially setting the leader to `nil`. 

Removing the `SetLeaderId` call altogether seems to have the same result so I now wonder if the test logic is correct.

## Issue

N/A (<4h of work)

## Type of change

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Removed `SetLeaderId` function from the interface and test.

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


consensus/state_sync/helpers.go
```diff
@@ -34,9 +34,9 @@ func (m *stateSync) sendToPeer(msg *anypb.Any, peerId cryptoPocket.Address) erro
 func getMessageType(msg *typesCons.StateSyncMessage) string {
 	switch msg.Message.(type) {
 	case *typesCons.StateSyncMessage_MetadataReq:
-		return "MetadataRequest"
```
Olshansk, line 37: This is how it's being used:

```go
	fields := map[string]any{
		"height":     height,
		"peerId":     peerId,
		"proto_type": getMessageType(stateSyncMsg),
	}

	m.logger.Info().Fields(fields).Msg("Sending StateSync Message")
	return m.sendToPeer(anyMsg, peerId)
```

I think we should leave this as `MetadataRequest` OR rename `MetadataRequest` to `PeerMetadataRequest`, but `StateSyncMetadataResponse` seems wrong.

Olshansk (changes requested):
@gokutheengineer 

1. Sorry for merging in broken code - no excuses on my emd

2. I didn't realize that the investigation to fix the logs was going to take more time such that a broken CI was going to rollover onto the next day. Let's split hotfix back into its own PR and do the followup work separately. Let's always go with this approach moving forward

consensus/debugging.go
```diff
@@ -194,7 +189,7 @@ func (m *consensusModule) sendGetMetadataStateSyncMessage(_ *messaging.DebugMess
 		}
 		valAddress := cryptoPocket.AddressFromString(val.GetAddress())
 		if err := m.stateSync.SendStateSyncMessage(stateSyncMetaDataReqMessage, valAddress, requestHeight); err != nil {
-			m.logger.Error().Err(err).Str("proto_type", "GetMetaDataRequest").Msg("failed to send StateSyncMessage")
+			m.logger.Error().Err(err).Str("proto_type", "StateSyncMetadataRequest").Msg("failed to send StateSyncMessage")
```
Olshansk, line 192: This isn't the actual proto type: `StateSyncMetadataRequest`.

consensus/state_sync/helpers.go
```diff
@@ -34,13 +34,13 @@ func (m *stateSync) sendToPeer(msg *anypb.Any, peerId cryptoPocket.Address) erro
...
+		return "PeerMetadataRequest"
 	case *typesCons.StateSyncMessage_MetadataRes:
-		return "MetadataResponse"
+		return "PeerMetadataResponse"
```
Olshansk, line 39: Let's be consistent with `StateSyncMetadataRequest` or `MetadataRequest` etc..

We're mixing things all over. What's your thinking?

Olshansk (changes requested):

consensus/state_sync/helpers.go
gokutheengineer: I think it is best to update all to actual proto names, to remove confusion and mixing, so:
- StateSyncMetadataRequest
- StateSyncMetadataResponse 
- GetBlockRequest 
- GetBlockResponse

Wdyt?

consensus/debugging.go
gokutheengineer: but it is: https://github.com/pokt-network/pocket/blob/main/consensus/types/proto/state_sync.proto

Olshansk: Good point

consensus/state_sync/helpers.go
Olshansk: Good idea, let's do it

gokutheengineer (approved):

Olshansk (approved):

PR #561, opened 2023-03-07
adds send and broadcast label to debug CLI actions

dylanlott:
## Issue

Fixes #<issue_number>

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Change #1
- Change #2
- ...

## Testing

- [ ] `make develop_test`

## Required Checklist

### If Applicable Checklist


shared/messaging/proto/debug_message.proto
```diff
@@ -22,12 +22,17 @@ enum DebugMessageAction {
...
 }
 
 message DebugMessage {
   DebugMessageAction action = 1;
-  google.protobuf.Any message = 2;
+  DebugMessageType type = 2;
+  google.protobuf.Any message = 3;
 }
+
+enum DebugMessageType {
```
bryanchriswhite, line 33: nit: How do you feel about something like `DebugMessageRoutingType`? I found it useful to think about overlay network routing in terms of "[routing / delivery schemes](https://en.wikipedia.org/wiki/Routing)".

Olshansk: Great suggestion Bryan! We could very well just go with `AnyCast`, `Multicast`, `Broadcast` and `Unicast`. 

Our `Send` is actually an `Anycast` per the formal list of delivery schemes.

Olshansk (changes requested):

shared/messaging/proto/debug_message.proto
bryanchriswhite: I interpreted `Send` being comparable to `Unicast`:

> Unicast delivers a message to a single specific node using a one-to-one association between a sender and destination: each destination address uniquely identifies a single receiver endpoint.

> Anycast delivers a message to any one out of a group of nodes, typically the one nearest to the source using a one-to-one-of-many association where datagrams are routed to any single member of a group of potential receivers that are all identified by the same destination address. The routing algorithm selects the single receiver from the group based on which is the nearest according to some distance or cost measure.

Unless I'm mistaken, this bit about "single member of a group of potential receivers that are all identified by the same destination address" in the `Anycast` definition doesn't sound quite right to me for our case.

Olshansk: **Current Usage**: The current usage of `Send` is only `PromptShowLatestBlockInStore`.

**Goal**: Get one of the nodes in the network to show the latest block

**Current implementation (arbitrarily selects 0)**:

Screenshot 2023-03-09 at 12 31 46 PM

I can argue it both ways:
- The goal was to send it to any node having the latest block (i.e. Anycast)
- The current implementation sends it to the 0th validators because it was hardcoded for no reason other than having something (i.e. Unicast)

I'm okay with w/e terminology you choose. Potentially adding a comment where ned be.

bryanchriswhite: Perhaps we've been thinking about this from different perspectives. I take your point about the relationship between the request and the network state; however, I think that the logic of the conclusion relies on a different categorical theoretic model than that of the network layer, which is what I thought we were talking about. The network layer is below (and therefore, unaware of) the protocol layer, which is where the concept of network state emerges. Therefore, I think to actually consider a model which resembles what you suggest would mean thinking about routing at the protocol layer, separate from the network layer. I'm not convinced that would be useful though, could it be that we're just mixing metaphors?

dylanlott: I like this suggestion and updated to reflect the name. I decided to go with Anycast and left a comment in the code describing that decision. 👍

shared/messaging/proto/debug_message.proto
```diff
@@ -22,12 +22,20 @@ enum DebugMessageAction {
...
 }
 
 message DebugMessage {
   DebugMessageAction action = 1;
-  google.protobuf.Any message = 2;
+  DebugMessageRoutingType type = 2;
+  google.protobuf.Any message = 3;
 }
+
+// NB: See https://en.wikipedia.org/wiki/Routing for more info on routing and delivery schemes.
```
Olshansk, line 33: Very @bryanchriswhite of you ;)

shared/messaging/proto/debug_message.proto
Olshansk: @bryanchriswhite We might be overthinking a bit but I think what Dylan has is more than good enough and pretty clear. You've given a lot of good food for thought though :)

Olshansk (approved):
@dylanlott Code LGTM.

1. Make sure to test locally (you'll probably be blocked by fixes coming in from @okdas).
2. See the failing CI.
3. When merging - we're currently doing a "squash and merge" and copy-pasta the Github description into the commit message.

Issue #506, opened 2023-02-13
[P2P] Peer Discovery - Enable Raintree


jessicadaugherty:
## Objective

Use Raintree as a router in peer discovery for messaging.

## Origin Document

Opportunity identified while integrating Pocket with LibP2P in https://github.com/pokt-network/pocket/issues/347

## Goals

- improve the efficiency, scalability, and security of message dissemination

## Deliverable

- [ ] Integrate Raintree with libp2p (does raintree need to be added as a module or library?)
- [ ] Modify LibP2P to use Raintree for routing

## Non-goals / Non-deliverables

- Refactoring addressbook

## General issue deliverables

- [ ] Update the appropriate CHANGELOG(s)
- [ ] Update any relevant local/global README(s)
- [ ] Update relevant source code tree explanations
- [ ] Add or update any relevant or supporting [mermaid](https://mermaid-js.github.io/mermaid/) diagrams

## Testing Methodology

- [ ] **All tests**: `make test_all`
- [ ] **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @jessicadaugherty

PR #576, opened 2023-03-10
[P2P] Refactor/consolidate P2P modules

### LINKED ISSUE #554: [P2P] Remove "legacy" P2P module & rename libp2p module directory
## Objective

Remove the legacy P2P module and rename the libp2p module directory and object names for clarity, while also simplifying the P2P module interface and cleaning up technical debt in the debug CLI and node startup.

## Origin Document

Next steps identified in https://github.com/pokt-network/pocket/pull/500

## Goals

- Remove unneeded legacy p2p module (`./p2p`)
- Improve libp2p clarity by renaming the module directory and object names (`./libp2p`)
- Simplify the p2p module interface by removing unnecessary or redundant code to improve maintainability
- Cleaning up technical debt in the debug CLI and node startup

## Deliverable

- [ ] Code associated with the legacy P2P module is removed from the codebase
- [ ] Libp2p module directory and associated object names are renamed to reflect their functionality and improve code clarity
- [ ] The p2p module interface is simplified and unnecessary or redundant code is removed.
- [ ] Technical debt introduced in the debug CLI and node startup is cleaned up
- [ ] Updated p2p documentation to reflect the module changes

- Adding new features or changing functionality to p2p business logic

## General issue deliverables

## Testing Methodology

---

**Co-owner**: @bryanchriswhite

bryanchriswhite:
## @Reviewer

This PR may be more digestible / reviewable on a commit-by-commit basis. Commits are organized logically and any given line is only modified in a single commit, with few exceptions.

---

## Description

Re-consolidates Libp2p and P2P modules and types.

## Issue

Fixes #554

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Moved peer & url conversion utils to `p2p/utils` package
- Refactor `getPeerIP` to use `net.DefaultResolver` for easier testing
- Moved & refactor libp2p `host.Host` setup util to `p2p/utils`
- Consolidated Libp2p & P2P `modules.Module` implementations
- Consolidated Libp2p & P2P `stdnetwork` `typesP2P.Network` implementations
- Refactored raintree `typesP2P.Network` implementation to use libp2p
- Moved `shared/p2p` package into `p2p/types` packages
- Removed `Trnasport` interface and implementations
- Removed `ConnectionFactory` type and related members
- Added libp2p `host.Host` mock generator
- Refactor debug CLI post P2P module re-consolidation
- Removed *temporary* `shared/p2p` package
- Removed `runtime/configs.Config#UseLibp2p` field
- Use pod IP for validator DNS resolution tilt localnet
- Add `LIBP2P_DEBUG` env var
- Debug logging improvements
- Set validator `POCKET_P2P_HOSTNAME` env var to the pod IP
- Set validator `p2p.hostname` config value to empty string so that the env var applies

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist


Olshansk:
@bryanchriswhite Are you still blocked on this or getting support from @deblasis?

bryanchriswhite:
> @bryanchriswhite Are you still blocked on this or getting support from @deblasis?

Still working through it and got some help from @deblasis. We paired for a bit this morning and thought through how to proceed. I put the [notes from our session in notion](https://www.notion.so/pocketnetwork/Raintree-test-triaging-0001e3d5d89a4c6fb2cffa42c256c570).

p2p/raintree/network.go
```diff
@@ -24,43 +27,51 @@ var _ typesP2P.Network = &rainTreeNetwork{}
...
-		peersManager:          pm,
 		nonceDeduper:          mempool.NewGenericFIFOSet[uint64, uint64](int(p2pCfg.MaxMempoolCount)),
 		pstoreProvider:        pstoreProvider,
 		currentHeightProvider: currentHeightProvider,
 		logger:                networkLogger,
 	}
 	n.SetBus(bus)
-	return typesP2P.Network(n)
+
+	if err := n.setupDependencies(); err != nil {
```
Olshansk, line 70: I think we should have a consistent style on whether `setupDependencies` fatals or returns an error. No preference on which one as long as its consistent.

Screenshot 2023-03-27 at 1 01 37 PM

Olshansk (changes requested):

p2p/utils/host.go
```diff
@@ -0,0 +1,52 @@
...
+	// TECHDEBT: consider more carefully and parameterize.
+	// TECHDEBT: unexport after consolidation of P2P modules
+	DefaultPeerTTL = 2 * week
+)
+
+// PopulateLibp2pHost iterates through peers in given `pstore`, converting peer
+// info for use with libp2p and adding it to the underlying libp2p host's peerstore.
+// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p@v0.26.2/core/host#Host)
+// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p@v0.26.2/core/peerstore#Peerstore)
+func PopulateLibp2pHost(host libp2pHost.Host, pstore p2p.Peerstore) error {
```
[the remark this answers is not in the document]
bryanchriswhite: > Do you commit locally things that don't compile?

Early in my workflow, usually not. I also find value in making gradual steps where the direction is clear (i.e. not breaking everything). However, I find that more often than not, the commits I produce initially aren't optimized for review (:hankey::sweat_smile:). Once I'm confident that I have the sufficient changes for what I'm working on (i.e. the code is working and doesn't look like trash), I'll look at the all the changes holistically and write up a story that explains them in the form of commit messages. I think that considering the boundaries between and ordering of commits at this stage saves additional (and more technically demanding) refinement later. I think it's also easier and clearer at this point to decide/notice whether and where a PR/branch should be split. Once I know how I would like to organizes the changes, I will repeatedly use interactive rebase to split, join and rearrange commits, iteratively, until it looks to me (like a PR that I would want to review).

> Personally, I leverage my local git environment as a way to snapshot things so I can see diffs and revert between different points in time when things break.

[`git bisect`](https://git-scm.com/docs/git-bisect) will still support scenarios where some commits can't be tested; see: [`git bisect skip`](https://git-scm.com/docs/git-bisect#_bisect_skip). In fact, I actually did exactly that on this branch, **after**, cleaning it up for review (i.e. with uncompilable commits). The result is just a larger range of possible offending commits. I would argue that if you take into consideration the fact that in this "review optimized form" each line (and sometimes entire file) is only modified in a single commit, any debugging related risks/downsides are pretty well mitigated.

I would also like to sing the praises of [`git reflog`](https://git-scm.com/docs/git-reflog) as an invaluable "emergency, break glass" move. For example, if I copy/pasted the wrong branch/commit hash while rebasing, or made a mistake in the editor during an interactive rebase, `git reflog` can tell you where you were before (and the sequence of events to the present), regardless of what branch labels have changed. It also just brings me piece of mind, knowing it's there.

p2p/module.go
```diff
@@ -1,119 +1,149 @@
...
 	"github.com/pokt-network/pocket/telemetry"
-	"google.golang.org/protobuf/proto"
-	"google.golang.org/protobuf/types/known/anypb"
 )
 
+const readStreamTimeoutDuration = time.Second * 10
+
+// TECHDEBT: configure timeouts. Consider security exposure vs. real-world conditions).
+// TECHDEBT: parameterize and expose via config.
+// readStreamTimeout is the duration to wait for a read operation on a
```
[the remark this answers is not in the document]
bryanchriswhite: Haha - indeed, thank you!

p2p/utils/host.go
```diff
@@ -0,0 +1,52 @@
...
+
+	libp2pHost "github.com/libp2p/go-libp2p/core/host"
+
+	"github.com/pokt-network/pocket/shared/p2p"
+)
+
+const (
+	week = time.Hour * 24 * 7
+	// TECHDEBT: consider more carefully and parameterize.
+	// TECHDEBT: unexport after consolidation of P2P modules
```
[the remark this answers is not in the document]
bryanchriswhite: Thanks for calling this out! This line is actually referring to this PR. :sweat_smile:

I opened a new issue for the line above: #629

p2p/module.go
```diff
@@ -1,119 +1,149 @@
...
-	bootstrapNodes []string
+	pstoreProvider        providers.PeerstoreProvider
+	identity              libp2p.Option
+	listenAddrs           libp2p.Option
+	// host represents a libp2p network node, it encapsulates a libp2p peerstore
+	// & connection manager. `libp2p.New` configures and starts listening
+	// according to options.
+	// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p#section-readme)
+	host libp2pHost.Host
+	// pubsub is used for broadcast communication
```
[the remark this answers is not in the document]
bryanchriswhite: So, I seem to have refactored out the need for pubsub and just didn't notice such that I came back and cleaned this up. :sweat_smile: We were using it in the libp2p `stdnetwork` analogue but during consolidation its usage reduced to:

```golang
for _, peer := range n.pstore.GetPeerList() {
	if err := utils.Libp2pSendToPeer(n.host, data, peer); err != nil {
		n.logger.Error().
			Err(err).
			Bool("TODO", true).
			Str("pokt address", peer.GetAddress().String()).
			Msg("broadcasting to peer")
		continue
	}
}
```

It would seem what's left over (and we don't need for now) is the module handling the receiving side of the pubsub.

It may make sense to think about this again in terms of pubsub if we get traction with #506 but for now it adds a dependency for no good reason.

p2p/module.go
```diff
@@ -123,31 +153,54 @@ func (m *p2pModule) Start() error {
...
-				continue
-			}
-			go m.handleNetworkMessage(data)
-		}
-	}()
+	if err = m.startHost(); err != nil {
+		return fmt.Errorf("starting libp2pHost: %w", err)
+	}
+
+	listenAddrLogEvent := m.logger.Info()
```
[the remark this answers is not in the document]
bryanchriswhite: It seemed like the most concise way to get the logger to print each listening address. The only alternative I'm aware of would be to use [`zerolog.Event#Array()`](https://pkg.go.dev/github.com/rs/zerolog@v1.27.0#Event.Array) which would require defining a [`LogArrayMarshaler`](https://pkg.go.dev/github.com/rs/zerolog@v1.27.0#LogArrayMarshaler), like in [`p2p/utils/url_conversion.go`](). I've added a `// TECHDEBT` comment to `stringLogArrayMarshaler` referencing #609.

(I also moved this to be inside of `p2pModule#startHost()` in response to the comment below)

p2p/module.go
```diff
@@ -123,31 +153,54 @@ func (m *p2pModule) Start() error {
...
 	return nil
 }
 
 func (m *p2pModule) Stop() error {
-	m.logger.Info().Msg("Stopping network module")
-	if err := m.listener.Close(); err != nil {
-		return err
-	}
-	return nil
+	return m.host.Close()
```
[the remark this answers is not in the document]
bryanchriswhite: 1. Yes - nice catch! :smile: However, as we're apparently not actually using pubsub at this time, I get to be lazy on this one. :sweat_smile: 
2. No, which is why I'm proposing to rename it to `setupNetwork` instead.

Now that I'm looking at it in context, I think we should also set `m.host` to `nil` in `#Stop()` so that `#Start()` will re-create it rather than try to use a closed host.

p2p/module.go
```diff
@@ -177,15 +230,181 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
 
-// TECHDEBT: Define what the node identity is throughout the codebase
+// TECHDEBT(#348): Define what the node identity is throughout the codebase
 func (m *p2pModule) GetAddress() (cryptoPocket.Address, error) {
 	return m.address, nil
 }
 
-func (m *p2pModule) handleNetworkMessage(networkMsgData []byte) {
-	appMsgData, err := m.network.HandleNetworkData(networkMsgData)
+func (m *p2pModule) setupDependencies() {
```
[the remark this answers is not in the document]
bryanchriswhite: Thanks for calling this out! I would prefer to rewrite it to return an error. `setupDependencies` needs to return an error anyways after moving `setupNetwork` into it.

p2p/module.go
```diff
@@ -177,15 +230,181 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
+	}
+
+	m.host, err = libp2p.New(opts...)
+	if err != nil {
+		return fmt.Errorf("unable to create libp2p host: %w", err)
+	}
+	return nil
+}
+
+func (m *p2pModule) isClientDebugMode() bool {
```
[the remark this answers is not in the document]
bryanchriswhite: @deblasis's suggestion from previous reviews IIRC :raised_hands:

p2p/module.go
```diff
@@ -177,15 +230,181 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
+	return m.GetBus().GetRuntimeMgr().GetConfig().ClientDebugMode
+}
+
+// handleStream is called each time a peer establishes a new stream with this
```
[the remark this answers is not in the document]
bryanchriswhite: It was not, just for clarity but it's not irrelevant as per the previous comments regarding pubsub.

p2p/module.go
```diff
@@ -177,15 +230,181 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
+	data, err := io.ReadAll(stream)
+	if err != nil {
+		m.logger.Error().Err(err).Msg("reading from stream")
+		if err := stream.Reset(); err != nil {
+			m.logger.Debug().Err(err).Msg("resetting stream")
+		}
+		return
+	}
+
+	m.handleNetworkData(data)
```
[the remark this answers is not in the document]
bryanchriswhite: Thanks for calling this out! :raised_hands: Since `#handleNetworkData()` is only being called by `#readStream()`, which is run in a go routine per stream, I hadn't noticed yet. I agree that it would be better to have it return an error. I think this moves all the error handling to one place, `readStream`, which is an improvement in maintainability and readability. :+1:

p2p/stdnetwork/network.go
```diff
@@ -19,45 +22,49 @@ var (
...
 func (n *network) NetworkBroadcast(data []byte) error {
 	for _, peer := range n.pstore.GetPeerList() {
-		if _, err := peer.GetStream().Write(data); err != nil {
-			n.logger.Error().Err(err).Msg("Error writing to one of the peers during broadcast")
+		if err := utils.Libp2pSendToPeer(n.host, data, peer); err != nil {
+			n.logger.Error().
+				Err(err).
+				Bool("TODO", true).
+				Str("pokt address", peer.GetAddress().String()).
+				Msg("broadcasting to peer")
```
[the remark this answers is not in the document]
bryanchriswhite: It seems to me like the distinction you're making between "broadcast" and "send" is a matter of routing given message to a given node. Where "broadcasting" is an indirect send to the target node. I.e.: it will also "send" but not directly to the target, the message will be propagated towards the target by other node(s) which receive copies of same message.

I don't think it makes sense to think of "broadcast" as targeting any specific node; but rather to think of its target as the network (or network segment, classically). As we don't currently have a conception of a "broadcast address" (nor do we need one right now), the "network address" degree of freedom reduces to zero. Subsequently, when I read this, I think something like "this is the peer this node is currently sending to in order to broadcast the given message". Does that make sense? Do you have any suggestions on how to better reflect that in the code, comments, and/or docs?

p2p/module_raintree_test.go
```diff
@@ -228,58 +232,66 @@ func testRainTreeCalls(t *testing.T, origNode string, networkSimulationConfig Te
...
 	// Create connection and bus mocks along with a shared WaitGroup to track the number of expected
 	// reads and writes throughout the mocked local network
 	var wg sync.WaitGroup
-	connMocks := make(map[string]typesP2P.Transport)
-	count := 0
-	for _, valId := range valIds {
+	for i, valId := range valIds {
 		expectedCall := networkSimulationConfig[valId]
-		expectedReads := expectedCall.numNetworkReads + 1
+		expectedReads := expectedCall.numNetworkReads
```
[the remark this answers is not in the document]
bryanchriswhite: I'm not fresh on the details but can get them if you need. The gist of it was that the read count was incremented by one because each channel being used in the mock transports would emit once during setup or something like that.

libp2p/network/network_test.go
```diff
@@ -1,140 +0,0 @@
-package network
```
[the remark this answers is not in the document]
bryanchriswhite: Fair point, we could port this to the `stdnetwork` package.

p2p/raintree/network.go
bryanchriswhite: :100: :+1:

p2p/raintree/peers_manager_test.go
```diff
@@ -270,3 +321,34 @@ func getAlphabetPeerstore(t *testing.T, n int) typesP2P.Peerstore {
 func strToAddrList(s string) string {
 	return hex.EncodeToString([]byte(s))
 }
+
+func mockAlphabetValidatorServiceURLsDNS(t *testing.T) (done func()) {
```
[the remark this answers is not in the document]
bryanchriswhite: Apparently not, I expected the linter to complain but it doesn't. Given that there's no context provided by the function signature I think it's useful to give it a name in the signature to ease readability on the consumer/caller side.

p2p/module_raintree_test.go
```diff
@@ -228,58 +232,66 @@ func testRainTreeCalls(t *testing.T, origNode string, networkSimulationConfig Te
...
+		p2pMod.host.SetStreamHandler(protocol.PoktProtocolID, func(stream libp2pNetwork.Stream) {
+			log.Printf("[valID: %s] Read\n", sURL)
+			(&mod).handleStream(stream)
+			wg.Done()
+		})
 	}
 
 	// Wait for completion
-	defer waitForNetworkSimulationCompletion(t, &wg)
+	defer func() {
```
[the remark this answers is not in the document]
bryanchriswhite: Nice! TIL: `t.Cleanup()` - thanks for sharing! :raised_hands:

p2p/utils/host.go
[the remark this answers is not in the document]
Olshansk: Thank you for the explanation.

Re local dev workflow - makes sense and think that finding the most logical/optimal for each individual is the way to go 👨‍💻 

Re git tools - on the same page 🤝 

Re dev environment - VSCode forever ✊

p2p/module_raintree_test.go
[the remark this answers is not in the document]
Olshansk: Makes a ton  of sense, thank you!

p2p/module.go
```diff
@@ -56,16 +56,8 @@ type p2pModule struct {
...
-	// pubsub is used for broadcast communication
-	// (i.e. multiple, unidentified receivers)
-	pubsub *pubsub.PubSub
-	// topic similar to pubsub but received messages are filtered by a "topic" string.
-	// Published messages are also given the respective topic before broadcast.
-	topic *pubsub.Topic
-	// subscription provides an interface to continuously read messages from.
-	subscription *pubsub.Subscription
-	network      typesP2P.Network
+	host    libp2pHost.Host
```
Olshansk, line 59: omg, so simple :D

p2p/module.go
```diff
@@ -318,9 +313,17 @@ func (m *p2pModule) startHost() (err error) {
 	if err != nil {
 		return fmt.Errorf("unable to create libp2p host: %w", err)
 	}
+
+	// TECHDEBT(#609): use `StringArrayLogMarshaler` post test-utilities refactor.
```
Olshansk, line 317: Nice

p2p/module_raintree_test.go
```diff
@@ -228,58 +233,65 @@ func testRainTreeCalls(t *testing.T, origNode string, networkSimulationConfig Te
...
 		err := p2pMod.Start()
 		require.NoError(t, err)
 
-		for _, peer := range p2pMod.network.GetPeerstore().GetPeerList() {
-			netPeer, ok := peer.(*typesP2P.NetworkPeer)
-			require.True(t, ok, "unexpected `peer` type: %T", peer)
-			netPeer.Transport = connMocks[peer.GetServiceURL()]
-		}
+		sURL := strings.Clone(serviceURL)
+		mod := *p2pMod
```
Olshansk, line 272: Nice trick with the callback below

p2p/module_test.go
```diff
@@ -127,3 +148,18 @@ func Test_Create_configureBootstrapNodes(t *testing.T) {
 		})
 	}
 }
+
+// TECHDEBT(#609): move & de-duplicate
```
Olshansk, line 152: I will start doing this as well. Definitely consider sharing `// TECHDEBT(#609): move & de-duplicate` with the team as a suggestion

p2p/raintree/network.go
```diff
@@ -25,49 +27,61 @@ var _ typesP2P.Network = &rainTreeNetwork{}
...
+	// selfAddr is the pocket address representing this host.
+	selfAddr              cryptoPocket.Address
+	peersManager          *rainTreePeersManager
+	pstoreProvider        peerstore_provider.PeerstoreProvider
+	currentHeightProvider providers.CurrentHeightProvider
+	nonceDeduper          *mempool.GenericFIFOSet[uint64, uint64]
 }
 
-func NewRainTreeNetwork(addr cryptoPocket.Address, bus modules.Bus, pstoreProvider providers.PeerstoreProvider, currentHeightProvider providers.CurrentHeightProvider) typesP2P.Network {
+// TECHDEBT: refactor signature to receive a config struct
```
Olshansk, line 44: NIT: Maybe suggestion to follow the `Create` factory pattern we already use?

p2p/raintree/network.go
```diff
@@ -122,24 +139,23 @@ func (n *rainTreeNetwork) NetworkSend(data []byte, address cryptoPocket.Address)
 	return n.networkSendInternal(bz, address)
 }
 
+// networkSendInternal sends `data` to the peer at pokt `address` if not self.
 func (n *rainTreeNetwork) networkSendInternal(data []byte, address cryptoPocket.Address) error {
 	// NOOP: Trying to send a message to self
+	n.logger.Debug().Str("self", n.selfAddr.String()).Str("target", address.String()).Msg("HEER!!!!")
```
Olshansk, line 145:
```suggestion
	n.logger.Debug().Str("self", n.selfAddr.String()).Str("target", address.String()).Msg("About to send message via RainTree")
```

p2p/utils/url_conversion.go
```diff
@@ -130,6 +131,9 @@ func newResolvePeerIPErr(hostname string, err error) error {
 }
 
 func getPeerIP(hostname string) (net.IP, error) {
+	// TECHDEBT(#595): receive `ctx` from caller.
+	ctx := context.Background()
```
Olshansk, line 135: Why not context.TODO?

p2p/utils/url_conversion.go
```diff
@@ -189,6 +193,7 @@ func getPeerIP(hostname string) (net.IP, error) {
 
 // stringLogArrayMarshaler implements the `zerolog.LogArrayMarshaler` interface
 // to marshal an array of strings for use with zerolog.
+// TECHDEBT(#609): move to test common utilities.
```
Olshansk, line 196: move and de-dup

p2p/utils/host.go
```diff
@@ -1,52 +1,101 @@
...
+	peerInfo, err := Libp2pAddrInfoFromPeer(peer)
+	if err != nil {
+		return err
+	}
+
+	host.Peerstore().RemovePeer(peerInfo.ID)
+	return host.Peerstore().RemoveProtocols(peerInfo.ID)
+}
+
+// Libp2pSendToPeer sends data to the given pocket peer from the given libp2p host.
```
Olshansk, line 78: Makes sense

Olshansk (changes requested):
@bryanchriswhite Just a couple more NITs, but I looked through all your responses and thank you for all the changes!

p2p/utils/url_conversion.go
bryanchriswhite: Missed it, thank you! 🙌

p2p/raintree/network.go
bryanchriswhite: Hehe, nice catch - thanks! 🙌😅

IIRC I was using this to debug and missy have gotten a little sloppy. My apologies. Unless you feel strongly about keeping this line, I would prefer to drop it as it will be very noisy (for no good reason IMO) in the logs.

p2p/raintree/network.go
bryanchriswhite: Great suggestion - thank you! :raised_hands: 

My apologies if I got a bit carried away (i.e. creating extra code to review) but I went ahead with the config struct refactor at the same time. :sweat_smile: As I was thinking about it though I had an idea about how we might formally define this such that we can enforce it - here's where I got to (will include in a PR at some point):

```golang
// e.g. shared/modules/module.go
type FactoryWithOptions[T interface{}] interface {
	Create(bus Bus, opts ...ModuleOption) (T, error)
}

type FactoryWithRequired[T interface{}, K interface{}] interface {
	Create(bus Bus, cfg K) (T, error)
}

type FactoryWithRequiredAndOptions[T interface{}, K interface{}] interface {
	Create(bus Bus, cfg K, opts ...ModuleOption) (T, error)
}
```

```golang
// e.g. p2p/raintree/network.go
var _ modules.FactoryWithRequired[typesP2P.Network, RainTreeConfig] = &rainTreeNetwork{}
```

bryanchriswhite:
I'm not sure what's up with the changelog validation - I haven't modifiied `consensus/docs/CHANGELOG.md` - perhaps it's confused about the merge commits?

p2p/raintree/network.go
```diff
@@ -24,6 +24,13 @@ import (
 
 var _ typesP2P.Network = &rainTreeNetwork{}
```
Olshansk, line 25: You can also enforce that it implements the integratable module interface

Olshansk (approved):
The bonus set of changes LGTM.

@bryanchriswhite The main thing I want to double-check is to make sure you validated this works with both docker-compose and k8s local net? If so, 🚢 !

Olshansk (changes requested):
Discussed offline. Unapproving for now until LocalNet issues are resolved.

Olshansk:
Linking to offline discussion here with the direction/solution for the bug: https://www.notion.so/pocketnetwork/Debugging-576-c5731427abef44ae9b8c05849a62b2b1?pvs=4#0525eff57a8d43a6834e8f9e69ba4521

@bryanchriswhite Let me know when this is ready and I'll take a look at the entire PR a second time.

app/client/cli/debug.go
```diff
@@ -105,13 +105,17 @@ func NewDebugCommand() *cobra.Command {
...
 			if err != nil {
 				logger.Global.Fatal().Err(err).Msg("Failed to create p2p module")
 			}
 
+			var ok bool
+			p2pMod, ok = mod.(modules.P2PModule)
+			if !ok {
+				logger.Global.Fatal().Msgf("unexpected P2P module type: %T", mod)
+			}
+
```
bryanchriswhite, line 118: NB: This cannot be simplified without additional changes because `p2pMod` is a global var which is referenced in other functions in this module.

build/localnet/manifests/cli-client.yaml
```diff
@@ -76,6 +76,8 @@ spec:
               # Any host that is visible and connected to the cluster can be arbitrarily selected as the RPC host
             - name: RPC_HOST
               value: v1-validator001
+            - name: POCKET_P2P_HOSTNAME
+              value: "127.0.0.1"
```
bryanchriswhite, line 80: I added this to resolve this error when using the debug cli in the tilt localnet:
```
{"level":"fatal","error":"parsing multiaddr from config: resolving peer IP for hostname: : resolving peer IP for hostname: , lookup : no such host","time":"2023-04-13T07:21:23Z","message":"Failed to create p2p module"}
command terminated with exit code 1
```

I read this as, it tried to resolve DNS for an empty string

I hypothesize that the reason this is necessary in tilt  and not in docker compose is because the config manifest  has `p2p.is_empty_connection_type: false` and is common to both the validator and the debug cli manifests. I'm attempting to set `POCKET_P2P_IS_CONNECTION_EMPTY` but it didn't seem to work straight away. :thinking: 

```yaml
  - name: POCKET_P2P_IS_EMPTY_CONNECTION_TYPE
    value: "true"
```

build/localnet/manifests/configs.yaml
```diff
@@ -36,6 +36,7 @@ data:
       },
       "p2p": {
         "port": 42069,
+        "hostname": "",
```
bryanchriswhite, line 39: ### Rationale

image

### Is there an issue for this?

p2p/CHANGELOG.md
```diff
@@ -7,6 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Refactor raintree constructor function signature to use new `RainTreeConfig` struct
+
+## [0.0.0.40] - 2023-04-12
+
+- Wrap IPv6 address in square brackets as per RFC3986 §3.2.2
+
+## [0.0.0.39] - 2023-04-12
+
+- Improve URL validation and error handling in Libp2pMultiaddrFromServiceURL function
+
```
bryanchriswhite, line 31: (transferred from the deleted libp2p module's changelog)

p2p/module.go
```diff
@@ -124,31 +153,35 @@ func (m *p2pModule) Start() error {
...
 func (m *p2pModule) Stop() error {
-	m.logger.Info().Msg("Stopping network module")
-	if err := m.listener.Close(); err != nil {
-		return err
-	}
-	return nil
+	err := m.host.Close()
+
+	// Don't reuse closed host, `#Start()` will re-create.
+	m.host = nil
```
bryanchriswhite, line 183: There's a bug here :roll_eyes:

If the module was created with a `WithHost()` option, `#Start()` will ignore if a P2P module is stopped and restarted (which is what the regression test for this will look like).

p2p/module.go
```diff
@@ -178,32 +211,247 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
-// TECHDEBT: Define what the node identity is throughout the codebase
+// TECHDEBT(#348): Define what the node identity is throughout the codebase
 func (m *p2pModule) GetAddress() (cryptoPocket.Address, error) {
 	return m.address, nil
 }
 
-func (m *p2pModule) handleNetworkMessage(networkMsgData []byte) {
-	appMsgData, err := m.network.HandleNetworkData(networkMsgData)
+// setupDependencies sets up the module's current height and peerstore
+// providers, and the network.
```
bryanchriswhite, line 220:
```suggestion
// setupDependencies sets up the module's current height and peerstore providers.
```

p2p/module.go
```diff
@@ -178,32 +211,247 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
+		pstoreProviderModule = persABP.NewPersistencePeerstoreProvider(m.GetBus())
+	} else if pstoreProviderModule != nil {
+		m.logger.Debug().Msg("loaded persistence peerstore...")
+	}
+
+	var ok bool
+	m.pstoreProvider, ok = pstoreProviderModule.(providers.PeerstoreProvider)
+	if !ok {
+		typeErr := fmt.Errorf("unknown peerstore provider type: %T", pstoreProviderModule)
+		return multierr.Append(err, typeErr)
```
bryanchriswhite, line 249:
```suggestion
		return fmt.Errorf("unknown peerstore provider type: %T", pstoreProviderModule)
```

p2p/module.go
```diff
@@ -178,32 +211,247 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
+		return errors.New("no current height provider or consensus module registered")
+	}
+
+	m.logger.Debug().Msg("loaded current height provider")
+
+	var ok bool
+	m.currentHeightProvider, ok = currentHeightProviderModule.(providers.CurrentHeightProvider)
+	if !ok {
+		typeErr := fmt.Errorf("unexpected current height provider type: %T", currentHeightProviderModule)
+		return multierr.Append(err, typeErr)
```
bryanchriswhite, line 273:
```suggestion
		return fmt.Errorf("unexpected current height provider type: %T", currentHeightProviderModule)
```

p2p/module.go
```diff
@@ -178,32 +211,247 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
+			m.host,
+			m.pstoreProvider,
+			m.currentHeightProvider,
+		)
+	}
+	return err
+}
+
+// setupHost creates a new libp2p host and assignes it to `m.host`, if one does
+// not already exist. Libp2p host starts listening upon instantiation.
```
bryanchriswhite, line 301:
```suggestion
// setupHost creates a new libp2p host and assignes it to `m.host`. Libp2p host
// starts listening upon instantiation.
```

p2p/module.go
```diff
@@ -178,32 +211,247 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
+		if err = stream.Reset(); err != nil {
+			m.logger.Error().Err(err).Msg("resetting stream")
+		}
 		return
 	}
 
+	if err := m.network.AddPeer(peer); err != nil {
+		m.logger.Error().Err(err).
+			Str("address", peer.GetAddress().String()).
+			Msg("adding remote peer to address book")
```
bryanchriswhite, line 361:
```suggestion
			Msg("adding remote peer to network")
```

p2p/module.go
```diff
@@ -178,32 +211,247 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
 
 	event := messaging.PocketEnvelope{
 		Content: networkMessage.Content,
 	}
 	m.GetBus().PublishEventToBus(&event)
+	return nil
+}
+
+// getMultiaddr returns a multiaddr constructed from the `hostname` and `port`
+// in the P2P config which pas provided upon creation.
```
bryanchriswhite, line 444:
```suggestion
// in the P2P config which was provided upon creation.
```

p2p/raintree/network.go
```diff
@@ -219,34 +248,85 @@ func (n *rainTreeNetwork) HandleNetworkData(data []byte) ([]byte, error) {
...
+	}
+	return nil
+}
+
+func (n *rainTreeNetwork) setupPeerManager(pstore typesP2P.Peerstore) (err error) {
+	n.peersManager, err = newPeersManager(n.selfAddr, pstore, true)
+	return err
+}
+
+func (cfg RainTreeConfig) isValid() error {
```
bryanchriswhite, line 319: Use `multierr`

p2p/raintree/network_test.go
```diff
@@ -1,59 +1,97 @@
...
 
 	// Start with a peerstore containing self.
-	selfAddr, err := cryptoPocket.GenerateAddress()
-	require.NoError(t, err)
-	selfPeer := &typesP2P.NetworkPeer{Address: selfAddr}
+	selfPeer, host := newTestPeer(t)
+	selfAddr := selfPeer.GetAddress()
 
 	expectedPStoreSize := 0
 	pstore := getPeerstore(nil, expectedPStoreSize)
```
bryanchriswhite, line 31:
```suggestion
	pstore := getPeerstore(t, expectedPStoreSize)
```

p2p/raintree/network_test.go
```diff
@@ -64,29 +102,39 @@ func TestRainTreeNetwork_RemovePeer(t *testing.T) {
 	expectedPStoreSize := 3
 	pstore := getPeerstore(nil, expectedPStoreSize)
```
bryanchriswhite, line 103:
```suggestion
	pstore := getPeerstore(t, expectedPStoreSize)
```

p2p/utils/host.go
```diff
@@ -0,0 +1,116 @@
...
+		return multierr.Append(
+			fmt.Errorf("writing to stream: %w", err),
+			stream.Reset(),
+		)
+	} else {
+		logger.Global.Debug().Int("bytes", n).Msg("written to peer stream")
+	}
+
+	// MUST USE `streamClose` NOT `stream.CloswWrite`; otherwise, outbound streams
+	// will accumulate until resource limits are hit (e.g.):
```
bryanchriswhite, line 113:
```suggestion
	// will accumulate until resource limits are hit; e.g.:
```

p2p/utils/logging.go
```diff
@@ -0,0 +1,54 @@
+package utils
+
+import (
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/pokt-network/pocket/p2p/types"
+	"github.com/pokt-network/pocket/shared/modules"
+	"github.com/rs/zerolog"
+	"net"
```
bryanchriswhite, line 8: group imports

p2p/utils/logging.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type scopeCallback func(scope network.ResourceScope) error
+
+// LogScopeStatFactory returns a function which prints the given scope stat fields
+// to the debug level of the provided logger.
+// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p@v0.27.0/core/network#ScopeStat)
+// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p@v0.27.0/core/network#ResourceScope)
+// TECHDEBT: would be nice to receive either a pocket logger object instead.
```
bryanchriswhite, line 17:
```suggestion
// TECHDEBT: would prefer receive a pocket logger object instead.
// Typical calls pass either `logger.Global` or a `*modules.Logger`.
```

p2p/module.go
```diff
@@ -1,121 +1,151 @@
...
 func Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, error) {
 	return new(p2pModule).Create(bus, options...)
 }
 
-func (*p2pModule) Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, error) {
-	log.Println("Creating network module")
-
-	m := &p2pModule{}
+// WithHostOption associates an existing (i.e. "started") libp2p `host.Host`
+// with this module, instead of creating a new one on `#Start()`.
```
Olshansk, line 74: Add a comment that it's primarily used for testing

p2p/module.go
```diff
@@ -1,121 +1,151 @@
...
+func WithHostOption(host libp2pHost.Host) modules.ModuleOption {
+	return func(m modules.InitializableModule) {
+		mod, ok := m.(*p2pModule)
+		if ok {
```
Olshansk, line 78: optional nit: I think you can single lign this

p2p/module.go
```diff
@@ -124,31 +154,44 @@ func (m *p2pModule) Start() error {
...
 
-	go func() {
-		for {
-			data, err := m.listener.ReadAll()
-			if err != nil {
-				m.logger.Error().Err(err).Msg("Error reading data from connection")
-				continue
-			}
-			go m.handleNetworkMessage(data)
+	// TECHDEBT: reconsider if this is acceptable as more `modules.ModuleOption`s
```
Olshansk, line 157: I think this is fine and pretty general purpose - we can remove the comment

p2p/module.go
```diff
@@ -178,32 +221,244 @@ func (m *p2pModule) Send(addr cryptoPocket.Address, msg *anypb.Any) error {
...
+		pstoreProviderModule = persABP.NewPersistencePeerstoreProvider(m.GetBus())
+	} else if pstoreProviderModule != nil {
+		m.logger.Debug().Msg("loaded persistence peerstore...")
+	}
+
+	var ok bool
+	m.pstoreProvider, ok = pstoreProviderModule.(providers.PeerstoreProvider)
+	if !ok {
+		return fmt.Errorf("unknown peerstore provider type: %T", pstoreProviderModule)
+	}
```
Olshansk, line 258: NIT: This just makes more sense to me.

```suggestion
	pstoreProvider, ok := pstoreProviderModule.(providers.PeerstoreProvider)
	if !ok {
		return fmt.Errorf("unknown peerstore provider type: %T", pstoreProviderModule)
	}
	m.pstoreProvider = pstoreProvicer
```

Olshansk: If you decided to go with it, apply in other places where you have `var ok bool`. If not, just resolve and ignore.

p2p/module.go
Olshansk: Can't we just do a thing where change the stop logic if there's a `WithHost` option provided?

p2p/module.go
```diff
@@ -373,21 +382,42 @@ func (m *p2pModule) readStream(stream libp2pNetwork.Stream) {
 		m.logger.Debug().Err(err).Msg("setting stream read deadline")
 	}
 
+	// debug logging: stream scope stats
+	// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p@v0.27.0/core/network#StreamScope)
+	// TECHDEBT: `logger.Global` is not a `*module.Logger`
```
Olshansk, line 387: Can we not go with the same approach you have elsewhere? `logScope := LogScopeStatFactory(&logger.Global.Logger, "host transient resource scope")`

Olshansk (changes requested):
@bryanchriswhite I'm getting to the point of "PR fatigue" from re-reviewing this PR so I just did a high level review and like the changes made to debugging.

Left a few nits, but assuming it works, and waiting on responses from @okdas on the outstanding questions, lgtm!

build/localnet/manifests/cli-client.yaml
okdas: @bryanchriswhite I think you don't see that behavior on docker-compose because by default CLI uses `client1.json` as a main config: https://github.com/pokt-network/pocket/blob/39d16cd7adaf03dfd329f06a8935399a9ec8b068/app/client/cli/debug.go#L80
Which has `node1.consensus` set as a hostname.

I need to look deeper to figure out why it tries to resolve the empty host string when the value is `127.0.0.1`, I wonder what'd happen if the value was `localhost`. I'd vote for `0.0.0.0` as a default value.

I think requiring such parameters (hostname, Postgres, etc.) is counterintuitive when the user only needs to use CLI. In the current v0 world, users often use CLI to query pocket RPC endpoints to get the latest height, information about nodes, etc. We definitely should not require unnecessary configs. I'll see if such a ticket exists to add more information or create a new one tomorrow.

p2p/module.go
bryanchriswhite: Unfortunately it doesn't seem quite that simple, and I'm realizing that the solution I described is still insufficient: 

1. Options are indistinguishable from one another and opaque with respect to what they modify on whatever gets passed to the option function.
1. A libp2p host can't be stopped, it can only be destroyed, after which it would need to be recreated. This is especially problematic in a "bring your own host" situation as we have no way to re-create that host.

TLDR; we should return an error in the case where a module `WithHost` is restarted. Apologies for the back and forth.

build/localnet/manifests/cli-client.yaml
bryanchriswhite: @okdas Apologies for any confusion, to be clear, setting  `POCKET_P2P_HOSTNAME` to `127.0.0.1` **does resolve the error**, I just wasn't confident that it was the correct or even temporarily acceptable solution. Given your improved explanation for the behavior, it seems like a change need to be made in the code rather than the configs. It shouldn't be necessary for the debug client to specify a hostname that it doesn't use (see #678).

build/localnet/manifests/configs.yaml
bryanchriswhite: :thinking: Unfortunately I don't think it's possible to add a comment inline with the JSON as that would produce invalid JSON. Comments were introduced in [JSON5](https://json5.org/), which has been slow to catch on :sweat_smile:

build/localnet/manifests/cli-client.yaml
bryanchriswhite: > 1. Isn't `is_empty_connection_type` only used for testing? The CLI still needs a non-empty connection type (i.e. TLS) to communicate with the network.

`runtime/configs.P2PConfig#ConnectionType` is only used to inform the listen address that is provided to the libp2p host. In the debug client case, we use `libp2p.NoListenAddrs` but at some other point, we try to construct a self service URL from the P2P config and parse it as a multiaddr anyways.

> 2. W/e solution we decide to go with, adding comments in the yaml file next to the config saying "it's a hack or needed because of Y" would help

:+1:

Olshansk: 👍

Olshansk (approved):
@bryanchriswhite I appreciate all the comments you added after the last PR.

Please do a last sanity check with localnet (docker-compose and k8s) before merging to avoid breaking main, but otherwise, let's merge this in!

bryanchriswhite (commented):
> Please do a last sanity check with localnet (docker-compose and k8s) before merging to avoid breaking main, but otherwise, let's merge this in!

@Olshansk so I have good news and less good news. 

Good news: both docker compose and tilt localnets were able to run in autonomous mode past 250 blocks before I shut them down. :raised_hands: 

Bad news: after merging the tilt validator refactor which introduces the validator helm chart (#658) there is a new class of error presenting with respect to DNS resolution. This error isn't present in `main` and it's not immediately clear to me how this issue wouldn't be presenting there as well as I've not modified genesis (where those hostnames come from) and both branches **are** resolving DNS and the diff there in the code is quite minimal ([`net.DefaultResolver#LookupHost`](https://pkg.go.dev/net#Resolver.LookupHost) instead of [`net.ResolveTCPAddr`](https://pkg.go.dev/net#ResolveTCPAddr)).

```
{"level":"fatal","error":"setting up network: converting peer info, pokt address: 00404a570febd061274f72b50d0a37f611dfe339: resolving peer IP for hostname: v1-validator004: resolving peer IP for hostname: v1-validator004, lookup v1-validator004 on 10.96.0.10:53: no such host; converting peer info, pokt address: 00104055c00bed7c983a48aac7dc6335d7c607a7: resolving peer IP for hostname: v1-validator001: resolving peer IP for hostname: v1-validator001, lookup v1-validator001 on 10.96.0.10:53: no such host; converting peer info, pokt address: 00204737d2a165ebe4be3a7d5b0af905b0ea91d8: resolving peer IP for hostname: v1-validator002: resolving peer IP for hostname: v1-validator002, lookup v1-validator002 on 10.96.0.10:53: no such host; converting peer info, pokt address: 00304d0101847b37fd62e7bebfbdddecdbb7133e: resolving peer IP for hostname: v1-validator003: resolving peer IP for hostname: v1-validator003, lookup v1-validator003 on 10.96.0.10:53: no such host","time":"2023-04-17T07:20:31Z","message":"Failed to start pocket node"}
```

Tapping @okdas once more for help. :sweat_smile: 

@Olshansk (or @okdas) I would like to ask if I may please delegate the responsibility of subsequent manual testing and merging of this PR after this (last last last) fix? It seems that due to the timezones there is always something new on `main` that needs to be merged and manually tested which more often than not leads to the need for more small changes in this PR, causing a cycle. That combined with the PR fatigue that everyone is feeling makes me feel like we should all push to merge this as soon as possible. Let's be catching these issues in other PRs instead of collecting them all here and perpetuating further divergence from `main`.

okdas (approved):
All the configuration changes look good.

PR #578, opened 2023-03-13
[CLI] add private keys embed to debug cli build

0xBigBoss:
## Description

SImplifies the debug CLI tooling by embedding private-keys.yaml manifest into the CLI binary when the `debug` build tag is present. It removes some code from the debug keybase that detected if the CLI was running in kubernetes and would source the debug private keys from a configmap secret.

## Issue

Fixes https://github.com/pokt-network/pocket/issues/525

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [x] Other 

## List of changes

See description.

## Testing

- [x] `make develop_test`

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/app/client/debug.go
+++ b/app/client/debug.go
@@ -3,7 +3,14 @@
 package main
 
 import (
+	// Importing the "debug" package from "github.com/pokt-network/pocket/app/client/keybase/debug".
 	_ "github.com/pokt-network/pocket/app/client/keybase/debug"
 )
 
-// silence is golden
+// This file serves as a feature flag based on the build tag "debug".
+// When the build tag "debug" is present, the init() function in keystore.go is triggered via the anonymous import above.
+// This functionality is intended for debugging purposes only.
+
+// Additional debug functionality in the client CLI could be included here.
+// For example, logging or error handling utilities that are useful for developers when debugging their applications.
+// To run the client CLI with the "debug" build tag, run the following command: go run -tags=debug app/client/*.go
--- a/app/client/debug.go
+++ b/app/client/debug.go
@@ -1,0 +1,9 @@
+//go:build debug
+
+package main
+
+import (
+	_ "github.com/pokt-network/pocket/app/client/keybase/debug"
+)
+
+// silence is golden
--- a/app/client/keybase/debug/keystore.go
+++ b/app/client/keybase/debug/keystore.go
@@ -1,31 +1,27 @@
+//go:build debug
+
 package debug
 
 import (
 	"fmt"
 	"os"
-	"path/filepath"
-	r "runtime"
+	"sync"
 
 	"github.com/pokt-network/pocket/app/client/keybase"
+	"github.com/pokt-network/pocket/build"
 	"github.com/pokt-network/pocket/logger"
-	"github.com/pokt-network/pocket/runtime"
 	cryptoPocket "github.com/pokt-network/pocket/shared/crypto"
-	pocketk8s "github.com/pokt-network/pocket/shared/k8s"
-	"github.com/pokt-network/pocket/shared/utils"
 	"gopkg.in/yaml.v2"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 )
 
 const (
 	// NOTE: This is the number of validators in the private-keys.yaml manifest file
-	numValidators       = 999
-	debugKeybaseSuffix  = "/.pocket/keys"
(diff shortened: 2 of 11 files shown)


app/client/debug.go
```diff
@@ -0,0 +1,9 @@
+//go:build debug
+
+package main
+
+import (
+	_ "github.com/pokt-network/pocket/app/client/keybase/debug"
+)
+
+// silence is golden
```
h5law, line 9: Much cleaner 👍🏼

app/client/keybase/debug/keystore.go
```diff
@@ -73,28 +66,55 @@ func initializeDebugKeybase() error {
...
-
-			// Encode KeyPair into []byte for value
-			keypairBz, err := keyPair.Marshal()
-			if err != nil {
-				return err
-			}
-			if err := wb.Set(addrKey, keypairBz); err != nil {
-				return err
-			}
+			go func(privHexString string) {
```
deblasis, line 82: Sweet

app/client/keybase/debug/keystore.go
```diff
@@ -108,48 +128,18 @@ func initializeDebugKeybase() error {
...
-	// BUG: When running the CLI using the build binary (i.e. `p1`), it searched for the private-keys.yaml file in `github.com/pokt-network/pocket/build/localnet/manifests/private-keys.yaml`
-	// Get private keys from manifest file
-	_, current, _, _ := r.Caller(0)
-	//nolint:gocritic // Use path to find private-keys yaml file from being called in any location in the repo
-	yamlFile := filepath.Join(current, privateKeysYamlFile)
-	if exists, err := utils.FileExists(yamlFile); !exists || err != nil {
-		return nil, fmt.Errorf("unable to find YAML file: %s", yamlFile)
-	}
+// fetchValidatorPrivateKeysFromYaml fetches the validator private keys from the private-keys.yaml manifest file.
+func fetchValidatorPrivateKeysFromYaml() (map[string]string, error) {
```
deblasis, line 132: Not a strong opinion, curious about your thoughts about renaming this to `parseValidatorPrivateKeysFromEmbeddedYaml` or something similar that conveys the intent of this func. 

It would make crystal clear what's going on above where we call this.

If you agree I would update the GoDoc comment above accordingly as well (even tho if we do a good job with naming, the comment should almost sound redundant IMHO).

app/client/debug.go
deblasis: 🤫💯

deblasis: Devil's advocate for newbies coming to this file:

```suggestion
// This file is only here to act as a feature flag based on the presence of the build tag `debug`.
// If that's set, we essentially trigger the `init()` func in `keystore.go` via the anonymous import above.
```

Of course feel free to change the wording as you see fit, I just think it would be good to provide some color on what's happening here.

Wdyt?

app/client/debug.go
```diff
@@ -0,0 +1,9 @@
+//go:build debug
```
deblasis, line 1: I was about to suggest:

```suggestion
//go:build debug
// +build debug
```

but then I realised that it seems that the `+build ...` syntax is obsolete since Go 1.17

Nice

deblasis (commented):
Nice, this is exactly what I meant 🙌

Just a couple of nits/comments/weak opinions but this is 🔥

app/client/debug.go
0xBigBoss: totally agree, @deblasis i forget sometimes to codify tribal knowledge.

app/client/keybase/debug/keystore.go
0xBigBoss: yup agree, another good refactor.

deblasis (approved):
LGTM! 👏

app/client/keybase/debug/keystore.go
```diff
@@ -73,28 +66,55 @@ func initializeDebugKeybase() error {
...
+				}
+			}(privHexString)
 		}
+
+		// Wait for all goroutines to finish
+		wg.Wait()
+
+		// Check if any goroutines returned an error
+		select {
+		case err := <-errCh:
```
Olshansk, line 113: For my personal understanding, did you go with an errChannel instead of a basic list to avoid concurrent writes?

0xBigBoss: Exactly. Without a channel it would have needed a mutex or similar to synchronize access.

PR #649, opened 2023-04-04
Added ChatGPT-CodeReview workflow

Olshansk:
As seen on https://github.com/anc95/ChatGPT-CodeReview.

Triggered by adding the `gpt review` label so its not too expensive and doesn't run every time.

### DIFF
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -1,0 +1,25 @@
+name: GPT Review
+
+permissions:
+  contents: read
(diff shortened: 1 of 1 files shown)


dylanlott (approved):
No showstoppers, approving for now.

h5law:
This is so cool.

Olshansk:
@dylanlott I updated the first suggestion but think we should ride on main as this thing is really new and I expect it to be improving quickly

Issue #634, opened 2023-03-29
[CONSENSUS] Create helper in consensus tests for waiting for FSM events


gokutheengineer:
## Objective

Create a helper function for waiting for FSM events in consensus tests. 

## Origin Document

![Error](https://user-images.githubusercontent.com/118421317/228469706-c92ae00a-1cd8-4cfd-9639-adac295f684c.png)

## Goals

- Ensure state sync tests show deterministic behavior.

## Deliverable

- [ ] Helper function `WaitForFSMEvents` for consensus tests.

## Non-goals / Non-deliverables

- Change business logic

## General issue deliverables

- [ ] Update the appropriate CHANGELOG(s)
- [ ] Update any relevant local/global README(s)
- [ ] Update relevant source code tree explanations
- [ ] Add or update any relevant or supporting [mermaid](https://mermaid-js.github.io/mermaid/) diagrams

## Testing Methodology

- [ ] **All tests**: `make test_all`
- [ ] **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @gokutheengineer 
**Co-Owners**: @Olshansk

Olshansk:
@gokutheengineer I think this is a priority over leader election.

It could also be good candidate for the "shared testing utilities" @bryanchriswhite is working on.

Olshansk:
As we will likely be removing the FSM, I'm closing this out for now

PR #686, opened 2023-04-17
[Consensus] State Sync - Active/Passive

gokutheengineer:
## Description
Includes changes that primarily focus on adding state sync channels to the consensus module, code refactoring, simplification, and improvements to the testing in the consensus module. 

## Issue

Fixes https://github.com/pokt-network/pocket/issues/634.

Fixes the following sub-issues of #352:
- PR1: [Testing] Channel for FSM events: `WaitForNetworkFSMEvents()` is implemented.
- PR4: [Consensus] Create Block and Metadata Response Channels.

## Type of change

- [ ] Bug fix
- [X] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes
- Update consensus testing framework with `WaitForNetworkFSMEvents()` and other helper functions.
- Implement state sync `Start()` and `Stop()` functions.
- Implement `blockApplicationLoop()` that accepts blocks from the network and applies.

## Testing

- [X] `make develop_test`; if any code changes were made

## Required Checklist

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added, or updated, [`godoc` format comments](https://go.dev/blog/godoc) on touched members (see: [tip.golang.org/doc/comment](https://tip.golang.org/doc/comment))
- [X] I have tested my changes using the available tooling
- [X] I have updated the corresponding CHANGELOG

### If Applicable Checklist


consensus/module_consensus_state_sync.go
```diff
@@ -22,13 +23,36 @@ func (m *consensusModule) GetNodeAddress() string {
...
+		}
+
+		err = m.verifyBlock(block)
+		if err != nil {
+			m.logger.Err(err).Msg("failed to verify block")
+			return
+		}
+
+		err = m.applyAndCommitBlock(block)
+		if err != nil {
```
Olshansk, line 49: Single line

consensus/module_consensus_state_sync.go
```diff
@@ -22,13 +23,36 @@ func (m *consensusModule) GetNodeAddress() string {
...
-	// commits the received block
+	for blockResponse := range m.blocksReceived {
+		block := blockResponse.Block
+		maxPersistedHeight, err := m.maxPersistedBlockHeight()
+		if err != nil {
+			m.logger.Err(err).Msg("couldn't query max persisted height")
+			return
+		}
+
+		if block.BlockHeader.Height <= maxPersistedHeight || block.BlockHeader.Height > m.CurrentHeight() {
```
Olshansk, line 37: Can you split this into 2 different if structures with specialized info messages?

consensus/module_consensus_state_sync.go
```diff
@@ -22,13 +23,36 @@ func (m *consensusModule) GetNodeAddress() string {
...
-	// listens on the blocksReceived channel
-	// commits the received block
+	for blockResponse := range m.blocksReceived {
+		block := blockResponse.Block
+		maxPersistedHeight, err := m.maxPersistedBlockHeight()
+		if err != nil {
+			m.logger.Err(err).Msg("couldn't query max persisted height")
+			return
+		}
+
```
Olshansk, line 36: `block.BlockHeader.Height > m.CurrentHeight()`

Does this mean we throw away future blocks? Doesn't this mean we need to re-request them again?

consensus/module_consensus_state_sync.go
```diff
@@ -39,3 +63,56 @@ func (m *consensusModule) metadataSyncLoop() {
...
+	if err != nil {
+		return 0, err
+	}
+
+	return maxHeight, nil
+}
+
+func (m *consensusModule) verifyBlock(block *coreTypes.Block) error {
+	return nil
+}
```
Olshansk, line 84: s/verify/validate

Can you add comments or TODOs on what kind of validation you want to do here?

consensus/state_sync/module.go
```diff
@@ -1,18 +1,25 @@
...
+	committedBlocsChannelSize = 100
 )
 
 type StateSyncModule interface {
 	modules.Module
 	StateSyncServerModule
+
+	Set(aggregatedMetaData *typesCons.StateSyncMetadataResponse)
+	CommittedBlock(uint64)
+	StartSyncing()
```
Olshansk, line 22: Why do we need this in addition to the `Start` function rather than using it directly?

consensus/state_sync/module.go
```diff
@@ -22,14 +29,22 @@ var (
 )
 
 type stateSync struct {
-	bus    modules.Bus
-	logger *modules.Logger
+	bus                    modules.Bus
+	logger                 *modules.Logger
+	validators             []*coreTypes.Actor
```
Olshansk, line 34: This changes on every height. Is it really the right thing to do it to store it locally rather than reading from the persistence module?

Remember: less state -> better

consensus/state_sync/module.go
```diff
@@ -1,18 +1,25 @@
...
+	stateSyncModuleName       = "stateSyncModule"
+	committedBlocsChannelSize = 100
 )
 
 type StateSyncModule interface {
 	modules.Module
 	StateSyncServerModule
+
+	Set(aggregatedMetaData *typesCons.StateSyncMetadataResponse)
+	CommittedBlock(uint64)
```
Olshansk, line 21: `CommittedBlock` sounds like something that happend rather than a function anyhoe would call

consensus/state_sync/module.go
```diff
@@ -1,18 +1,25 @@
...
-	stateSyncModuleName = "stateSyncModule"
+	stateSyncModuleName       = "stateSyncModule"
+	committedBlocsChannelSize = 100
 )
 
 type StateSyncModule interface {
 	modules.Module
 	StateSyncServerModule
+
+	Set(aggregatedMetaData *typesCons.StateSyncMetadataResponse)
```
Olshansk, line 20: You're specifying the type and name here but only the type of the variable below. It's inconsistent so please fix.

Olshansk, line 20: `SetXXX()`

consensus/state_sync/module.go
```diff
@@ -22,14 +29,22 @@ var (
...
+	aggregatedMetaData     *typesCons.StateSyncMetadataResponse
+	committedBlocksChannel chan uint64
 }
 
 func CreateStateSync(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, error) {
 	return new(stateSync).Create(bus, options...)
 }
 
+// CommittedBlock is called by the consensus module when a block received by the network is committed by blockApplicationLoop() function
+func (m *stateSync) CommittedBlock(height uint64) {
```
Olshansk, line 44: Why don't we use the event bus for this?

consensus/module_consensus_state_sync.go
```diff
@@ -39,3 +63,56 @@ func (m *consensusModule) metadataSyncLoop() {
...
+	m.paceMaker.NewHeight()
+
+	m.logger.Info().Msgf("New block is committed, current height is :%d", m.height)
+	return nil
+}
+
+func (m *consensusModule) getAggregatedStateSyncMetadata() typesCons.StateSyncMetadataResponse {
+	minHeight, maxHeight := uint64(1), uint64(1)
+
+	chanLen := len(m.metadataReceived)
```
Olshansk, line 101: WHy don't we just loop through the channel directly?

consensus/state_sync/module.go
```diff
@@ -40,25 +55,83 @@ func (*stateSync) Create(bus modules.Bus, options ...modules.ModuleOption) (modu
...
+	defer readCtx.Release()
+
+	// get the current validators
+	m.validators, err = readCtx.GetAllValidators(int64(currentHeight))
+	if err != nil {
+		return err
+	}
+
+	// requests blocks from the current height to the aggregated metadata height
+	for currentHeight <= m.aggregatedMetaData.MaxHeight {
```
Olshansk, line 92: So this whole function is blocking?

I thought we wanted a background channel that always does this asynchronously, no?

Olshansk (changes requested):
@gokutheengineer I did a quick first round, but I think this PR will need some iteration.

consensus/state_sync/module.go
gokutheengineer: removed, using event bus.

consensus/state_sync/module.go
gokutheengineer: added a background channel process for passive state sync. Active state sync is still like this though, added a note to consider changing this.

consensus/module_consensus_state_sync.go
gokutheengineer: I removed this function for this PR. But, the reason why we don't loop thorugh the channel is we want to get all metadata stored at that moment, and return, without waiting forever.

consensus/module_consensus_state_sync.go
gokutheengineer: yes. Added a todo to consider not throw away future blocks

consensus/state_sync/module.go
gokutheengineer: removed

consensus/e2e_tests/utils_test.go
```diff
@@ -663,8 +690,8 @@ func broadcastMessages(t *testing.T, msgs []*anypb.Any, pocketNodes IdToNodeMapp
 // WaitForNodeToSync waits for a node to sync to a target height
 // For every missing block for the unsynced node:
 //
-//	first, waits for the node to request a missing block via `waitForNodeToRequestMissingBlock()` function,
-//	then, waits for the node to receive the missing block via `waitForNodeToReceiveMissingBlock()` function,
+//	first, waits for the unsynced node to request a missing block via `waitForNodeToRequestMissingBlock()` function,
+//	then, waits for other nodes to send the requested block via `waitForNodesToReplyToBlockRequest()` function,
 //	finally, wait for the node to catch up to the target height via `waitForNodeToCatchUp()` function.
 func WaitForNodeToSync(
```
Olshansk, line 696: Double checking that this needs to be publically exposed.

consensus/e2e_tests/utils_test.go
```diff
@@ -673,68 +700,78 @@ func WaitForNodeToSync(
 	unsyncedNode *shared.Node,
 	allNodes IdToNodeMapping,
 	targetHeight uint64,
-) error {
+) {
 	currentHeight := unsyncedNode.GetBus().GetConsensusModule().CurrentHeight()
```
Olshansk, line 704: Optional improvement: You can call `t.Helper()` to improve logging.

https://pkg.go.dev/testing#B.Helper

Ditto elsewhere but not a must.

consensus/hotstuff_handler.go
```diff
@@ -22,6 +23,18 @@ func (m *consensusModule) handleHotstuffMessage(msg *typesCons.HotstuffMessage)
 	// Pacemaker - Liveness & safety checks
 	if shouldHandle, err := m.paceMaker.ShouldHandleMessage(msg); !shouldHandle {
 		m.logger.Debug().Fields(loggingFields).Msg("Not handling hotstuff msg...")
+		// if the message height is higher than node's current height, node needs to start active state sync
+		if msg.Height > m.height {
+			// if the message is a decide message (which is the final consensus cycle, and it means proposed block is persisted), set active sync height to message height
```
Olshansk, line 28:
```suggestion
			// if the message is a DECIDE message (the final consensus cycle leading to block finality), set active sync height to message height
```

consensus/hotstuff_handler.go
```diff
@@ -22,6 +23,18 @@ func (m *consensusModule) handleHotstuffMessage(msg *typesCons.HotstuffMessage)
...
+			// else, set active sync height to message height - 1.
+			if msg.Step == Decide {
+				m.stateSync.SetActiveSyncHeight(msg.Height)
+			} else {
+				m.stateSync.SetActiveSyncHeight(msg.Height - 1)
+			}
+			err := m.GetBus().GetStateMachineModule().SendEvent(coreTypes.StateMachineEvent_Consensus_IsUnsynced)
+			return err
```
Olshansk, line 36:
```suggestion
			return m.GetBus().GetStateMachineModule().SendEvent(coreTypes.StateMachineEvent_Consensus_IsUnsynced)
```

consensus/hotstuff_handler.go
```diff
@@ -22,6 +23,18 @@ func (m *consensusModule) handleHotstuffMessage(msg *typesCons.HotstuffMessage)
 	// Pacemaker - Liveness & safety checks
 	if shouldHandle, err := m.paceMaker.ShouldHandleMessage(msg); !shouldHandle {
 		m.logger.Debug().Fields(loggingFields).Msg("Not handling hotstuff msg...")
+		// if the message height is higher than node's current height, node needs to start active state sync
+		if msg.Height > m.height {
+			// if the message is a decide message (which is the final consensus cycle, and it means proposed block is persisted), set active sync height to message height
+			// else, set active sync height to message height - 1.
+			if msg.Step == Decide {
```
Olshansk, line 30: Can you explain why we have this difference?

consensus/module.go
```diff
@@ -77,10 +74,13 @@ type consensusModule struct {
 	hotstuffMempool map[typesCons.HotstuffStep]*hotstuffFIFOMempool
 
 	// block responses received from peers are collected in this channel
-	blocksReceived chan *typesCons.GetBlockResponse
+	blocksResponsesReceived chan *typesCons.GetBlockResponse
 
 	// metadata responses received from peers are collected in this channel
-	metadataReceived chan *typesCons.StateSyncMetadataResponse
+	// metadataReceived chan *typesCons.StateSyncMetadataResponse
```
Olshansk, line 80: Is this intentionally commented out? SHould it be removed?

consensus/module.go
```diff
@@ -77,10 +74,13 @@ type consensusModule struct {
...
+
+	// wait group to wait for blockApplicationLoop() to finish
+	wg sync.WaitGroup
```
Olshansk, line 83: Consider renaming to `blockApplicableWG`

persistence/module.go
```diff
@@ -231,6 +231,27 @@ func (m *persistenceModule) NewWriteContext() modules.PersistenceRWContext {
...
+		return false, err
+	}
+	defer readCtx.Release()
+
+	validators, err := readCtx.GetAllValidators(int64(height))
+	if err != nil {
+		return false, err
+	}
+
+	for _, actor := range validators {
```
Olshansk, line 246: If you use `GetActor` you don't have to use the forloop

consensus/state_sync_handler.go
```diff
@@ -44,7 +47,7 @@ func (m *consensusModule) handleStateSyncMessage(stateSyncMessage *typesCons.Sta
 		return m.stateSync.HandleStateSyncMetadataRequest(stateSyncMessage.GetMetadataReq())
 	case *typesCons.StateSyncMessage_MetadataRes:
 		m.logger.Info().Str("proto_type", "MetadataResponse").Msg("Handling StateSyncMessage MetadataRes")
-		m.metadataReceived <- stateSyncMessage.GetMetadataRes()
+		m.stateSync.HandleStateSyncMetadataResponse(stateSyncMessage.GetMetadataRes())
```
Olshansk, line 50: We have two different paradigms here:
1. Sending a message to channel
2. Sending message to function

Can you pick one and be consistent in both?

Screenshot 2023-05-01 at 8 16 42 PM

consensus/module.go
```diff
@@ -194,13 +193,24 @@ func (m *consensusModule) Start() error {
...
+	go func() {
+		defer m.wg.Done()
+		m.blockApplicationLoop()
+	}()
 
 	return nil
 }
 
 func (m *consensusModule) Stop() error {
+	close(m.blocksResponsesReceived)
```
Olshansk, line 206: 1. Is this the only channel we need to close?
2. Is there a way to log if there are still other messages inside it when we close it?

consensus/module.go
```diff
@@ -194,13 +193,24 @@ func (m *consensusModule) Start() error {
 		return err
 	}
 
-	go m.metadataSyncLoop()
-	go m.blockApplicationLoop()
+	m.wg.Add(1)
+	go func() {
+		defer m.wg.Done()
+		m.blockApplicationLoop()
+	}()
```
Olshansk, line 200: I don't fully understand why we have a wait group here at all and don't just do 

`go m.blockApplicationLoop()`

and inside of the loop, we'd do something like

```
	for {
		select {
		case blockResponse := <-blocksResponsesReceived:
```

consensus/e2e_tests/utils_test.go
```diff
@@ -673,68 +700,78 @@ func WaitForNodeToSync(
...
+		// waiting for unsynced node to request missing block
+		blockRequests, err := waitForNodeToRequestMissingBlock(t, clck, eventsChannel)
+		require.NoError(t, err)
 
-		blockResponse, err := waitForNodeToReceiveMissingBlock(t, clck, eventsChannel, allNodes, blockRequest)
-		if err != nil {
-			return err
-		}
+		// broadcast one of the requests to all nodes
+		P2PBroadcast(t, allNodes, blockRequests[0])
```
Olshansk, line 712: Why are we only broadcasting `[0]`? DItto below

- Should we be looping over all of them and sending each one?
- Should we assert there's only only response if we're only using the 0th one?

consensus/e2e_tests/utils_test.go
```diff
@@ -673,68 +700,78 @@ func WaitForNodeToSync(
...
+
+	return msgs, err
 }
 
-// TODO(#352): implement this function.
-// waitForNodeToReceiveMissingBlock requests block request of the unsynced node
-// for given node to node to catch up to the target height by sending the requested block.
-func waitForNodeToReceiveMissingBlock(
+// waitForNodesToReplyToBlockRequest waits for nodes to send back requested block
+func waitForNodesToReplyToBlockRequest(
```
Olshansk, line 746: The business logic seems to be the same between this function and the one above it. What am I missing?

consensus/fsm_handler.go
```diff
@@ -84,7 +82,20 @@ func (m *consensusModule) HandleUnsynced(msg *messaging.StateMachineTransitionEv
...
+	isValidator, err := m.GetBus().GetPersistenceModule().IsValidator(int64(m.height), m.GetNodeAddress())
+	if err != nil {
+		m.logger.Err(err).Msg("failed to check if the node is a validator")
+		return err
+	}
+
+	// if the node is validator, start active state sync with StartActiveSync
+	// if the node is not validator, it will always stay in the sync mode, and it will eventually catch up to the final state with passive state sync
+	if isValidator {
+		m.logger.Debug().Msg("Validator node is starting active state sync")
```
Olshansk, line 94: I think this shold be an Info, not a debug

consensus/module_consensus_state_sync.go
```diff
@@ -22,20 +23,69 @@ func (m *consensusModule) GetNodeAddress() string {
...
 
-// TODO(#352): Implement this function, currently a placeholder.
-// commitReceivedBlocks commits the blocks received from the blocksReceived channel
-// it is intended to be run as a background process
+// blockApplicationLoop commits the blocks received from the blocksResponsesReceived channel in state sync processes
 func (m *consensusModule) blockApplicationLoop() {
-	// runs as a background process in consensus module
-	// listens on the blocksReceived channel
-	// commits the received block
+	for blockResponse := range m.blocksResponsesReceived {
```
Olshansk, line 28: Per my comment above, I think this should be similar to what we have in `waitForEventsInternal`, not a forloop

consensus/module_consensus_state_sync.go
```diff
@@ -22,20 +23,69 @@ func (m *consensusModule) GetNodeAddress() string {
 	return m.nodeAddress
 }
 
-// TODO(#352): Implement this function, currently a placeholder.
-// commitReceivedBlocks commits the blocks received from the blocksReceived channel
-// it is intended to be run as a background process
+// blockApplicationLoop commits the blocks received from the blocksResponsesReceived channel in state sync processes
 func (m *consensusModule) blockApplicationLoop() {
```
Olshansk, line 27: Can you add a subcontexted logger here: https://github.com/pokt-network/pocket/wiki/Logger-Readme#submodule--subcontext-logging

consensus/module_consensus_state_sync.go
```diff
@@ -22,20 +23,69 @@ func (m *consensusModule) GetNodeAddress() string {
...
-	// runs as a background process in consensus module
-	// listens on the blocksReceived channel
-	// commits the received block
+	for blockResponse := range m.blocksResponsesReceived {
+		block := blockResponse.Block
+		maxPersistedHeight, err := m.maxPersistedBlockHeight()
+		if err != nil {
+			m.logger.Err(err).Msg("couldn't query max persisted height")
+			return
```
Olshansk, line 33: If we really want `blockApplicationLoop` to be a backroudn process, shouldn't these be `continue` statements?

consensus/state_sync/module.go
```diff
@@ -1,18 +1,39 @@
...
 	StateSyncServerModule
+
+	HandleStateSyncBlockCommittedEvent(message *anypb.Any) error
+	HandleStateSyncMetadataResponse(metadataRes *typesCons.StateSyncMetadataResponse)
+
+	// SetActiveSyncHeight sets the active sync height
+	SetActiveSyncHeight(height uint64)
+
+	// StartActiveSync starts active state sync from current height to activeSyncHeight
+	StartActiveSync()
```
Olshansk, line 36: Wouldn't it be simpler if we consolidate it with `StartActiveSync(height uint64)`?

Olshansk (changes requested):
I haven't gotten to the "real core" of the PR yet, but just submitting what I have reviewed up until now

Olshansk:
The WIP state sync has undergone quite a bit of changes in #812 so we are closing out this PR as it is stale.

PR #691, opened 2023-04-19
Update changelog-verify.yml

Olshansk:
# Description

Only add a changelog validation comment if the `cl validate` label is added.

## Context

The following output is useful but can be noisy, especially for draft PRs.

This change allows the author / reviewer to request a check for changelog validation explicitly.

For example:

Screenshot 2023-04-18 at 7 53 31 PM


bryanchriswhite:
@Olshansk amazing - thank you for hearing my cries for help. :joy: 

My only concern is how this interacts with our ability to merge. For example, does this make it possible to merge without changelog validation having run at all? I think it should still remain useful as "check" but I'm not sure how much additional complexity that begs.

Olshansk:
@bryanchriswhite I moved it from `validate` to `review` so it always validates but optionally reviews.

bryanchriswhite (approved):
Thank you sir! :partying_face::raised_hands:

Issue #505, opened 2023-02-13
[P2P] Background gossip using libp2p-pubsub


jessicadaugherty:
## Objective

Add "random gossip" broadcast layer (i.e. in addition to raintree) in P2P module using [`go-libp2p-pubsub`](https://github.com/libp2p/go-libp2p-pubsub) to support full-node participation and provide basic raintree redundancy. 

## Origin Document

Opportunity identified while integrating Pocket with LibP2P in https://github.com/pokt-network/pocket/issues/347

P2P currently only supports network broadcasting via raintree which in insufficient for full-node participation. The raintree algorithm requires trust in (and/or an effective disincentive against) the entire set of recipients in order to guarantee complete message propagation. For this reason an alternative broadcast mechanism must be used to support full-node participation in P2P (and, by extension, higher-level things like consensus). 

## Goals

- Determine a reliable, scalable, and customizable broadcasting protocol that has minimal network overhead
- Install and configure a broadcast protocol in the libp2p integration for full node random gossip

## Deliverable

- [ ] Research the tradeoffs between gossipsub and randomsub (ADR)
- [ ] Integrate the background router with the P2P module
  - [ ] handle incoming messages from background router sources
  - [ ] include background router in P2P module broadcast implementation
- [ ] Wrap background router messages in new `BackgroundMessage` protobuf type

### Integration / Architecture

Legends:
```mermaid
flowchart
subgraph Legend
    m[[`Method`]]
    c[Component]

    m -- "unconditional usage" --> c
    m -. "conditional usage" .-> c
end
```

```mermaid
classDiagram
class ConcreteType {
  +ExportedField
  -unexportedField
  +ExportedMethod(argType) returnType
  -unexportedMethod()
}

class InterfaceType {
    <<interface>>
    +Method(argType) (returnType1, returnType2)
}

ConcreteType --|> InterfaceType : Interface realization

ConcreteType --> OtherType : Direct usage
ConcreteType --o OtherType : Composition
ConcreteType --* OtherType : Aggregatation
ConcreteType ..*  "(cardinality)" OtherType : Indirect (via interface)
```

Outgoing messages:
```mermaid
flowchart
    subgraph p2p[P2P Module]
        ps[[`Send`]]
        pb[[`Broadcast`]]

        subgraph rt[RainTree Router]
            rtu[UnicastRouter]
        end

        subgraph bg[Background Router]
            bgu[UnicastRouter]
            bgg[Gossipsub Topic]
        end
    end
 
[...]

Issue #588, opened 2023-03-15
[Core, Tooling] upgrade to go 1.20.x


bryanchriswhite:
## Objective

Upgrade the version of  golang that we use in the repo to the  latest stable version of the 1.2 major version branch.

## Origin Document

![image](https://user-images.githubusercontent.com/600733/225300219-987d8f47-3794-4537-af30-119d3e6d84b4.png)

## Goals

- Minimize future upgrade friction
- Enable new native features (e.g. stdlib multi-error support)

## Deliverable

- [x]  Update go version in the repo & debug any issues which arise.
- [x] Update go version in infrastructure code (e.g. [Dockerfile(s)](https://github.com/pokt-network/pocket/blob/main/build/Dockerfile.localdev#L1), k8s manifest(s)/template(s), etc.)
- [ ] Replace usage of `go.uber.org/multierr` to use `err.join` (https://stackoverflow.com/a/74645028/768439)

## Non-goals / Non-deliverables

- Refactoring outside the scope of debugging any issues encountered.

## General issue deliverables

- [ ] Update the appropriate CHANGELOG(s)
- [ ] Update any relevant local/global README(s)

## Testing Methodology

- [ ] Updated version should be reflected in the [go.mod](https://github.com/pokt-network/pocket/blob/main/go.mod#L3)
- [ ] **All tests**: `make test_all`
- [ ] **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)

---

**Creator**: @bryanchriswhite 
**Co-Owners**: @Olshansk

Olshansk:
@bryanchriswhite I added this to iteration 13 and added removing uber's multierr as a dependency.

Olshansk:
Completed by @okdas and @dylanlott in the prior iteration.

PR #707, opened 2023-04-28
[P2P] feat: add background router

bryanchriswhite:
## Description

Implements a "background router" to work alongside raintree (to be integrated in a future PR) which uses  a kademlia DHT for peer discovery and a gossipsub pubsub router for broadcasting and listening.

## Issue

Part 1 of #505 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Removed unused `Transport` interface
- Moved and renamed `raintree.RainTreeConfig` to `util.RouterConfig`
- Renamed `protocol.DefaultTopicStr` to `protocol.BackgroundTopicStr`
- Added `protocol.PeerDiscoveryNamespace`
- Added kademlia peer discovery baseline test
- Added background router (kad + gossipsub)

## Testing

## Required Checklist

### If Applicable Checklist


build/docs/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.0.0.39] - 2023-04-28
 
+- Updated Dockerfiles using outdated go version to 1.19
+
+## [0.0.0.39] - 2023-04-28
+
 - Added a `fisherman_per_session` governance parameter
 - Updated the default `blocks_per_session` from `4` to `1`
```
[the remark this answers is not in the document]
bryanchriswhite: Just plain wrong :thinking:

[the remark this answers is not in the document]
bryanchriswhite: I  think this is what happens if you forget to update the version number :sweat_smile:

p2p/background/router_test.go
```diff
@@ -0,0 +1,355 @@
...
+	// simulate network links between each to every other
+	// (i.e. fully-connected network)
+	err := libp2pMockNet.LinkAll()
+	require.NoError(t, err)
+
+	bootstrap(t, ctx, testHosts)
+
+	go func() {
+		// wait for hosts to listen and peer discovery
+		time.Sleep(time.Second * 2)
```
dylanlott, line 167: If possible, I would prefer waiting on a signal channel instead to prevent a load bearing sleep.

p2p/background/router_test.go
```diff
@@ -0,0 +1,355 @@
...
+	// waitgroup done or timeout
+	select {
+	case <-testTimeout:
+		t.Fatalf(
+			"timed out waiting for message: got %d; wanted %d",
+			len(seenMessages),
+			numPeers,
+		)
+	case <-done:
+	}
```
dylanlott, line 190: 🙌

build/Dockerfile.client
```diff
@@ -1,4 +1,4 @@
-ARG GOLANG_IMAGE_VERSION=golang:1.18.3-alpine3.16
+ARG GOLANG_IMAGE_VERSION=golang:1.19-alpine3.16
```
dylanlott, line 1: nit: why not latest of [1.20.3](https://hub.docker.com/layers/library/golang/1.20.3/images/sha256-9413aaefd182f9499f7e17b7400709c34eb1edac9ca71c37cdc73a63691c5f95?context=explore)?

bryanchriswhite: My intention was only to synchronize the current version of go that we're using, which is 1.19.x. The reason was an error produced by go-libp2p-kad-dht or go-libp2p-pubsub IIRC, complaining about needing go 1.19 or later.

I think we anticipate that the 1.20 migration may require some effort, see #588.

p2p/background/router_test.go
bryanchriswhite: I did my best; with the addition of notify/notifee I was able to reduce it to ~100ms on my machine, so I set it to 250ms to be on the safe side.

build/Dockerfile.client
Olshansk: I'm usually weary of upgrading language versions & business logic in the same PR (harder to revert, things can break, etc...) but assuming you're confident its safe in this scenario?

bryanchriswhite: `Dockerfile.client` (this file) is the image that provides the debug CLI in docker compose localnet. The other dockerfile that I modified ([`Dockerfile.localdev`](https://github.com/pokt-network/pocket/blob/main/build/Dockerfile.localdev)) is also a compose localnet only image (which includes `dlv` debugger :call_me_hand:). See the [docker-compose.yaml](https://github.com/pokt-network/pocket/blob/main/build/deployments/docker-compose.yaml). I compared these to what we're using in production ([`Dockerfile.debian.prod`](https://github.com/pokt-network/pocket/blob/main/build/Dockerfile.debian.prod)) which is already using go 1.19. See [`build-images` CI job](https://github.com/pokt-network/pocket/blob/main/.github/workflows/main.yml#L149) for how `Dockerfile.debian.prod` is used by CI to build production the image.

p2p/background/router_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+	// construct mock host
+	return newMockNetHostFromPeer(t, mockNet, privKey, peer)
+}
+
+func readSubscription(
+	t *testing.T,
+	ctx context.Context,
+	wg *sync.WaitGroup,
```
Olshansk, line 350: Can you reflect the same name of the wg as what you passed in. It'll help with the flow of the code to the reader.

p2p/background/kad_discovery_baseline_test.go
```diff
@@ -0,0 +1,128 @@
...
+	require.Equalf(t, addr4.String(), host2DiscoveredHost4Addrs[0].String(), "host2 did not discover host4")
+
+	host3DiscoveredHost4Addrs := host3.Peerstore().Addrs(host4.ID())
+	require.Lenf(t, host3DiscoveredHost4Addrs, 1, "host3 did not discover host4")
+	require.Equalf(t, addr4.String(), host3DiscoveredHost4Addrs[0].String(), "host3 did not discover host4")
+
+	require.ElementsMatchf(t, expectedPeerIDs, host4.Peerstore().Peers(), "host4 peer IDs don't match")
+}
+
+//nolint:unparam // DHT must exist but is otherwise "unused" (i.e. its API)
```
Olshansk, line 78: Would it be clearer if we just refacto to `setupHost` and call `setupDHT(t, ctx, host, bootstrapAddr)` in place in the test above?

We can group the host creation and make it explicit when we need the bootstrap addr or not.

Lmk if you can "visualize" the alternate (potentially clearer) implementation

p2p/background/kad_discovery_baseline_test.go
```diff
@@ -0,0 +1,128 @@
...
+	bootstrapAddr, err := multiaddr.NewMultiaddr(fmt.Sprintf("%s/p2p/%s", addr1, host1.ID().String()))
+	require.NoError(t, err)
+
+	addr2, host2, _ := setupHostAndDiscovery(t, ctx, defaults.DefaultP2PPort+1, bootstrapAddr)
+	addr3, host3, _ := setupHostAndDiscovery(t, ctx, defaults.DefaultP2PPort+2, bootstrapAddr)
+
+	expectedPeerIDs := []libp2pPeer.ID{host1.ID(), host2.ID(), host3.ID()}
+
+	// TECHDEBT: consider using `host.ConnManager().Notifee()` to avoid sleeping here
+	// delay assertions for 500ms
```
Olshansk, line 35: Update comment saying you're waiting for the DHT to fill up

p2p/background/kad_discovery_baseline_test.go
```diff
@@ -0,0 +1,128 @@
...
+	time.Sleep(time.Millisecond * 500)
+
+	// assert that host2 has host3 in its peerstore
+	host2DiscoveredAddrs := host2.Peerstore().Addrs(host3.ID())
+	require.Lenf(t, host2DiscoveredAddrs, 1, "did not discover host3")
```
Olshansk, line 40: This might be a huge NIT, but what if it's (somehow) it's greater than 1? Maybe we should update the error to be "expected exactly 1 address for host3 but found %d"

p2p/background/kad_discovery_baseline_test.go
```diff
@@ -0,0 +1,128 @@
...
+	require.Equalf(t, addr3.String(), host2DiscoveredAddrs[0].String(), "did not discover host3")
+	require.ElementsMatchf(t, expectedPeerIDs, host2.Peerstore().Peers(), "host2 peer IDs don't match")
+
+	// assert that host3 has host2 in its peerstore
+	host3DiscoveredHost2Addrs := host3.Peerstore().Addrs(host2.ID())
+	require.Lenf(t, host3DiscoveredHost2Addrs, 1, "host3 did not discover host2")
+	require.Equalf(t, addr2.String(), host3DiscoveredHost2Addrs[0].String(), "host3 did not discover host2")
+	require.ElementsMatchf(t, expectedPeerIDs, host3.Peerstore().Peers(), "host3 peer IDs don't match")
+
+	// add another peer to network...
```
Olshansk, line 50: Missed opportunity to say "A wild peer appears in the network"

tumblr_e19b0725ffd8972d59be1a9950b78229_8f06cfe2_640

p2p/background/kad_discovery_baseline_test.go
```diff
@@ -0,0 +1,128 @@
...
+	addr4, host4, _ := setupHostAndDiscovery(t, ctx, defaults.DefaultP2PPort+3, bootstrapAddr)
+	expectedPeerIDs = append(expectedPeerIDs, host4.ID())
+
+	// TECHDEBT: consider using `host.ConnManager().Notifee()` to avoid sleeping here
```
Olshansk, line 54: Ditto as above as to why we need a sleep.

Maybe move it to a constant ith a comment.

```go
// ...
const dhtUpdateSleepTime := time.Millisecond * 500
```

Olshansk (changes requested):
Nicely done on all fronts: tests, documentation, clarity, etc!

p2p/README.md
```diff
@@ -15,21 +16,39 @@ This document is meant to be a supplement to the living specification of [1.0 Po
...
+
+A specific ["pubsub"](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub) router implementation / protocol which uses information (referred to internally as 'gossip', **distinct from our definition**) about which peers have seen which messages to facilitate "gossip" (as per our definition).
+
+### "raintree gossip"
+
+A "gossip" protocol (and implementation) which uses the raintree algorithm for peer selection. Used between all staked actors to optimize for "gossip" speed.
+
+### "background gossip"
+
+A "gossip" protocol (implementation TBD) which facilitates "gossip" to all P2P participants, including non-staked actors (e.g. full-nodes).
```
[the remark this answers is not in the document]
bryanchriswhite: Yes. My thinking on the TBD bit is that we haven't done a comprehensive comparison between gossipsub and randomsub routers yet. I figured that might shake out in an ADR.

p2p/background/router.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// NOTE_TO_SELF: `pubsub.NewRandomSub` requires a `size` arg.
+	gossipSub, err := pubsub.NewGossipSub(ctx, cfg.Host)
+	if err != nil {
+		return nil, fmt.Errorf("creating gossip pubsub: %w", err)
+	}
+
+	dhtMode := dht.ModeAutoServer
+	// NB: don't act as a bootstrap node in peer discovery in client debug mode
```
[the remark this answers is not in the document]
bryanchriswhite: I don't think so. To me "discoverable" would imply that this is related to the nodes ability to advertise (being discovered). Rather this option affects the node's ability to discover other peers **while being connected to**.

p2p/background/router.go
```diff
@@ -0,0 +1,168 @@
...
+
+// GetPeerstore implements the respective `typesP2P.Router` interface  method.
+func (rtr *backgroundRouter) GetPeerstore() typesP2P.Peerstore {
+	return rtr.pstore
+}
+
+// AddPeer implements the respective `typesP2P.Router` interface  method.
+func (rtr *backgroundRouter) AddPeer(peer typesP2P.Peer) error {
+	// Noop if peer with the pokt address already exists in the peerstore.
+	// TECHDEBT: add method(s) to update peers.
```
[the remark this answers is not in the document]
bryanchriswhite: Hopefully that day won't come. I'm anticipating the next `Router` interface refactor will remove the need for these.

p2p/background/router_test.go
```diff
@@ -0,0 +1,373 @@
...
+	existingPeerMultiaddr, err := utils.Libp2pMultiaddrFromServiceURL(existingPeer.GetServiceURL())
+	require.NoError(t, err)
+	require.Equal(t, existingPeerstoreAddrs[0].String(), existingPeerMultiaddr.String())
+
+	err = testRouter.RemovePeer(existingPeer)
+	require.NoError(t, err)
+
+	require.Len(t, testRouter.pstore, 0)
+
+	// NB: libp2p peerstore implementations only remove peer keys and metadata
```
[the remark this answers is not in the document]
bryanchriswhite: It was a bit confusing at first but fortunately it was caught early and we've had the benefit of toting this comment around to remind us. :raised_hands:

p2p/background/router_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+	actualPeerIDs = testutil.GetKeys[string](seenMessages)
+	require.ElementsMatchf(t, expectedPeerIDs, actualPeerIDs, "peerIDs don't match")
+}
+
+// bootstrap connects each host to one other except for the arbitrarily chosen "bootstrap host"
+func bootstrap(t *testing.T, ctx context.Context, testHosts []libp2pHost.Host) {
+	t.Helper()
+
+	t.Log("bootstrapping...")
```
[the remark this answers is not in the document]
bryanchriswhite: Line 236 is logging the ID of the node which we're connecting to. Is that what you had in mind?

p2p/background/kad_discovery_baseline_test.go
bryanchriswhite: Good catch! :+1:

p2p/utils/config.go
```diff
@@ -0,0 +1,45 @@
...
+// RouterConfig is used to configure `Router` implementations using the given
+// libp2p host and current height and peerstore providers.
+// TECHDEBT: I would prefer for this to be in p2p/types/router.go but this causes
+// an import cycle between `typesP2P` and `providers`.
+type RouterConfig struct {
+	Addr                  crypto.Address
+	CurrentHeightProvider providers.CurrentHeightProvider
+	Host                  host.Host
+	Hostname              string
+	MaxMempoolCount       uint64
```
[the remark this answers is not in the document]
bryanchriswhite: > 1. Do we need Hostname? I feel like it's something that should be retrievable from Host

This is only used for logging and can be recovered from the P2P config via the bus. The libp2p `Host` will only have a reference to the resolved IP address that the hostname points to.

> 2. Feels weird that MaxMempoolCount is part of the router. I realize this is probably a legacy thing but feels

I think it was just named poorly. It's used by Raintree to determine the capacity of the nonce deduper queue. I've refactored the configs to define `RouterConfig` as an interface - lmk what you think.

p2p/background/kad_discovery_baseline_test.go
bryanchriswhite: :rofl:

p2p/background/kad_discovery_baseline_test.go
bryanchriswhite: I see what you're getting at. May I suggest that we leave it as is for now and see what API requirements emerge as we consolidate testutils and get into peer discovery testing? My thinking is that we will definitely need access to the DHT object at that point and I suspect there will be sufficient opportunity to refactor and reuse some of these testutils by then as well.

p2p/config/config.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/pokt-network/pocket/p2p/providers"
+	"github.com/pokt-network/pocket/shared/crypto"
+	"go.uber.org/multierr"
+)
+
+// baseConfig implements `RouterConfig` using the given libp2p host and current
+// height and peerstore providers. Intended for internal use by other `RouterConfig`
+// implementations with common config parameters.
+//
+// NB: intentionally *not* embedding `baseConfig` to improve readability of usages
```
Olshansk, line 15: I don't fully understand this. So when would you embed a struct inside of another one?

To me this seems like a worthwhile usecase

Olshansk (changes requested):
Just a couple questions but otherwise almost g2g

p2p/config/config.go
bryanchriswhite: I'm suggesting never. It doesn't look/feel good from the usage side:

```golang
	bgCfg := BackgroundConfig{
		BaseConfig: BaseConfig{
			Host:                  nil,
			Addr:                  nil,
			CurrentHeightProvider: nil,
			PeerstoreProvider:     nil,
		},
	}

	rtCfg := RainTreeConfig{
		BaseConfig: BaseConfig{
			Host:                  nil,
			Addr:                  nil,
			CurrentHeightProvider: nil,
			PeerstoreProvider:     nil,
		},
		MaxNonces: 100,
	}
```

Seeing this :point_up_2: makes me want to write a `NewRainTreeConfig()` function which would be a bit ridiculous IMHO given that that the purpose of the config originally was to simplify the `NewRainTreeRouter()` function signature.

Additionally, I expect significant simplification may be around the corner for the router interface. I'm anticipating being able to eliminate the need for the background router to hold a reference to either provider (maybe even the pokt addr as  well). This would cause further divergence between `BackgroundConfig` and `RainTreeConfig`, likely making `baseConfig` another temporary step.

runtime/defaults/defaults.go
```diff
@@ -47,11 +47,11 @@ var (
 	DefaultPersistencePostgresURL    = "postgres://postgres:postgres@pocket-db:5432/postgres"
 	DefaultPersistenceBlockStorePath = "/var/blockstore"
 	// p2p
-	DefaultUseLibp2p          = false
-	DefaultP2PPort            = uint32(42069)
-	DefaultP2PUseRainTree     = true
-	DefaultP2PConnectionType  = types.ConnectionType_TCPConnection
-	DefaultP2PMaxMempoolCount = uint64(1e5)
```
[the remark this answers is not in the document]
bryanchriswhite: I'm less clear on the original intention but  based on the current state of the codebase, as far as I could tell, the only thing that config field has to do with mempool is the fact that it uses a generic implementation of a FIFO queue provided by the `mempool` package. It uses this queue to track, at maximum, the specified number of seen `RainTreeMessage#Nonce`s. Am I missing something?

p2p/config/config.go
bryanchriswhite: I feel strongly about not taking embedding seriously as an option because of the impact on readability, given the context. An alternative to defining `baseConfig`  would be to duplicate the `IsValid()` implementation across both `RainTreeConfig` and `BackgroundConfig`; however, I see the way it's currently written as a simplification of that.

runtime/defaults/defaults.go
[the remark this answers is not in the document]
Olshansk: The number of Nonce was a direct mapping of the P2P mempool, so semantically that name was more descriptive of what it was capturing.

Either way, I don't feel strongly about this and will approve.

We can always change back if we choose to.

p2p/config/config.go
Olshansk: giphy

Olshansk (approved):

Issue #613, opened 2023-03-24
[Core] Deploying all-the-protocol-actors


Olshansk:
## Objective

Enable deploying different types of protocol actors.

## Origin Document

[The utility specification](https://github.com/pokt-network/pocket-network-protocol/tree/main/utility) outlines several different types of actors:
1. `Validator`
2. `Servicer`
3. `Fisherman`
4. `Portal`
5. `Full Nodes` (a full node that is not a protocol actor)

As of writing, the V1 repo only supports running validators and the debug client is a "makeshift" full node. As we are about to start working on M3 and have a live devnet, there needs to be an easy way to build and scale different types of actors.

_Note that `Application` and `Actor` do not fall into the scope of this ticket._

### V0 - Swagger Documentation (as a reference)

_Note: It is slightly outdated_

The V0 RPC spec can be found [here](https://editor.swagger.io/?url=https://raw.githubusercontent.com/pokt-network/pocket-core/staging/doc/specs/rpc-spec.yaml).

For example, the height can be queried like so:

```bash
curl --location -X 'POST' -H 'Content-Type: application/json' 'https://<user>:<pass>@<hostname>:<port>/v1/query/height' | jq

{
  "height": 89554
}
```

### V1 - Swagger Documentation (as a reference)

_Note: It is still in the early stages_

The V1 RPC spec can be found [here](https://editor.swagger.io/?url=https://raw.githubusercontent.com/pokt-network/pocket/main/rpc/v1/openapi.yaml).

<img width="1307" alt="Screenshot 2023-04-11 at 4 54 13 PM" src="https://user-images.githubusercontent.com/1892194/231313431-81d49217-beb3-4657-bc97-e948747c879c.png">

## Goals

- Enable deploying different types of protocol actors from a single binary
- Enable scaling different types of protocol actors
- Enable iterating on the business logic for M3 (locally & and in devnet)
- Set a foundation for full nodes, which will extend to light clients

## Deliverable

Note that there are a lot of notes and deliverables below. These should be used as a guide but the implementor is expected to use their best judgment and creativity to achieve the goals above.

- [ ] Update `config.go` to contain the type of actor being deployed
		- **Note to implementor**: Creating empty `XXXConfig` (e.g. `FishermanConfig`) protos that we can build on in the future would help
- [ ] Update/add `config.json` files for each actor(s) being deployed on this node
- [ ] Add empty placeholder logic (e.g. a print statement in the `Utility Module`) that is specific to each protocol ator
[...]

PR #710, opened 2023-04-30
[Core] Deploying all-the-protocol-actors

0xBigBoss:
## Description

Adds necessary updates to begin deploying different kind of actor nodes for Pocket Network. It is still incomplete according to #613 .

## Issue

Fixes https://github.com/pokt-network/pocket/issues/613

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [x] Major breaking change
- [x] Documentation
- [x] Development Tools

## List of changes

- Updates localnet to support deploying different actors
- Adds new config protos for enabling various actors
- Creates and starts new actor specific utility module
- Adds new RPC routes
  - /v1/query/nodeRoles
- Adds new CLI command `bin/p1 query noderoles` for querying a nodes roles
- Stub `HandleRelay`
- Converts "" to ":memory:" to be explicit about using in memory stores
  
## Testing

## Required Checklist

### If Applicable Checklist


rpc/handlers_servicer.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"net/http"
+
+	"github.com/labstack/echo/v4"
+)
+
+func (s *rpcServer) PostV1ServicerRelay(ctx echo.Context) error {
+
+	// TODO: Move to the servicer module
```
Olshansk, line 11: Any reason not to do it in this PR?

rpc/handlers_servicer.go
```diff
@@ -0,0 +1,26 @@
+package rpc
+
+import (
+	"net/http"
+
+	"github.com/labstack/echo/v4"
+)
+
+func (s *rpcServer) PostV1ServicerRelay(ctx echo.Context) error {
+
```
Olshansk, line 10:
```suggestion
```

rpc/handlers_servicer.go
```diff
@@ -0,0 +1,26 @@
...
+	// TODO: Move to the servicer module
+	servicerModuleName := "servicer"
+	found := false
+	for _, am := range s.GetBus().GetUtilityModule().GetActorModules() {
```
Olshansk, line 14: NIT: I think you can simply this using `slices.IndexFunc` https://stackoverflow.com/a/38654444/768439

rpc/v1/openapi.yaml
```diff
@@ -91,6 +91,122 @@ paths:
           content:
             text/plain:
               example: "description of failure"
+  /v1/servicer/relay:
```
Olshansk, line 94: Make sure to rebase on top of @h5law's recent changes since I anticipate merge conflicts.

utility/portal/module.go
```diff
@@ -0,0 +1,75 @@
+package portal
```
Olshansk, line 1: Let's remove the portal package for now because we added it to the spec after we started working on V1. 

I anticipate there will be a separate piece of work to add it in (the enum, this package, etc...) in the future, so I want to reduce the confusion.

utility/module.go
```diff
@@ -1,35 +1,54 @@
...
 import (
+	"errors"
+
 	"github.com/pokt-network/pocket/logger"
 	"github.com/pokt-network/pocket/runtime/configs"
 	"github.com/pokt-network/pocket/shared/mempool"
 	"github.com/pokt-network/pocket/shared/modules"
 	"github.com/pokt-network/pocket/shared/modules/base_modules"
+	"github.com/pokt-network/pocket/utility/fisherman"
+	fisherman_module "github.com/pokt-network/pocket/utility/fisherman"
```
Olshansk, line 12: I think we should go with just `fisherman` instead of `fisherman_module`; ditto elsewhere

https://go.dev/blog/package-names

Screenshot 2023-05-03 at 7 09 53 PM

utility/module.go
```diff
@@ -44,12 +63,94 @@ func (*utilityModule) Create(bus modules.Bus, options ...modules.ModuleOption) (
...
+			m.logger.Error().Err(err).Msg("failed to create portal module")
+			return err
+		}
+		m.portal = &portal
+		m.actorModules = append(m.actorModules, portal)
+	}
 
-	return m, nil
+	actors := m.GetActorModuleNames()
+	if len(m.actorModules) > 1 {
```
Olshansk, line 119: Can you move this into a helper function call `validateActorModuleExclusivity` with a `TODO` comment to figure out all the rules we want to enable?

utility/module.go
```diff
@@ -44,12 +63,94 @@ func (*utilityModule) Create(bus modules.Bus, options ...modules.ModuleOption) (
...
+		// only case where this is allowed is if the node is a validator and a servicer
+		if !(validatorCfg.Enabled && servicerCfg.Enabled) {
+			m.logger.Error().Strs("actors", actors).Msg(ErrInvalidActorsEnabled)
```
Olshansk, line 122: We also have a `StringLogArrayMarshaler` helper that might be useful here

runtime/configs/proto/fisherman_config.proto
```diff
@@ -0,0 +1,14 @@
+
+syntax = "proto3";
+
+package configs;
+
+option go_package = "github.com/pokt-network/pocket/runtime/configs";
+
+// FishermanConfig defines the configuration for the node acting as a fisherman. Fishermen grade and enforce the quality of the Web3 access provided by Servicers
+// TODO(0xbigboss): Add fields
```
Olshansk, line 9: If you can't think of it right now, let's remove the TODOs/INVESTIGATE and add things as we implement e2e features. Ditto for the other configs.

I'm working on those in a notion doc for now

Screenshot 2023-05-03 at 7 36 52 PM

utility/validator/module.go
```diff
@@ -0,0 +1,75 @@
...
+
+type validator struct {
+	base_modules.IntegratableModule
+	logger *modules.Logger
+}
+
+// type assertions for validator module
+var (
+	_ ValidatorModule = &validator{}
+	// TODO: add validator module functions here
```
Olshansk, line 33: As long as you do `_ ValidatorUtility = &validator{}` it'll enforce all the functions

utility/validator/module.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/pokt-network/pocket/logger"
+	"github.com/pokt-network/pocket/shared/modules"
+	"github.com/pokt-network/pocket/shared/modules/base_modules"
+)
+
+const (
+	ValidatorModuleName = "validator"
+)
+
+type ValidatorModule interface {
```
Olshansk, line 13: I see the approach you took here. Still thinking about, but I do think it's a good start!

runtime/configs/config.go
```diff
@@ -28,6 +29,10 @@ type Config struct {
 	Logger      *LoggerConfig      `json:"logger"`
 	RPC         *RPCConfig         `json:"rpc"`
 	Keybase     *KeybaseConfig     `json:"keybase"` // Determines and configures which keybase to use, `file` or `vault`. IMPROVE(#626): See for rationale around proto design. We have proposed a better config design, but did not implement it due to viper limitations
+	Validator   *ValidatorConfig   `json:"validator"`
+	Servicer    *ServicerConfig    `json:"servicer"`
+	Fisherman   *FishermanConfig   `json:"fisherman"`
+	Portal      *PortalConfig      `json:"portal"`
```
Olshansk, line 35: I think we should either remove `Portal` from this PR completely or make sure it's account for everything.

For example:
- `runtime/genesis/proto/genesis.proto`
- `ActorType` in `actor.proto`

Personally, I think one PR to add Portal everywhere later is the way to go

utility/fisherman/module.go
```diff
@@ -0,0 +1,75 @@
...
+
+type FishermanModule interface {
+	modules.Module
+
+	// TODO: add fisherman module functions here
+	FishermanUtility
+}
+
+// TODO_IN_THIS_COMMIT: exists to help with type assertions, drop once you've added your own functions
+type FishermanUtility interface {
```
Olshansk, line 21: Let's move the `XXInterface` of each actor in `shared/utility_module.go` so its easy to understand and add functionality everywhere.

Olshansk (changes requested):
Looks pretty good to me for the most part. Good foundation to add things to in the future.

Note that you'll have to rebase on top of @h5law's changes (already in) and @okdas has #727 in progress. @gokutheengineer is also going to submit a version of state sync for review soon.

Everything is in 🛫

rpc/handlers_servicer.go
0xBigBoss: For sure, I meant to leave it in as a TODO_IN_THIS_COMMIT. I just ran out time before adding the new `GetXXXModule` on the Bus interface to make it easy to do that.

h5law: @Olshansk @0xBigBoss Not sure the purpose of this function but is it not a duplicate of `PostV1ClientRelay` from the RPC spec ima merge ina bit?

Makefile
```diff
@@ -319,10 +319,11 @@ generate_rpc_openapi: go_oapi-codegen ## (Re)generates the RPC server and client
 	oapi-codegen  --config ./rpc/client.gen.config.yml ./rpc/v1/openapi.yaml > ./rpc/client.gen.go
 	echo "OpenAPI client and server generated"
 
+SWAGGER_PORT=127.0.0.1:8080
 .PHONY: swagger-ui
 swagger-ui: ## Starts a local Swagger UI instance for the RPC API
 	echo "Attempting to start Swagger UI at http://localhost:8080\n\n"
-	docker run -p 8080:8080 -e SWAGGER_JSON=/v1/openapi.yaml -v $(shell pwd)/rpc/v1:/v1 swaggerapi/swagger-ui
+	docker run --name pocket-swagger-ui --rm -p $(SWAGGER_PORT):8080 -e SWAGGER_JSON=/v1/openapi.yaml -v $(shell pwd)/rpc/v1:/v1 swaggerapi/swagger-ui
```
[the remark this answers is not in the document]
0xBigBoss: ahhh very cool, yes that is actually helpful lol better than googling which command to use on mac vs ubuntu vs alpine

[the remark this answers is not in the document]
0xBigBoss: > i should learn rust
☕ 🐱

okdas (commented):
Please excuse my lack of understanding. While reviewing the config file, I noticed that it seems to allow one process to serve as different actors. Is this the intended behavior? I was under the impression that we were not going to support this functionality.

For example, if one process has fisherman.enabled=true and servicer.enabled=true, would that be considered valid? Or is the configuration file looks that way to follow the module approach, where each module has its own property in the config file?

build/localnet/README.md
```diff
@@ -159,6 +159,30 @@ If you need to add a new parameter – feel free to modify the section in place.
 
 Please refer to helm charts documentation for more details.
 
+### Overriding default values for localnet with Tilt
```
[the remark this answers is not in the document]
0xBigBoss: I do not believe they will vary greatly. It is more for experimentation or adding specific configurations depending on a developers choice of environment. It's a small escape hatch for when you need to deviate a little bit from the defaults.

I only needed to enable CORS for nodes so I could use swagger. It could be used to change any part of the helm chart.

Perhaps you had something else in mind for customizing one's localnet env?

build/localnet/Tiltfile
```diff
@@ -159,25 +169,75 @@ k8s_resource('dev-cli-client', labels=['client'])
...
+for x in range(localnet_config["servicers"]["count"]):
+    actor_number = actor_number + 1
+    formatted_number = formatted_actor_number(actor_number)
+
+    k8s_yaml(helm(chart_dir,
+        name="servicer-%s-pocket" % formatted_number,
         set=[
             "global.postgresql.auth.postgresPassword=LocalNetPassword",
             "image.repository=validator-image",
             "privateKeySecretKeyRef.name=validators-private-keys",
```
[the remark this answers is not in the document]
0xBigBoss: They do currently. The actor_number variable increments for each validator, servicer, and fisherman(watcher hehe) which ensures that they do not grab the same key. But they do share all of the keys in the existing `private-keys.yaml` file.

build/localnet/manifests/cli-client.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: dev-cli-client
-  namespace: default
```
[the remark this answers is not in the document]
0xBigBoss: Isn't the default namespace inherent if you don't specify one with `kubectl` or in your kubeconfig? 

Yes, it's exactly that, I deploy localnet to another namespace for easy cleanup.

utility/module.go
0xBigBoss: I think they accomplish the same thing.

0xBigBoss:
> Please excuse my lack of understanding. While reviewing the config file, I noticed that it seems to allow one process to serve as different actors. Is this the intended behavior? I was under the impression that we were not going to support this functionality.
> 
> For example, if one process has fisherman.enabled=true and servicer.enabled=true, would that be considered valid? Or is the configuration file looks that way to follow the module approach, where each module has its own property in the config file?

AFAIK only a validator and servicer can co-exist in the same node. All other roles are exclusive to one another (a validator cannot be a fisherman, a servicer cannot be a fisherman). This will all be enforced in the utility module.

Olshansk:
@0xBigBoss Can you rebase on top of @okdas's PR (#727) once it is merged in?

Olshansk (commented):
@0xBigBoss I'm thinking of reviewing and merging the config work (even though it's not prod ready yet) since overlapping work is happening in other PRs (e.g. #789). Wdyt?

0xBigBoss:
> I'm thinking of reviewing and merging the config work (even though it's not prod ready yet) since overlapping work is happening in other PRs (e.g. #789). Wdyt?

Sounds good @Olshansk. I did my best to merge and rebase the current PR off of main. Also, needed to patch the tests after using the temp config method I added.

utility/module.go
```diff
@@ -44,12 +60,75 @@ func (*utilityModule) Create(bus modules.Bus, options ...modules.ModuleOption) (
 
 	m.config = utilityCfg
 	m.mempool = types.NewTxFIFOMempool(utilityCfg.MaxMempoolTransactionBytes, utilityCfg.MaxMempoolTransactions)
+	m.logger = logger.Global.CreateLoggerForModule(m.GetModuleName())
+
+	return m, enableActorModules(cfg, m, bus)
+}
+
+// enableActorModules enables the actor-specific modules and adds them to the utility module's actorModules to be started later.
+func enableActorModules(cfg *configs.Config, m *utilityModule, bus modules.Bus) error {
```
Olshansk, line 69: If you make this a receiver function of the module, there would be no need to bus in any parameters

```go
func (u *utilityModule) enableActorModules() error {
```

utility/module.go
```diff
@@ -60,3 +139,39 @@ func (u *utilityModule) GetModuleName() string {
...
+	return u.actorModules
+}
+
+func (u *utilityModule) GetServicerModule() modules.Module {
+	return u.svc
+}
+
+// validateActorModuleExclusivity validates that the actor modules are enabled in a valid combination.
+// TODO: There are probably more rules that need to be added here.
+func validateActorModuleExclusivity(m *utilityModule, cfg *configs.Config) error {
```
Olshansk, line 157: DItto above.

1. Do not pass in `utilityModule`, make it a receiver
2. Config is already part of the `utilityModule` so access that directly

utility/module_enable_actors_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+			bus, err := runtime.CreateBus(mockRunMgr)
+			assert.NoError(t, err)
+
+			// Call enableActorModules with the test config
+			m, err := Create(bus)
+
+			// Verify error output
+			if test.expectedError == "" {
+				assert.NoError(t, err)
```
Olshansk, line 81: Use `require` instead of `assert`.

Screenshot 2023-05-31 at 6 43 40 PM

https://github.com/stretchr/testify#require-package

Olshansk (changes requested):
Apologize for the long turnaround on this @0xBigBoss, but PTAL at the comments I left. Everything should be pretty straight forward and the whole team is looking forward to getting this merged in!

app/client/cli/query.go
```diff
@@ -884,6 +884,33 @@ func queryCommands() []*cobra.Command {
...
+					return err
+				}
+				response, err := client.PostV1QueryNodeRoles(cmd.Context())
+				if err != nil {
+					return unableToConnectToRpc(err)
+				}
+				statusCode := response.StatusCode
+				body, err := io.ReadAll(response.Body)
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "❌ Error reading response body: %s\n", err.Error())
```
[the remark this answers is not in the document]
0xBigBoss: I believe I was just following the example in this file. Plus it seems fitting to ouput to stderr on an error.

persistence/blockstore/block_store.go
```diff
@@ -35,7 +35,7 @@ type blockStore struct {
 // NewBlockStore initializes a new blockstore with the given path.
 // * If "" is provided as the path, an in-memory store is used.
 func NewBlockStore(path string) (BlockStore, error) {
-	if path == "" {
+	if path == ":memory:" {
```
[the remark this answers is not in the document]
0xBigBoss: u noticed 😏🤠

0xBigBoss:
@Olshansk @adshmh I tried to take care of the lil changes, will address and rebase the utlity module changes tom. 🫡🤠

app/client/cli/query.go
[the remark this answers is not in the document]
h5law: @Olshansk yeah this was a change I made when removing the `log` prints to users as stderr for errs makes sense

runtime/manager_test.go
```diff
@@ -4229,7 +4240,10 @@ func TestNewManagerFromReaders(t *testing.T) {
 						Timeout: 30000,
 						UseCors: false,
 					},
-					Keybase: defaultCfg.Keybase,
+					Keybase:   defaultCfg.Keybase,
+					Servicer:  &configs.ServicerConfig{Enabled: true},
+					Validator: &configs.ValidatorConfig{Enabled: true},
```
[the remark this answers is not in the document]
0xBigBoss: Not sure about this one, as the default config is not enabled where as this test is checking a node that is staked as a validator and a servicier.

utility/module.go
0xBigBoss: We need the root config for this bit, and didn't want to add that as part of the module.

build/deployments/docker-compose.yaml
```diff
@@ -143,6 +135,78 @@ services:
...
+    image: pocket/servicer:latest
+    command: >
+      sh -c '
+      if [ "$NODE5_SERVICER_ENABLED" = "true" ]; then
+        build/scripts/watch.sh \
+          build/config/config.node5.servicer.json \
+          build/config/genesis.json;
+      else
+        echo "😴 Going to sleep. Servicer is not enabled.";
+        sleep infinity;
```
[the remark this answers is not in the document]
0xBigBoss: ah you are right, exit 0 is okay to docker compose here, i confused it with k8s scheduling thinking it would just crash loop.

utility/session.go
```diff
@@ -21,15 +21,11 @@ const (
 
 // TODO: Implement this
 func (u *utilityModule) HandleRelay(relay *coreTypes.Relay) (*coreTypes.RelayResponse, error) {
-
-	if u.svc == nil {
+	if u.servicer == nil {
+		// if u.GetServicerModule() == nil {
```
Olshansk, line 25: remove?

utility/module.go
```diff
@@ -31,11 +31,9 @@ type utilityModule struct {
 
 	mempool mempool.TXMempool
 
-	actorModules []modules.Module
-
-	vld validator.ValidatorModule
-	svc servicer.ServicerModule
-	fsh fisherman.FishermanModule
+	actorModules map[string]modules.Module
+	// TODO: initialize
```
Olshansk, line 35: Still TODO?

Olshansk:
@0xBigBoss Left a couple small NITS, but otherwise the changes LGTM as long as the tests pass. I've added the `e2e-devnet-test` label to spin up a devnet for this.

utility/module.go
```diff
@@ -143,16 +141,32 @@ func (u *utilityModule) GetActorModules() map[string]modules.Module {
 	return u.actorModules
 }
 
-func (u *utilityModule) GetServicerModule() modules.ServicerModule {
-	return u.actorModules[servicer.ServicerModuleName]
+func (u *utilityModule) GetServicerModule() (modules.ServicerModule, error) {
+	m, ok := u.actorModules[servicer.ServicerModuleName].(modules.ServicerModule)
+	if !ok {
```
Olshansk, line 146: OPTIONAL NIT:

I think doing the following is more readable and shorter, but not a blocker

```go
if m, ok := u.actorModules[servicer.ServicerModuleName].(modules.ServicerModule); ok {
	return m, nil
}
u.logger.Warn(...)
...
```

utility/module.go
```diff
@@ -143,16 +141,32 @@ func (u *utilityModule) GetActorModules() map[string]modules.Module {
...
+		u.logger.Warn().Err(errors.New("failed to cast servicer module")).Msg("failed to get servicer module")
+		return nil, errors.New("failed to cast servicer module")
+	}
+	return m, nil
 }
 
-func (u *utilityModule) GetFishermanModule() modules.FishermanModule {
-	return u.actorModules[fisherman.FishermanModuleName]
+func (u *utilityModule) GetFishermanModule() (modules.FishermanModule, error) {
+	m, ok := u.actorModules[fisherman.FishermanModuleName].(modules.FishermanModule)
```
Olshansk, line 154: Can we ever be in a situation where the map does not contain the key so we get a panic because we try to typeast a nill?

utility/servicer/module.go
```diff
@@ -28,30 +45,325 @@ func CreateServicer(bus modules.Bus, options ...modules.ModuleOption) (modules.S
...
+	}
+
+	return fmt.Errorf("relay block height %d not within session ID %s starting block %d and last block %d",
+		relayMeta.BlockHeight,
+		session.Id,
+		sessionStartingBlock,
+		sessionLastBlock)
+}
+
+// TECHDEBT: These structures were copied as placeholders from v0 and need to be updated to reflect changes in v1
```
Olshansk: Can you move these into `service/relay.go` and make the TECHDEBT more explicit. We are likely going to change all of these but I want to keep deleting them outside the scope of this PR. 

 cc @adshmh for context

utility/session.go
```diff
@@ -21,11 +22,13 @@ const (
 
 // TODO: Implement this
 func (u *utilityModule) HandleRelay(relay *coreTypes.Relay) (*coreTypes.RelayResponse, error) {
-	if u.servicer == nil {
-		// if u.GetServicerModule() == nil {
-		return nil, fmt.Errorf(NodeIsNotServicerErr)
+	sm, err := u.GetServicerModule()
+
```
Olshansk, line 26:
```suggestion
```

Olshansk (changes requested):
A few more minor NITS but otherwise LGTM. Will approve after latest changes.

1. @adshmh Please prioritize taking a look at this as it conflicts with your work but I want us to merge it in sooner rather than later.

2. @0xBigBoss Please make sure unit tests and E2E tests are passing (green CI)

utility/module.go
0xBigBoss: Currently not, but you're right, it is sort of shoot urself in the foot material. Added more checks.

0xBigBoss:
@Olshansk should be GTG, i had to update some of the stuff for e2e tests due to not using the default namespace.

Olshansk (approved):
Let's merge it in once CI is all 🟢 !

0xBigBoss:
Thanks @Olshansk, gonna merge now since the checks failing are for off by one day for the changelogs and the codecov plugin failing on a lot of the code that got merged in from main.

Issue #438, opened 2023-01-16
[P2P] P2P Module Knowledge Transfer


Olshansk:
## Objective

This is an umbrella ticket to capture references and discussion around knowledge transfer of the P2P module. The primary focus is to integrate Pocket V1 with LibP2P and answer outstanding questions in RainTree.

## Origin Document

The results of the [V1 P2P Research](https://docs.google.com/document/d/10rHFr8HpUfpmHzXbyZWzq60aJDY1lYMR9kHNgdfgbNE/edit#heading=h.e1rj36wlfr45) ultimately resulted in the development of a structured gossip algorithm called [RainTree, presented here](https://github.com/pokt-network/pocket/files/9853354/raintree.pdf), and also summarized in the [V1 P2P specification](https://github.com/pokt-network/pocket-network-protocol/blob/main/p2p/README.md).

### RainTree

A [Python simulator](https://github.com/pokt-network/rain-tree-sim/blob/main/python/README.md) was developed based on the RainTree specifications outlined [here](https://github.com/pokt-network/pocket/files/9853354/raintree.pdf). Pocket is a Network with many nodes but not transmit large blobs of data (i.e. akin to smart contract blockchains), so whenever a new block is being committed or proposed, our goal is to distribute it throughout the network as quickly as possible. RainTree was designed to achieve this via a structured gossip algorithm with `O(log3(n))` complexity. Though it was not formally proven, here were some preliminary stats collected:

![Screenshot 2023-01-16 at 2 50 13 PM](https://user-images.githubusercontent.com/1892194/212756098-ab17cb74-e561-43e5-b05a-494533621383.png)

[This comment](https://github.com/pokt-network/pocket/pull/405#discussion_r1051198643) summarizes well the different types of broadcasts necessary in Pocket.

![Screenshot 2023-01-16 at 3 08 51 PM](https://user-images.githubusercontent.com/1892194/212758382-2f3c1da4-4be3-40cf-84ef-169a7c40756c.png)

The current implementation of RainTree and the P2P module lives [here](https://github.com/pokt-network/pocket/tree/main/p2p).

### Redundancy & Cleanup

[...]

PR #732, opened 2023-05-05
[P2P] Integrate background router

### LINKED ISSUE #505: [P2P] Background gossip using libp2p-pubsub

bryanchriswhite:
## @Reviewer
This PR may be more digestible / reviewable on a commit-by-commit basis. Commits are organized logically and any given line is only modified in a single commit, with few exceptions*.

*(In the interest of preserving the git-time-continuum :police_officer::rotating_light:, this applies in batches of commits between comments or reviews *by humans*, only once "in review")

---

### Before

While `backgroundRouter` exists and implements the `Router` interface, it had yet to be connected to anything. Additionally, it was not able to handle incoming unicast messages.

```mermaid
classDiagram
    class p2pModule {
        -router Router
        -handlePocketEnvelope([]byte) error
    }

    class P2PModule {
        <<interface>>
        GetAddress() (Address, error)
        HandleEvent(*anypb.Any) error
        Send([]byte, Address) error
        Broadcast([]byte) error
    }
    p2pModule --|> P2PModule

    class RainTreeRouter {
        UnicastRouter
        -handler MessageHandler
        +Broadcast([]byte) error
        -handleRainTreeMsg([]byte) error
    }

    class BackgroundRouter {
        UnicastRouter
        -handler MessageHandler
        +Broadcast([]byte) error
        -handleBackgroundMsg([]byte) error
        -readSubscription(subscription *pubsub.Subscription)
    }

    class UnicastRouter {
        -messageHandler MessageHandler
        -peerHandler PeerHandler
        +Send([]byte, Address) error
        -handleStream(libp2pNetwork.Stream)
        -readStream(libp2pNetwork.Stream)
    }
    RainTreeRouter --* UnicastRouter : (embedded)

    p2pModule --o "1" Router
    p2pModule ..* RainTreeRouter : (`router`)
    
    class Router {
        <<interface>>
        +Send([]byte, Address) error
        +Broadcast([]byte) error
    }
    BackgroundRouter --|> Router
    RainTreeRouter --|> Router
```

### After

`backgroundRouter` embeds `UnicastRouter` and is integrated into `p2pModule` such that it is considered when calling `P2PModule#Send()` and `P2PModule#Broadcast()`.

```mermaid
classDiagram
    class p2pModule {
        -stakedActorRouter Router
        -unstakedActorRouter Router
        -handlePocketEnvelope([]byte) error
    }

    class P2PModule {
        <<interface>>
        GetAddress() (Address, error)
        HandleEvent(*anypb.Any) error
        Send([]byte, Address) error
        Broadcast([]byte) error
    }
    p2pModule --|> P2PModule

    class RainTreeRouter {
        UnicastRouter
        -handler MessageHandler
        +Broadcast([]byte) error
        -handleRainTreeMsg([]byte) error
    }

    class BackgroundRouter {
        UnicastRouter
        -handler MessageHandler
        +Broadcast([]byte) error
        -handleBackgroundMsg([]byte) error
        -readSubscription(subscription *pubsub.Subscription)
    }

    class UnicastRouter {
        -messageHandler MessageHandler
        -peerHandler PeerHandler
        +Send([]byte, Address) error
        -handleStream(libp2pNetwork.Stream)
        -readStream(libp2pNetwork.Stream)
    }
    RainTreeRouter --* UnicastRouter : (embedded)
    BackgroundRouter --* UnicastRouter : (embedded)

    p2pModule --o "2" Router
    p2pModule ..* RainTreeRouter : (`stakedActorRouter`)
    p2pModule ..* BackgroundRouter : (`unstakedActorRouter`)
    
    class Router {
        <<interface>>
        +Send([]byte, Address) error
        +Broadcast([]byte) error
    }
    BackgroundRouter --|> Router
    RainTreeRouter --|> Router
```

## Issue

- #505 

## Dependencies

- #844 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

* added background message protobuf type
* added `Router#Close()` interface method
* separated raintree & bg protocol IDs
* move `PocketEnvelope` nonce generation to `PackMessage()`
* added `Handler` to router configs
* refactored raintree router to support P2P module & `Router` interface changes
* refactored raintree router bootstrapping
* refactored background router to support P2P module integration & `Router` interface changes
* integrated background router into P2P module
* refactored staked actor router (raintree) peer discovery
* updated tests, post-refactoring

## Testing

## Required Checklist

### If Applicable Checklist


bryanchriswhite:
I don't seem to be able to replicate the failing E2E test locally. I also tried merging in main, which had no effect other than to confuse GitHub's diff (as this branch is based off of `refactor/unicast-router`).

p2p/module.go
```diff
@@ -62,6 +67,7 @@ func Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, e
 	return new(p2pModule).Create(bus, options...)
 }
 
+// TODO_THIS_COMMIT: rename (WithHost) & consider moving to testutil file
```
bryanchriswhite: (Done subsequently in 5c59cd9edb065e)

Olshansk:
Sharing an observation (not a callout a request, but just what I experience as a reader).

In #844, the "before and after" views of the interfaces & structure is very helpful because of how small and simple it is. In this PR, it's a bit less useful and introduces a lot of cognitive load because seeing changes in such a large diagram is hard. I realize it is much more useful to the implementor for tackling the problem, but as the reader, I would benefit from one of two things:

1. A `git diff` like color coding with red & green
2. Having them side by side, which I did manually myself :)

Again, this is just me sharing.

p2p/event_handler.go
```diff
@@ -23,21 +24,25 @@ func (m *p2pModule) HandleEvent(event *anypb.Any) error {
...
-		updatedPeerstore, err := m.pstoreProvider.GetStakedPeerstoreAtHeight(consensusNewHeightEvent.Height)
-		if err != nil {
+		if isStaked, err := m.isStakedActor(); err != nil {
 			return err
-		}
-
-		added, removed := oldPeerList.Delta(updatedPeerstore.GetPeerList())
-		for _, add := range added {
-			if err := m.router.AddPeer(add); err != nil {
+		} else if isStaked {
```
Olshansk, line 29: OPTIONAL PERSONAL STYLE NIT:  Reduce indent for readability

```go
isStaked, err := m.isStakedActor()
if err != nil {
   return err
}
if !isStaked {
   return nil
}
// business logic
```

Olshansk (changes requested):
Not a complete review (e.g. didn't cover the second half of `p2p/background/router.go`) but wanted to post what I have so far

p2p/background/router.go
```diff
@@ -9,31 +9,50 @@ import (
...
 	logger *modules.Logger
+	// handler is the function to call when a message is received.
+	handler typesP2P.MessageHandler
 	// host represents a libp2p network node, it encapsulates a libp2p peerstore
 	// & connection manager. `libp2p.New` configures and starts listening
 	// according to options.
 	// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p#section-readme)
 	host libp2pHost.Host
+	// cancelCtx is the cancel function for the context which is provided to the
+	// gossipsub subscription. Used to terminate the `readSubscription()` go routine.
```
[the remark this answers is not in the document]
bryanchriswhite: Correct.

p2p/testutil.go
```diff
@@ -0,0 +1,40 @@
...
+
+package p2p
+
+import (
+	libp2pHost "github.com/libp2p/go-libp2p/core/host"
+	typesP2P "github.com/pokt-network/pocket/p2p/types"
+	"github.com/pokt-network/pocket/shared/modules"
+)
+
+// WithHost associates an existing (i.e. "started") libp2p `host.Host`
```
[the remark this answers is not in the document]
bryanchriswhite: :100: Thanks for keeping the bar high! :raised_hands::fire:

p2p/utils_test.go
```diff
@@ -313,11 +341,32 @@ func prepareEventMetricsAgentMock(t *testing.T, valId string, wg *sync.WaitGroup
 	ctrl := gomock.NewController(t)
 	eventMetricsAgentMock := mockModules.NewMockEventMetricsAgent(ctrl)
 
-	eventMetricsAgentMock.EXPECT().EmitEvent(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
-	eventMetricsAgentMock.EXPECT().EmitEvent(gomock.Any(), gomock.Any(), gomock.Eq(telemetry.P2P_RAINTREE_MESSAGE_EVENT_METRIC_SEND_LABEL), gomock.Any()).Do(func(n, e any, l ...any) {
+	// DISCUSS_THIS_COMMIT: The number of times each telemetry event is expected
```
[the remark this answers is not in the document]
bryanchriswhite: TL;DR 
1. This test needs to be overhauled
2. It should be much easier to do after #810 TECHDEBT in P2P is resolved

IIRC, previously we were measuring redundant and non-redundant messages in the same place, and quite significantly, **before deduplication**. While refactoring this test lost the ability to hook into raintree router in the same way, instead relying entirely on the telemetry module.

> `redundant messages` -> isn't this configurable

For clarity, I'm **not** talking about redundancy in the #85 sense, as as such wouldn't be configurable AFAIK. My observation was that at specific network sizes, some nodes receive redundant messages at level 1. My assumption was that this was the result of raintree doing its thing on an unbalanced tree but I didn't rigorously test that.

p2p/background/router.go
```diff
@@ -166,6 +192,213 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
 	return rtr.pstore.RemovePeer(peer.GetAddress())
 }
 
+func (rtr *backgroundRouter) setupUnicastRouter() error {
```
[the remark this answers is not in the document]
bryanchriswhite: "unicast" here is referencing the name of the field, not characterizing the background router itself. Adding a comment similar to `rainTreeRouter#setupUnicastRouter()`. :+1:

p2p/background/router.go
```diff
@@ -47,96 +66,103 @@ type backgroundRouter struct {
...
+
 	// TECHDEBT(#595): add ctx to interface methods and propagate down.
-	return rtr.topic.Publish(context.TODO(), data)
+	return rtr.topic.Publish(context.TODO(), backgroundMsgBz)
 }
 
 // Send implements the respective `typesP2P.Router` interface  method.
-func (rtr *backgroundRouter) Send(data []byte, address cryptoPocket.Address) error {
+func (rtr *backgroundRouter) Send(pocketEnvelopeBz []byte, address cryptoPocket.Address) error {
+	rtr.logger.Warn().Str("address", address.String()).Msg("sending background message to peer")
```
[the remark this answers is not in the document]
bryanchriswhite: TL;DR I think state sync depends on unstaked actors supporting unicast send. I would demote this to a debug log.

---

I was actually considering converting this to a debug log (similar to [`rainTreeRouter#sendInternal()`](https://github.com/pokt-network/pocket/blob/main/p2p/raintree/router.go#L147)).

One case where we "need" to support unstaked actor to unstaked actor unicast is for the debug CLI. Granted, it's a weak case to make the argument; however, should lead to reduced complexity in the debug CLI (and possibly raintree's peer manager) as it no longer has to consider a scenario where it's not part of the staked actor set.

The stronger case to make the argument would be that, if my understanding is correct, state sync depends on unicast for responses. I'm under the impression that this is the case given [the current state](https://github.com/pokt-network/pocket/pull/812/files) of #812. Please correct me if I'm mistaken.

Again, correct me if I'm mistaken, but I think it may be the case that background router

p2p/utils_test.go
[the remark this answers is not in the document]
bryanchriswhite: (converting this to a `TECHDEBT` comment)

p2p/utils_test.go
```diff
@@ -199,6 +219,14 @@ func createMockBus(t *testing.T, runtimeMgr modules.RuntimeMgr) *mockModules.Moc
 	mockModulesRegistry.EXPECT().GetModule(peerstore_provider.ModuleName).Return(nil, runtime.ErrModuleNotRegistered(peerstore_provider.ModuleName)).AnyTimes()
 	mockModulesRegistry.EXPECT().GetModule(current_height_provider.ModuleName).Return(nil, runtime.ErrModuleNotRegistered(current_height_provider.ModuleName)).AnyTimes()
 	mockBus.EXPECT().GetModulesRegistry().Return(mockModulesRegistry).AnyTimes()
+	mockBus.EXPECT().PublishEventToBus(gomock.AssignableToTypeOf(&messaging.PocketEnvelope{})).
```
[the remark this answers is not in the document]
bryanchriswhite: Sorry, which TODO were you referring to?

p2p/background/router.go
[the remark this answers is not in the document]
bryanchriswhite: (relates to the tables in the [P2P Module / Router Decoupling](https://github.com/pokt-network/pocket/blob/bf96542fa6620d338b42184f7bad497f277eabe9/p2p/README.md#p2p-module--router-decoupling) section of the updated P2P README)

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+
+		if err := rtr.host.Connect(ctx, libp2pAddrInfo); err != nil {
+			return fmt.Errorf("connecting to peer: %w", err)
+		}
+	}
+	return nil
+}
+
+// topicValidator is used in conjunction with libp2p-pubsub's notion of "topic
+// validaton". It is usefed for arbitrary and concurrent pre-propagation validation
```
dylanlott, line 345:
```suggestion
// validaton". It is used for arbitrary and concurrent pre-propagation validation
```

shared/messaging/envelope.go
```diff
@@ -12,7 +14,10 @@ func PackMessage(message proto.Message) (*PocketEnvelope, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &PocketEnvelope{Content: anyMsg}, nil
+	return &PocketEnvelope{
+		Content: anyMsg,
+		Nonce:   cryptoPocket.GetNonce(),
```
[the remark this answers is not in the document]
dylanlott: 😰

p2p/background/router.go
```diff
@@ -9,31 +9,50 @@ import (
...
+	// handler is the function to call when a message is received.
+	handler typesP2P.MessageHandler
 	// host represents a libp2p network node, it encapsulates a libp2p peerstore
 	// & connection manager. `libp2p.New` configures and starts listening
 	// according to options.
 	// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p#section-readme)
 	host libp2pHost.Host
+	// cancelReadSubscription is the cancel function for the context which is
+	// monitored in the `#readSubscription()` go routine. Call to terminate it.
+	cancelReadSubscription context.CancelFunc
```
Olshansk, line 52:
```suggestion
	// only one read subscription exists per router at any point in time
	cancelReadSubscription context.CancelFunc
```

p2p/background/router.go
```diff
@@ -47,96 +66,106 @@ type backgroundRouter struct {
...
+	rtr.cancelReadSubscription()
+	rtr.subscription.Cancel()
+
+	var topicCloseErr error
+	if err := rtr.topic.Close(); err != context.Canceled {
+		topicCloseErr = err
 	}
 
-	return rtr, nil
+	return multierr.Append(
```
Olshansk, line 123: Great usage of multierr everywhere!

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+		return fmt.Errorf("setting up topic: %w", err)
+	}
+
+	if err := rtr.setupSubscription(); err != nil {
+		return fmt.Errorf("setting up subscription: %w", err)
+	}
+
+	if err := rtr.setupPeerstore(
+		ctx,
+		cfg.PeerstoreProvider, cfg.CurrentHeightProvider); err != nil {
```
Olshansk, line 240: If we're multi-lining args, we gotta do it right

```suggestion
		cfg.PeerstoreProvider,
		 cfg.CurrentHeightProvider,
	 ); err != nil {
```

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+	}
+
+	if err := rtr.bootstrap(ctx); err != nil {
+		return fmt.Errorf("bootstrapping peerstore: %w", err)
+	}
+
+	return nil
+}
+
+func (rtr *backgroundRouter) setupPeerDiscovery(ctx context.Context) (err error) {
```
Olshansk, line 266:
```suggestion
// setupPeerDiscovery sets up the Kademlia Distributed Hash Table (DHT)
func (rtr *backgroundRouter) setupPeerDiscovery(ctx context.Context) (err error) {
```

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+	// NB: don't act as a bootstrap node in peer discovery in client debug mode
+	if isClientDebugMode(rtr.GetBus()) {
+		dhtMode = dht.ModeClient
+	}
+
+	rtr.kadDHT, err = dht.New(ctx, rtr.host, dht.Mode(dhtMode))
+	return err
+}
+
+func (rtr *backgroundRouter) setupPubsub(ctx context.Context) (err error) {
```
Olshansk, line 277:
```suggestion
// setupPubsub sets up a new gossip sub topic using libp2p
func (rtr *backgroundRouter) setupPubsub(ctx context.Context) (err error) {
```

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+	unicastRouter, err := unicast.Create(rtr.GetBus(), &unicastRouterCfg)
+	if err != nil {
+		return fmt.Errorf("setting up unicast router: %w", err)
+	}
+
+	rtr.UnicastRouter = *unicastRouter
+	return nil
+}
+
+func (rtr *backgroundRouter) setupDependencies(ctx context.Context, cfg *config.BackgroundConfig) error {
```
Olshansk, line 217:
```suggestion
func (rtr *backgroundRouter) setupDependencies(ctx context.Context, cfg *config.BackgroundConfig) error {
	// NB: The order in which the internal components are setup below is important
```

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+
+func (rtr *backgroundRouter) setupPubsub(ctx context.Context) (err error) {
+	// TECHDEBT(#730): integrate libp2p tracing via `pubsub.WithEventTracer()`.
+
+	// CONSIDERATION: If switching to `NewRandomSub`, there will be a max size
+	rtr.gossipSub, err = pubsub.NewGossipSub(ctx, rtr.host)
+	return err
+}
+
+func (rtr *backgroundRouter) setupTopic() (err error) {
```
Olshansk, line 285: Why are we using named `err` return values in this file?

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+	// > output buffer. The default length is 32 but it can be configured to avoid
+	// > dropping messages if the consumer is not reading fast enough.
+	// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#WithBufferSize)
+	rtr.subscription, err = rtr.topic.Subscribe()
+	return err
+}
+
+func (rtr *backgroundRouter) bootstrap(ctx context.Context) error {
+	// CONSIDERATION: add `GetPeers` method to `PeerstoreProvider` interface
+	// to avoid this loop.
```
Olshansk, line 317:
```suggestion
	// CONSIDERATION: add `GetPeers` method which returns a map
	// to the `PeerstoreProvider` interface to avoid this loop.
```

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+//
+// Also note: https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#BasicSeqnoValidator
+func (rtr *backgroundRouter) topicValidator(_ context.Context, _ libp2pPeer.ID, msg *pubsub.Message) bool {
+	var backgroundMsg typesP2P.BackgroundMessage
+	if err := proto.Unmarshal(msg.Data, &backgroundMsg); err != nil {
+		return false
+	}
+
+	if backgroundMsg.Data == nil {
+		return false
```
Olshansk, line 359: Do we want/need a warning/error in the other `false` cases?

p2p/background/router.go
```diff
@@ -166,6 +195,219 @@ func (rtr *backgroundRouter) RemovePeer(peer typesP2P.Peer) error {
...
+	networkMessage := messaging.PocketEnvelope{}
+	if err := proto.Unmarshal(backgroundMsg.Data, &networkMessage); err != nil {
+		rtr.logger.Error().Err(err).Msg("Error decoding network message")
+		return false
+	}
+
+	return true
+}
+
+func (rtr *backgroundRouter) readSubscription(ctx context.Context) {
```
Olshansk, line 371:
```suggestion
// readSubscription is a while loop for receiving and handling messages from the
// subscription. It is intended to be called as a goroutine.
func (rtr *backgroundRouter) readSubscription(ctx context.Context) {
```

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
+| Staked Actor   | Staked Actor   | Raintree only   | Consensus (state sync) messages (to validators only) |
+| Unstaked Actor | Staked Actor   | Background only | Consensus (state sync) messages (to validators only) |
+| Unstaked Actor | Unstaked Actor | Background only | Consensus (state sync) & Debug (CLI) messages        |
+
+**Broadcast**
+
+| Broadcaster    | Receiver       | Router                | Example Usage                              |
+|----------------|----------------|-----------------------|--------------------------------------------|
+| Staked Actor   | Staked Actor   | Raintree + Background | Utility tx messages                        |
+| Unstaked Actor | Staked Actor   | Background only       | Utility tx messages (gossipsub redundancy) |
```
Olshansk, line 129:
```suggestion
| Unstaked Actor | Staked Actor   | Background only       | Utility tx messages (raintree redundancy) |
```

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
+
+Both router submodule implementations embed a `UnicastRouter` which enables them to send and receive messages directly to/from a single peer.
+
+**Class Diagram**
 
 ```mermaid
-flowchart TD
-  subgraph lMod[Local P2P Module]
-    subgraph lHost[Libp2p `Host`]
+classDiagram
```
Olshansk, line 137: Makes 💯 sense!

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
-subgraph rBG[Background Router]
-subgraph rBGPS[Background Peerstore]
-rNetPS([all P2P participants])
-end
+  lhost --> rhost
+```
 
-subgraph rGossipSub[GossipSub]
-end
+### Message Deduplication
```
Olshansk, line 334: Can you add a note on how/when the deduplication is cleared? On node restart? Configuration? etcc

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
 
-hs --4a--> rRT
-hs --4b--> rBG
-rBG  --"5a (cont. propagation)"--> rHost
-linkStyle 11 stroke:#ff3
-rRT  --"5b (cont. propagation)"--> rHost
-linkStyle 12 stroke:#ff3
-end
+### Peer Discovery
+Peer discovery involves pairing peer IDs to their network addresses (multiaddr).
```
Olshansk, line 424:
```suggestion

Peer discovery involves pairing peer IDs to their network addresses (multiaddr).
```

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
+This pairing always has an associated TTL (time-to-live), near the end of which it must
+be refreshed.
+
+In the background gossip overlay network (`backgroundRouter`), peers will re-advertise themselves 7/8th through their TTL.
```
Olshansk, line 428: What do you mean by 7/8th?

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
+
+      bgu -- "add if new" --> rBGPS
+      bgr -- "add if new" --> rBGPS
+      rDHT -- "continuous import" --> rBGPS
+         
+      bgu[UnicastRouter]
+    end
+
+  end
 ```
```
Olshansk, line 488: I glossed over the remaining diagrams and makes sense to me (at a high level) but just notifying that there might be micro-typos I haven't caught.

It'll be super useful for anyone (myself included) when debugging and jumping into P2P

Olshansk (changes requested):
Reviewed everything except for `p2p/background/router_test.go` !

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
+    <<interface>>
+    +Method(argType) (returnType1, returnType2)
+}
+
+ConcreteType --|> InterfaceType : Interface realization
+
+ConcreteType --> OtherType : Direct usage
+ConcreteType --o OtherType : Composition
+ConcreteType --* OtherType : Aggregatation
+ConcreteType ..*  "(cardinality)" OtherType : Indirect (via interface)
```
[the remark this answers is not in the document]
bryanchriswhite: With respect to cardinality, this example is intended to convey a 1:n, `ConcreteType`:`OtherType` relationship where "n" could be a constant or a range.

This line is mainly intended to illustrate "indirect" relation between two concrete types. The addition of cardinality was to avoid adding another relation just for that as it's getting a bit crowded already.

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
-    <li>4a. <code>handleStream</code> propagates message via <code>Raintree Router</code></li>
-    <li>4b. <code>handleStream</code> propagates message via <code>Background Router</code></li>
-    <li>5a. Repeat step 1a from <code>Remote P2P Module</code>'s perspective targeting its next peers</li>
-    <li>5b. Repeat step 1b from <code>Remote P2P Module</code>'s perspective targeting its next peers</li>
-</ul>
+**Unicast**
+
+| Sender         | Receiver       | Router          | Example Usage                                        |
+|----------------|----------------|-----------------|------------------------------------------------------|
+| Staked Actor   | Staked Actor   | Raintree only   | Consensus (state sync) messages (to validators only) |
```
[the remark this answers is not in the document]
bryanchriswhite: @Olshansk would you mind elaborating on this suggestion?

p2p/README.md
bryanchriswhite: Thanks for calling this out. I intended to provide a reference here.

When I wrote that I was thinking of [this discovery util](https://github.com/libp2p/go-libp2p/blob/v0.28.1/p2p/discovery/util/util.go#L50) which would mean literally 7/8 = 0.875 = 87.5% of the way through their TTL. However, upon further inspection, I don't think this is actually what's being used in the context of the DHT. It looks like  peers which are "DHT providers" (currently all by default, I think) get a 24 hour TTL by default but they are expected to re-advertise every 3 hours.

```suggestion
In the background gossip overlay network (`backgroundRouter`), peers will re-advertise themselves every 3 hours through their TTL (see: [`RoutingDiscovery#Advertise()`](https://github.com/libp2p/go-libp2p/blob/87c2561238cb0340ddb182c61be8dbbc7a12a780/p2p/discovery/routing/routing.go#L34) and [`ProviderManager#AddProvider()`](https://github.com/libp2p/go-libp2p-kad-dht/blob/v0.24.2/providers/providers_manager.go#L255)).
```

p2p/README.md
bryanchriswhite: I was attempting to make a distinction here between the "raintree redundancy layer" (as per #85) and the redundancy provided by gossipsub, out of band with respect to raintree's protocol. Do you have a suggestion for how to better articulate this distinction, perhaps more definitions?

p2p/README.md
```diff
@@ -67,105 +66,431 @@ flowchart TD
...
-`Routers` is where [RainTree](https://github.com/pokt-network/pocket/files/9853354/raintree.pdf) (or the simpler basic approach) is implemented. See `raintree/router.go` for the specific implementation of RainTree, but please refer to the [specifications](https://github.com/pokt-network/pocket-network-protocol/tree/main/p2p) for more details.
+`Routers` is where [RainTree](https://github.com/pokt-network/pocket/files/9853354/raintree.pdf) is implemented.
+See [`raintree/router.go`](./raintree/router.go) for the specific implementation of RainTree, but please refer to the [specifications](https://github.com/pokt-network/pocket-network-protocol/tree/main/p2p) for more details.
 
-### Raintree Router Architecture
+## Module Architecture
 
-_DISCUSS(team): If you feel this needs a diagram, please reach out to the team for additional details._
-_TODO(olshansky, BenVan): Link to RainTree visualizations once it is complete._
+### Legends
```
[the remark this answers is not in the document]
bryanchriswhite: I don't think this is a trivial amount of effort. I suspect that different diagrams are using slightly different design languages and that the legends and diagrams will have to be refactored to a common one.

I've attempted to represent things according to UML to the extent applicable and possible, given mermaid's constraints. Regardless of the language choice, I think we should always include a legend which articulates the relevant aspects.

p2p/background/router.go
bryanchriswhite: Thanks for catching this! This change got caught in-between two branches and didn't quite make it all the way into either. :sweat_smile:

p2p/background/router.go
bryanchriswhite: To simplify line 296:

```go
if rtr.topic, err = ...
```

It saves a line. I think there are valid scenarios where we definitely do want named return args; e.g. when they're ambiguous or if we need to trace an error in a defer. While this isn't one of those scenarios, that's what makes me feel comfortable converting this to one.

(I **do not** condone "naked returns" though.)

p2p/module.go
```diff
@@ -180,38 +176,102 @@ func (m *p2pModule) Start() (err error) {
...
 func (m *p2pModule) Broadcast(msg *anypb.Any) error {
-	c := &messaging.PocketEnvelope{
+	isStaked, err := m.isStakedActor()
+	if err != nil {
+		return err
+	}
+
+	if isStaked {
+		if m.stakedActorRouter == nil {
+			return fmt.Errorf("staked actor router not started")
```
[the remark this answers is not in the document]
bryanchriswhite: I've been trying to be fairly consistent with my error message structure such that it adds context without being too wordy. I've found that using a present progressing verb to describe what's happening followed by the wrapped and/or specific error strikes a good balance of readability and context when reading the code as well as the log output itself.

How do you feel about something like this?

```suggestion
			return fmt.Errorf("broadcasting: staked actor router not started")
```

p2p/utils_test.go
```diff
@@ -313,11 +341,32 @@ func prepareEventMetricsAgentMock(t *testing.T, valId string, wg *sync.WaitGroup
 	ctrl := gomock.NewController(t)
 	eventMetricsAgentMock := mockModules.NewMockEventMetricsAgent(ctrl)
 
-	eventMetricsAgentMock.EXPECT().EmitEvent(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
-	eventMetricsAgentMock.EXPECT().EmitEvent(gomock.Any(), gomock.Any(), gomock.Eq(telemetry.P2P_RAINTREE_MESSAGE_EVENT_METRIC_SEND_LABEL), gomock.Any()).Do(func(n, e any, l ...any) {
+	// TECHDEBT: The number of times each telemetry event is expected
+	// (below) is dependent on the number of redundant messages all validators see,
+	// which is a function of the network size. Until this function is derived and
+	// implemented, we cannot predict the number of times each event is expected.
+	_ = expectedNumNetworkWrites
```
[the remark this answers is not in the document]
bryanchriswhite: It is :100:% concerning to me, make no mistake 😔. As I stated in [the previous thread](https://github.com/pokt-network/pocket/pull/732#discussion_r1246670466), this test needs to be overhauled as a ton of architectural changes have happened around it and it's original form relied heavily on its ability to interact with unexported objects and members. Additionally, it exists at the "module level" which IMHO is the least well defined in terms of conventions / best practices as it sits in-between the unit and e2e levels.

p2p/utils_test.go
```diff
@@ -199,6 +219,14 @@ func createMockBus(t *testing.T, runtimeMgr modules.RuntimeMgr) *mockModules.Moc
...
 	mockModulesRegistry.EXPECT().GetModule(current_height_provider.ModuleName).Return(nil, runtime.ErrModuleNotRegistered(current_height_provider.ModuleName)).AnyTimes()
 	mockBus.EXPECT().GetModulesRegistry().Return(mockModulesRegistry).AnyTimes()
+	mockBus.EXPECT().PublishEventToBus(gomock.AssignableToTypeOf(&messaging.PocketEnvelope{})).
+		Do(func(envelope *messaging.PocketEnvelope) {
+			fmt.Println("[valId: unknown] Read")
+			fmt.Printf("content type: %s\n", envelope.Content.GetTypeUrl())
+			if readWriteWaitGroup != nil {
+				readWriteWaitGroup.Done()
+			}
+		}).AnyTimes() // TODO: specific times
```
bryanchriswhite, line 229:
```suggestion
		}).AnyTimes() // TECHDEBT: assert number of times. Consider `waitForEventsInternal` or similar as in consensus.
```

p2p/README.md
```diff
@@ -71,7 +71,22 @@ See [`raintree/router.go`](./raintree/router.go) for the specific implementation
 
 ## Module Architecture
 
+### Architecture Design Language
```
Olshansk, line 74: Please add a TODO to move this (and the definitions you have below) into `shared` since this text is not just relevant to `p2p`

p2p/README.md
```diff
@@ -106,9 +121,48 @@ ConcreteType --* OtherType : Aggregatation
 ConcreteType ..*  "(cardinality)" OtherType : Indirect (via interface)
 ```
 
+#### Interface Realization
+
```
Olshansk, line 125:
```suggestion

_tl;dr An instance (i.e. client) implements an interface (i.e. supplier)_

```

p2p/README.md
```diff
@@ -106,9 +121,48 @@ ConcreteType --* OtherType : Aggregatation
...
+> Realization is a specialized abstraction relationship between two sets of model elements, one representing a specification (the supplier) and the other represents an implementation of the latter (the client).
+
+> Realization can be used to model stepwise refinement, optimizations, transformations, templates, model synthesis, framework composition, etc.
+
+_(see: [UML Realization](https://www.uml-diagrams.org/realization.html))_
+
+#### Direct Usage
+
```
Olshansk, line 133:
```suggestion

_tl;dr One instance (i.e. client) is dependant on another instance (i.e. supplier) to function properly. _ 

```

p2p/README.md
```diff
@@ -106,9 +121,48 @@ ConcreteType --* OtherType : Aggregatation
...
+> Dependency is a directed relationship which is used to show that some UML element or a set of elements requires, needs or depends on other model elements for specification or implementation. Because of this, dependency is called a supplier - client relationship, where supplier provides something to the client, and thus the client is in some sense incomplete while semantically or structurally dependent on the supplier element(s). Modification of the supplier may impact the client elements.
+
+> Usage is a dependency in which one named element (client) requires another named element (supplier) for its full definition or implementation.
+
+_(see: [UML Dependency](https://www.uml-diagrams.org/dependency.html))_
+
+#### Composition
+
```
Olshansk, line 141:
```suggestion

_tl;dr Deleting an instance also deletes all the instances it holds_

```

p2p/README.md
```diff
@@ -106,9 +121,48 @@ ConcreteType --* OtherType : Aggregatation
...
+> A "strong" form of aggregation
+
+> If a composite (whole) is deleted, all of its composite parts are "normally" deleted with it.
+
+_(see: [UML Shared composition](https://www.uml-diagrams.org/composition.html))_
+
+#### Aggregation
+
```
Olshansk, line 149:
```suggestion

_tl;dr Deleting an instance does not necessarily delete instances it depends on_


```

p2p/README.md
```diff
@@ -106,9 +121,48 @@ ConcreteType --* OtherType : Aggregatation
...
+> A "weak" form of composition
+
+> Shared part could be included in several composites, and if some or all of the composites are deleted, shared part may still exist.
+
+_(see: [UML Shared aggregation](https://www.uml-diagrams.org/aggregation.html))_
+
+#### Cardinality
+
```
Olshansk, line 157:
```suggestion

_tl;dr An M:N relationship defines how M instances of one type work with N instances of another type._

```

p2p/README.md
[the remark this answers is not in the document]
Olshansk: As noted in the P2P knowledge transfer (#438) ticket, for hotstuff consensus, we don't use RainTree for messages propagation between staked actors (e.g. state sync) but for consensus, it's a star pattern (i.e. direct)

p2p/utils_test.go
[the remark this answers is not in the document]
Olshansk: Sgtm. One small request: Can you take a screenshot of this thread and use it in a github issue as part of your techdebt comment?

I think it's important enough to have a separate issue for and document the temporary regression.

p2p/module.go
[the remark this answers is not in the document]
Olshansk: LGTM

p2p/README.md
Olshansk: Got it. Let's go with `libp2p gossipub redundancy` then - closer to what you had before

p2p/README.md
Olshansk: LGTM, thanks for the update!

p2p/README.md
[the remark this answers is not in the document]
bryanchriswhite: Gotcha, thanks for the additional context - I see what you mean now. I think your suggestion of "NA (Direct Send)" is orthogonal to the "router" degree of freedom as it would apply to the entire table. It seems like the "unicast" label above the table was insufficient in the context I intended it to provide. :sweat_smile:

I think where tripping over the distinction between "raintree router" and the "raintree algorithm/propogation". What I was originally trying to communicate here is that the "raintree router" (hence, the column name) will be used for staked-to-staked actor communication. This also aligns with the structure of the component diagrams under "Message Propagation & Handling". 

Do you have a suggestion for how to better structure this?

p2p/utils_test.go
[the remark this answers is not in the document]
bryanchriswhite: - #886

p2p/README.md
bryanchriswhite: I've only ever seen cardinality expressed in terms of "<client> has <type> association with <cardinality> number of <supplier(s)>".

I think `M:N` is an unconventional way to think about it. For example, in both composition or aggregation 
 it doesn't make sense to consider a ratio where `M` is anything other than `1` as in both association types, only one end can be the aggregation (as per the UML spec). 

I'm more than happy to take a pass at a tl;dr for this one that you can criticize. :sweat_smile:

bryanchriswhite: Here's what I went with, wdyt?:

---

#### Cardinality

_TL;DR indicates a number, or range of instances associated (i.e. supplier(s))_

Cardinality indicates the number or range of simultaneous instances of supplier that are associated with the client. Applicable to multiple association types. Can be expressed arbitrarily (e.g. wildcards, variable, equation, etc.)

p2p/event_handler.go
```diff
@@ -50,7 +57,7 @@ func (m *p2pModule) HandleEvent(event *anypb.Any) error {
 		m.logger.Debug().Fields(messaging.TransitionEventToMap(stateMachineTransitionEvent)).Msg("Received state machine transition event")
 
 		if stateMachineTransitionEvent.NewState == string(coreTypes.StateMachineState_P2P_Bootstrapping) {
-			if m.router.GetPeerstore().Size() == 0 {
+			if m.stakedActorRouter.GetPeerstore().Size() == 0 {
```
bryanchriswhite, line 60: need to check `m.isStakedActor()` here...

p2p/background/router_test.go
```diff
@@ -114,6 +123,73 @@ func TestBackgroundRouter_RemovePeer(t *testing.T) {
 	require.Len(t, existingPeerstoreAddrs, 1)
 }
 
+func TestBackgroundRouter_Validation(t *testing.T) {
+	ctx := context.Background()
+	libp2pMockNet := mocknet.New()
+
+	invalidWireFormatData := []byte("test message")
```
Olshansk, line 130:
```suggestion
	invalidWireFormatData := []byte("not a serialized proto")
```

p2p/background/router_test.go
```diff
@@ -114,6 +123,73 @@ func TestBackgroundRouter_RemovePeer(t *testing.T) {
...
+	invalidPocketEnvelope := &anypb.Any{
+		TypeUrl: "/test",
```
Olshansk, line 132:
```suggestion
		TypeUrl: "/notADefinedProtoMessage",
```

p2p/background/router_test.go
```diff
@@ -114,6 +123,73 @@ func TestBackgroundRouter_RemovePeer(t *testing.T) {
...
+
+	for _, invalidMessageBz := range invalidMessages {
+		err = topic.Publish(ctx, invalidMessageBz)
+		require.NoError(t, err)
+	}
+
+	select {
+	case <-time.After(time.Second * 2):
+		// TECHDEBT: find a better way to prove that pre-propagation validation
+		// works as expected. Ideally, we should be able to distinguish which
```
Olshansk, line 182: Wdym by `Ideally, we should be able to distinguish which invalid message was received in the event of failure.`

Can't we filter based on the message type/content?

p2p/background/router_test.go
```diff
@@ -114,6 +123,73 @@ func TestBackgroundRouter_RemovePeer(t *testing.T) {
...
+		// invalid message was received in the event of failure.
+		//
+		// CONSIDERATION: we could use the telemetry module mock to set expectations
+		// for validation failure telemetry calls, which would probably be useful in
+		// their own right.
+	case <-receivedChan:
+		t.Fatal("expected message to not be received")
```
Olshansk, line 189:
```suggestion
		t.Fatal("no messages should have been received by the host")
```

p2p/background/router_test.go
```diff
@@ -138,30 +214,44 @@ func TestBackgroundRouter_Broadcast(t *testing.T) {
...
 		host := newTestHost(t, libp2pMockNet, privKey)
 		testHosts = append(testHosts, host)
 		expectedPeerIDs[i] = host.ID().String()
-		rtr := newRouterWithSelfPeerAndHost(t, selfPeer, host)
-		go readSubscription(t, ctx, &broadcastWaitgroup, rtr, &seenMessagesMutext, seenMessages)
+		newRouterWithSelfPeerAndHost(t, peer, host, func(data []byte) error {
+			seenMessagesMutext.Lock()
+			broadcastWaitgroup.Done()
+			seenMessages[host.ID().String()] = struct{}{}
+			seenMessagesMutext.Unlock()
```
Olshansk, line 239:
```suggestion
			seenMessagesMutext.Lock()
			defer seenMessagesMutext.Unlock()
			seenMessages[host.ID().String()] = struct{}{}
			broadcastWaitgroup.Done()
```

p2p/README.md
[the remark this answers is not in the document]
Olshansk: How about this:

```
 Sender         | Receiver       | Router          | Example Usage                                        |
|----------------|----------------|-----------------|------------------------------------------------------|
| Staked Actor   | Staked Actor   | No router  | Consensus (hotstuff) messages (to validators only) |
| Staked Actor   | Staked Actor   | Raintree only   | Consensus (state sync) messages (to validators only) |
| Unstaked Actor | Staked Actor   | Background only | Consensus (state sync) messages (to validators only) |
| Unstaked Actor | Unstaked Actor | Background only | Consensus (state sync) & Debug (CLI) messages        |
```

p2p/README.md
Olshansk: @bryanchriswhite Can you please apply these changes?

Olshansk (changes requested):

p2p/README.md
Olshansk: I like the tl;dr let's do it!

p2p/background/router_test.go
bryanchriswhite: Thanks for pressing this, there has been a lot of shifting ground around these parts so it's good that you made me revisit! :pray: 

The issue here currently is that we're passing a method of `BackgroundRouter` to [`Gossipsub#RegisterTopicValidator()`](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#section-readme) which doesn't leave any room for the test to hook in. After looking at the docs again though, I'm realizing that I can use `#UnregisterTopicValidator()` in the test and subsequently re-register a wrapper function which the test can hook into. :+1:

bryanchriswhite: I re-wrote the test, apologize the the additional review effort. I think it's much higher quality now though - I hope you agree. :sweat_smile:

p2p/README.md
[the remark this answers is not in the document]
bryanchriswhite: :thinking: I think we're still coming at this from different perspectives, I think I was also confused about the "validator only" distinction.

It seems like you're trying to articulate a couple of points: 

1. That hotstuff messages are unicast rather than broadcast
2. Unicast routing does not use the raintree gossip algorithm

In my view, 1. is implied by the fact that this is the "unicast" table and therefore applies to all these. If we applied your perspective consistently to the entire table, I would expect to see "no router" in the "router" column for every row, which seems wrong to me.

My intention with this table was not to speak to which "gossip algorithm" is used, but rather which `Router` implementation is handling the control flow, matching the diagrams that follow. This is important because the raintree and background routers have distinct message types, protocol IDs, and stream handlers.

I hope that helps clarify my perspective. Please correct my understanding of yours if I've missed something or misunderstood.

I've updated both tables to try to improve clarity as well as fix some mistakes / fill in gaps, here's my new proposal:

**Unicast**

| Sender         | Receiver       | Router          | Example Usage                                                        |
|----------------|----------------|-----------------|----------------------------------------------------------------------|
| Staked Actor   | Staked Actor   | Raintree only   | Consensus hotstuff messages (validators only) & state sync responses |
| Staked Actor   | Untaked Actor  | Background only | Consensus state sync responses                                       |
| Unstaked Actor | Staked Actor   | Background only | Consensus state sync responses, debug messages                       |
| Unstaked Actor | Unstaked Actor | Background only | Consensus state sync responses, debug messages                       |

**Broadcast**

| Broadcaster    | Receiver       | Router                | Example Usage                                                   |
|----------------|----------------|-----------------------|-----------------------------------------------------------------|
| Staked Actor   | Staked Actor   | Raintree + Background | Utility tx messages, consensus state sync requests              |
| Staked Actor   | Untaked Actor  | Background only       | Utility tx messages (redundancy), consensus state sync requests |
| Unstaked Actor | Staked Actor   | Background only       | Utility tx messages (redundancy), consensus state sync requests |
| Unstaked Actor | Unstaked Actor | Background only       | Utility tx messages, consensus state sync requests              |

p2p/background/router_test.go
```diff
@@ -124,69 +127,112 @@ func TestBackgroundRouter_RemovePeer(t *testing.T) {
...
+	}{
+		{
+			name: "invalid BackgroundMessage",
+			// NB: `msgBz` would normally be a serialized `BackgroundMessage`.
+			msgBz: mustMarshal(t, &invalidProtoMessage),
+		},
+		{
+			name: "empty PocketEnvelope",
+			msgBz: mustMarshal(t, &typesP2P.BackgroundMessage{
+				// NB: `Data` is normally a serialized `PocketEnvelope`.
```
Olshansk, line 147: Appreciate the comments 🙌

p2p/background/router_test.go
```diff
@@ -124,69 +127,112 @@ func TestBackgroundRouter_RemovePeer(t *testing.T) {
...
 	receiverHost := newTestHost(t, libp2pMockNet, receiverPrivKey)
-	receiverRouter := newRouterWithSelfPeerAndHost(t, receiverPeer, receiverHost, func(data []byte) error {
-		receivedChan <- struct{}{}
-		return nil
-	})
+	receiverRouter := newRouterWithSelfPeerAndHost(
+		t, receiverPeer,
+		receiverHost,
+		func(data []byte) error {
+			receivedChan <- data
```
Olshansk, line 171: Nice change

Olshansk (approved):
Let's 🚢 this!!

p2p/README.md
```diff
@@ -67,105 +68,506 @@ flowchart TD
...
+| Unstaked Actor | Staked Actor   | Background only | Consensus (state sync) messages (to validators only) |
+| Unstaked Actor | Unstaked Actor | Background only | Consensus (state sync) & Debug (CLI) messages        |
+
+**Broadcast**
+
+| Broadcaster    | Receiver       | Router                | Example Usage                                     |
+|----------------|----------------|-----------------------|---------------------------------------------------|
+| Staked Actor   | Staked Actor   | Raintree + Background | Utility tx messages                               |
+| Unstaked Actor | Staked Actor   | Background only       | Utility tx messages (libp2p gossipsub redundancy) |
+| Unstaked Actor | Unstaked Actor | Background only       | Utility tx messages                               |
```
bryanchriswhite, line 201:
```suggestion
**Unicast**

| Sender         | Receiver       | Router          | Example Usage                                                        |
|----------------|----------------|-----------------|----------------------------------------------------------------------|
| Staked Actor   | Staked Actor   | Raintree only   | Consensus hotstuff messages (validators only) & state sync responses |
| Staked Actor   | Untaked Actor  | Background only | Consensus state sync responses                                       |
| Unstaked Actor | Staked Actor   | Background only | Consensus state sync responses, debug messages                       |
| Unstaked Actor | Unstaked Actor | Background only | Consensus state sync responses, debug messages                       |

**Broadcast**

| Broadcaster    | Receiver       | Router                | Example Usage                                                   |
|----------------|----------------|-----------------------|-----------------------------------------------------------------|
| Staked Actor   | Staked Actor   | Raintree + Background | Utility tx messages, consensus state sync requests              |
| Staked Actor   | Untaked Actor  | Background only       | Utility tx messages (redundancy), consensus state sync requests |
| Unstaked Actor | Staked Actor   | Background only       | Utility tx messages (redundancy), consensus state sync requests |
| Unstaked Actor | Unstaked Actor | Background only       | Utility tx messages, consensus state sync requests              |
```

PR #742, opened 2023-05-12
Refactoring Request Body Initialization in HTTP Handlers

innocent-saeed36:
Previously, the request bodies body were declared as value types, and then their addresses were passed to the ctx.Bind(&body) function.

In this commit, I have streamlined this process by declaring body as a pointer upfront. The ctx.Bind() function call now directly uses body without requiring the & operator.

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -83,8 +83,8 @@ func (s *rpcServer) PostV1ClientGetSession(ctx echo.Context) error {
 // TECHDEBT: This will need to be changed when the HandleRelay function is actually implemented
 // because it copies data structures from v0. For example, AATs are no longer necessary in v1.
 func (s *rpcServer) PostV1ClientRelay(ctx echo.Context) error {
-	var body RelayRequest
-	if err := ctx.Bind(&body); err != nil {
+	body := new(SessionRequest)
+	if err := ctx.Bind(body); err != nil {
 		return ctx.String(http.StatusBadRequest, "bad request")
 	}
 


Olshansk (approved):
Makes sense to me and happy to approve. Curious what prompted this change from you?

h5law (changes requested):
So, I think we should either go one way or the other. Essentially this is entirely style based but I am a big fan of patterns so why not make this one consistent?

We have a few handlers that use `new()` and others that pass the `&` reference. As this PR only addresses a single instance of these we should probably update:
1) All other handlers to follow this pattern
2) Refactor the few handlers not using the `var xyz; ctx.Bind(&xyz)` pattern to use that instead.

Either way this PR will need to address more than these two lines to make everything consistent IIRC.

innocent-saeed36:
> Makes sense to me and happy to approve. Curious what prompted this change from you?

Hi Daniel. I'm a dev beginner and I'm playing around POKT codebase. I was going through code by the help of AI and it pinpointed me this area which can be improved. I'm unfortunately incapable of helping much further from here at a current knowledge level, so I'm happy if this helped rethinking improvements on this matter throughout the entire codebase. I'm a POKT supporter in the first place. I'll be playing around POKT codebase in the future as well, so let me know if you want me to pinpoint such things for rethinking in the future or keep them for myself. Thank you and have a great day all.

Olshansk:
@innocent-saeed36 Please close this out if you don't plan to follow the recommended changes from @h5law

Olshansk:
Per @h5law's comment, going to close this out. @innocent-saeed36 If you still believe we should do this, please apply it across the whole codebase

innocent-saeed36:
Thanks @Olshansk for the comment. I'll not be working on it, at least not for now, but I'm happy if it open up ideas for @h5law to work on it. I'm happy if it helped in any way.

PR #743, opened 2023-05-13
Improved Error Handling in TransactionExists Function

innocent-saeed36:
### Description:

This commit introduces an important update to the `TransactionExists` function in the persistence module, aiming to provide a more accurate return value and better error handling.

### Changes:

Previously, the function had an issue where it would return `true` for the boolean result even if an error occurred during the transaction search process. This was due to the function checking if the result (`res`) was `nil` before checking if an error (`err`) had occurred.

The updated version of the function first checks if the result (`res`) is not `nil` (indicating that a transaction was found), and then checks if an error occurred during the process. If an error occurred, it now correctly returns `false` for the boolean value (indicating that the transaction was not found due to the error).

### Impact:

This change is crucial as it prevents potential confusion or misleading outcomes from the `TransactionExists` function. By ensuring that the boolean result correctly reflects whether the transaction was found or not, we can provide more accurate information to the function's callers and handle potential errors more effectively.

### DIFF
--- a/persistence/block.go
+++ b/persistence/block.go
@@ -18,14 +18,17 @@ func (p *persistenceModule) TransactionExists(transactionHash string) (bool, err
 		return false, err
 	}
 	res, err := p.txIndexer.GetByHash(hash)
-	if res == nil {
-		// check for not found
-		if err != nil && err.Error() == kvstore.BadgerKeyNotFoundError {
-			return false, nil
+	if res != nil {
+		if err != nil {
+			return false, err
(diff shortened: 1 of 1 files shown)


Olshansk (changes requested):
@innocent-saeed36 Any chance you can add a unit test that would've failed before this changed but passed after?

Olshansk:
@innocent-saeed36 Please reopen if you'd still like to move forward with this change.

PR #744, opened 2023-05-13
[Persistence] Removing code duplication, improving error handling in shouldHydrateGenesisDb function and refactoring Create function

innocent-saeed36:
This is the last PR I'll be submitting not to spam GitHub area with non consistent PRs. If you still find them helpful, feel free to let me know.

### Change 1:

The existing code has two identical lines that serve the same purpose, which is to check at compile-time that `persistenceModule` implements the modules.`PersistenceModule` interface. Including this line twice doesn't provide any additional benefit.

### Change 2:

In the current code, if there is an error when calling `readCtx.GetMaximumBlockHeight()`, the function simply returns `true` for `shouldHydrateGenesisDb` and ignores the error. This might lead to incorrect program behavior because the error could indicate a problem that needs to be resolved. In the revised code, the function returns the error itself, allowing the caller to handle it appropriately.

New codes comes with improved error handling in `shouldHydrateGenesisDb` function. If there's an error when calling r`eadCtx.GetMaximumBlockHeight()`, the function now correctly returns the error to the caller. This change enhances the robustness of the code by ensuring that potential errors are not silently ignored and can be properly dealt with by the calling function.

### Change 3:

In the current code, there is an unnecessary `else` statement after the `if` block that checks `shouldHydrateGenesis`. As the `if` block returns if there's an error, the following code can be moved outside of the `else` block. Removing the `else` statement makes the code simpler, easier to read, and more "Go idiomatic", as Go encourages avoiding unnecessary `else` after `return` in `if`.

### DIFF
--- a/persistence/module.go
+++ b/persistence/module.go
@@ -17,7 +17,6 @@ import (
 
 var (
 	_ modules.PersistenceModule = &persistenceModule{}
-	_ modules.PersistenceModule = &persistenceModule{}
 
 	_ modules.PersistenceRWContext = &PostgresContext{}
 )
@@ -120,16 +119,18 @@ func (*persistenceModule) Create(bus modules.Bus, options ...modules.ModuleOptio
 	// 		     this forces the genesis state to be reloaded on every node startup until state
 	//           sync is implemented.
 	// Determine if we should hydrate the genesis db or use the current state of the DB attached
-	if shouldHydrateGenesis, err := m.shouldHydrateGenesisDb(); err != nil {
+	shouldHydrateGenesis, err := m.shouldHydrateGenesisDb()
+	if err != nil {
 		return nil, err
-	} else if shouldHydrateGenesis {
+	}
+	if shouldHydrateGenesis {
 		m.populateGenesisState(genesisState) // fatal if there's an error
 	} else {
 		// This configurations will connect to the SQL database and key-value stores specified
 		// in the configurations and connected to those.
 		logger.Global.Info().Msg("Loading state from disk...")
 	}
-
+	
 	return m, nil
 }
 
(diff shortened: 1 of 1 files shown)


Olshansk (approved):

Olshansk:
This looks good to me! @innocent-saeed36, do you mind sharing what tools/prompts you used to find these changes? It would be really awesome to work together on automating it!

persistence/module.go
```diff
@@ -262,7 +263,7 @@ func (m *persistenceModule) shouldHydrateGenesisDb() (bool, error) {
 
 	blockHeight, err := readCtx.GetMaximumBlockHeight()
 	if err != nil {
-		return true, nil
+		return false, err
```
bryanchriswhite, line 266: I'm not super familiar with the logic here but from a cursory read of the context, my interpretation is that if there was an error loading the state from the DB then we assume that we "should hydrate" it, which is what (`persistendeModule#populateGenesisState()`)[https://github.com/pokt-network/pocket/blob/main/persistence/genesis.go#L15] seems to be responsible for. **The tests pass again locally when I revert this change.**

In a future refactor, I would expect that we may make some assertion about the error and have another branch which does return it if it is an unexpected error.

Olshansk (changes requested):
@innocent-saeed36

1. See @bryanchriswhite's comments
2. Are you part of our discord server? If so, would be great to chat there to understand what kind of contributions you'd like to make to the pocket repo in the future

Olshansk:
Closing per the broken build and @bryanchriswhite's comment.

innocent-saeed36:
> This looks good to me! @innocent-saeed36, do you mind sharing what tools/prompts you used to find these changes? It would be really awesome to work together on automating it!

Hello @Olshansk , to help you how to detect possible improvements and potential bugs, it's useful to use  GPT-4 for the entire .go file and use it for almost any file in the entire codebase. It unfortunately cannot understand interconnections among files and functions as it cannot use multiple files, but prompts that can help a bit is:
`extensively analyze the following code, tell me if there are any bugs, potential bugs or places for improvements:`

and then copy paste the entire .go file content. It can be automated I guess for all .go (and other) files to see where it makes sense to make improvements. Thanks and sorry for the slower response. Feel free to ping me if you need me for something else or if you want to discuss further on Discord.

Issue #762, opened 2023-05-17
[P2P] Support multiple simultaneous routers


bryanchriswhite:
## Objective

Enable the P2P module to handle incoming messages from multiple router implementations simultaneously.

## Origin Document

Currently, the message handling interaction between the P2P and the router looks like this:
```mermaid
flowchart LR
    subgraph p2p[P2P Module]
        subgraph rt[RainTree Router]
            nd[Nonce Deduper]
            rth[[`RainTreeMessage` Handler]]
        end

        p2ph[[`PocketEnvelope` message Handler]]
    end

rth --> nd
rth --> p2ph
```

```mermaid
classDiagram
    class RainTreeMessage {
        <<protobuf>>
        +Level uint32
        +Data []byte
        +Nonce uint64
    }
    
    class PocketEnvelope {
        <<protobuf>>
        +Content *anypb.Any   
    }

    RainTreeMessage --* PocketEnvelope : serialized as `Data`
    
      
    class P2PModule {
        -handleNetworkData([]byte) error
        -handleStream(stream libp2pNetwork.Stream)
        -readStream(stream libp2pNetwork.Stream)
    }

    class RainTreeRouter {
        +HandleNetworkData func([]byte) ([]byte, error)
    }

    RainTreeRouter --> P2PModule
    P2PModule --> RainTreeRouter
    RainTreeRouter --o RainTreeMessage
    RainTreeRouter --* NonceDeduper
    P2PModule --o PocketEnvelope
```

## Goals

- Support handling message from multiple routers simultaneously
- Support message deduplication across all messages (i.e. not just raintree)
- Improve separation of concerns between P2P module and router(s)

## Deliverable

```mermaid
flowchart LR
    subgraph p2p[P2P Module]
        subgraph rt[RainTree Router]
            rth[[`RainTreeMessage` Handler]]
        end

        nd[Nonce Deduper]
    
        p2ph[[`PocketEnvelope` message Handler]]
    end

p2ph --> nd
rth --> p2ph
```

```mermaid
classDiagram
    class RainTreeMessage {
        <<protobuf>>
        +Level uint32
        +Data []byte
    }
    
    class PocketEnvelope {
        <<protobuf>>
        +Content *anypb.Any
        +Nonce uint64
    }

    RainTreeMessage --* PocketEnvelope : serialized as `Data`
    
    class P2PModule {
        -handleAppData([]byte) error
    }

    class RainTreeRouter {
        -appDataHandler RouterHandler
[...]

PR #763, opened 2023-05-17
[P2P] refactor: message handling

bryanchriswhite:
### Before

The P2P module is responsible for handling incoming messages from the router's host. When it receives a message, it calls into the router to unpack the `RainTreeMessage` (and facilitate further broadcast propagation). The resulting `PocketEnvelope` is returned to the P2P module to be emitted over the application event bus.

```mermaid
classDiagram
    class RainTreeMessage {
        <<protobuf>>
        +Level uint32
        +Data []byte
        +Nonce uint64
    }
    
    class PocketEnvelope {
        <<protobuf>>
        +Content *anypb.Any   
    }

    class P2PModule {
        -handleNetworkData([]byte) error
    }

    class RainTreeRouter {
        +HandleNetworkData func([]byte) ([]byte, error)
    }

    RainTreeRouter --> P2PModule
    P2PModule --> RainTreeRouter
    P2PModule --o PocketEnvelope
```

### After

The router encapsulates handling incoming `RainTreeMessage`s, unpacking them and passing them along to the P2P module as serialized `PocketEnvelope`s. The P2P module then deserializes and emits them over the application event bus.

```mermaid
classDiagram
    class RainTreeMessage {
        <<protobuf>>
        +Level uint32
        +Data []byte
        +Nonce uint64
    }
    
    class PocketEnvelope {
        <<protobuf>>
        +Content *anypb.Any
    }

    class P2PModule {
        -handlePocketEnvelope([]byte) error
    }

    class RainTreeRouter {
        -handler RouterHandler
        -handleRainTreeMsg([]byte) ([]byte, error)
    }

    RainTreeRouter --> P2PModule : `handler` == `handlePocketEnvelope`

    P2PModule --o PocketEnvelope
```

## Issue

Second deliverable in #762 

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Added `Handler` field to `RainTreeConfig` & `BackgroundConfig`
- Refactored `rainTreeRouter` logging methods
- Renamed `rainTreeRouter#HandleNetworkData()` to `#handleRainTreeMsg()`
- Renamed `p2pModule#handleNetworkData()` to `#handleAppData()`
- Added -tags=test to all test make targets
- Fixed mockdns usage in `TestP2PModule_Insecure_Error` test
- Moved message handling from p2p module to router

## Testing

## Required Checklist

### If Applicable Checklist


p2p/raintree/logging.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/pokt-network/pocket/p2p/utils"
+)
+
+// logStream logs the incoming stream and its scope stats
+func (rtr *rainTreeRouter) logStream(stream libp2pNetwork.Stream) {
+	rtr.logStreamScopeStats(stream)
+
+	remotePeer, err := utils.PeerFromLibp2pStream(stream)
+	if err != nil {
+		rtr.logger.Debug().Err(err).Msg("getting remote remotePeer")
```
Olshansk, line 16: Why did you go with `Debug` instead of `Error` here?

Ditto below

p2p/raintree/testutil.go
```diff
@@ -0,0 +1,13 @@
+//go:build test
+
+package raintree
+
+import libp2pNetwork "github.com/libp2p/go-libp2p/core/network"
+
+// RainTreeRouter exports `rainTreeRouter` for testing purposes.
```
Olshansk, line 7: I like this design 💯

p2p/raintree/router.go
```diff
@@ -282,6 +298,79 @@ func (rtr *rainTreeRouter) Size() int {
...
+
+// readStream reads the incoming stream, extracts the serialized `PocketEnvelope`
+// data from the incoming `RainTreeMessage`, and passes it to the application by
+// calling the configured `rtr.handler`.
+func (rtr *rainTreeRouter) readStream(stream libp2pNetwork.Stream) {
+	// Time out if no data is sent to free resources.
+	if err := stream.SetReadDeadline(newReadStreamDeadline()); err != nil {
+		// NB: tests using libp2p's `mocknet` rely on this not returning an error.
+		// `SetReadDeadline` not supported by `mocknet` streams.
+		rtr.logger.Debug().Err(err).Msg("setting stream read deadline")
```
Olshansk, line 334: Ditto. Noticed you're using Debug instead of Err for a lot of these. Why is that?

p2p/raintree/router.go
```diff
@@ -282,6 +298,79 @@ func (rtr *rainTreeRouter) Size() int {
...
+	go rtr.readStream(stream)
+}
+
+// readStream reads the incoming stream, extracts the serialized `PocketEnvelope`
+// data from the incoming `RainTreeMessage`, and passes it to the application by
+// calling the configured `rtr.handler`.
+func (rtr *rainTreeRouter) readStream(stream libp2pNetwork.Stream) {
+	// Time out if no data is sent to free resources.
+	if err := stream.SetReadDeadline(newReadStreamDeadline()); err != nil {
+		// NB: tests using libp2p's `mocknet` rely on this not returning an error.
```
Olshansk, line 332: I think you should move this to the method header and add that the function is intended to be called in a goroutine.

p2p/raintree/router.go
```diff
@@ -282,6 +298,79 @@ func (rtr *rainTreeRouter) Size() int {
...
+	data, err := io.ReadAll(stream)
+	if err != nil {
+		rtr.logger.Error().Err(err).Msg("reading from stream")
+		if err := stream.Reset(); err != nil {
+			rtr.logger.Debug().Err(err).Msg("resetting stream (read-side)")
+		}
+		return
+	}
+
+	// done reading; reset to signal this to remote peer
```
Olshansk, line 350: Add a note that if we don't do this, we can easily max out the number of open network connections on the receiver's side

p2p/raintree/router.go
```diff
@@ -282,6 +298,79 @@ func (rtr *rainTreeRouter) Size() int {
...
+	if err := stream.Reset(); err != nil {
+		rtr.logger.Debug().Err(err).Msg("resetting stream (read-side)")
+	}
+
+	// extract `PocketEnvelope` from `RainTreeMessage` (& continue propagation)
+	appMsgData, err := rtr.handleRainTreeMsg(data)
```
Olshansk, line 356: `s/appMsgData/pocketEnvelopeBz` ?

I realize you were staying consistent with what we had before but doesn't hurt to be more explicit seeing how we're already here.

Olshansk (changes requested):
Wdyt about adding the mermaid diagram in the github description to the README for p2p?

p2p/module_raintree_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build test
```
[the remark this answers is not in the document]
bryanchriswhite: It's only necessary to add the test build tag to:
1. **non-test** files in a non-internal package which exports a private member for testing purposes.
2. **test files** which depend on something exported by a file which uses the build tag (assumes at least one instance of case 1). 


Are you suggesting a linter which enforces the `test` build tag on all test files or something?

I don't suppose there would really be a downside to adding the test tag in places where it's not strictly necessary. I can imagine an argument for either but am leaning towards "ubiquitous build tags" at the moment (unexpectedly):

- ubiquitous build tags
  - pros
    - consistent
    - simpler
    - better justifies necessary updates to the makefile (e.g. `-tags test`; necessary in either case)
  - cons
    - slightly unconventional
    - requires running tests via make targets OR adding tags manually (e.g. IDE run/test configs)
  - UX
    - `go test` runs no tests
- only necessary build tags
  - pros
    - minimal impact on the codebase
  - cons
    - more mental overhead
    - less obvious when test(s) didn't run because of a missing tag
  - UX
    - `go test` runs some but not all tests

bryanchriswhite:
> Wdyt about adding the mermaid diagram in the github description to the README for p2p?

@Olshansk I was planning on incorporating the diagrams from #505, which represents the composition of this and #769, into #732. My thinking was that the diagrams in the PRs are trimmed down versions which only show what's relevant for understanding the change, whereas what I expect we want to keep in the repo is the representation of the desired state with sufficient context.

Do you imagine a step in between?

p2p/raintree/router.go
bryanchriswhite: :100: :fire: Perfect!

p2p/raintree/router.go
bryanchriswhite: Thank you for scrutinizing this, please continue to do so! :raised_hands: 

I was trying to be conservative with what we log to non-debug loggers. In this particular case, my thinking was that being unable to set the read deadline isn't actionable and isn't really useful as a signal for end-users (unlike the errors on lines 327 & 340).

Looking over it once more, you've convinced me that the `stream.Reset()` error should probably go to the `Error()` logger instead but I think only in the case where we're not already handling another error (i.e. `io.ReadAll()` on line 312). Does that seem reasonable?

p2p/raintree/logging.go
bryanchriswhite: (see below)

In this case, my thinking was that this error only happens in the context of this logging helper function which is not a critical function. I didn't imagine that this logging helper producing an error would be useful to the end user (i.e. not actionable nor a useful signal).

Makefile
```diff
@@ -339,102 +339,102 @@ generate_node_state_machine_diagram: ## (Re)generates the Node State Machine dia
...
 
 .PHONY: test_app
 test_app: ## Run all go app module unit tests
-	go test ${VERBOSE_TEST} -p=1 -count=1  ./app/...
+	go test ${VERBOSE_TEST} -p=1 -count=1 -tags=test ./app/...
 
 .PHONY: test_utility
 test_utility: ## Run all go utility module unit tests
-	go test ${VERBOSE_TEST} -p=1 -count=1  ./utility/...
+	go test ${VERBOSE_TEST} -p=1 -count=1 -tags=test./utility/...
```
dylanlott, line 363: I think you need a space between test and ./utility

Olshansk (approved):

p2p/raintree/router.go
```diff
@@ -282,6 +298,81 @@ func (rtr *rainTreeRouter) Size() int {
 	return rtr.peersManager.GetPeerstore().Size()
 }
 
+// handleStream ensures the peerstore contains the remote peer and then reads
+// the incoming stream in a new go routine.
+func (rtr *rainTreeRouter) handleStream(stream libp2pNetwork.Stream) {
+	rtr.logger.Debug().Msg("handling incoming stream")
```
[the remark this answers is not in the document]
bryanchriswhite: Oh neat, thanks for sharing! :raised_hands:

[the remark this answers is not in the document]
bryanchriswhite: I took a look at both `#handleStream()` and `#readStream()` (assuming that's what you were referring to) and it looks like this line (:304) is the only one that's not an error level log. If there were several debug level logs, I agree that it would be worth it but am not convinced it adds any value at the moment. Will definitely keep this API and use case in mind though in the future! :wink:

PR #767, opened 2023-05-18
[Helm] Add ServiceMonitor to the helm chart

okdas:
## Description

Adds `ServiceMonitor` to the helm chart, so our software can be monitored by different monitoring stacks that support this CRD, which is just about an industry standard now.

We are going to use it in our infrastructure as well.

Also, included changes to run e2e tests within one, `main` GitHub action workflow because the other option wasn't working on non-default branches.

## Type of change

## List of changes

- Add a new 'ServiceMonitor' object to the helm chart, which will be disabled by default because not everyone has this CRD installed on their cluster.
- Adjusted e2e test workflow so it kicks off on pull requests.

## Testing

## Required Checklist

### If Applicable Checklist


.github/workflows/main.yml
```diff
@@ -6,13 +6,7 @@ on:
   workflow_dispatch:
   push:
     branches: [main]
-    paths-ignore:
```
okdas, line 9: Given that we now depend on the most recent gitsha from the pull request (PR) to include a container image, we should not hinder the continuous integration (CI) process from generating an image in cases where only a documentation change is made.

.github/workflows/main.yml
```diff
@@ -151,3 +145,57 @@ jobs:
           cache-to: type=gha,mode=max
           build-args: |
             TARGET_GOLANG_VERSION=${{ env.TARGET_GOLANG_VERSION }}
+
+
+  # Run e2e tests on devnet IF the PR has a label "e2e-devnet-test"
+  e2e-tests:
```
okdas, line 151: I missed a clarification in the Github Actions documentation that this would only work on the default branch, so having a separate workflow did not work. Going to have a larger main workflow instead.

okdas:
In addition to the changes related to this PR, I had to adjust a GitHub Action as they didn't work as expected.
But now you can actually see the e2e test passed! :)

.github/workflows/main.yml
```diff
@@ -151,3 +145,58 @@ jobs:
           cache-to: type=gha,mode=max
           build-args: |
             TARGET_GOLANG_VERSION=${{ env.TARGET_GOLANG_VERSION }}
+
+
+  # Run e2e tests on devnet IF the PR has a label "e2e-devnet-test"
```
Olshansk, line 150:
```suggestion
  # Run e2e tests on devnet if the PR has a label "e2e-devnet-test"
```

.github/workflows/main.yml
Olshansk: Can you leave an inline comment. Maybe something like `OPTIMIZE: We generate new images even on non src code changes, but this cost is okay for now`

charts/pocket/values.yaml
```diff
@@ -170,6 +170,8 @@ service:
   annotations: {}
```
Olshansk, line 170: Was going to ask you some stuff, but learnt everything from ChatGPT.

Screenshot 2023-05-22 at 4 07 50 PM

Olshansk (approved):
One minor need but otherwise LGTM

Issue #754, opened 2023-05-16
[Utility] A1: E2E Relay- Trustless Relay


Olshansk:
## Objective

Implement MVP E2E Feature Path A1: E2E Relay- Trustless Relay

## Origin Document

**Purpose:** A staked Application can make a direct relay and receive a response from a staked Servicer, selected by the Session, based on the GeoZone and ChainID.

**Actors**: Check all of the protocol actors involved in the feature

- [ ]  Validator
- [x]  Application
- [x]  Servicer
- [ ]  Fisherman
- [ ]  Portal

**Data Structures**

- `shared/core/types/proto/relay.proto` - will need to be updated to reflect the V1 design. For example:
    - AATs are no longer needed since the signature will define the validation
- `utility/service/service.go` will need to be updated to reflect V1-specific logic in servicing a request per the [[README](https://github.com/pokt-network/pocket/blob/main/utility/doc/PROTOCOL_RELAY.md)](https://github.com/pokt-network/pocket/blob/main/utility/doc/PROTOCOL_RELAY.md)

**Interfaces**

- `app/client/cli/` will need to be updated to reflect the changes made to `relay.proto`
- `shared/modules/utility_module.go`  may need to be updated with Servicer specific logic for handling a relay

**Diagram**

```mermaid
sequenceDiagram
    actor A as Application
    actor FN as Full/Dispatch Node
    actor S as Servicer
    actor D as Data Node

		title E2E Trustless Relay

    loop Repeats During Session
				A ->> +FN: GetSessionData
				FN ->> -A: SessionData
        A ->> +S: Relay Request
				S ->> S: Validate Relay
				S ->> +D: RPC Request
				D ->> -S: RPC Response
				S ->> S: Store signed RPC (Request, Response) 
				S ->>-A: Relay Response
    end
```

**User Stories as Tests**

- Happy test cases
    - An Application requests the account balance of a specific address at a specific height from a Servicer staked for the Ethereum RelayChain in the same GeoZone, and receives a successful response.
- Sad test cases
    - An Application requests the account balance of a specific address at a specific height from a Servicer staked for the Ethereum RelayChain in the same GeoZone, and the request times out without a response.

**Blockers**

- E2E feature path dependencies:
    - `E.1: Actor Staking`
- Infrastructure dependencies
    - Running an Ethereum data node

## Goals

- Complete the MVP implementation of the E2E Feature Path outlined in the objective
- Identify future tasks and test requirements to transition the feature to production

## Deliverable

[...]

PR #773, opened 2023-05-19
[Docs] add code review guidelines

bryanchriswhite:
## Description

This document was originally exported from my notes into [this notion doc](https://www.notion.so/pocketnetwork/Code-Review-Guidelines-proposal-5e22dbb0d40d44f69b5ccab42f64bfe5). I started it a while back and it represents a collaborative effort across multiple teams I've been a part of. This PR incorporates all the immediately actionable feedback in an attempt to get an initial, Pocket specific, version in the repo and move the conversation/iteration to GitHub.

### Summary generated by Reviewpad on 20 Jul 23 00:14 UTC
This pull request includes the addition of a new file called CODE_REVIEW_GUIDELINES.md in the docs/development directory. It outlines code quality and code review guidelines, including best practices and expectations for pull request authors and reviewers. It also provides information on review comments, finishing a review, and merging. Additionally, references to the CODE_REVIEW_GUIDELINES.md file were added in the README.md file. Two new images were also added in the assets directory.

## Issue

N/A

## Type of change

Please mark the relevant option(s):

- [ ] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

- Adds code review guidelines docs
- Adds code review guidelines doc links to main README

## Testing

- [ ] `make develop_test`; if any code changes were made
- [ ] `make test_e2e` on [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any code changes were made
- [ ] `e2e-devnet-test` passes tests on [DevNet](https://pocketnetwork.notion.site/How-to-DevNet-ff1598f27efe44c09f34e2aa0051f0dd); if any code was changed
- [ ] [Docker Compose LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md); if any major functionality was changed or introduced
- [ ] [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any infrastructure or configuration changes were made

## Required Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added, or updated, [`godoc` format comments](https://go.dev/blog/godoc) on touched members (see: [tip.golang.org/doc/comment](https://tip.golang.org/doc/comment))
- [ ] I have tested my changes using the available tooling
- [ ] I have updated the corresponding CHANGELOG

[...]

PR #778, opened 2023-05-22
[Utility] Add trustless relay to CLI

adshmh:
### Summary generated by Reviewpad on 19 Jun 23 16:17 UTC
This pull request contains various changes including updates to shared types, implementation of new validation functions, modifications to RPC documentation and API endpoint schema, implementation of a new servicer command in the CLI package, and changes to the JSON payload types. There is also a suggestion for improving the `Session` message and added `TECHDEBT` for signature field.

## Issue

Part of work on #754 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Added a "Relay" sub-command to CLI's "Servicer" command.

## Testing

## Required Checklist

### If Applicable Checklist


Olshansk:
@adshmh One of the things we discussed in today's backlog refinement is that if you need preliminary feedback on a PR that's unfinished (e.g. no tests), keep it as a draft, but add the reviewers. If you don't hear back within 48-72 hours, that's when pinging them in a discord channel is considered 👌. We're planning to add this to #773.

Olshansk:
@adshmh I updated the PR metadata on this. Check it out and feel free to ask questions if you don't know what to select in future PRs.

Screenshot 2023-05-25 at 1 38 25 PM

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+func servicerCommands() []*cobra.Command {
+	cmdDef := actorCmdDef{"Servicer", coreTypes.ActorType_ACTOR_TYPE_SERVICER}
+	cmds := []*cobra.Command{
+		newStakeCmd(cmdDef),
+		newEditStakeCmd(cmdDef),
+		newUnstakeCmd(cmdDef),
+		newUnpauseCmd(cmdDef),
+		{
```
Olshansk, line 47: Can you either implement or add a TODO to accept a payload from a file via `[--input_file]`. For example, see `cli/keys.go`.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+			Use:   "Relay <servicer> <application> <relayChainID> <payload>",
```
Olshansk, line 48: `s/payload/payloadSerializedProto`?

Please rename `payload` so its self explanatory on what we expect

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+			Short: "Relay <servicer> <application> <relayChainID> <payload>",
```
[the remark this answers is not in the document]
Olshansk: You will need to update the code as well if we reorder the two.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	hasher := sha.SHA3_256.New()
+	if _, err := hasher.Write(data); err != nil {
+		return nil, fmt.Errorf("Error hashing data: %w", err)
+	}
+
+	return hasher.Sum(nil), nil
+}
+
+// TODO: remove use of package-level variables
+func getPrivateKey(address string) (crypto.PrivateKey, error) {
```
Olshansk, line 265: s/getPrivateKey/getPrivateKeyFromKeybase

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+Will prompt the user for the *application* account passphrase`,
+			Aliases: []string{},
+			Args:    cobra.ExactArgs(4),
+			RunE: func(cmd *cobra.Command, args []string) error {
+				servicerAddr := args[0]
+				applicationAddr := args[1]
+				chain := args[2]
+				relayPayload := args[3]
+
+				// TODO: (SUGGESTION) refactor to decouple the client logic from the CLI/command
```
Olshansk, line 60: Can you update the comment so its clearer what part is the client and what's in the CLI. In general, I support the sentiment but don't have an understanding of the suggested changes.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+
+// fetchServicer returns the servicer specified by the <servicer> argument.
+// It validates the following conditions:
+//
+//	A. The <application> argument is the address of an active application
+//	B. The <servicer> is the address of a servicer that is active in the application's current session.
+//
+// TODO: (SUGGESTION) use a package-internal interface for servicer and application?
+// TODO: (SUGGESTION) use a struct as input to combine all fields (same for output)
+func fetchServicer(ctx context.Context, appAddress, chain, servicerAddress string) (rpc.Session, rpc.ProtocolActor, error) {
```
Olshansk, line 104: Why not return `*rpc.Session` so we can return `nil` if there's an error?

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	session, err := getCurrentSession(ctx, appAddress, chain)
+	if err != nil {
+		return rpc.Session{}, rpc.ProtocolActor{}, fmt.Errorf("Error getting servicer: %w", err)
+	}
+
+	var (
+		servicer rpc.ProtocolActor
+		found    bool
+	)
+	// TODO: a map may be a better choice for storing servicers
```
Olshansk, line 115: Can you move this TODO next to the appropriate field in `session.proto`?

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	for _, s := range session.Servicers {
+		if s.Address == servicerAddress {
+			servicer = s
+			found = true
+			break
+		}
+	}
+
+	// TODO: cover with unit tests
```
Olshansk, line 124: s/unit tests/gherkin tests

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	if !found {
+		return rpc.Session{}, rpc.ProtocolActor{}, fmt.Errorf("Error getting servicer: address %s does not match any servicers in the session", servicerAddress)
+	}
+
+	// TODO: cover with unit tests
```
Olshansk, line 129: You shouldn't need to do this. getCurrentSession accepts a chain, so the protocol wouldn't return a servicer unless it supports the chain for that height.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+// TODO: (QUESTION): do we need/want a cli subcommand for fetching servicers?
+
+// fetchServicer returns the servicer specified by the <servicer> argument.
+// It validates the following conditions:
+//
+//	A. The <application> argument is the address of an active application
+//	B. The <servicer> is the address of a servicer that is active in the application's current session.
+//
+// TODO: (SUGGESTION) use a package-internal interface for servicer and application?
+// TODO: (SUGGESTION) use a struct as input to combine all fields (same for output)
```
Olshansk, line 103: Personally I think it's overkill for this context. We pass individual parameters in many different places and I'd rather not overengineer unless we need to.

If we do end up needing this, it'll come up repeatedly, so I'd bias not to have an explicit TODO for it.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+
+// TODO: (QUESTION): do we need/want a cli subcommand for fetching servicers?
+
+// fetchServicer returns the servicer specified by the <servicer> argument.
+// It validates the following conditions:
+//
+//	A. The <application> argument is the address of an active application
+//	B. The <servicer> is the address of a servicer that is active in the application's current session.
+//
+// TODO: (SUGGESTION) use a package-internal interface for servicer and application?
```
Olshansk, line 102: We used to have many package specific interfaces for every type (e.g. the Validator interface in consensus, the Validator interface in persistence, etc) and had 6 different Validator interfaces at one point.

I'm open to either approach but would push back against this just because it got really confusing in the past.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+
+				return nil
+			},
+		},
+	}
+
+	return cmds
+}
+
+// TODO: (QUESTION): do we need/want a cli subcommand for fetching servicers?
```
Olshansk, line 94: 1. Is that not what retrieving the session does?
2. What would the interface look like?
3. If/when we need it something custom, let's add it.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+				chain := args[2]
+				relayPayload := args[3]
+
+				// TODO: (SUGGESTION) refactor to decouple the client logic from the CLI/command
+				pk, err := getPrivateKey(applicationAddr)
+				if err != nil {
+					return fmt.Errorf("error getting application's private key: %w", err)
+				}
+
+				session, servicer, err := fetchServicer(cmd.Context(), applicationAddr, chain, servicerAddr)
```
Olshansk, line 66: Thoughts on decoupling "getSession" and "getServicer" or getting the session in place and passing that to `fetchServicer`?

Feels weird to get back a session from a function called `fetchServicer`

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+// fetchServicer returns the servicer specified by the <servicer> argument.
+// It validates the following conditions:
+//
+//	A. The <application> argument is the address of an active application
+//	B. The <servicer> is the address of a servicer that is active in the application's current session.
+//
+// TODO: (SUGGESTION) use a package-internal interface for servicer and application?
+// TODO: (SUGGESTION) use a struct as input to combine all fields (same for output)
+func fetchServicer(ctx context.Context, appAddress, chain, servicerAddress string) (rpc.Session, rpc.ProtocolActor, error) {
+	// TECHDEBT: cache session data
```
Olshansk, line 105: I think this might event warrant creating a ticket. Then you can ` // TECHDEBT(#XXX): cache session data`

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+
+	if !found {
+		return rpc.Session{}, rpc.ProtocolActor{}, fmt.Errorf("Error getting servicer: service %s does not support chain %s", servicerAddress, chain)
+	}
+
+	return *session, servicer, nil
+}
+
+func getCurrentSession(ctx context.Context, appAddress, chain string) (*rpc.Session, error) {
+	// TODO: passing 0 as the height value to get the current session seems more optimal than this.
```
Olshansk, line 146:
```suggestion
	// CONSIDERATION: passing 0 as the height value to get the current session seems more optimal than this.
```

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	currentHeight, err := getCurrentHeight(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting current session: %w", err)
+	}
+
+	req := rpc.SessionRequest{
+		AppAddress: appAddress,
+		Chain:      chain,
+		// TODO: Geozone
```
Olshansk, line 155:
```suggestion
		// TODO(#697): Geozone
```

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+
+	if resp.JSON200 == nil {
+		return 0, fmt.Errorf("Error getting current height: Unexpected response %v", resp)
+	}
+
+	return resp.JSON200.Height, nil
+}
+
+// TODO: (localnet) Publish Servicer(s) Host and Port as env. vars in K8s: similar to Validators
+// TODO: (REFACTOR) should we move package-level variables (e.g. remoteCLIURL) to a cli object?
```
Olshansk, line 202: We have a `REFACTOR` tag (see Makefile). +1 to the refactor

cc @h5law for the idea

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+func sendTrustlessRelay(ctx context.Context, servicerUrl string, relay rpc.RelayRequest) (*rpc.PostV1ClientRelayResponse, error) {
```
Olshansk, line 203: See suggestion from linters

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+
+	relayMeta := rpc.RelayRequestMeta{
+		BlockHeight: session.SessionHeight,
+		// TODO: use Identifiable for Chain in Session (or string for Chain in Relay Meta)
+		Chain: rpc.Identifiable{
+			Id: session.Chain,
+		},
+		ServicerPubKey: servicer.PublicKey,
+		// TODO: Geozone
+		// TODO: Token
```
Olshansk, line 230: Remove token

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	relayMeta := rpc.RelayRequestMeta{
+		BlockHeight: session.SessionHeight,
+		// TODO: use Identifiable for Chain in Session (or string for Chain in Relay Meta)
+		Chain: rpc.Identifiable{
+			Id: session.Chain,
+		},
+		ServicerPubKey: servicer.PublicKey,
+		// TODO: Geozone
```
Olshansk, line 229: Add the same github issue related to geozones as what I pasted to above

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return client.PostV1ClientRelayWithResponse(ctx, relay)
+}
+
+// TODO: (NICE) allow reading the relay request from the command line arguments AND from a file
+func buildRelay(payload string, appPrivateKey crypto.PrivateKey, session rpc.Session, servicer rpc.ProtocolActor) (rpc.RelayRequest, error) {
+	// TECHDEBT: This is mostly COPIED from pocket-go: we should refactor pocket-go code and import this functionality from there instead.
```
Olshansk, line 214: Noted

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	relayPayload := rpc.Payload{
```
Olshansk, line 215: Just FYI. As you're working on this, can you also refactor `relay.proto` to enable other request types?

I was thinking something like

```proto
message Relay {
    oneof RelayPayload {
        JSONRPCPayload json_rpc_payload = 1;
        RESTPayload rest_payload = 2;
        GRPCPayload grpc_payload = 3;
        GraphQLPayload graphql_payload = 4;
        WebSocketsPayload websockets_payload = 5;
    }
	...
}
```

Olshansk (changes requested):

app/client/cli/servicer.go
adshmh: Thank you for the review. I will update the PR to return a pointer here to be consistent with the rest of the code-base.
My 2 cents is that we may want to consider passing copies rather than pointers in general, as it helps improve readability by making it clear that no modification is intended (and maybe as an added protection measure against accidental changes by a receiver of the value)

app/client/cli/servicer.go
adshmh: Thank you for the review. Removed.
The main reason for the suggestion was improving readability when looking at a call to a function like this that has the same type for all/most of the arguments. One thing that can make structs-as-input obsolete is if we end up using specific, separate types to identify a chain, an application, etc, as GO will enforce the function signature.

app/client/cli/servicer.go
adshmh: Thanks, makes sense. This was a TODO from the first time I read the code. Doesn't look necessary on a second/third pass.

app/client/cli/servicer.go
adshmh: > 1. Is that not what retrieving the session does?
> 
>     2. What would the interface look like?
> 
>     3. If/when we need it something custom, let's add it.

1. Yes, it is.
2. I think you answered this already somewhere else. The idea is whether we need/want to add a CLI command to fetch a session (with the main goal of getting a list of servicers), the output from which can be used to call the Trustless Relay command. I don't see any commands under `app/client` that performs this.

3. Sure, I'll just mark this as a nice-to-have.

app/client/cli/servicer.go
adshmh: Thank you for catching this. Yes, the `fetchServicer` signature looked strange. Fixed and it reads a lot better now IMO.

Olshansk:
@adshmh When this needs a review or follow up on the responses, please re-request it on GitHub so I get a notification.

Screenshot 2023-05-26 at 10 50 17 AM

app/client/cli/servicer.go
adshmh: Done. https://github.com/pokt-network/pocket/issues/791

app/client/cli/actor.go
```diff
@@ -169,6 +168,7 @@ func newEditStakeCmd(cmdDef actorCmdDef) *cobra.Command {
 				return err
 			}
 
+			// TODO: should skip if non-ineractive
```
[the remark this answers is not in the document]
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/pokt-network/pocket/rpc"
+	coreTypes "github.com/pokt-network/pocket/shared/core/types"
+	"github.com/pokt-network/pocket/shared/crypto"
+)
+
+func init() {
+	rootCmd.AddCommand(NewServicerCommand())
+}
+
+// TECHDEBT: (unittest) unit test the command: e.g. on number of arguments
```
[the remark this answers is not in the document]
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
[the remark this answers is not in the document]
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	signature, err := appPrivateKey.Sign(hashedReq)
+	if err != nil {
+		return relay, fmt.Errorf("Error signing relay: %w", err)
+	}
+	relay.Meta.Signature = hex.EncodeToString(signature)
+
+	return relay, nil
+}
+
+func hash(data []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,284 @@
...
+	hasher := sha.SHA3_256.New()
+	if _, err := hasher.Write(data); err != nil {
+		return nil, fmt.Errorf("Error hashing data: %w", err)
+	}
+
+	return hasher.Sum(nil), nil
+}
+
+// TODO: remove use of package-level variables
```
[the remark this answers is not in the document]
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

app/client/cli/servicer.go
adshmh: Thanks. Fixed.

shared/core/types/proto/relay.proto
```diff
@@ -6,24 +6,37 @@ option go_package = "github.com/pokt-network/pocket/shared/core/types";
...
-    map<string, string> headers = 4;
+message RESTPayload {
+    string contents = 1;
+}
+
+message JSONRPCPayload {
+    bytes id = 1;
+    string json_rpc = 2;
+    string method = 3;
+    bytes parameters = 4;
```
Olshansk, line 28: should this be `repeated bytes parameters`?

shared/core/types/proto/relay.proto
```diff
@@ -6,24 +6,37 @@ option go_package = "github.com/pokt-network/pocket/shared/core/types";
...
-    string method = 2;
-    string http_path = 3;
-    map<string, string> headers = 4;
+message RESTPayload {
+    string contents = 1;
+}
+
+message JSONRPCPayload {
+    bytes id = 1;
+    string json_rpc = 2;
```
Olshansk, line 26: `json_rpc_version`?

shared/core/types/proto/relay.proto
```diff
@@ -6,24 +6,37 @@ option go_package = "github.com/pokt-network/pocket/shared/core/types";
...
-    string data = 1;
-    string method = 2;
-    string http_path = 3;
-    map<string, string> headers = 4;
+message RESTPayload {
+    string contents = 1;
+}
+
+message JSONRPCPayload {
+    bytes id = 1;
```
Olshansk, line 25: What does the `id` represent?

shared/core/types/proto/relay.proto
```diff
@@ -6,24 +6,37 @@ option go_package = "github.com/pokt-network/pocket/shared/core/types";
...
+    string json_rpc = 2;
+    string method = 3;
+    bytes parameters = 4;
+    // DISCUSS: is HttpPath needed and is it a Meta field or Payload field?
```
Olshansk, line 29: I don't really see/understand how `http_path` would be used in json-rpc because the method captures the functionality we are calling.

Can you:
1. Move http_path from `JSONRPCPayload` to `RESTPayload`
2. Add an enum in `RESTPayload` for the request type; `GET/SET/DELETE/etc`

shared/core/types/relay.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+var (
+	errInvalidRelayInvalidPayload  = errors.New("invalid relay payload")
+	errInvalidJSONRPCInvalidRPC    = errors.New("invalid value for JSONRPC field")
```
Olshansk, line 11: NIT: Do we need the second `Invalid` in the naming of these variables? I think can be omitted.

shared/core/types/relay.go
```diff
@@ -0,0 +1,53 @@
+package types
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+var (
+	errInvalidRelayInvalidPayload  = errors.New("invalid relay payload")
```
Olshansk, line 10:
```suggestion
	// TODO: Once the proto structures are finalized, add an exhaustive list of errors and tests
	errInvalidRelayInvalidPayload  = errors.New("invalid relay payload")
```

shared/core/types/proto/relay.proto
```diff
@@ -6,24 +6,37 @@ option go_package = "github.com/pokt-network/pocket/shared/core/types";
 
 // TECHDEBT: These structures were copied as placeholders from v0 and need to be updated to reflect changes in v1
 message Relay {
-    RelayPayload payload = 1;
-    RelayMeta meta = 2;
+    RelayMeta meta = 1;
+    oneof RelayPayload {
```
Olshansk, line 10: Should we `snake_case` `RelayPayload` to stay consistent with other proto fields? I believe it gets auto camel-cased when converted to a go struct.

shared/core/types/relay.go
```diff
@@ -0,0 +1,53 @@
...
+	errInvalidJSONRPCInvalidRPC    = errors.New("invalid value for JSONRPC field")
+	errInvalidJSONRPCMissingMethod = errors.New("Method field not set")
+	errInvalidRESTPayload          = errors.New("invalid REST payload")
+)
+
+// INCOMPLETE: perform any possible metadata validation
+// Validate performs validation on the relay payload
+func (r Relay) Validate() error {
```
Olshansk, line 18: Question: Why not do a switch-case statement on `switch r.RelayPayload.(type) {`

shared/core/types/relay.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	return fmt.Errorf("%w: %v", errInvalidRelayInvalidPayload, r)
+}
+
+// Validate performs validation on JSONRPC payload. More specifically, it verifies that:
+//  1. The JSONRPC field is set to "2.0" as per the JSONRPC spec requirement, and
+//  2. The Method field is not empty
+func (p JSONRPCPayload) Validate() error {
+	if p.JsonRpc != "2.0" {
```
Olshansk, line 34: Consider moving "2.0" into a local constant

shared/core/types/relay.go
```diff
@@ -0,0 +1,53 @@
...
+		return fmt.Errorf("%w: %s", errInvalidJSONRPCInvalidRPC, p.JsonRpc)
+	}
+
+	// DISCUSS: do we need/want chain-specific validation? Potential for reusing the existing logic of Portal V2/pocket-go
```
Olshansk, line 38: We still haven't fully defined what relay chains will look like.

For example, what if they're `JWT` like tokens/strings that can hold more data to support general purpose RPC? `<Type>.<ID>.<Config>`. See slide 26 form https://docs.google.com/presentation/d/19NqPV0e6JbQzzwqEnfJ_WiMqpjaPwWo-qui6Xr3pNSQ/edit#slide=id.g23e8b9d22f1_0_134.

I think updating the TODO with some extra details and leaving as is should be good.

shared/core/types/relay.go
```diff
@@ -0,0 +1,53 @@
...
+		return errInvalidJSONRPCMissingMethod
+	}
+	return nil
+}
+
+// Validate verifies that the payload is valid REST, i.e. valid JSON
+func (p RESTPayload) Validate() error {
+	var parsed json.RawMessage
+	err := json.Unmarshal([]byte(p.Contents), &parsed)
+	if err != nil {
```
Olshansk, line 49:
```suggestion
	if err := json.Unmarshal([]byte(p.Contents), &parsed); err != nil {
```

shared/core/types/relay_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+func TestRelayValidate(t *testing.T) {
+	testCases := []struct {
+		name     string
+		relay    Relay
+		expected error
+	}{
+		{
+			name: "valid Relay: JSONRPC",
+			// IMPROVE: use a factory function to build test relays
```
Olshansk, line 16: I think this `IMPROVE` belongs in `relay.go` instead of test, no?

shared/core/types/relay_test.go
```diff
@@ -0,0 +1,137 @@
...
+			expected: errInvalidRESTPayload,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			err := testCase.relay.Validate()
+			switch {
+			case err == nil && testCase.expected != nil:
+				t.Fatalf("Expected error %v, got: nil", testCase.expected)
```
Olshansk, line 60: We use the `testify` package, which does automatic "expected vs got" type comments. https://pkg.go.dev/github.com/stretchr/testify/require#Error

You should be able to do:

```go
if testCase.expected == nil {
	require.NoError(t, err)
} else {
	require.ErrorIs(t, err, testCase.expected)
}
```

Ditto below

Olshansk (changes requested):
@adshmh Feel free to resolve / close out any comments where the change / suggestion is trivial or minor. There's a level of trust we got to have with others :)

Leave open comments where we need to continue the discussion.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	if resp.JSON200 == nil {
+		return 0, fmt.Errorf("Error getting current height: Unexpected response %v", resp)
+	}
+
+	return resp.JSON200.Height, nil
+}
+
+// IMPROVE: [K8s][LocalNet] Publish Servicer(s) Host and Port as env. vars in K8s: similar to Validators
```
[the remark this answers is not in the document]
adshmh: Done: #823

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+			Id: session.Chain,
+		},
+		ServicerPubKey: servicer.PublicKey,
+		// TODO(#697): Geozone
+	}
+
+	relay := &rpc.RelayRequest{
+		Payload: relayPayload,
+		Meta:    relayMeta,
+		// DISCUSS: why is there no Proof field in v1 struct?
```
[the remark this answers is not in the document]
adshmh: Thank you for the clarification. Fixed.

rpc/handlers.go
```diff
@@ -229,3 +206,37 @@ func (s *rpcServer) GetV1P2pStakedActorsAddressBook(ctx echo.Context, params Get
 
 	return ctx.JSON(http.StatusOK, response)
 }
+
+// TECHDEBT: handle other relay payload types
+func buildRelayPayload(body RelayRequest) *coreTypes.Relay {
```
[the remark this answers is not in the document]
adshmh: Thank you, fixed.

rpc/handlers.go
```diff
@@ -229,3 +206,37 @@ func (s *rpcServer) GetV1P2pStakedActorsAddressBook(ctx echo.Context, params Get
...
+		payload.JsonRpcPayload.HttpPath = *body.Payload.Path
+	}
+
+	if body.Payload.Parameters != nil {
+		payload.JsonRpcPayload.Parameters = *body.Payload.Parameters
+	}
+
+	if body.Payload.Headers != nil {
+		headers := make(map[string]string)
+		for _, header := range *body.Payload.Headers {
```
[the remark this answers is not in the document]
adshmh: This is needed, because `Headers` is a pointer to an array in the generated code.

rpc/handlers.go
```diff
@@ -229,3 +206,37 @@ func (s *rpcServer) GetV1P2pStakedActorsAddressBook(ctx echo.Context, params Get
...
+			headers[header.Name] = header.Value
```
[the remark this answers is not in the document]
adshmh: I double-checked and the `Headers` field is not being initialized to an empty map. Please correct me if this is not accurate.

shared/core/types/proto/relay.proto
adshmh: Not based on my understanding of the `parameters` field in [JSONRPC spec](https://www.jsonrpc.org/specification#parameter_structures). I have updated the proto file with a comment on this. Please let me know whether that addresses this review item.

shared/core/types/proto/relay.proto
adshmh: This is matching the `jsonrpc` field in [JSONRPC spec](https://www.jsonrpc.org/specification#request_object). I have updated the proto file with a comment on this. Please let me know whether that addresses this review item. We could still name this json_rpc_version and map to a different field name in serialization/deserialization, but that may be overkill considering how infrequent the version changes are.

shared/core/types/relay.go
adshmh: Thank you, fixed.

shared/core/types/relay.go
adshmh: Great suggestion, thank you. Fixed.

shared/core/types/relay.go
adshmh: Thank you, fixed.

shared/core/types/relay.go
adshmh: Thank you for catching this. Fixed.

shared/core/types/relay_test.go
adshmh: Yes it does. Thank you. Fixed.

shared/core/types/relay_test.go
adshmh: Thank you for the review. Fixed.

shared/core/types/relay.go
adshmh: Thank you for the review. Fixed.

shared/core/types/relay.go
adshmh: Thank you for the reminder. Fixed.

shared/core/types/proto/relay.proto
adshmh: Thank you for the review. Fixed (As a side-note, on AVAX chain there is currently a path being used on portal, but that is more of a work-around, and a node can just set its URL to include the path instead)

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+				applicationAddr := args[0]
+				servicerAddr := args[1]
+				chain := args[2]
+				relayPayload := args[3]
+
+				// REFACTOR: decouple the client logic from the CLI
+				//	The client will: send the trustless relay and return the response (using a single function as entrypoint)
+				//	The CLI will:
+				//		1) extract the required input from the command arguments
+				//		2) call the client function that performs the trustless relay
```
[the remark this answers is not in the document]
adshmh: Thanks, fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+				if err != nil {
+					return fmt.Errorf("error getting servicer for the relay: %w", err)
+				}
+
+				relay, err := buildRelay(relayPayload, pk, session, servicer)
+				if err != nil {
+					return fmt.Errorf("error building relay from payload: %w", err)
+				}
+
+				fmt.Printf("sending trustless relay for %s to %v with payload: %s\n", applicationAddr, servicer, relayPayload)
```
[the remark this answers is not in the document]
adshmh: Thank you, fixed.

shared/core/types/proto/session.proto
```diff
@@ -20,6 +20,7 @@ message Session {
     // CONSIDERATION: Should a single session support multiple geo zones?
     string geo_zone = 6; // the target geographic region where the actors are present
     core.Actor application = 7; // the application that is being served
+    // IMPROVE: a map may be a better choice for storing servicers/fishermen
```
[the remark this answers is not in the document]
adshmh: Thank you for the review. Will keep this in mind. Fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+	client, err := rpc.NewClientWithResponses(remoteCLIURL)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting current session for app/chain/height: %s/%s/%d: %w", appAddress, chain, currentHeight, err)
+	}
+
+	resp, err := client.PostV1ClientGetSessionWithResponse(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting current session with request %v: %w", req, err)
+	}
+	// TODO: refactor boiler-plate code
```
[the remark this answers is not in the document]
adshmh: It is a comment on processing HTTP response/codes. I updated the TODO comment to add more detail.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+
+				return nil
+			},
+		},
+	}
+
+	return cmds
+}
+
+// DECIDE: do we need/want a cli subcommand for fetching sessions?
```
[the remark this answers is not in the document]
adshmh: Thank you, fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+// validateServicer returns the servicer specified by the <servicer> argument.
+// It validates that the <servicer> is the address of a servicer that is active in the passed session.
```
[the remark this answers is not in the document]
adshmh: Thanks, fixed.

shared/core/types/proto/relay.proto
```diff
@@ -6,24 +6,37 @@ option go_package = "github.com/pokt-network/pocket/shared/core/types";
 
 // TECHDEBT: These structures were copied as placeholders from v0 and need to be updated to reflect changes in v1
```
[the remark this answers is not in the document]
adshmh: Thanks, fixed.

rpc/v1/openapi.yaml
```diff
@@ -123,11 +123,12 @@ paths:
             text/plain:
               example: "description of failure"
   # TODO: Update this handler and its schemas when the HandleRelay function has been implemented
+  # DISCUSS: do we need separate JSON and JSONRPC endpoints?
```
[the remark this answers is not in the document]
adshmh: Thanks, fixed.

shared/core/types/proto/relay.proto
```diff
@@ -6,24 +6,37 @@ option go_package = "github.com/pokt-network/pocket/shared/core/types";
...
 }
 
 message RelayMeta {
     int64 block_height = 1;
     string servicer_public_key = 2;
     Identifiable relay_chain = 3;
     Identifiable geo_zone = 4;
-    AAT token = 5;
-    string signature = 6;
+    string signature = 5;
```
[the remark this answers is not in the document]
adshmh: Thanks, fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+
+	client, err := rpc.NewClientWithResponses(remoteCLIURL)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting current session for app/chain/height: %s/%s/%d: %w", appAddress, chain, currentHeight, err)
+	}
+
+	resp, err := client.PostV1ClientGetSessionWithResponse(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting current session with request %v: %w", req, err)
+	}
```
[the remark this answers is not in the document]
adshmh: Thanks, fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+
+	if resp.JSON200 == nil {
+		return nil, fmt.Errorf("Error getting current session: Unexpected response %v", resp)
+	}
+
+	return resp.JSON200, nil
+}
+
+// TODO: reuse this function in the query commands
```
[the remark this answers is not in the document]
adshmh: Thanks, fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+
+	if resp.JSON200 == nil {
+		return 0, fmt.Errorf("Error getting current height: Unexpected response %v", resp)
+	}
+
+	return resp.JSON200.Height, nil
+}
+
+// IMPROVE: [K8s][LocalNet] Publish Servicer(s) Host and Port as env. vars in K8s: similar to Validators
+// REFACTOR: move package-level variables (e.g. remoteCLIURL) to a cli object.
```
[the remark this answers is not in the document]
adshmh: Thank you, fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+		ServicerPubKey: servicer.PublicKey,
+		// TODO(#697): Geozone
+	}
+
+	relay := &rpc.RelayRequest{
+		Payload: relayPayload,
+		Meta:    relayMeta,
+		// DISCUSS: why is there no Proof field in v1 struct?
+	}
+	reqBytes, err := json.Marshal(relay)
```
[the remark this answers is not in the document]
adshmh: Thank you, fixed.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+	signature, err := appPrivateKey.Sign(hashedReq)
+	if err != nil {
+		return relay, fmt.Errorf("Error signing relay: %w", err)
+	}
+	relay.Meta.Signature = hex.EncodeToString(signature)
+
+	return relay, nil
+}
+
+// TECHDEBT: remove use of package-level variables
```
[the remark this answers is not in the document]
adshmh: Thank you, fixed.

rpc/handlers.go
```diff
@@ -229,3 +206,37 @@ func (s *rpcServer) GetV1P2pStakedActorsAddressBook(ctx echo.Context, params Get
...
+			JsonRpc: body.Payload.Jsonrpc,
+			Method:  body.Payload.Method,
+		},
+	}
+
+	if body.Payload.Id != nil {
+		payload.JsonRpcPayload.Id = []byte(*body.Payload.Id)
+	}
+
+	if body.Payload.Path != nil {
```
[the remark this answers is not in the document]
adshmh: Marking this as unresolved to make sure this is acceptable.

adshmh:
> @adshmh There are still some comments that:
> 
>     1. You did not tend to
> 
>     2. You mentioned are resolved but are not
> 
> 
> Can you please:
> 
>     1. Double check and resolve all outstanding comments
> 
>     2. Click "Resolve" once you have tended to a comment
> 
>     3. Push the latest changes
> 
> 
> Re-request review once the above are taken care of.
> 
> Screenshot 2023-06-14 at 2 07 09 PM
> 
> Screenshot 2023-06-14 at 2 06 37 PM

Thank you for the reminder. All review comments are addressed now, except for items that may need further discussion.

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+
+				return nil
+			},
+		},
+	}
+
+	return cmds
+}
+
+// TODO: add a cli command for fetching sessions
```
Olshansk, line 103: I think I already asked this but why is `/v1/client/get_session:` insufficient?

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Error getting current session for app/chain/height: %s/%s/%d: %w", appAddress, chain, currentHeight, err)
+	}
+
+	resp, err := client.PostV1ClientGetSessionWithResponse(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting current session with request %v: %w", req, err)
+	}
+
+	// IMPROVE: move the HTTP response processing code to a separate function to enable reuse.
```
Olshansk, line 141:
```suggestion
	// CLEANUP: move the HTTP response processing code to a separate function to enable reuse.
```

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	if resp.JSON200 == nil {
+		return 0, fmt.Errorf("Error getting current height: Unexpected response %v", resp)
+	}
+
+	return resp.JSON200.Height, nil
+}
+
+// IMPROVE: [K8s][LocalNet] Publish Servicer(s) Host and Port as env. vars in K8s: similar to Validators: #823
```
Olshansk, line 174:
```suggestion
// IMPROVE(#823): [K8s][LocalNet] Publish Servicer(s) Host and Port as env. vars in K8s: similar to Validators
```

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+// REFACTOR: move package-level variables (e.g. remoteCLIURL) to a cli object and consider storing it in the context
```
Olshansk, line 175:
```suggestion
// CONSIDERATION: move package-level variables (e.g. remoteCLIURL) to a cli object and consider storing it in the context
```

@bryanchriswhite recently refactored it to consolidate and export them from on eplace

app/client/cli/servicer.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, fmt.Errorf("Error marshalling relay request %v: %w", relay, err)
+	}
+	hashedReq := crypto.SHA3Hash(reqBytes)
+	signature, err := appPrivateKey.Sign(hashedReq)
+	if err != nil {
+		return relay, fmt.Errorf("Error signing relay: %w", err)
+	}
+	relay.Meta.Signature = hex.EncodeToString(signature)
+
+	return relay, nil
```
Olshansk, line 220: I'm pretty sure we're going to hit problems/issues with this function, but we'll deal with it when we hit it in the E2E test

rpc/v1/openapi.yaml
```diff
@@ -127,9 +127,9 @@ paths:
     post:
       tags:
         - client
-      summary: Sends a relay to the servicer to receive a response
+      summary: Sends a relay to the servicer to receive a response. The payload is the serialized form of one of the supported formats.
```
Olshansk, line 130:
```suggestion
      summary: Sends a relay to the servicer to receive a response. The payload is the serialized form of one of the supported formats in a protobuf.
```

shared/core/types/relay_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			err := testCase.payload.Validate()
+			require.ErrorIs(t, err, testCase.expected)
+		})
+	}
+}
+
+func TestRESTValidate(t *testing.T) {
```
Olshansk, line 93: NIT: Can you update all of these to `TestRelay_ValidateREST`, `TestRelay_ValidJSONRPC`, etc...so we can run them with `go test -run TestRelay`

Olshansk (approved):
@adshmh Just a few minor comments, but approving regardless.

Please see & then to them and then feel free to squash & merge.

app/client/cli/servicer.go
adshmh: Yes, there was a [review comment](https://github.com/pokt-network/pocket/pull/778#discussion_r1210830146) on this. Technically the `/v1/client/get_session` is sufficient in terms of actually having access to the functionality. Like the rest of the CLI, it is a matter ease of use. Considering we will be addressing #791 soon, i.e. there will be a client-side cache for sessions, adding a command for fetching sessions makes sense IMO.

app/client/cli/servicer.go
adshmh: Thanks, fixed.

app/client/cli/servicer.go
adshmh: Thanks, fixed.

app/client/cli/servicer.go
adshmh: Noted. Work on E2E tests will start shortly.

rpc/v1/openapi.yaml
adshmh: Thanks, fixed.

shared/core/types/relay_test.go
adshmh: Great suggestion, thank you. Fixed.

Olshansk:
Using my admin powers to bypass CI checks due to the following reasons:

Screenshot 2023-06-20 at 2 23 10 PM

Olshansk (approved):

Olshansk:
Tests were confirmed to pass. COdecov is being disabled. Ggshield issue is being handled in #848. 

I will merged this in.

PR #804, opened 2023-06-05
[P2P] refactor: peerstore provider

bryanchriswhite:
## Issue

Related to #505 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- ...

## Testing

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/p2p/providers/peerstore_provider/persistence/p2p.go
+++ b/p2p/providers/peerstore_provider/persistence/p2p.go
@@ -1,0 +1,72 @@
+package persistence
+
+import (
+	"fmt"
+
+	"github.com/pokt-network/pocket/p2p/providers/peerstore_provider"
+	typesP2P "github.com/pokt-network/pocket/p2p/types"
+	"github.com/pokt-network/pocket/shared/modules"
+	"github.com/pokt-network/pocket/shared/modules/base_modules"
+)
+
+var (
+	_ peerstore_provider.PeerstoreProvider = &p2pPeerstoreProvider{}
+	_ p2pPStoreProviderFactory             = &p2pPeerstoreProvider{}
+)
+
+type p2pPStoreProviderFactory = modules.FactoryWithConfig[peerstore_provider.PeerstoreProvider, modules.P2PModule]
+
+type p2pPeerstoreProvider struct {
+	base_modules.IntegratableModule
+	persistencePeerstoreProvider
+
+	p2pModule modules.P2PModule
+}
+
+// TODO_THIS_COMMIT: refactor/move (?)
+type UnstakedPeerstoreProvider interface {
+	GetUnstakedPeerstore() (typesP2P.Peerstore, error)
+}
+
+func NewP2PPeerstoreProvider(
+	bus modules.Bus,
+	p2pModule modules.P2PModule,
+) (peerstore_provider.PeerstoreProvider, error) {
+	return new(p2pPeerstoreProvider).Create(bus, p2pModule)
+}
+
+func (*p2pPeerstoreProvider) Create(
+	bus modules.Bus,
+	p2pModule modules.P2PModule,
+) (peerstore_provider.PeerstoreProvider, error) {
+	if bus == nil {
+		return nil, fmt.Errorf("bus is required")
+	}
+
+	if p2pModule == nil {
+		return nil, fmt.Errorf("p2p module is required")
+	}
+
+	p2pPSP := &p2pPeerstoreProvider{
+		IntegratableModule: *base_modules.NewIntegratableModule(bus),
+		p2pModule:          p2pModule,
+	}
+
+	return p2pPSP, nil
+}
+
+func (p2pPSP *p2pPeerstoreProvider) Start() error {
+	return nil
+}
+
+func (*p2pPeerstoreProvider) GetModuleName() string {
+	return peerstore_provider.ModuleName
+}
+
+func (p2pPSP *p2pPeerstoreProvider) GetUnstakedPeerstore() (typesP2P.Peerstore, error) {
+	unstakedPStoreProvider, ok := p2pPSP.p2pModule.(UnstakedPeerstoreProvider)
+	if !ok {
+		return nil, fmt.Errorf("p2p module does not implement UnstakedPeerstoreProvider")
+	}
+	return unstakedPStoreProvider.GetUnstakedPeerstore()
+}
--- a/p2p/providers/peerstore_provider/rpc/provider.go
+++ b/p2p/providers/peerstore_provider/rpc/provider.go
@@ -101,6 +101,10 @@ func (rabp *rpcPeerstoreProvider) GetP2PConfig() *configs.P2PConfig {
 	return rabp.p2pCfg
 }
 
+func (rabp *rpcPeerstoreProvider) GetUnstakedPeerstore() (typesP2P.Peerstore, error) {
+	return nil, fmt.Errorf("unstaked peerstore not supported by rpc peerstore provider")
+}
+
 func (rabp *rpcPeerstoreProvider) initRPCClient() {
 	rpcClient, err := rpc.NewClientWithResponses(rabp.rpcURL)
 	if err != nil {
--- a/p2p/providers/peerstore_provider/peerstore_provider.go
+++ b/p2p/providers/peerstore_provider/peerstore_provider.go
@@ -5,7 +5,6 @@ package peerstore_provider
 import (
 	"github.com/pokt-network/pocket/logger"
 	typesP2P "github.com/pokt-network/pocket/p2p/types"
-	"github.com/pokt-network/pocket/runtime/configs"
 	coreTypes "github.com/pokt-network/pocket/shared/core/types"
 	cryptoPocket "github.com/pokt-network/pocket/shared/crypto"
 	"github.com/pokt-network/pocket/shared/modules"
@@ -19,7 +18,6 @@ type PeerstoreProvider interface {
 	modules.IntegratableModule
 
 	GetStakedPeerstoreAtHeight(height uint64) (typesP2P.Peerstore, error)
-	GetP2PConfig() *configs.P2PConfig
 }
 
 func ActorsToPeerstore(abp PeerstoreProvider, actors []*coreTypes.Actor) (pstore typesP2P.Peerstore, errs error) {
--- a/p2p/providers/peerstore_provider/peerstore_provider.go
+++ b/p2p/providers/peerstore_provider/peerstore_provider.go
@@ -16,7 +16,7 @@ const ModuleName = "peerstore_provider"
 
 // PeerstoreProvider is an interface that provides Peerstore accessors
 type PeerstoreProvider interface {
-	modules.Module
+	modules.IntegratableModule
 
 	GetStakedPeerstoreAtHeight(height uint64) (typesP2P.Peerstore, error)
 	GetP2PConfig() *configs.P2PConfig
--- a/p2p/providers/peerstore_provider/peerstore_provider.go
+++ b/p2p/providers/peerstore_provider/peerstore_provider.go
@@ -18,6 +18,7 @@ type PeerstoreProvider interface {
 	modules.IntegratableModule
 
 	GetStakedPeerstoreAtHeight(height uint64) (typesP2P.Peerstore, error)
+	GetUnstakedPeerstore() (typesP2P.Peerstore, error)
 }
 
 func ActorsToPeerstore(abp PeerstoreProvider, actors []*coreTypes.Actor) (pstore typesP2P.Peerstore, errs error) {
--- a/p2p/providers/peerstore_provider/persistence/persistence.go
+++ b/p2p/providers/peerstore_provider/persistence/persistence.go
@@ -1,6 +1,8 @@
 package persistence
 
 import (
+	"fmt"
+
 	"github.com/pokt-network/pocket/p2p/providers/peerstore_provider"
 	typesP2P "github.com/pokt-network/pocket/p2p/types"
 	"github.com/pokt-network/pocket/runtime/configs"
@@ -15,6 +17,7 @@ type persistencePeerstoreProvider struct {
 	base_modules.InterruptableModule
 }
 
+// TECHDEBT: refactor
 func NewPersistencePeerstoreProvider(bus modules.Bus, options ...func(*persistencePeerstoreProvider)) *persistencePeerstoreProvider {
 	pabp := &persistencePeerstoreProvider{
 		IntegratableModule: *base_modules.NewIntegratableModule(bus),
@@ -27,10 +30,12 @@ func NewPersistencePeerstoreProvider(bus modules.Bus, options ...func(*persisten
 	return pabp
 }
 
+// TECHDEBT: remove
 func Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, error) {
 	return new(persistencePeerstoreProvider).Create(bus, options...)
 }
 
+// TECHDEBT: refactor
 func (*persistencePeerstoreProvider) Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, error) {
 	return NewPersistencePeerstoreProvider(bus), nil
 }
@@ -55,4 +60,8 @@ func (pabp *persistencePeerstoreProvider) GetStakedPeerstoreAtHeight(height uint
 
 func (pabp *persistencePeerstoreProvider) GetP2PConfig() *configs.P2PConfig {
 	return pabp.GetBus().GetRuntimeMgr().GetConfig().P2P
+}
+
+func (pabp *persistencePeerstoreProvider) GetUnstakedPeerstore() (typesP2P.Peerstore, error) {
+	return nil, fmt.Errorf("persistence peerstore provider does not support unstaked peerstore")
 }


bryanchriswhite:
I threw together a mindmap of my initialthinking, which was mainly focusd on whether this new API should be a debug-only thing (as it's a "debug CLI" scenario). TLDR; I think "no" is the answer I landed on (plus some additional thinking):

```mermaid
mindmap
Q: should p2p peerstore provider be debug only?
    NO
        REASONING
            Will likely fit better with future versions / refactors
                r1(**Perhaps it would and/or already does make sense to split PeerstoreProvider into StakedPeerstoreProvider and UnstakedPeerstoreProvider)
        CONSEQUENCES
            c1[MUST design a non-debug way to access p2p module's peerstores]
                eg[e.g.: add GetUnstakedPeerstore method to modules.P2PModule]
                    Does this fight against peerstore provider / modular design?
                        YES because...
                            p[provides multiple ways to access the same thing: un/staked-actor-router peerstore]
                        no[**NO because...**]
                            allows any module to get either peerstore without having to know anything about p2p module
                            no other single module can support this interface
                                justified by the need in the CLI
                    Alternatives
                        ?
    YES
        REASONING
            r1[Only currently needed in a low-level / debug scenarios]
        CONSEQUENCES
            c1[Simpler implementation; no need for PeerstoreProvider interface refactor #804]
            c2[Harder to maintain and more complex to reason about; would likely rely on build tags]
```

** Here's what splitting `PeerstoreProvider` into `StakedPeerstoreProvider` and `UnstakedPeerstoreProvider` would look like:

```mermaid
classDiagram 

class IntegratableModule {
    <<interface>>
    +SetBus(bus Bus)
    +GetBus() Bus
}

class PeerstoreProvider {
    <<interface>>
    
}

class StakedPeerstoreProvider {
    <<interface>>
    +GetStakedPeerstoreAtHeight(height int) (Peerstore, error)
}

class UnstakedPeerstoreProvider {
    <<interface>>
    +GetUnstakedPeerstore() (Peerstore, error)
}

class persistencePeerstoreProvider
class rpcPeerstoreProvider
class p2pPeerstoreProvider {
  - persistencePeerstoreProvider
  - p2pModule P2PModule
}
class p2pModule

p2pPeerstoreProvider --o p2pModule
p2pPeerstoreProvider --* persistencePeerstoreProvider
p2pPeerstoreProvider --|> PeerstoreProvider

rpcPeerstoreProvider --|> StakedPeerstoreProvider
persistencePeerstoreProvider --|> StakedPeerstoreProvider

PeerstoreProvider --* StakedPeerstoreProvider
PeerstoreProvider --* UnstakedPeerstoreProvider

PeerstoreProvider --|> IntegratableModule
StakedPeerstoreProvider --|> IntegratableModule
UnstakedPeerstoreProvider --|> IntegratableModule
```

p2p/providers/peerstore_provider/persistence/p2p.go
```diff
@@ -0,0 +1,72 @@
...
+func (*p2pPeerstoreProvider) Create(
+	bus modules.Bus,
+) (peerstore_provider.PeerstoreProvider, error) {
+	if bus == nil {
+		return nil, fmt.Errorf("bus is required")
+	}
+
+	p2pModule := bus.GetP2PModule()
+	if p2pModule == nil {
+		return nil, fmt.Errorf("p2p module is not registered to bus and is required")
```
dylanlott, line 51: Nit: consider passing the bus to the Errrof for debugging purposes.

shared/modules/factory.go
```diff
@@ -4,23 +4,29 @@ package modules
...
 // and an error.
 // TECHDEBT: apply enforcement across applicable "sub-modules"
 type FactoryWithOptions[T interface{}, O interface{}] interface {
 	Create(bus Bus, opts ...O) (T, error)
 }
 
-// FactoryWithConfigAndOptions implements a `#Create()` factory method which
-// takes both a required "config" argument of type K and a variadic "optional"
+// FactoryWithConfigAndOptions implements a `#Create()` factory method which takes
+// a bus and both a required "config" argument of type K and a variadic "optional"
```
dylanlott, line 29: Nit: I think this should read as `a bus, a required "config", and a variadic "optional"`

shared/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.0.60] - 2023-06-05
+
+- Added `Factory` generic type
```
dylanlott, line 12: 💯 Great spot for a good generic 🚀

shared/modules/factory.go
```diff
@@ -4,23 +4,29 @@ package modules
 // a variadic `ModuleOption` argument(s) and returns a `Module`and an error.
 type ModuleFactoryWithOptions FactoryWithOptions[Module, ModuleOption]
 
-// FactoryWithConfig implements a `#Create()` factory method which takes a
-// required "config" argument of type K and returns a value of type T and an error.
+// Factory implements a `#Create()` factory method which takes a bus and returns
+// a value of type T and an error.
+type Factory[T interface{}] interface {
```
dylanlott, line 9: Would this be better declared as an any instead of an interface? I think they're functionally identical, but any seems more appropriate in the context of generics.

dylanlott (commented):
First round feedback, waiting for E2E test results 👍

p2p/providers/peerstore_provider/persistence/p2p.go
bryanchriswhite: The bus doesn't implement `Stringer`, what do you expect/want the DX to look/feel like?

shared/modules/factory.go
bryanchriswhite: Thanks for the reminder, I keep forgetting thats a thing! :raised_hands:

bryanchriswhite: Thanks for the reminder, I keep forgetting thats a thing! :raised_hands:

p2p/providers/peerstore_provider/persistence/p2p.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/pokt-network/pocket/shared/modules/base_modules"
+)
+
+var (
+	_ peerstore_provider.PeerstoreProvider = &p2pPeerstoreProvider{}
+	_ p2pPStoreProviderFactory             = &p2pPeerstoreProvider{}
+)
+
+// unstakedPeerstoreProvider is an interface which the p2p module supports in
+// order to allow access to the unstaked-actor-router's peerstore
```
Olshansk, line 18:
```suggestion
// order to allow access to the unstaked-actor-router's (i.e. actors, full and light nodes) peerstore
```

shared/modules/p2p_module.go
```diff
@@ -14,6 +14,8 @@ type P2PModule interface {
 
 	// Returns the public P2P address of this node
 	GetAddress() (cryptoPocket.Address, error)
+	// TECHDEBT(#xxx): uncomment after moving `typesP2P.Peerstore` interface to a shared package
```
Olshansk, line 17: Can you create a ticket to track this and update all the #xxx?

shared/modules/factory.go
```diff
@@ -4,25 +4,31 @@ package modules
...
 
-// FactoryWithConfig implements a `#Create()` factory method which takes a
-// required "config" argument of type K and returns a value of type T and an error.
+// Factory implements a `#Create()` factory method which takes a bus and returns
+// a value of type T and an error.
+type Factory[T any] interface {
+	Create(bus Bus) (T, error)
+}
+
+// FactoryWithConfig implements a `#Create()` factory method which takes a bus and
```
Olshansk, line 13: NIT: Thoughts on `s/T/M/g` and `s/K/C/g/` to represent `M`odule and `C`onfig respectively

shared/modules/factory.go
```diff
@@ -4,25 +4,31 @@ package modules
 // a variadic `ModuleOption` argument(s) and returns a `Module`and an error.
 type ModuleFactoryWithOptions FactoryWithOptions[Module, ModuleOption]
```
Olshansk, line 5: Should we add other types like `type ModuleFactory Factory[Module]` as well?

p2p/providers/peerstore_provider/persistence/p2p.go
```diff
@@ -0,0 +1,72 @@
...
+//
+// TECHDEBT(#xxx): will become unnecessary after `modules.P2PModule#GetUnstakedPeerstore` is added.`
+// CONSIDERATION: split `PeerstoreProvider` into `StakedPeerstoreProvider` and `UnstakedPeerstoreProvider`.
```
Olshansk, line 21: I didn't have time to dive into your mind map (too many other reviews right now) but at a high-level, getting access to staked & unstaked peers should not be debug-only IMO.

p2p/providers/peerstore_provider/persistence/p2p.go
```diff
@@ -0,0 +1,72 @@
...
+	GetUnstakedPeerstore() (typesP2P.Peerstore, error)
+}
+
+type p2pPStoreProviderFactory = modules.Factory[peerstore_provider.PeerstoreProvider]
+
+type p2pPeerstoreProvider struct {
+	base_modules.IntegratableModule
+	persistencePeerstoreProvider
+
+	p2pModule modules.P2PModule
```
Olshansk, line 33: Storing a reference to `p2pModule` feels like an antipattern. You should always be able to access it via the bus and everyone stores a reference to the bus.

p2p/providers/peerstore_provider/persistence/p2p.go
```diff
@@ -0,0 +1,72 @@
...
+// order to allow access to the unstaked-actor-router's peerstore
+//
+// TECHDEBT(#xxx): will become unnecessary after `modules.P2PModule#GetUnstakedPeerstore` is added.`
+// CONSIDERATION: split `PeerstoreProvider` into `StakedPeerstoreProvider` and `UnstakedPeerstoreProvider`.
+// (see: https://github.com/pokt-network/pocket/pull/804#issuecomment-1576531916)
+type unstakedPeerstoreProvider interface {
+	GetUnstakedPeerstore() (typesP2P.Peerstore, error)
+}
+
+type p2pPStoreProviderFactory = modules.Factory[peerstore_provider.PeerstoreProvider]
```
Olshansk, line 27: Why not embed this in the struct as well?

p2p/providers/peerstore_provider/persistence/p2p.go
```diff
@@ -0,0 +1,72 @@
...
+
+type p2pPeerstoreProvider struct {
+	base_modules.IntegratableModule
+	persistencePeerstoreProvider
+
+	p2pModule modules.P2PModule
+}
+
+func NewP2PPeerstoreProvider(
```
Olshansk, line 36: Optional NIT: Personally, I think the multilined method header here makes it harder to read (given only one parameter), but it's such a subjective preference feel free to ignore.

p2p/providers/peerstore_provider/persistence/p2p.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	p2pModule := bus.GetP2PModule()
+	if p2pModule == nil {
+		return nil, fmt.Errorf("p2p module is not registered to bus and is required")
+	}
+
+	p2pPSP := &p2pPeerstoreProvider{
+		IntegratableModule: *base_modules.NewIntegratableModule(bus),
+		p2pModule:          p2pModule,
```
Olshansk, line 56: Note above: I don't think we should be doing this.

Olshansk (changes requested):

p2p/providers/peerstore_provider/persistence/p2p.go
bryanchriswhite: :100: nice simplification, thanks!

shared/modules/factory.go
bryanchriswhite: :thinking: I could imagine an argument for either case but without being exhaustive here, I think it's nice that all modules conform to the `ModuleFactoryWithOptions` interface so that it can be embedded in `Module` (and therefore consistent between modules). Additionally, for any module which has a required config, it should load that from the base config file (unless I'm mistaken, I believe every module currently does this). This eliminates the possibility of needing a `ModuleFactoryWithConfig` or `ModuleFactoryWithConfigAndOptions`, and since the variadic arg on `ModuleFactoryWithOptions` can be omitted, the DevEx would be equivalent on the consumer side as compared to using a `ModuleFactory` interface like you suggest. The only difference then would be on the module side, we're basically requiring that each module accept and optionally consider a slice of `ModuleOption`s. I think it's also more prudent to be more forward compatible at the module-level (as compared with submodules), which makes me prefer this consistency accross modules.

shared/modules/factory.go
bryanchriswhite: I'm in.

p2p/providers/peerstore_provider/persistence/p2p.go
bryanchriswhite: So I know this instance is no longer applicable as I refactored this type out of existence in future commits but I'm noticing something that will continue to apply to the `persistencePeerstoreProvider` which was just me being confused initially.

As this is an alias to an interface type with the intention of being used for a compile-time check, it doesn't make sense for it to be embedded as that have the same effect. I had a conversation with chatGPT about this, here's the summary:

> Embedding an interface in a struct is a statement of intent that the struct should satisfy the interface, but it doesn't provide compile-time safety.
>
> The var _ SomeInterface = &myStruct{} check provides compile-time safety by ensuring that myStruct does indeed satisfy SomeInterface.

shared/modules/factory.go
bryanchriswhite: To add one more bit of detail; unlike `Module`s, requiring all submodules to accept and optionally consider a slice of options adds more boilerplace in the context of a submodule because each submodule will have to define its own option type for use with the `Factory...` generic type params, which it then may ignore.

p2p/providers/peerstore_provider/rpc/provider.go
```diff
@@ -37,6 +38,8 @@ type rpcPeerstoreProvider struct {
 	rpcClient *rpc.ClientWithResponses
 }
 
+// TECHDEBT(#810): refactor to be consistent with `persistencePeerstoreProvider`
+// (i.e. `NewRPCPeerstoreProvider` calls `rpcPeerstoreProvider#Create()`.
```
Olshansk, line 42: Given that we have `func Create` in our factory functions, I think all `func New` should be deprecated & removed.

p2p/providers/peerstore_provider/persistence/provider.go
```diff
@@ -0,0 +1,59 @@
+package persistence
```
Olshansk: Please try to do `mv` in the future when possible. Seeing a file deleted and a new file created requires the reviewer to identify what the diff was or simply review all the business logic from scratch which takes a non-trivial amount of effort.

Olshansk (changes requested):

shared/modules/factory.go
bryanchriswhite: > > . Additionally, for any module which has a required config, it should load that from the base config file (unless I'm mistaken, I believe every module currently does this).
> 
> It "should", but there is no enforcement, correct? Not a request to change it, just confirming.

Correct. A module may or may not use any part of the base config via the runtime manager, via the bus but there is no enforcement that any particular module consults any particular part of the config (or doesn't).

> > I think it's also prudent to be more forward compatible at the module-level (as compared with submodules), which makes me prefer this consistency accross modules.
> 
> In #810, can you talk more about the need to differentiate modules and submodules? The only difference I currently see is where the interface is defined (shared or non-shared) so am thinking that an internal type cast might be able to consolidate the business logic.

:thinking: The main hangup I have is forcing submodules to conform to the `ModuleFactoryWithOptions`, this is what causes the need for the additional boilerplate around submodule dependency injection. Perhaps this could be addressed simply by removing the `Create()` method from `InitializableModule`  and defining a new `Submodule` interface type which doesn't embed `ModuleFactoryWithOptions` (nor `InterruptableModule`):

```go
type InitializableModule interface {
	// GetModuleName returns the name of the module.
	GetModuleName() string
}

type Module interface {
	InitializableModule
	IntegratableModule
	InterruptableModule
	ModuleFactoryWithOptions
}

type Submodule interface {
  InitializableModule
  IntegratableModule
}
```

Then `Submodule` can be enforced on the submodule alongside its local "factory" type. Wdyt?

p2p/providers/peerstore_provider/persistence/provider.go
bryanchriswhite: :+1: I will make an effort to stage & commit file renames / moves independently of changes within the file; sometimes it's not obvious that such a change is upon you until it is. I think that's likely more representative of what happened in this case as I use my IDE to rename/move files and very rarely do `mv` or `git mv` in a terminal.

shared/modules/factory.go
bryanchriswhite: I updated #810 to include the `Submodule` interface now that you got me to think about it a bit more. I also updated/added some deliverables.

p2p/providers/peerstore_provider/persistence/provider.go
```diff
@@ -0,0 +1,59 @@
...
+	_ persistencePStoreProviderFactory     = &persistencePeerstoreProvider{}
+)
+
+type persistencePStoreProviderOption func(*persistencePeerstoreProvider)
+type persistencePStoreProviderFactory = modules.FactoryWithOptions[peerstore_provider.PeerstoreProvider, persistencePStoreProviderOption]
+type persistencePeerstoreProvider struct {
+	base_modules.IntegratableModule
+}
+
+func NewPersistencePeerstoreProvider(bus modules.Bus, options ...persistencePStoreProviderOption) (peerstore_provider.PeerstoreProvider, error) {
```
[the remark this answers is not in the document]
bryanchriswhite: :+1: I added a comment above the struct.

shared/modules/factory.go
Olshansk: Closing out this comment in favor to continue the conversation in #810

p2p/providers/peerstore_provider/persistence/provider.go
```diff
@@ -3,19 +3,28 @@ package persistence
...
 	"github.com/pokt-network/pocket/shared/modules/base_modules"
 )
 
-var _ peerstore_provider.PeerstoreProvider = &persistencePeerstoreProvider{}
+var (
+	_ peerstore_provider.PeerstoreProvider = &persistencePeerstoreProvider{}
+	_ persistencePStoreProviderFactory     = &persistencePeerstoreProvider{}
+)
+
+type persistencePStoreProviderOption func(*persistencePeerstoreProvider)
```
Olshansk, line 15: On the topic of "not being afraid to change everything", take into consideration that maybe "providers" aren't the best approach with our bus based system.

Not making a comment on removing/extending them, but just sharing that it is not set in stone so don't feel constrained.

Olshansk (approved):

PR #756, opened 2023-05-16
[Persistence] TreeStore Refactor

dylanlott:
## Description

Refactors the stateTrees component behind the TreeStore interface.

## Issue

Part of a refactor effort towards implementing #562

Part of a series of refactors in the persistence package, with #736 being the first.

## Type of change

Please mark the relevant option(s):

- [ ] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Pulls the stateTrees component down into its own component behind a dedicated interface for managing the SMT store.

## Testing

- [x] `make develop_test`; if any code changes were made
- [ ] [Docker Compose LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md); if any major functionality was changed or introduced
- [x] [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any infrastructure or configuration changes were made

## Required Checklist

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added, or updated, [`godoc` format comments](https://go.dev/blog/godoc) on touched members (see: [tip.golang.org/doc/comment](https://tip.golang.org/doc/comment))
- [ ] I have tested my changes using the available tooling
- [ ] I have updated the corresponding CHANGELOG

### If Applicable Checklist

- [ ] I have updated the corresponding README(s); local and/or global
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added, or updated, [mermaid.js](https://mermaid-js.github.io) diagrams in the corresponding README(s)
- [ ] I have added, or updated, documentation and [mermaid.js](https://mermaid-js.github.io) diagrams in `shared/docs/*` if I updated `shared/*`README(s)

PR #808, opened 2023-06-07
[Persistence] Refactors TreeStore to Use Module Pattern

dylanlott:
## Description

Refactors the TreeStore to use the Module pattern and register it on the Bus.
Addresses comment from #756 by @Olshansk about making the TreeStore a full module.

## Issue

image

Addresses refactor from #756 mentioned above.

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Adds the TreeStore module plumbing
- Creates the TreeStoreModule interface in shared/modules

## Testing

## Required Checklist

### If Applicable Checklist


persistence/trees/trees.go
```diff
@@ -90,6 +91,8 @@ const (
 // of the underlying trees by utilizing the lazy loading
 // functionality provided by the underlying smt library.
 type treeStore struct {
+	modules.Module
```
Olshansk, line 94: Add type assertion to guarantee that `treeStore` implements the module interface like we do with `var _ = `

persistence/trees/trees.go
```diff
@@ -90,6 +91,8 @@ const (
...
+
 	treeStoreDir string
 	merkleTrees  map[merkleTree]*smt.SMT
 	nodeStores   map[merkleTree]kvstore.KVStore
```
Olshansk, line 98: The fact that we still have `NewStateTrees` but also introduced `Create` seems like an anti-pattern: I don't know which one to use.

I think we should:
1.  rename `trees.go` to `module.go`
2. Consolidate all the business logic
3. Make the `dir` param part of the configs
4. Remove `NewStateTrees` altogether and use `Create` instead

persistence/trees/trees.go
Olshansk: Also, embed the appropriate factory interface we use introduced by Bryan

Olshansk (changes requested):

persistence/trees/trees.go
dylanlott: This is a good point about the clarity, I agree with refactoring the creation logic into the Create function. However, I would politely pushback on renaming trees.go to module.go, I think those two files should stay separate, trees.go should instead concern only the update and query functionality instead, and the creation logic it currently has should go in module.go.

persistence/trees/trees.go
Olshansk: Yessir

persistence/trees/trees.go
Olshansk: Pushback politely received and accepted 🤝

persistence/trees/trees.go
dylanlott: Done 👍

persistence/trees/module.go
```diff
@@ -0,0 +1,25 @@
...
+	return m, nil
+}
+
+func Create(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, error) {
+	return new(treeStore).Create(bus, options...)
+}
+
+func (t *treeStore) GetModuleName() string {
+	return modules.TreeStoreModuleName
+}
```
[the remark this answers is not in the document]
Olshansk: @dylanlott Do you get value from these?

shared/modules/treestore_module.go
```diff
@@ -0,0 +1,33 @@
...
+	IntegratableModule
+
+	// Update returns the new state hash for a given height.
+	// * Height is passed through to the Update function and is used to query the TxIndexer for transactions
+	// to update into the merkle tree set
+	// * Passing a higher height will cause a change but repeatedly calling the same or a lower height will
+	// not incur a change.
+	// * By nature of it taking a pgx transaction at runtime, Update inherits the pgx transaction's read view of the
+	// database.
+	// TECHDEBT: Remove the TxIndexer pass here and get it from the bus
```
Olshansk, line 29: Is this going to be done in a separate PR?

Also, the previous comment we have in the other file is clearer IMO:

```
// TODO(#808): Change interface to `Update(pgtx pgx.Tx, height uint64) (string, error)`
```

Can you make sure the other text/comments are properly mirrored as well.

Olshansk (changes requested):

persistence/trees/trees.go
```diff
@@ -78,12 +79,17 @@ const (
...
 
+// Ensure treeStore implements TreeStore
+var _ modules.TreeStoreModule = &treeStore{}
+
 // treeStore stores a set of merkle trees that
 // it manages. It fulfills the modules.TreeStore interface.
 // * It is responsible for atomic commit or rollback behavior
 // of the underlying trees by utilizing the lazy loading
 // functionality provided by the underlying smt library.
 type treeStore struct {
```
h5law, line 90: @dylanlott Wdyt of adding a field here such as `rootTree *smt.SMT` which contains the different statetree name-hash pairings. This way we no longer need to keep track of the order of the trees we just insert their root into this tree and this trees root is our stateHash. Updates in any order to the SMT will produce the same tree.

h5law: Root store doesn't need to not hash values as it will simply be used for 1) root hash calculation and 2) inclusion proofs of a subtree in the state
 - the IBC state tree will have no value hashing which is where the difference will be but this is still the same `*smt.SMT` type and poses no real changes except for how the IBC state tree will be made
 
 As the root store is going to be made of the hashes of the sub trees - it makes sense to me to use the `KVStore` Postgres seems overkill. It could even be an in memory `smt.SimpleMap` as the main purpose will be root calculation - and the only other thing that will use it is the IBC light client verification system - this will build the tree using the hashes of the sub trees which will be added to the block header. So in reality an in memory ephemeral tree is probably best
 
The benefits of this approach is that we no longer have to track the order of the trees as the SMT is not order dependent for its insertions

shared/modules/treestore_module.go
dylanlott: > @dylanlott Do you get value from these? 

No, good catch; I was able to remove that function because of the way @bryanchriswhite and I settled on to properly handle the TreeStore as a sub-module of Persistence.

persistence/trees/trees.go
dylanlott: > The benefits of this approach is that we no longer have to track the order of the trees as the SMT is not order dependent for its insertions

 This will be good, the order dependent var list of trees was an eye sore and a half 💩

h5law: Infact it doesnt even need a node store

```go
lazy := smt.NewSparseMerkleTree(nil, sha256.New())
_ = lazy.Update([]byte("foo"), []byte("oof"))
_ = lazy.Update([]byte("bar"), []byte("rab"))
_ = lazy.Update([]byte("baz"), []byte("zab"))
lazyRoot := lazy.Root()
fmt.Println(lazyRoot)

lazy = smt.NewSparseMerkleTree(nil, sha256.New())
_ = lazy.Update([]byte("baz"), []byte("zab"))
_ = lazy.Update([]byte("bar"), []byte("rab"))
_ = lazy.Update([]byte("foo"), []byte("oof"))
lazyRoot = lazy.Root()
fmt.Println(lazyRoot)
```
```
[194 210 162 176 162 41 147 217 75 19 160 224 30 159 210 181 33 239 178 208 203 236 240 248 145 48 127 62 53 46 139 169]
[194 210 162 176 162 41 147 217 75 19 160 224 30 159 210 181 33 239 178 208 203 236 240 248 145 48 127 62 53 46 139 169]
```

persistence/trees/module.go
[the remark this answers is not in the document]
dylanlott: @Olshansk occasionally, I meant to turn it off earlier though, it's less useful when they're repeated

persistence/trees/trees.go
Olshansk: Done in #843

Olshansk (approved):
LGTM but merge conflicts need to be resolved

persistence/trees/module.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func (*treeStore) Create(bus modules.Bus, options ...modules.TreeStoreOption) (modules.TreeStoreModule, error) {
+	m := &treeStore{}
+
+	for _, option := range options {
+		option(m)
+	}
+
+	m.SetBus(bus)
```
h5law, line 18:
```suggestion
	bus.RegisterModule(m)
```

Is there a difference in these calls?

dylanlott: @bryanchriswhite would be able to explain the difference better. I just followed his pointers on how to get my module working.

shared/modules/persistence_module.go
```diff
@@ -25,10 +25,6 @@ type PersistenceModule interface {
 	// BlockStore maps a block height to an *coreTypes.IndexedTransaction
 	GetBlockStore() blockstore.BlockStore
 
-	// TreeStore manages atomic access to a set of merkle trees
-	// that compose the state hash.
-	GetTreeStore() TreeStore
```
h5law, line 30: As we are removing this from the persistence context how can we access the `TreeStore` module from outside of the persistence package if needed?

persistence/module.go
```diff
@@ -237,7 +235,7 @@ func (m *persistenceModule) GetTxIndexer() indexer.TxIndexer {
 	return m.txIndexer
 }
 
-func (m *persistenceModule) GetTreeStore() modules.TreeStore {
+func (m *persistenceModule) GetTreeStore() modules.TreeStoreModule {
```
h5law, line 238: I am confused as this is present but has been removed from the `PersistenceModule` interface?

persistence/module.go
```diff
@@ -18,7 +18,6 @@ import (
 
 var (
 	_ modules.PersistenceModule = &persistenceModule{}
-	_ modules.PersistenceModule = &persistenceModule{}
```
h5law, line 21: NIT: Can we also delete the line after

persistence/module.go
dylanlott: Good catch, this was dead code, it's removed now. 👍

shared/modules/persistence_module.go
dylanlott: It can't be accessed from the outside, which is the desired outcome. Only the PersistenceContext should be exposed for external calls.

PR #844, opened 2023-06-19
[P2P] refactor: unicast router

bryanchriswhite:
## Description

Factor out "unicast routing" concerns which will be common to both RainTree and background routers. To implement the `Router` interface (which I believe is applied appropriately), **each** must be able to both send and receive messages directly to/from individual peers. In libp2p this is done via streams.

### Before

```mermaid
classDiagram
    class RainTreeMessage {
        <<protobuf>>
        +Level uint32
        +Data []byte
    }

    class PocketEnvelope {
        <<protobuf>>
        +Content *anypb.Any
        +Nonce uint64
    }

    %% class p2pModule {
    %%     -handlePocketEnvelope([]byte) error
    %% }

    %% class P2PModule {
    %%     <<interface>>
    %%     GetAddress() (Address, error)
    %%     HandleEvent(*anypb.Any) error
    %%     Send([]byte, address Address) error
    %%     Broadcast([]byte) error
    %%     BroadcastStaked([]byte) error
    %% }
    %% p2pModule --|> P2PModule

    class RainTreeRouter {
        -handler RouterHandler
        +Broadcast([]byte) error
        +Send([]byte, address Address) error
        -handleRainTreeMsg([]byte) ([]byte, error)
    }

    %% p2pModule --* RainTreeRouter
    %% RainTreeRouter --> p2pModule : `handler` == `handlePocketEnvelope`


    %% p2pModule --o PocketEnvelope
    %% p2pModule --* NonceDeduper

    class Router {
        <<interface>>
        +Send([]byte, address Address) error
        +Broadcast([]byte) error
    }
    RainTreeRouter --|> Router
```

### After

```mermaid
classDiagram
    class RainTreeMessage {
        <<protobuf>>
        +Level uint32
        +Data []byte
    }

    class PocketEnvelope {
        <<protobuf>>
        +Content *anypb.Any
        +Nonce uint64
    }

    %% class p2pModule {
    %%     -handlePocketEnvelope([]byte) error
    %% }

    %% class P2PModule {
    %%     <<interface>>
    %%     GetAddress() (Address, error)
    %%     HandleEvent(*anypb.Any) error
    %%     Send([]byte, address Address) error
    %%     Broadcast([]byte) error
    %%     BroadcastStaked([]byte) error
    %% }
    %% p2pModule --|> P2PModule

    class RainTreeRouter {
        UnicastRouter
        -handler MessageHandler
        +Broadcast([]byte) error
        -handleRainTreeMsg([]byte) error
    }

    class UnicastRouter {
        -messageHandler MessageHandler
        -peerHandler PeerHandler
        +Send([]byte, address Address) error
    }
    RainTreeRouter --* UnicastRouter : (embedded)
    %% UnicastRouter --> RainTreeRouter : via `messageHandler`


    %% p2pModule --* RainTreeRouter
    %% RainTreeRouter --> p2pModule : `handler` == `handlePocketEnvelope`


    %% p2pModule --o PocketEnvelope
    %% p2pModule --* NonceDeduper

    class Router {
        <<interface>>
        +Send([]byte, address Address) error
        +Broadcast([]byte) error
    }
    RainTreeRouter --|> Router
```

See #505 "Integration / Architecture" class diagrams for more context.

### Summary generated by Reviewpad on 29 Jun 23 06:52 UTC
This pull request includes several changes related to improving testability, reducing technical debt, and implementing unicast functionality.

The overall changes aim to improve the code structure, make it more maintainable and testable, and add support for handling unicast messages in a peer-to-peer network.

Here is a summary of the changes per file:

1. In the file `setup.go`, changes were made to register providers and routers to the module registry and improve testability.
2. In the file `router.go`, changes were made to improve peer discovery, introduce encapsulated approaches, and share interfaces among modules. Additionally, type and function names were updated.
3. The file `logging.go` underwent a rename and changes were made to the package name, logging functionality, and function names.
4. In the file `network.go`, changes were made to imports, type names, function names, and the implementation of the `handleRainTreeMsg` function to support the unicast functionality.
5. The file `testutil.go` was added as a new file to provide testing utilities for the unicast functionality.
6. In the file `p2p/raintree/testutil.go`, changes were made to imports and a function to call the `HandleStream` method from the `UnicastRouter` struct.

These changes collectively improve the testability, code structure, and unicast functionality in the project.

## Issue

- #505 

## Dependants

- #732 

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Refactored stream handling logic (i.e. unicast routing) out of `rainTreeRouter` and into new `UnicastRouter` type

## Testing

## Required Checklist

### If Applicable Checklist


p2p/unicast/router.go
```diff
@@ -0,0 +1,139 @@
...
+
+// TECHDEBT(#629): configure timeouts. Consider security exposure vs. real-world conditions.
+// TECHDEBT(#629): parameterize and expose via config.
+// readStreamTimeout is the duration to wait for a read operation on a
+// stream to complete, after which the stream is closed ("timed out").
+const readStreamTimeout = time.Second * 10
+
+var _ unicastRouterFactory = &UnicastRouter{}
+
+// TODO_THIS_COMMIT: consider defining/(re)using `RouterFactory` type
```
bryanchriswhite, line 25:
```suggestion
```

p2p/unicast/router.go
```diff
@@ -0,0 +1,139 @@
...
+	logger *modules.Logger
+	// messageHandler is the function to call when a message is received.
+	// host represents a libp2p network node, it encapsulates a libp2p peerstore
+	// & connection manager. `libp2p.New` configures and starts listening
+	// according to options.
+	// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p#section-readme)
+	host           libp2pHost.Host
+	messageHandler typesP2P.MessageHandler
+	// TODO_THIS_COMMIT: consider defining alongside `MessageHandler` type
+	// OR removing `MessageHandler` type
```
bryanchriswhite, line 40:
```suggestion
```

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+
+var _ unicastRouterFactory = &UnicastRouter{}
+
+type unicastRouterFactory = modules.FactoryWithConfig[*UnicastRouter, *config.UnicastRouterConfig]
+
+type UnicastRouter struct {
+	base_modules.IntegratableModule
+
+	logger *modules.Logger
+	// messageHandler is the function to call when a message is received.
```
dylanlott, line 31: nit: should this line not be directly above messageHandler field?

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+	host           libp2pHost.Host
+	messageHandler typesP2P.MessageHandler
+	peerHandler func(peer typesP2P.Peer) error
+}
+
+func Create(bus modules.Bus, cfg *config.UnicastRouterConfig) (*UnicastRouter, error) {
+	return new(UnicastRouter).Create(bus, cfg)
+}
+
+func (*UnicastRouter) Create(bus modules.Bus, cfg *config.UnicastRouterConfig) (*UnicastRouter, error) {
```
dylanlott, line 45: Question: What motivates the preference for accepting configs vs options in the module pattern?

p2p/unicast/logging.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// TECHDEBT(#830): it would be nice to have at least one more degree of freedom with which
+// to limit logging in areas where it is known to be excessive / high frequency.
+// Especially applicable to debug log lines which only contribute in edge cases,
+// unusual circumstances, or regressions (e.g. hitting OS resource limits because
+// of too many concurrent streams).
+//
+// This could ultimately be actuated from the CLI via flags, configs, and/or env
+// vars. Initially, weo could consider coupling to a `--verbose` persistent flag.
```
dylanlott, line 17: 🙌

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+
+// TECHDEBT(#629): configure timeouts. Consider security exposure vs. real-world conditions.
+// TECHDEBT(#629): parameterize and expose via config.
+// readStreamTimeout is the duration to wait for a read operation on a
+// stream to complete, after which the stream is closed ("timed out").
+const readStreamTimeout = time.Second * 10
+
+var _ unicastRouterFactory = &UnicastRouter{}
+
+type unicastRouterFactory = modules.FactoryWithConfig[*UnicastRouter, *config.UnicastRouterConfig]
```
dylanlott, line 25: @h5law for some examples of the Facotry generic pattern

p2p/raintree/router.go
```diff
@@ -207,25 +200,36 @@ func (rtr *rainTreeRouter) handleRainTreeMsg(data []byte) ([]byte, error) {
...
-			return nil, err
+			return err
 		}
 	}
 
-	// Return the data back to the caller so it can be handled by the app specific bus
-	return rainTreeMsg.Data, nil
+	// There was no error, but we don't need to forward this to the app-specific bus.
+	// For example, the message has already been handled by the application.
+	if rainTreeMsg.Data == nil {
```
dylanlott, line 224: Maybe worth debug logging this condition?

dylanlott (commented):
Just some minor feedback and questions, overall looks pretty good.

p2p/unicast/router.go
bryanchriswhite: The fact that everything in the config [is required](https://github.com/pokt-network/pocket/blob/refactor/unicast-router/p2p/config/config.go#L82).

p2p/raintree/router.go
```diff
@@ -207,25 +201,37 @@ func (rtr *rainTreeRouter) handleRainTreeMsg(data []byte) ([]byte, error) {
...
 	}
 
+	// TECHDEBT(#763): refactor as "pre-propagation validation"
 	networkMessage := messaging.PocketEnvelope{}
 	if err := proto.Unmarshal(rainTreeMsg.Data, &networkMessage); err != nil {
 		rtr.logger.Error().Err(err).Msg("Error decoding network message")
-		return nil, err
+		return err
 	}
+	// --
```
Olshansk, line 214:
```suggestion
```

p2p/raintree/router.go
```diff
@@ -207,25 +201,37 @@ func (rtr *rainTreeRouter) handleRainTreeMsg(data []byte) ([]byte, error) {
...
-	// Return the data back to the caller so it can be handled by the app specific bus
-	return rainTreeMsg.Data, nil
+	// There was no error, but we don't need to forward this to the app-specific bus.
+	// For example, the message has already been handled by the application.
+	if rainTreeMsg.Data == nil {
+		rtr.logger.Debug().Msg("no data in RainTree message")
+		return nil
+	}
+
+	// call configured handler to forward to app-specific bus
```
Olshansk, line 230: This comment is hard to understand from my perspective.

Is it implying: `Forward the data in the message to the app-specific bus`?

p2p/raintree/router.go
```diff
@@ -207,25 +201,37 @@ func (rtr *rainTreeRouter) handleRainTreeMsg(data []byte) ([]byte, error) {
...
+	if err := rtr.handler(rainTreeMsg.Data); err != nil {
+		rtr.logger.Error().Err(err).Msg("handling raintree message")
```
Olshansk, line 232: Leave a comment (assuming it's intended): `// intentionally not returning the error to avoid propagating it up`

p2p/raintree/router.go
```diff
@@ -270,94 +276,41 @@ func (rtr *rainTreeRouter) Size() int {
...
-
-	// read stream
-	rainTreeMsgBz, err := io.ReadAll(stream)
-	if err != nil {
-		rtr.logger.Error().Err(err).Msg("reading from stream")
-		if err := stream.Reset(); err != nil {
-			rtr.logger.Error().Err(err).Msg("resetting stream (read-side)")
-		}
-		return
+func (rtr *rainTreeRouter) setupUnicastRouter() error {
```
Olshansk, line 284:
```suggestion
// setupUnicastRouter configures, creates and assigns the `rtr.UnicastRouter`
func (rtr *rainTreeRouter) setupUnicastRouter() error {
```

p2p/unicast/logging.go
```diff
@@ -7,8 +7,18 @@ import (
 	"github.com/pokt-network/pocket/p2p/utils"
 )
 
+// TECHDEBT(#830): it would be nice to have at least one more degree of freedom with which
```
Olshansk, line 10: 💯

p2p/unicast/logging.go
```diff
@@ -7,8 +7,18 @@ import (
...
+// to limit logging in areas where it is known to be excessive / high frequency.
+// Especially applicable to debug log lines which only contribute in edge cases,
+// unusual circumstances, or regressions (e.g. hitting OS resource limits because
+// of too many concurrent streams).
+//
+// This could ultimately be actuated from the CLI via flags, configs, and/or env
+// vars. Initially, we could consider coupling to a `--verbose` persistent flag.
+//
```
Olshansk, line 18:
```suggestion
```

p2p/raintree/testutil.go
```diff
@@ -2,12 +2,30 @@
...
-	rtr.handleStream(stream)
+	rtr.UnicastRouter.HandleStream(stream)
+}
+func (rtr *rainTreeRouter) HandlerProxy(
+	t gocuke.TestingT,
+	handlerProxyFactory routerHandlerProxyFactory,
+) {
+	t.Helper()
+
+	// pass original handler to proxy factory & replace it with the proxy
```
Olshansk, line 28: Can you expand on this? I don't fully understand the need for this (without spending time diving deeper).

p2p/config/config.go
```diff
@@ -22,6 +32,14 @@ type baseConfig struct {
 	PeerstoreProvider     providers.PeerstoreProvider
 }
 
+type UnicastRouterConfig struct {
+	Logger         *modules.Logger
```
Olshansk, line 36: This is the first:
1. Package exposed `Logger` in a struct
2. Config without an associated proto

No problems/concerns, but wanted to understand if this is the approach we're taking with submodules?

p2p/unicast/testutil.go
```diff
@@ -0,0 +1,10 @@
+//go:build test
+
+package unicast
+
+import libp2pNetwork "github.com/libp2p/go-libp2p/core/network"
+
+// HandleStream exports `unicastRouter#handleStream` for testing purposes.
```
Olshansk, line 7: ❤️ this pattern!

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/pokt-network/pocket/p2p/utils"
+	"github.com/pokt-network/pocket/shared/modules"
+	"github.com/pokt-network/pocket/shared/modules/base_modules"
+)
+
+// TECHDEBT(#629): configure timeouts. Consider security exposure vs. real-world conditions.
+// TECHDEBT(#629): parameterize and expose via config.
+// readStreamTimeout is the duration to wait for a read operation on a
+// stream to complete, after which the stream is closed ("timed out").
+const readStreamTimeout = time.Second * 10
```
Olshansk, line 21: one const => tons of work :)

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+	// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p#section-readme)
+	messageHandler typesP2P.MessageHandler
+	peerHandler    func(peer typesP2P.Peer) error
+}
+
+func Create(bus modules.Bus, cfg *config.UnicastRouterConfig) (*UnicastRouter, error) {
+	return new(UnicastRouter).Create(bus, cfg)
+}
+
+func (*UnicastRouter) Create(bus modules.Bus, cfg *config.UnicastRouterConfig) (*UnicastRouter, error) {
```
Olshansk, line 45:
```suggestion
// All fields in the cfg are required
func (*UnicastRouter) Create(bus modules.Bus, cfg *config.UnicastRouterConfig) (*UnicastRouter, error) {
```

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+var _ unicastRouterFactory = &UnicastRouter{}
+
+type unicastRouterFactory = modules.FactoryWithConfig[*UnicastRouter, *config.UnicastRouterConfig]
+
+type UnicastRouter struct {
+	base_modules.IntegratableModule
+
+	logger *modules.Logger
+	host   libp2pHost.Host
+	// messageHandler is the function to call when a message is received.
```
Olshansk, line 32: This is a great comment!

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+	// host represents a libp2p network node, it encapsulates a libp2p peerstore
+	// & connection manager. `libp2p.New` configures and starts listening
+	// according to options.
+	// (see: https://pkg.go.dev/github.com/libp2p/go-libp2p#section-readme)
+	messageHandler typesP2P.MessageHandler
+	peerHandler    func(peer typesP2P.Peer) error
```
Olshansk, line 38: Can you add a comment that this helps with managing peers during peer discovery?

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, err
+	}
+
+	rtr := &UnicastRouter{
+		logger:         cfg.Logger,
+		host:           cfg.Host,
+		messageHandler: cfg.MessageHandler,
+		peerHandler:    cfg.PeerHandler,
+	}
+	rtr.SetBus(bus)
```
Olshansk, line 56: Not `RegisterModule`, correct?

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+// the incoming stream in a new go routine.
+func (rtr *UnicastRouter) handleStream(stream libp2pNetwork.Stream) {
+	rtr.logger.Debug().Msg("handling incoming stream")
+	peer, err := utils.PeerFromLibp2pStream(stream)
+	if err != nil {
+		rtr.logger.Error().Err(err).
+			Str("address", peer.GetAddress().String()).
+			Msg("parsing remote peer identity")
+
+		if err = stream.Reset(); err != nil {
```
Olshansk, line 76:
```suggestion
		// Reseting the stream is required to avoid keeping an open connection on the other node
		// NB: failing to reset the stream can easily max out the number of available
		// network connections on the receiver's side.		
		if err = stream.Reset(); err != nil {
```

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+		return
+	}
+
+	if err := rtr.peerHandler(peer); err != nil {
+		rtr.logger.Error().Err(err).
+			Str("address", peer.GetAddress().String()).
+			Msg("adding remote peer to router")
+	}
+
+	go rtr.readStream(stream)
```
Olshansk, line 88:
```suggestion
        // asynchronously read the data from the libp2p stream and handle the incoming data
	go rtr.readStream(stream)
```

p2p/unicast/router.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// readStream reads the message bytes out of the incoming stream and passes it to
+// the configured `rtr.messageHandler`. Intended to be called in a go routine.
+func (rtr *UnicastRouter) readStream(stream libp2pNetwork.Stream) {
+	// Time out if no data is sent to free resources.
+	// NB: tests using libp2p's `mocknet` rely on this not returning an error.
+	if err := stream.SetReadDeadline(newReadStreamDeadline()); err != nil {
+		// `SetReadDeadline` not supported by `mocknet` streams.
+		rtr.logger.Error().Err(err).Msg("setting stream read deadline")
```
Olshansk, line 98: Should we return after logging the error?

Olshansk (changes requested):

p2p/unicast/router.go
bryanchriswhite: That would break tests which use the libp2p mocknet as per the comments. I think the only reason an error would happen is if a node was using an incompatible transport (like in the case of mocknet). Since each node explicitly supports specific transports via the multiaddr(s) it advertises, it is in control of which transports are used in this context. 

TL;DR this should only return an error if a node listens via an unsupported transport protocol (and in tests).

p2p/unicast/router.go
bryanchriswhite: Correct. I'll add a comment to clarify. This is **not** a submodule and as a result is not dependency injectable, nor do we want it to be. It's intended to be embedded by other submodules (e.g. `RainTreeRouter`).

p2p/raintree/router.go
bryanchriswhite: Yea, I'm not sure about this Frankenstein's monster anymore :nut_and_bolt::drooling_face::nut_and_bolt:. Lemme try again (also the comment above which mentions "app-specific bus"). :sweat_smile:

p2p/raintree/testutil.go
bryanchriswhite: Apologies for this garbage. It should've been cleaned up, we don't need this. :broom:

p2p/config/config.go
bryanchriswhite: > 1. Package exposed Logger in a struct

The structs in this package are all config structs which are used to pass required parameters to constructor functions (most of which are submodule constructors). They must export their fields in order for consumers to be able to configure the unexported fields of the object under construction, via the respective constructor.

I could understand the argument that it may make more sense for these configs to live in same package as the constructor which they accompany; however, even then, these fields would still need to be exported for package-external consumers to use them (e.g. `p2p` package). The rationale behind the current code organization was to deduplicate `baseConfig#IsValid()` between `RainTreeConfig` and `BackgroundConfig`. To do this without creating an import cycle requires an additional package. At that point, this package seemed most appropriate for `UnicastConfig` to live here as well. I don't have a strong opinion about it, I just felt like this was the most readable and least complex option at the time.

> 2. Config without an associated proto

`RainTreeRouter` and `BackgroundRouter` configs (also defined in this package) do not have an associated protobuf config type, nor do I think it would make sense for them to. In the specific cases of the routers, only one of the required parameters are serializable and it is (appropriately) derived from a `P2PConfig` field (`Address`). This makes it a poor candidate, in my view, to be converted into a protobuf type such that it can be embedded in the P2P config type; however, you've made me aware of the possibility of such a case an potential consistent integration approach. :raised_hands:

```mermaid
mindmap
    root((Submodules))
        (options)
            ASSUMPTIONS
                (MUST be optional, effecting the submodules state and/or behavior)
        (configs)
            (Consideration:
            embedding submodule configs in respective module's protobuf config]
                ASSUMPTIONS
                    (Must be required for a submodule to function properly)
                    ("Multiple submodules may share a common dependency which MAY submodule, 'module-level shared dependencies'")
                        %% ("Initializaiton of such shared depencencies SHOULD be initialized and/or registered (in the case of submodules) in a parent module  in each submodule, ideally via some common and/or embedded implementation.")
                PROS
                    ("Provides clear and consistent convention with respect to how to structure configs in both modules and submodules")
                    ("Supports shared submodules; i.e. such configs could be aggregated with cardinality greater than 1 in one or more modules simultaneously without conflict")
                    ("Encourages usage of dependecy injection for submodules.")
                            ("For example, the P2P module currently sets up a peerstore provider and then passes it to router submdoules via their configs.
                            Since the providers (pstore & height) are submodules, the P2P SHOULD instead register them to the bus and let the routers dendency-inject them.")
                CONS
                    ("Protobuf config can only contain primitive/serializable types")
                        ("Functions and non-submodule, non-serializable dependencies become problematic")
                            ("For example, the P2P module currently sets up a libp2p host and passes it along with a reference to one of its methods to router submdoules via their configs.
                            Forcing submodules to rely only on seralizable configs seems insufficient.")
                ALTERNATIVES
                    (Support either and/or both; embed submodule-specific protobuf configs, where applicable, AND allow for an additional layer of abstraction to support passing high-level objects.]
                        (A protobuf config type MAY be embedded in the submodule's "config type")
                OPEN Qs
                    (Should submodules be constrained in whether they can consider a protobuf config type directly?)
        

```

p2p/raintree/router.go
bryanchriswhite: I think this is actually a mistake and it should indeed return this error instead of log it - nice catch! :sunglasses::raised_hands:

p2p/unicast/router.go
bryanchriswhite: Do you feel strongly about adding this comment? After the submodule formalization and subsequent documentation updates it will be conventional that configs contain required values by definition. Additionally, these configs implement an `#IsValid()` method which may be enforced by an interface in the future.

bryanchriswhite: Perhaps an ADR around the module system would be appropriate?

Olshansk: Don't feel strongly about this. Resolving

p2p/unicast/router.go
Olshansk: Thoughts on leaving a comment: `// not returning an error for testing purposes. This should only produce an error if a node listens via an unsupported transport protocol, which should never happen in prod.`

p2p/config/config.go
Olshansk: Makes sense to me. My tl;dr takeaway is:

1. **Separate package**: share configs across `RainTreeConfig` and `BackgroundConfig`
2. **No protobuf**: not exposed to the user / node runner.

Olshansk (approved):
Aside from this one comment, lgtm: https://github.com/pokt-network/pocket/pull/844#discussion_r1242850117

PR #845, opened 2023-06-19
[IBC] Implement ICS-23 CommitmentProof verification for the SMT

### LINKED ISSUE #841: [IBC] Implement ICS-23 allowing for the verification of SMT proofs
## Objective

Implement ICS-23 (vector commitments) including the serialisable data types that will allow for the proof generation and verification of membership/non-membership proofs based on elements in the IBC stores.

The implementation of ICS-23 should utilise [`cosmos/ics23`](https://github.com/cosmos/ics23) and the types it defines to carry out verification of serialisable proofs without the need for the underlying tree structure.

## Origin Document

Issue #794 

## Goals

- Integrate `cosmos/ics23` proof verification into Pocket V1s IBC module
- Create helper functions to convert `SparseMerkleProof` types into the relevent `cosmos/ics23` types

## Deliverable

- [ ] Implement ICS-23 fully with its various methods and types
  - [ ] Utilise `cosmos/ics23` where possible
- [ ] Create helper functions to convert `SparseMerkleProof` objects into the relevant `cosmos/ics23` types
- [ ] Add unit tests covering this functionality

- Change any existing SMT proof logic

## General issue deliverables

## Testing Methodology

- [ ] **Task specific tests or benchmarks**: `make ...`
- [ ] **New tests or benchmarks**: `make ...`
- [ ] **k8s LocalNet**: verify a `k8s LocalNet` is still functioning correctly by following the instructions [here](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md)

---

**Creator**: @h5law 
**Co-Owners**: @h5law

h5law:
### Summary generated by Reviewpad on 29 Jun 23 20:46 UTC
This pull request includes various changes across multiple files. 

In the `README.md` file, a new section titled "ICS-23 Vector Commitments" was added under the "Components" section, along with a link to the ICS-23 specification. Additionally, the link to the ICS-24 specification was updated to point to the `ics24.md` file.

In the `go.sum` file, new dependencies were added for the libraries `github.com/cosmos/gogoproto` and `github.com/h5law/ics23/go`, with specific version tags. The existing dependency on `github.com/pokt-network/smt` was also updated to a newer version.

The `ics23.md` file provides implementation details of ICS-23 Vector Commitments, including the benefits of using the `cosmos/ics23` library and the changes made for Pocket's implementation. It also explains the proof verification process for membership and non-membership proofs, accompanied by flowchart diagrams. The implementation logic can be found in the file `proofs_ics23.go`.

A new file `proofs_ics23_test.go` was added to the `ibc/store` package, containing test functions for generating and verifying commitment proofs using the ICS23 library. The tests cover membership and non-membership proofs for a sparse Merkle tree.

The `go.mod` file shows various changes, including the addition of replace directives for modules `github.com/cosmos/ics23/go` and `github.com/regen-network/gocuke`, updates to module versions, and the addition of a new module `github.com/cosmos/gogoproto`.

The file `proofs_ics23.go` is a new addition to the `ibc/store` package, containing functions and constants for verifying membership and non-membership in a Sparse Merkle Tree.

Lastly, in the file where the `NextCode` constant is defined, it was updated to a new value and a new constant and error function were added.

Please review these changes in the pull request.

## Issue

Fixes #841 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Add proof generation and verification in accordance with `cosmos/ics23`
- Add SMT proof -> ics23 proof conversion

## Testing

## Required Checklist

### If Applicable Checklist


ibc/docs/ics23.md
```diff
@@ -0,0 +1,78 @@
...
+1. Use the key-value pair to generate a leaf hash
+2. Hash the leaf with the `SideNodes` found in the `path` field of the `ExistenceProof` to generate the root hash
+3. Compare the root hash with the one provided
+
+For non-membership the logic is as follows:
+
+1. If the `ActualValueHash` field in the `ExclusionProof` is the SMT's placeholder value, then use the placeholder value as the leaf node hash and follow steps 3-4
+2. If the `ActualValueHash` field is not the placeholder value, then use the `ActualPath` and `ActualValueHash` fields to generate the leaf node hash - do not hash these values before hashing the node as they are populated from the SMT proof's `NonMembershipLeafData` field and thus are already hashed
+3. Hash the leaf node hash with the `SideNodes` found in the `Path` field of the `ExclusionProof` to generate the root hash
+4. Compare the root hash with the one provided
```
Olshansk, line 72: because we're using `ActualValueHash`, we also verify that the root hash is the same as the one computed even though it's different from the value the user gives?

Can you clarify this point

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,159 @@
...
+
+type position int
+
+const (
+	left  position = iota // 0
+	right                 // 1
+)
+
+var (
+	// Custom SMT spec as the store does not hash values
```
Olshansk, line 19: I keep bringing this up because I'm still personally not 100% sure myself (but trust your judgment), but you're confident that we should keep hashing the other values even though we're not hashing these ones?

Lmk if this is better for an offline dicussion.

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		InnerSpec: &ics23.InnerSpec{
+			ChildOrder:      []int32{0, 1},
+			ChildSize:       32,
+			MinPrefixLength: 1,
+			MaxPrefixLength: 1,
+			EmptyChild:      make([]byte, 32),
+			Hash:            ics23.HashOp_SHA256,
+		},
+		MaxDepth:                   256,
```
Olshansk, line 36: Cool. Should be enough leaves

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+type position int
+
+const (
+	left  position = iota // 0
+	right                 // 1
+)
+
+var (
```
Olshansk, line 18: Should probably mention that this has a direct dependency on constants /values from `pokt-network/smf`

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,159 @@
...
+				Key:   key,
+				Value: value,
+				Leaf:  smtSpec.LeafSpec,
+				Path:  steps,
+			},
+		},
+	}
+}
+
+// convertSMPToExistenceProof converts a SparseMerkleProof to an ics23
```
Olshansk, line 97:
```suggestion
// convertSMPToExclusionProof converts a SparseMerkleProof to an ics23
```

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,159 @@
...
+		LeafSpec: &ics23.LeafOp{
+			Hash:         ics23.HashOp_SHA256,
+			PrehashKey:   ics23.HashOp_SHA256,
+			PrehashValue: ics23.HashOp_NO_HASH,
+			Length:       ics23.LengthOp_NO_PREFIX,
+			Prefix:       []byte{0},
+		},
+		InnerSpec: &ics23.InnerSpec{
+			ChildOrder:      []int32{0, 1},
+			ChildSize:       32,
```
Olshansk, line 30: Move `32` into a local constant and reuse it everywhere you use `32`

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,159 @@
...
+		// Do not re-hash already hashed fields from NonMembershipLeafData
+		PrehashKey:   ics23.HashOp_NO_HASH,
+		PrehashValue: ics23.HashOp_NO_HASH,
+		Length:       ics23.LengthOp_NO_PREFIX,
+		Prefix:       []byte{0},
+	}
+	actualPath := path[:]
+	actualValue := defaultValue
+	if proof.NonMembershipLeafData != nil {
+		actualPath = proof.NonMembershipLeafData[1:33]
```
Olshansk, line 113:
```suggestion
		actualPath = proof.NonMembershipLeafData[1:33] // len(prefix): len(prefix)+`newConstSize`
```

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+		steps = append(steps, op)
+	}
+	return steps
+}
+
+// getPathBit determines whether the hash of a node at a certain depth in the tree is the
+// left or the right child of its parent
+func getPathBit(data []byte, position int) position {
+	if int(data[position/8])&(1<<(8-1-uint(position)%8)) > 0 {
```
Olshansk, line 155: Can you provide an example of what this does so others avoid spending the time figuring out what you already have

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// convertSideNodesToSteps converts the SideNodes field in the SparseMerkleProof
+// into a list of InnerOps for the ics23 CommitmentProof
+func convertSideNodesToSteps(sideNodes [][]byte, path []byte) []*ics23.InnerOp {
+	steps := make([]*ics23.InnerOp, 0, len(sideNodes))
+	for i := 0; i < len(sideNodes); i++ {
+		var prefix, suffix []byte
+		prefix = append(prefix, innerPrefix...)
+		if getPathBit(path, len(sideNodes)-1-i) == left {
```
Olshansk, line 136: Can you add a comment on what (and why) we're updating the suffic on left and prefix on right?

ibc/store/proofs_ics23_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	testCases := []struct {
+		key           []byte
+		value         []byte
+		nonmembership bool
+		fails         bool
+		expected      error
+	}{
+		{
+			// Successfully generates a membership proof for a key stored
```
Olshansk, line 37: You should name these and use the name below. This way, when a test fails, the name shows up in terminal (or CI)

ibc/store/proofs_ics23_test.go
```diff
@@ -0,0 +1,178 @@
...
+	require.NoError(t, err)
+	err = tree.Update([]byte("testKey"), []byte("testValue"))
+	require.NoError(t, err)
+	err = tree.Update([]byte("testKey2"), []byte("testValue2"))
+	require.NoError(t, err)
+
+	testCases := []struct {
+		key           []byte
+		value         []byte
+		nonmembership bool
```
Olshansk, line 32: s/nonmembership/isNonMembership

ibc/store/proofs_ics23_test.go
```diff
@@ -0,0 +1,178 @@
...
+		var proof *ics23.CommitmentProof
+		if tc.nonmembership {
+			proof, err = createNonMembershipProof(tree, tc.key)
+		} else {
+			proof, err = createMembershipProof(tree, tc.key, tc.value)
+		}
+		if tc.fails {
+			require.EqualError(t, err, tc.expected.Error())
+			require.Nil(t, proof)
+			return
```
Olshansk, line 72: s/return/continue

ibc/store/proofs_ics23_test.go
```diff
@@ -0,0 +1,178 @@
...
+			valid = VerifyNonMembership(root, proof, tc.key)
+		} else {
+			valid = VerifyMembership(root, proof, tc.key, tc.value)
+		}
+
+		if tc.valid {
+			require.True(t, valid)
+		} else {
+			require.False(t, valid)
+		}
```
Olshansk, line 173: require.Equal(t, tc.valid, valid)

Olshansk (changes requested):

ibc/docs/ics23.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. It is already SDK agnostic, so can be used by Pocket (a non-cosmos chain) without any issues.
+2. The functions defined for proof verification are separate from the tree structures themselves
+   - This means we do not need to interact with an SMT instance in order to verify proofs.
+
+However, there were some changes made specifically for Pocket's implementation of ICS-23.
+
+### Custom SMT `ProofSpec`
+
+The `ProofSpec` type in `cosmos/ics23` is used to define the steps needed to verify a proof, what hashing functions should be used, any node prefixes, etc. This is then passed into the verification functions in order to verify a proof, instead of having to interact with the tree itself. This is useful as proofs must be verified on a light-client and as such being able to verify a proof without creating a tree is much more memory efficient.
```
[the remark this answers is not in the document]
h5law: No so remember the way that we construct a proof requires us to provide the hashes of the neighbouring nodes so that when we hash the key-value pair in conjunction with the hashes in the proof (as long as we get their left/right order correct) we can reconstruct the root hash. The tree itself does this method without going back "into" the tree. What the using the tree does is that it allows us to use the hashing functions the tree was set up with without providing them again.

So when we do this without the tree it means
1. We do not need to create a tree
2. We follow the exact same steps
3. We have to provide the hashing functions
4. In the "steps" we have to say the order of the hashes

If we look at how we define the order of the hashes its kinda simple:
```go
func convertSideNodesToSteps(sideNodes [][]byte, path []byte) []*ics23.InnerOp {
	steps := make([]*ics23.InnerOp, 0, len(sideNodes))
	for i := 0; i < len(sideNodes); i++ {
		var prefix, suffix []byte
		prefix = append(prefix, innerPrefix...)
		if getPathBit(path, len(sideNodes)-1-i) == left {
			suffix = make([]byte, 0, len(sideNodes[i]))
			suffix = append(suffix, sideNodes[i]...)
		} else {
			prefix = append(prefix, sideNodes[i]...)
		}
		op := &ics23.InnerOp{
			Hash:   ics23.HashOp_SHA256,
			Prefix: prefix,
			Suffix: suffix,
		}
		steps = append(steps, op)
	}
	return steps
```

Then when we hash using this step with either a prefix or a suffix we add the current hash into the missing space and we have an inner node! This is then hashed to produce the next current hash and so on until we have the root.

ibc/docs/ics23.md
```diff
@@ -0,0 +1,78 @@
...
+	MaxDepth:                   256,
+	PrehashKeyBeforeComparison: true,
+}
+```
+
+The main change from the `cosmos/ics23` `SmtSpec` object is that the `PrehashValue` field is set to not hash values prior to hashing the key-value pair.
+
+### Converting `SparseMerkleProof` to `CommitmentProof`
+
+In order to convert the proofs generated by the SMT into a serialisable proof used by `cosmos/ics23`, the `SideNodes` field of the `SparseMerkleProof` must be converted into a list of `InnerOp` types which define the order of the hashes. The order of the hashes is important as depending on whether the next hash is the left or right neighbour of the current hash, they will be hashed in a different order, ultimately creating a different root hash. This conversion allows the verification to produce the same root hash as the SMT would have produced when verifying the proof.
```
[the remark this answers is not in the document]
h5law: There were old discussions in celestia's repo on this. It is possible and we could do it. It would be a heft task though. A cool idea would be setting `k` on creation. But as stated in those previous discussions we would need to benchmark it appropriately.

I am thinking of creating a suite of benchmark tests in the SMT we could then use as a baseline for any changes moving forward

ibc/store/proofs_ics23.go
h5law: We dont hash any other values. The other side nodes are already hashed. When we insert into the tree there are 3 steps:
1. hash the key - this gives us the path (we use the `PathHasher`)
  - The `PathHasher` can be customised with the option `WithPathHasher()` when creating the tree, this cannot be `nil`
2. hash the value - this gives us the valueHash (we use the `ValueHasher`)
  - The `ValueHasher` can be customised with the option `WithValueHasher()` if the value hasher is `nil` we do a `noop` function when hashing the value leaving us with the raw bytes
3. We then hash both of these using the `TreeHasher` this is required when we create the tree - this gives us the nodes digest
  - The `TreeHasher` is used as the `PathHasher` and `ValueHasher` if they are not provided
  - The `TreeHasher` cannot be nil

From this we can see that we dont want to prehash the value as we need the raw bytes but the node is still encoded and hashed to give the digest whether the value is or not. The side nodes are digests, they are already hashed. What we do with them is create a new inner node like this:
```
innerNode = []byte{[]byte{1}, leftChild.digest, rightChild.digest}
```
And then we hash that to give us the next digest.

h5law:
🚨 I HAVE REWRITTEN HISTORY 🚨

@Olshansk I went through all the comments, and rebased off of main. The PR is now much smaller without all the unecissary files from other PRs.

ibc/docs/ics23.md
```diff
@@ -0,0 +1,89 @@
...
+
+Non-membership proofs are verified as follows:
+
+1. If the `ActualValueHash` field in the `ExclusionProof` is the SMT's placeholder value, then use the placeholder value as the leaf node hash and skip to step 3 below
+2. If the `ActualValueHash` field is not the placeholder value, then use the `ActualPath` and `ActualValueHash` fields to generate the leaf node hash.
+   - **IMPORTANT**: DO NOT hash these values before hashing the node as they are populated from the SMT proof's `NonMembershipLeafData` field and thus are already hashed
+3. Hash the leaf node hash with the `SideNodes` found in the `Path` field of the `ExclusionProof` to generate the root hash
+4. Compare the root hash with the one provided
+   - As the non-membership proof uses the `ActualValueHash` field to generate the leaf node hash, the non-membership proof is actually proving membership of either a placeholder key or an unrelated key in the tree.
+   - This means that if the root hash computed and the one provided are equal then the key we were looking for was not in the tree. If they are not equal the proof is invalid and the key is in the tree.
```
Olshansk, line 83: SOrry for the back & forth, but I'm personally still struggling to understand this. I tried to reword it so it's easier for the reader:

Is this correct?
```suggestion
   - if `computedRooHash == providedRootHash`: `key` not in tree -> `Proof` is valid -> exclusion QED
   - if `computedRooHash != providedRootHash`: `key is in tree` -> `Proof` is invalid  -> exclusion QED
```

ibc/docs/ics23.md
```diff
@@ -0,0 +1,78 @@
+# ICS-23 Vector Commitments <!-- omit in toc -->
```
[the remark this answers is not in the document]
Olshansk: @h5law Stupid question below.

> . Hash the leaf node hash with the `SideNodes` found in the `Path` field of the `ExclusionProof` to generate the root hash

This looks like a merkle proof, so do you know why we call it a vector commitment?

ibc/docs/ics23.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Proof Verification
+
+Membership proofs are verified as follows:
+
+1. Use the key-value pair to generate a leaf hash
+2. Hash the leaf with the `SideNodes` found in the `path` field of the `ExistenceProof` to generate the root hash
+3. Compare the root hash with the one provided and expect them to be identical
+
+Non-membership proofs are verified as follows:
```
Olshansk, line 75: @h5law I don't know why I'm struggling to understand this documentation, but I put together this mermaid diagram. Can you edit it based on your understanding?

I still feel like I have a gap in my knowledge somewhere.

```mermaid

---
title: Exclusion Proof 
---
flowchart TD
    I["Proof,Key"]
    NMD{"proof.NonMembershipLeafData == nil ?"}
    KP1["actualPath = sha256(key) \n actualValue = placeholderNode"]
    KP2["actualPath = ProvidedKeyHash \n actualValue = ProvidedValueHash"]
    Compare{"ComputedRootHash == ProvidedRootHash ?"}
    R1{"actualValue == placeholderNode ?\n\n Tree matches - exclusion \n requires empty node for path(key)"}
    R2{"actualValue == providedValueHash ?\n\n Tree does not match - exclusion \n requires provided value at the path given"}    
    EV["Exclusion Prove VALID"]
    EI["Exclusion Prove INVALID"]

    I --> NMD
    NMD -- Yes --> KP1
    NMD -- No --> KP2

    KP1 -- Key,Value --> Compare
    KP2 -- Key,Value --> Compare
    
    Compare -- Yes --> R1
    Compare -- No --> R2

    R1 -- Yes --> EV
    R1 -- No --> EI
    R2 -- Yes --> EV
    R2 -- No --> EI    
```

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,177 @@
...
+
+import (
+	"crypto/sha256"
+
+	ics23 "github.com/cosmos/ics23/go"
+	coreTypes "github.com/pokt-network/pocket/shared/core/types"
+	"github.com/pokt-network/smt"
+)
+
+// position refers to whether the node is either the left or right child of its parent
```
Olshansk, line 11:
```suggestion
// position refers to whether the node is either the left or right child of its parent for the binary smt
```

ibc/store/proofs_ics23.go
Olshansk: The hashing and nonhashing makes sense to me in the context of IBC and SMT in independence.

It's not a blocker for this PR, but my question/concern is moreso for the larger context of the blockchain to make sure that I'm personally not missing/misunderstanding something. For the rest of the persistence module, question is: `Should we hash the values for the other trees or not?`

@dylanlott do you have thoughts/opinion here?

ibc/store/proofs_ics23_test.go
```diff
@@ -0,0 +1,180 @@
...
+			var proof *ics23.CommitmentProof
+			if tc.isNonMembership {
+				proof, err = createNonMembershipProof(tree, tc.key)
+			} else {
+				proof, err = createMembershipProof(tree, tc.key, tc.value)
+			}
+			if tc.fails {
+				require.EqualError(t, err, tc.expected.Error())
+				require.Nil(t, proof)
+				return
```
Olshansk, line 74: Noticed this was resolved without a comment. 

I'm okay with the pushback but would like to understand why.


Screenshot 2023-06-28 at 3 22 07 PM

ibc/store/proofs_ics23_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// Proof generation cannot fail but verification can
+func TestICS23Proofs_GenerateCommitmentProofs(t *testing.T) {
+	nodeStore := kvstore.NewMemKVStore()
+	tree := smt.NewSparseMerkleTree(nodeStore, sha256.New(), smt.WithValueHasher(nil))
+	require.NotNil(t, tree)
+
+	// Set a value in the store
```
Olshansk, line 19:
```suggestion
	// Prepare a tree with a predetermined set of key-value pairs
```

ibc/store/proofs_ics23_test.go
```diff
@@ -0,0 +1,180 @@
...
+	err = nodeStore.Stop()
+	require.NoError(t, err)
+}
+
+func TestICS23Proofs_VerifyCommitmentProofs(t *testing.T) {
+	nodeStore := kvstore.NewMemKVStore()
+	tree := smt.NewSparseMerkleTree(nodeStore, sha256.New(), smt.WithValueHasher(nil))
+	require.NotNil(t, tree)
+
+	// Set a value in the store
```
Olshansk, line 99:
```suggestion
	// Prepare a tree with a predetermined set of key-value pairs
```

ibc/store/proofs_ics23.go
```diff
@@ -0,0 +1,177 @@
...
+			Hash:   ics23.HashOp_SHA256,
+			Prefix: prefix,
+			Suffix: suffix,
+		}
+		steps = append(steps, op)
+	}
+	return steps
+}
+
+// getPathBit takes the hash of a key (the path) and a position (depth) and returns whether at
```
Olshansk, line 157: See the ref - `Ref: https://github.com/pokt-network/smt/blob/main/utils.go`

Yea, this is definitely non-trivial to implement...

This business logic is VERY closely coupled with our specific SMT implementation. I'd suggest the following:
1. exposing `GetPathBit` in the smt and reusing it here
2. Move the comments you added to the smt
3. Add an `isLeftChild` helper and use it here

Olshansk (changes requested):

ibc/docs/ics23.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. If the `ActualValueHash` field in the `ExclusionProof` is the SMT's placeholder value, then use the placeholder value as the leaf node hash and skip to step 3 below
+2. If the `ActualValueHash` field is not the placeholder value, then use the `ActualPath` and `ActualValueHash` fields to generate the leaf node hash.
+   - **IMPORTANT**: DO NOT hash these values before hashing the node as they are populated from the SMT proof's `NonMembershipLeafData` field and thus are already hashed
+3. Hash the leaf node hash with the `SideNodes` found in the `Path` field of the `ExclusionProof` to generate the root hash
+4. Compare the root hash with the one provided
+   - As the non-membership proof uses the `ActualValueHash` field to generate the leaf node hash, the non-membership proof is actually proving membership of either a placeholder key or an unrelated key in the tree.
```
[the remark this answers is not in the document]
h5law: Yeah my bad good spot!

ibc/docs/ics23.md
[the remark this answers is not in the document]
h5law: This is a merkle proof but they call them vector commitments in ICS-23 its the name of the spec heres how they define it:
> A vector commitment is a construction that produces a constant-size, binding commitment to an indexed vector of elements and short membership and/or non-membership proofs for any indices & elements in the vector.

AFAIK no one is actually using vector commitments they are all using merkle proofs for ICS-23

ibc/docs/ics23.md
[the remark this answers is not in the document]
h5law: Yeah I'd like to explore this, as I mentioned it was a topic of discussion in the original SMT repo and would be cool to find our own optimisation - potentially for different node stores we could provide a optimal `k` constant:
```go
const (
  SimpleMapStoreK = 2
  BadgerK = 4
)
```
This could be cool.

ibc/docs/ics23.md
h5law: Have updated the docs with an accurate diagram

ibc/docs/ics23.md
```diff
@@ -0,0 +1,89 @@
...
+
+Membership proofs are verified as follows:
+
+1. Use the key-value pair to generate a leaf hash
+2. Hash the leaf with the `SideNodes` found in the `path` field of the `ExistenceProof` to generate the root hash
+3. Compare the root hash with the one provided and expect them to be identical
+
+Non-membership proofs are verified as follows:
+
+1. If the `ActualValueHash` field in the `ExclusionProof` is the SMT's placeholder value, then use the placeholder value as the leaf node hash and skip to step 3 below
```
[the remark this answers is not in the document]
h5law: Not made this change but instead clarified the statement

ibc/store/proofs_ics23.go
h5law: So in my ICS-24 PR I made the decision to not hash any values in all of the trees (except root tree as its created differently) this is just to make things easier during creation. But open to ideas here

ibc/store/proofs_ics23_test.go
h5law: My bad this was actually an error from reverting out of the `t.Run()` way of things with the name. Now its back the return should be there.

ibc/store/proofs_ics23.go
h5law: I have added a techdebt comment here as I have opened [smt#14](https://github.com/pokt-network/smt/issues/14) which will cover this. As I have just release `v0.6.0` for the SMT i think that its best to bundle these small changes into releases to save on constantly updating the `go.mod`. As the functionality will not change until this issue in in progress I think its alright to leave it until then.

LMK wdyt

ibc/store/proofs_ics23.go
Olshansk: Interesting. Would like to discuss this more (as a team). 

Moving the thread to https://github.com/pokt-network/pocket/pull/847#discussion_r1245898305

ibc/docs/ics23.md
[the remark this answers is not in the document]
Olshansk: Got it. So the tl;dr of the point you're making is that merkle proofs are verifiable without the tree. Correct?

ibc/store/proofs_ics23_test.go
Olshansk: Got it. Makes more sense now.

ibc/store/proofs_ics23.go
Olshansk: Going to push back on this.

I think https://github.com/pokt-network/smt/issues/14 is a 1+ month research and development project, whereas exposing the `GetPathBit` function is something we can do before EOD.

Olshansk (changes requested):

ibc/docs/ics23.md
```diff
@@ -0,0 +1,116 @@
...
+   - **IMPORTANT**: DO NOT hash these values before hashing the node as they are populated from the SMT proof's `NonMembershipLeafData` field and thus are already hashed
+3. Hash the leaf node hash with the `SideNodes` found in the `Path` field of the `ExclusionProof` to generate the root hash
+4. Compare the root hash with the one provided
+   - if `computedRootHash == providedRootHash`
+     - `key` not in tree -> `Proof` is valid -> exclusion QED
+   - if `computedRootHash != providedRootHash`
+     - `key` is in tree -> `Proof` is invalid -> exclusion QED
+
+```mermaid
+flowchart TD
```
Olshansk, line 88: Appreciate the new diagram!

ibc/store/proofs_ics23.go
h5law: @Olshansk https://github.com/pokt-network/smt/pull/15

Olshansk (approved):
ICS23!!!!!

PR #827, opened 2023-06-15
[Configs] Cleanup private keys and genesis file

okdas:
## Description

A DNS resolution error was reported by @bryanchriswhite. I discovered that the service URL, in contrast to v0, was utilized as an endpoint for peer-to-peer (P2P) connections. This particular practice resulted in a certain level of confusion. The service URL, prefixed with the `http://` scheme, was subsequently incorporated into the genesis file. This action, unfortunately, disrupted DNS resolution as `http://` cannot constitute a segment of the DNS name.

As part of the troubleshooting process, I regenerated the private keys for the fishermen, ensuring they align with the existing pattern of addresses. The mentioned issue concerning the service URL is comprehensively documented in another ticket - #818.

### Summary generated by Reviewpad on 15 Jun 23 23:47 UTC
This pull request includes changes to various files:
- One file has been renamed from "909aa7caea408519a7ed73df70ca3e65.md5" to "8b7cc7e70ff0cabc68abd1e691a6a047.md5"
- The Tiltfile has been updated to include provisions for servicer and fishermen nodes in addition to the existing validator nodes
- A binary file "debug_keybase.bak" has been changed but without additional details
- A change has been made to the container spec of a file to remove the `args` section and use the image's default command instead
- Two new Kubernetes services "pocket-fishermen" and "pocket-servicers" have been added with annotations for Prometheus monitoring and ports for RPC and metrics on each service.

## Issue

Fixes DNS resolution errors.

## Type of change

Please mark the relevant option(s):

- [ ] New feature, functionality or library
- [x] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Regenerated keys for fishermen to align with an existing address pattern (addresses begin with XXXYY, where YY is an ID of the actor)
- Changed a list of addresses that have balance to not include a 1k validators, but 100 first addresses that belong for each actor.
- Modified service urls to fix DNS resolution.
- Add DevNet genesis to the helm chart.

## Testing

- [x] `make develop_test`; if any code changes were made
- [x] `make test_e2e` on [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any code changes were made
- [ ] `e2e-devnet-test` passes tests on [DevNet](https://pocketnetwork.notion.site/How-to-DevNet-ff1598f27efe44c09f34e2aa0051f0dd); if any code was changed
[...]

PR #803, opened 2023-06-02
[Utility] trustless relays servicer token validation

adshmh:
## Description

Add validation of application's session tokens to the servicer.

## Issue

Part of work on #754 

## Type of change

Please mark the relevant option(s):

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Servicer now validates the availability of tokens in the current session before executing relays.

## Testing

- [x] `make develop_test`; if any code changes were made
- [ ] `make test_e2e` on [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any code changes were made
- [ ] `e2e-devnet-test` passes tests on [DevNet](https://pocketnetwork.notion.site/How-to-DevNet-ff1598f27efe44c09f34e2aa0051f0dd); if any code was changed
- [ ] [Docker Compose LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md); if any major functionality was changed or introduced
- [ ] [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any infrastructure or configuration changes were made

## Required Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [] I have added, or updated, [`godoc` format comments](https://go.dev/blog/godoc) on touched members (see: [tip.golang.org/doc/comment](https://tip.golang.org/doc/comment))
- [x] I have tested my changes using the available tooling
- [x] I have updated the corresponding CHANGELOG

### If Applicable Checklist

- [ ] I have updated the corresponding README(s); local and/or global
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added, or updated, [mermaid.js](https://mermaid-js.github.io) diagrams in the corresponding README(s)
- [ ] I have added, or updated, documentation and [mermaid.js](https://mermaid-js.github.io) diagrams in `shared/docs/*` if I updated `shared/*`README(s)

PR #848, opened 2023-06-20
[General] Ignore false positive secret uncovered by `ggshield` on main

Olshansk:
### Summary generated by Reviewpad on 20 Jun 23 22:06 UTC
This pull request includes changes to add .gitguardian.yaml file to ignore secrets and gitignore patterns for ggshield. There are also updates to the Makefile to include new targets for ggshield to scan and ignore secrets, as well as an update to make the path relative in the .gitguardian.yaml file.

## Issue

NA

## Type of change

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Added `ggshiled` helpers to scan and ignore secrets in the future
- Ignoring a secret that was introduced by #827 that was blocking main seen [here](https://github.com/pokt-network/pocket/pull/803/checks?check_run_id=14359233756)
- 
Screenshot 2023-06-20 at 2 50 59 PM

## Testing

NA

### DIFF
--- a/.gitguardian.yaml
+++ b/.gitguardian.yaml
@@ -1,5 +1,5 @@
 secret:
   ignored-matches:
-  - match: d1585b49ffa864bfb76b2432b26933a866c3cf621a9d5dd41690c1568edd85ed
-    name: Generic High Entropy Secret - /Users/olshansky/workspace/pocket/pocket/charts/pocket/templates/configmap-genesis.yaml
+    - match: d1585b49ffa864bfb76b2432b26933a866c3cf621a9d5dd41690c1568edd85ed
+      name: Generic High Entropy Secret - charts/pocket/templates/configmap-genesis.yaml
 version: 2
--- a/.gitguardian.yaml
+++ b/.gitguardian.yaml
@@ -1,0 +1,5 @@
+secret:
+  ignored-matches:
+  - match: d1585b49ffa864bfb76b2432b26933a866c3cf621a9d5dd41690c1568edd85ed
+    name: Generic High Entropy Secret - /Users/olshansky/workspace/pocket/pocket/charts/pocket/templates/configmap-genesis.yaml
+version: 2
--- a/Makefile
+++ b/Makefile
@@ -579,5 +579,5 @@ ggshield_secrets_scan: ## Scans the project for secrets using ggshield
 	ggshield secret scan path --recursive .
 
 .PHONY: ggshield_secrets_add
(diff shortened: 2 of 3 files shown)


.gitguardian.yaml
```diff
@@ -0,0 +1,5 @@
+secret:
+  ignored-matches:
+  - match: d1585b49ffa864bfb76b2432b26933a866c3cf621a9d5dd41690c1568edd85ed
+    name: Generic High Entropy Secret - /Users/olshansky/workspace/pocket/pocket/charts/pocket/templates/configmap-genesis.yaml
```
adshmh, line 4: Do we want to commit the part that starts with `/Users` ?

Makefile
```diff
@@ -573,3 +573,11 @@ search_interfaces: ## Greps and outputs all of the structs in the project (exclu
 .PHONY: search_protos
 search_protos: ## Finds all of the proto files in the project (excluding vendor)
 	find . -name "*.proto" -not -path "./vendor/*"
+
+.PHONY: ggshield_secrets_scan
+ggshield_secrets_scan: ## Scans the project for secrets using ggshield
+	ggshield secret scan path --recursive .
+
+.PHONY: ggshield_secrets_add
+ggshield_secrets_add: ## Ignore recently foudn secrets using ggshield
```
adshmh, line 582:
```suggestion
ggshield_secrets_add: ## Ignore recently found secrets using ggshield
```

Makefile
```diff
@@ -573,3 +573,11 @@ search_interfaces: ## Greps and outputs all of the structs in the project (exclu
...
+	ggshield secret ignore --last-found
```
adshmh, line 583: Doesn't this effectively disable `ggshield`? i.e. it won't complain even if we actually include a secret in the diff by mistake.

adshmh (commented):
Not sure if the comments I have added should be blockers, so added them all as review comments.

Makefile
adshmh: Correction to my comment above: it would perhaps make it a bit too easy to miss a secret by ignoring it. Perhaps there is a mechanism to add a comment to the line that could be interpreted as a secret to get `ggshield` to ignore it.

Olshansk: Wdyt of the updated comment?

adshmh (approved):

okdas (approved):

Issue #562, opened 2023-03-08
[Utility][Persistence][Consensus] Implement SavePoints & Rollbacks


deblasis:
## Objective

This issue is going to be an "umbrella" ticket/epic used to wrap and link all the intermediate steps/deliverables that are necessary to build `Savepoints` and `Rollbacks` as highlighted in the [Design Document](https://github.com/pokt-network/pocket/pull/533/files#diff-265cd51f4b9ede6b146d014f04d6c494d813a0c186a37667c9b5ec86e88ed763) - Issue #493 

## Origin Document

This is the result of R&D work on #508 which was essentially asking the assignee to build savepoints and rollbacks, just with a different wording thanks to a more "practical" problem to solve.

## Goals

- Implement an abstraction that can contain the operations performed on the various persistence datastores (as described in the design document) and apply/rollback them atomically
- Implement a way to serialize and deserialize the whole persistence state so that's possible to create snapshots and rehydrate them
- Implement the Moonshot parts of the design document (perhaps out-of-scope for now? I won't include the deliverables below but please let me know if I should)
- Ensure that the state of the modules is restored in case of a rollback 

## Deliverable

- [ ] Refactor `UtilityContext` into `UtilityUnitOfWork`
- [ ] Implement KISS SavePoints - Serialize WorldState from `Persistence` #327 
- [ ] Implement change tracking in `UtilityUnitOfWork` to allow ephemeral state and remove side effects
- [ ] Implement KISS Rollbacks - Deserialize WorldState to `Persistence` #435 
- [ ] Implement Tooling (CLI) entrypoints for Snapshots/Rollbacks
- [ ] Implement Snapshots - Persist SavePoints
- [ ] Implement Rollback from Snapshot
- [ ] Integrate Snapshots and Rollbacks with CLI
- [ ] Moonshots **TBD**

## Non-goals / Non-deliverables

- Changes in logic with particular focus to `Consensus`

## General issue deliverables

- [ ] Update the appropriate CHANGELOG(s)
- [ ] Update any relevant local/global README(s)
- [ ] Update relevant source code tree explanations
- [ ] Add or update any relevant or supporting [mermaid](https://mermaid-js.github.io/mermaid/) diagrams

## Testing Methodology

- [ ] **Task specific tests or benchmarks**: `make ...`
- [ ] **New tests or benchmarks**: `make ...`
- [ ] **All tests**: `make test_all`
[...]

PR #855, opened 2023-06-23
chore: introduce `Submodule` interface

bryanchriswhite:
## Description

Introduces the `Submodule` interface and applies it to the P2P module's peerstore provider.

## Issue

Deliverables 1, 2, 3, & 5:
- #810

## Type of change

Please mark the relevant option(s):

- [ ] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [x] Documentation
- [ ] Other 

## List of changes

* fixed typo; renamee `IntegratableModule` to `IntegrableModule`
* renamed `InitializableModule` to `InjectableModule`
* removed `InjectableModule#Create()` as `Moudle` also embeds `ModuleFactoryWithConfig`
* added `Submodule` interface type
* refactored peerstore providers as submodules
* updated module README

## Testing

- [x] `make develop_test`; if any code changes were made
- [x] `make test_e2e` on [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any code changes were made
- [x] `e2e-devnet-test` passes tests on [DevNet](https://pocketnetwork.notion.site/How-to-DevNet-ff1598f27efe44c09f34e2aa0051f0dd); if any code was changed
- [ ] [Docker Compose LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md); if any major functionality was changed or introduced
- [x] [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any infrastructure or configuration changes were made

## Required Checklist

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added, or updated, [`godoc` format comments](https://go.dev/blog/godoc) on touched members (see: [tip.golang.org/doc/comment](https://tip.golang.org/doc/comment))
- [ ] I have tested my changes using the available tooling
- [ ] I have updated the corresponding CHANGELOG

### If Applicable Checklist

- [x] I have updated the corresponding README(s); local and/or global
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added, or updated, [mermaid.js](https://mermaid-js.github.io) diagrams in the corresponding README(s)
- [ ] I have added, or updated, documentation and [mermaid.js](https://mermaid-js.github.io) diagrams in `shared/docs/*` if I updated `shared/*`README(s)

PR #861, opened 2023-06-27
[Persistence] Adds atomic Update for TreeStore

dylanlott:
## Description

Adds atomic savepoint and rollback creation to the TreeStore

## Issue

Relates to #562 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- adds savepoints and rollbacks to TreeStore Update function

## Testing

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/persistence/context.go
+++ b/persistence/context.go
@@ -78,6 +78,11 @@ func (p *PostgresContext) Commit(proposerAddr, quorumCert []byte) error {
 		return err
 	}
 
+	// Commit the tree store updates
+	if err := p.stateTrees.Commit(); err != nil {
+		return err
+	}
+
 	// Commit the SQL transaction
 	ctx := context.TODO()
 	if err := p.tx.Commit(ctx); err != nil {
--- a/persistence/context.go
+++ b/persistence/context.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pokt-network/pocket/persistence/indexer"
 	coreTypes "github.com/pokt-network/pocket/shared/core/types"
 	"github.com/pokt-network/pocket/shared/modules"
+	"go.uber.org/multierr"
 )
 
 var _ modules.PersistenceRWContext = &PostgresContext{}
@@ -47,15 +48,18 @@ func (p *PostgresContext) SetSavePoint() error {
 // RollbackToSavepoint triggers a rollback for the current pgx transaction and the underylying submodule stores.
 func (p *PostgresContext) RollbackToSavePoint() error {
 	ctx, _ := p.getCtxAndTx()
-	err := p.tx.Rollback(ctx)
-	p.stateTrees.Rollback()
-	return err
+	pgErr := p.tx.Rollback(ctx)
+	treesErr := p.stateTrees.Rollback()
+	return multierr.Combine(pgErr, treesErr)
 }
 
 // Full details in the thread from the PR review: https://github.com/pokt-network/pocket/pull/285#discussion_r1018471719
 func (p *PostgresContext) ComputeStateHash() (string, error) {
 	stateHash, err := p.stateTrees.Update(p.tx, uint64(p.Height))
 	if err != nil {
+		return "", err
+	}
+	if err := p.stateTrees.Commit(); err != nil {
 		return "", err
 	}
 	p.stateHash = stateHash
--- a/persistence/context.go
+++ b/persistence/context.go
@@ -46,7 +46,7 @@ func (p *PostgresContext) NewSavePoint() error {
 
 // RollbackToSavepoint triggers a rollback for the current pgx transaction and the underylying submodule stores.
 func (p *PostgresContext) RollbackToSavePoint() error {
-	ctx := context.TODO()
+	ctx, _ := p.getCtxAndTx()
 	err := p.tx.Rollback(ctx)
 	p.stateTrees.Rollback()
 	return err
--- a/persistence/context.go
+++ b/persistence/context.go
@@ -36,8 +36,8 @@ type PostgresContext struct {
 	networkId string
 }
 
-// NewSavePoint generates a new Savepoint for this context.
-func (p *PostgresContext) NewSavePoint() error {
+// SetSavePoint generates a new Savepoint for this context.
+func (p *PostgresContext) SetSavePoint() error {
 	if err := p.stateTrees.Savepoint(); err != nil {
 		return err
 	}
--- a/persistence/context.go
+++ b/persistence/context.go
@@ -81,11 +81,6 @@ func (p *PostgresContext) Commit(proposerAddr, quorumCert []byte) error {
 		return err
 	}
 
-	// Commit the tree store updates
-	if err := p.stateTrees.Commit(); err != nil {
-		return err
-	}
-
 	// Commit the SQL transaction
 	ctx := context.TODO()
 	if err := p.tx.Commit(ctx); err != nil {
--- a/persistence/context.go
+++ b/persistence/context.go
@@ -36,18 +36,22 @@ type PostgresContext struct {
 	networkId string
 }
 
-func (p *PostgresContext) NewSavePoint(bytes []byte) error {
-	p.logger.Info().Bool("TODO", true).Msg("NewSavePoint not implemented")
+// NewSavePoint generates a new Savepoint for this context.
+func (p *PostgresContext) NewSavePoint() error {
+	if err := p.stateTrees.Savepoint(); err != nil {
+		return err
+	}
 	return nil
 }
 
-// TECHDEBT(#327): Guarantee atomicity betweens `prepareBlock`, `insertBlock` and `storeBlock` for save points & rollbacks.
-func (p *PostgresContext) RollbackToSavePoint(bytes []byte) error {
-	p.logger.Info().Bool("TODO", true).Msg("RollbackToSavePoint not fully implemented")
-	return p.tx.Rollback(context.TODO())
+// RollbackToSavepoint triggers a rollback for the current pgx transaction and the underylying submodule stores.
+func (p *PostgresContext) RollbackToSavePoint() error {
+	ctx := context.TODO()
+	err := p.tx.Rollback(ctx)
+	p.stateTrees.Rollback()
+	return err
 }
 
-// IMPROVE(#361): Guarantee the integrity of the state
 // Full details in the thread from the PR review: https://github.com/pokt-network/pocket/pull/285#discussion_r1018471719
 func (p *PostgresContext) ComputeStateHash() (string, error) {
 	stateHash, err := p.stateTrees.Update(p.tx, uint64(p.Height))
@@ -58,7 +62,6 @@ func (p *PostgresContext) ComputeStateHash() (string, error) {
 	return p.stateHash, nil
 }
 
-// TECHDEBT(#327): Make sure these operations are atomic
 func (p *PostgresContext) Commit(proposerAddr, quorumCert []byte) error {
 	p.logger.Info().Int64("height", p.Height).Msg("About to commit block & context")
 
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -1,0 +1,72 @@
+package trees
+
+import (
+	"encoding/hex"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	mock_types "github.com/pokt-network/pocket/persistence/types/mocks"
+	mockModules "github.com/pokt-network/pocket/shared/modules/mocks"
+	"github.com/rs/zerolog"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestTreeStore_AtomicUpdates(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	mockTxIndexer := mock_types.NewMockTxIndexer(ctrl)
+	mockBus := mockModules.NewMockBus(ctrl)
+	mockPersistenceMod := mockModules.NewMockPersistenceModule(ctrl)
+
+	mockBus.EXPECT().GetPersistenceModule().AnyTimes().Return(mockPersistenceMod)
+	mockPersistenceMod.EXPECT().GetTxIndexer().AnyTimes().Return(mockTxIndexer)
+
+	ts := &TreeStore{
+		logger:       &zerolog.Logger{},
+		Bus:          mockBus,
+		TXI:          mockBus.GetPersistenceModule().GetTxIndexer(),
+		TreeStoreDir: ":memory:",
+	}
+	ts.setupTrees()
+	require.NotEmpty(t, ts)
+
+	hash0 := ts.getStateHash()
+	require.NotEmpty(t, hash0)
+
+	err := ts.Savepoint()
+	require.NoError(t, err)
+
+	for _, treeName := range stateTreeNames {
+		err := ts.merkleTrees[treeName].tree.Update([]byte("foo"), []byte("bar"))
+		require.NoError(t, err)
+	}
+
+	hash1 := ts.getStateHash()
+	require.NotEmpty(t, hash1)
+	require.NotEqual(t, hash0, hash1)
+
+	err = ts.Savepoint()
+	require.NoError(t, err)
+	require.NotEmpty(t, ts.Prev.MerkleTrees)
+	require.NotEmpty(t, ts.Prev.RootTree)
+	require.Equal(t, hash1, hex.EncodeToString(ts.Prev.RootTree.tree.Root()))
+
+	hash2 := ts.getStateHash()
+	require.Equal(t, hash2, hash1)
+
+	for _, treeName := range stateTreeNames {
+		require.NotEmpty(t, ts.merkleTrees[treeName])
+		require.NotEmpty(t, ts.Prev.MerkleTrees[treeName])
+	}
+
+	for _, treeName := range stateTreeNames {
+		err := ts.merkleTrees[treeName].tree.Update([]byte("fiz"), []byte("buz"))
+		require.NoError(t, err)
+	}
+
+	ts.Rollback()
+
+	hash3 := ts.getStateHash()
+	require.Equal(t, hash3, hash2)
+}
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -26,29 +26,26 @@ func TestTreeStore_AtomicUpdates(t *testing.T) {
 		logger:       &zerolog.Logger{},
 		treeStoreDir: ":memory:",
 	}
-	err := ts.setupTrees()
-	require.NoError(t, err)
-	require.NotEmpty(t, ts)
+	require.NoError(t, ts.setupTrees())
+	require.NotEmpty(t, ts.merkleTrees[TransactionsTreeName])
 
 	hash0 := ts.getStateHash()
 	require.NotEmpty(t, hash0)
 
-	err = ts.Savepoint()
-	require.NoError(t, err)
+	require.NoError(t, ts.Savepoint())
 
 	for _, treeName := range stateTreeNames {
 		err := ts.merkleTrees[treeName].tree.Update([]byte("foo"), []byte("bar"))
 		require.NoError(t, err)
 	}
-	err = ts.Commit()
-	require.NoError(t, err)
+
+	require.NoError(t, ts.Commit())
 
 	hash1 := ts.getStateHash()
 	require.NotEmpty(t, hash1)
 	require.NotEqual(t, hash0, hash1)
 
-	err = ts.Savepoint()
-	require.NoError(t, err)
+	require.NoError(t, ts.Savepoint())
 	require.NotEmpty(t, ts.Prev.MerkleTrees)
 	require.NotEmpty(t, ts.Prev.RootTree)
 	require.Equal(t, hash1, hex.EncodeToString(ts.Prev.RootTree.tree.Root()))
@@ -62,8 +59,7 @@ func TestTreeStore_AtomicUpdates(t *testing.T) {
 	}
 
 	for _, treeName := range stateTreeNames {
-		err := ts.merkleTrees[treeName].tree.Update([]byte("fiz"), []byte("buz"))
-		require.NoError(t, err)
+		require.NoError(t, ts.merkleTrees[treeName].tree.Update([]byte("fiz"), []byte("buz")))
 	}
 
 	ts.Rollback()
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -40,6 +40,8 @@ func TestTreeStore_AtomicUpdates(t *testing.T) {
 		err := ts.merkleTrees[treeName].tree.Update([]byte("foo"), []byte("bar"))
 		require.NoError(t, err)
 	}
+	err = ts.Commit()
+	require.NoError(t, err)
 
 	hash1 := ts.getStateHash()
 	require.NotEmpty(t, hash1)
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -82,7 +82,8 @@ func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 	}
 
 	// rollback the changes made to the trees above BEFORE anything was committed
-	ts.Rollback()
+	err := ts.Rollback()
+	require.NoError(t, err)
 
 	// validate that the state hash is unchanged after new data was inserted but rolled back before commitment
 	hash3 := ts.getStateHash()
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -13,6 +13,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	// the root hash of a tree store where each tree is empty but present and initialized
+	h0 = "302f2956c084cc3e0e760cf1b8c2da5de79c45fa542f68a660a5fc494b486972"
+	// the root hash of a tree store where each tree has has key foo value bar added to it
+	h1 = "7d5712ea1507915c40e295845fa58773baa405b24b87e9d99761125d826ff915"
+)
+
 func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 	ctrl := gomock.NewController(t)
 
@@ -32,6 +39,7 @@ func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 
 	hash0 := ts.getStateHash()
 	require.NotEmpty(t, hash0)
+	require.Equal(t, hash0, h0)
 
 	require.NoError(t, ts.Savepoint())
 
@@ -48,6 +56,7 @@ func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 	hash1 := ts.getStateHash()
 	require.NotEmpty(t, hash1)
 	require.NotEqual(t, hash0, hash1)
+	require.Equal(t, hash1, h1)
 
 	// set a new savepoint
 	require.NoError(t, ts.Savepoint())
@@ -56,8 +65,10 @@ func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 	// assert that savepoint creation doesn't mutate state hash
 	require.Equal(t, hash1, hex.EncodeToString(ts.PrevState.RootTree.tree.Root()))
 
+	// verify that creating a savepoint does not change state hash
 	hash2 := ts.getStateHash()
 	require.Equal(t, hash2, hash1)
+	require.Equal(t, hash2, h1)
 
 	// validate that state tree was updated and a previous savepoint is created
 	for _, treeName := range stateTreeNames {
@@ -76,4 +87,5 @@ func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 	// validate that the state hash is unchanged after new data was inserted but rolled back before commitment
 	hash3 := ts.getStateHash()
 	require.Equal(t, hash3, hash2)
+	require.Equal(t, hash3, h1)
 }
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -22,19 +22,18 @@ func TestTreeStore_AtomicUpdates(t *testing.T) {
 	mockBus.EXPECT().GetPersistenceModule().AnyTimes().Return(mockPersistenceMod)
 	mockPersistenceMod.EXPECT().GetTxIndexer().AnyTimes().Return(mockTxIndexer)
 
-	ts := &TreeStore{
+	ts := &treeStore{
 		logger:       &zerolog.Logger{},
-		Bus:          mockBus,
-		TXI:          mockBus.GetPersistenceModule().GetTxIndexer(),
-		TreeStoreDir: ":memory:",
+		treeStoreDir: ":memory:",
 	}
-	ts.setupTrees()
+	err := ts.setupTrees()
+	require.NoError(t, err)
 	require.NotEmpty(t, ts)
 
 	hash0 := ts.getStateHash()
 	require.NotEmpty(t, hash0)
 
-	err := ts.Savepoint()
+	err = ts.Savepoint()
 	require.NoError(t, err)
 
 	for _, treeName := range stateTreeNames {
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -60,10 +60,10 @@ func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 
 	// set a new savepoint
 	require.NoError(t, ts.Savepoint())
-	require.NotEmpty(t, ts.PrevState.MerkleTrees)
-	require.NotEmpty(t, ts.PrevState.RootTree)
+	require.NotEmpty(t, ts.prevState.merkleTrees)
+	require.NotEmpty(t, ts.prevState.rootTree)
 	// assert that savepoint creation doesn't mutate state hash
-	require.Equal(t, hash1, hex.EncodeToString(ts.PrevState.RootTree.tree.Root()))
+	require.Equal(t, hash1, hex.EncodeToString(ts.prevState.rootTree.tree.Root()))
 
 	// verify that creating a savepoint does not change state hash
 	hash2 := ts.getStateHash()
@@ -73,7 +73,7 @@ func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 	// validate that state tree was updated and a previous savepoint is created
 	for _, treeName := range stateTreeNames {
 		require.NotEmpty(t, ts.merkleTrees[treeName])
-		require.NotEmpty(t, ts.PrevState.MerkleTrees[treeName])
+		require.NotEmpty(t, ts.prevState.merkleTrees[treeName])
 	}
 
 	// insert additional test data into all of the trees
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -31,7 +31,7 @@ func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 	mockPersistenceMod.EXPECT().GetTxIndexer().AnyTimes().Return(mockTxIndexer)
 
 	ts := &treeStore{
-		logger:       logger.Global.CreateLoggerForModule(modules.TreeStoreSubmoduleName), // TODO
+		logger:       logger.Global.CreateLoggerForModule(modules.TreeStoreSubmoduleName),
 		treeStoreDir: ":memory:",
 	}
 	require.NoError(t, ts.setupTrees())
--- a/persistence/trees/atomic_test.go
+++ b/persistence/trees/atomic_test.go
@@ -5,14 +5,15 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	"github.com/pokt-network/pocket/logger"
 	mock_types "github.com/pokt-network/pocket/persistence/types/mocks"
+	"github.com/pokt-network/pocket/shared/modules"
 	mockModules "github.com/pokt-network/pocket/shared/modules/mocks"
-	"github.com/rs/zerolog"
 
 	"github.com/stretchr/testify/require"
 )
 
-func TestTreeStore_AtomicUpdates(t *testing.T) {
+func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
 	ctrl := gomock.NewController(t)
 
 	mockTxIndexer := mock_types.NewMockTxIndexer(ctrl)
@@ -23,7 +24,7 @@ func TestTreeStore_AtomicUpdates(t *testing.T) {
 	mockPersistenceMod.EXPECT().GetTxIndexer().AnyTimes().Return(mockTxIndexer)
 
 	ts := &treeStore{
-		logger:       &zerolog.Logger{},
+		logger:       logger.Global.CreateLoggerForModule(modules.TreeStoreSubmoduleName), // TODO
 		treeStoreDir: ":memory:",
 	}
 	require.NoError(t, ts.setupTrees())
@@ -34,36 +35,45 @@ func TestTreeStore_AtomicUpdates(t *testing.T) {
 
 	require.NoError(t, ts.Savepoint())
 
+	// insert test data into every tree
 	for _, treeName := range stateTreeNames {
 		err := ts.merkleTrees[treeName].tree.Update([]byte("foo"), []byte("bar"))
 		require.NoError(t, err)
 	}
 
+	// commit the above changes
 	require.NoError(t, ts.Commit())
 
+	// assert state hash is changed
 	hash1 := ts.getStateHash()
 	require.NotEmpty(t, hash1)
 	require.NotEqual(t, hash0, hash1)
 
+	// set a new savepoint
 	require.NoError(t, ts.Savepoint())
-	require.NotEmpty(t, ts.Prev.MerkleTrees)
-	require.NotEmpty(t, ts.Prev.RootTree)
-	require.Equal(t, hash1, hex.EncodeToString(ts.Prev.RootTree.tree.Root()))
+	require.NotEmpty(t, ts.PrevState.MerkleTrees)
+	require.NotEmpty(t, ts.PrevState.RootTree)
+	// assert that savepoint creation doesn't mutate state hash
+	require.Equal(t, hash1, hex.EncodeToString(ts.PrevState.RootTree.tree.Root()))
 
 	hash2 := ts.getStateHash()
 	require.Equal(t, hash2, hash1)
 
+	// validate that state tree was updated and a previous savepoint is created
 	for _, treeName := range stateTreeNames {
 		require.NotEmpty(t, ts.merkleTrees[treeName])
-		require.NotEmpty(t, ts.Prev.MerkleTrees[treeName])
+		require.NotEmpty(t, ts.PrevState.MerkleTrees[treeName])
 	}
 
+	// insert additional test data into all of the trees
 	for _, treeName := range stateTreeNames {
 		require.NoError(t, ts.merkleTrees[treeName].tree.Update([]byte("fiz"), []byte("buz")))
 	}
 
+	// rollback the changes made to the trees above BEFORE anything was committed
 	ts.Rollback()
 
+	// validate that the state hash is unchanged after new data was inserted but rolled back before commitment
 	hash3 := ts.getStateHash()
 	require.Equal(t, hash3, hash2)
 }
--- a/persistence/trees/main_test.go
+++ b/persistence/trees/main_test.go
@@ -1,0 +1,12 @@
+//go:build test
+
+package trees
+
+import (
+	"crypto/sha256"
+	"hash"
+)
+
+type TreeStore = treeStore
+
+var SMTTreeHasher hash.Hash = sha256.New()
--- a/persistence/trees/module.go
+++ b/persistence/trees/module.go
@@ -17,8 +17,6 @@ func (*treeStore) Create(bus modules.Bus, options ...modules.TreeStoreOption) (m
 		option(m)
 	}
 
-	bus.RegisterModule(m)
-
 	if err := m.setupTrees(); err != nil {
 		return nil, err
 	}
--- a/persistence/trees/module.go
+++ b/persistence/trees/module.go
@@ -17,6 +17,8 @@ func (*treeStore) Create(bus modules.Bus, options ...modules.TreeStoreOption) (m
 		option(m)
 	}
 
+	bus.RegisterModule(m)
+
 	if err := m.setupTrees(); err != nil {
 		return nil, err
 	}
--- a/persistence/trees/module_test.go
+++ b/persistence/trees/module_test.go
@@ -142,7 +142,6 @@ func validatorId(i int) string {
 func createMockBus(t *testing.T, runtimeMgr modules.RuntimeMgr) *mockModules.MockBus {
 	t.Helper()
 	ctrl := gomock.NewController(t)
-
 	mockBus := mockModules.NewMockBus(ctrl)
 	mockModulesRegistry := mockModules.NewMockModulesRegistry(ctrl)
 
@@ -171,5 +170,6 @@ func createMockBus(t *testing.T, runtimeMgr modules.RuntimeMgr) *mockModules.Moc
 	mockBus.EXPECT().
 		PublishEventToBus(gomock.Any()).
 		AnyTimes()
+
 	return mockBus
 }
--- a/persistence/trees/module_test.go
+++ b/persistence/trees/module_test.go
@@ -48,7 +48,6 @@ func TestTreeStore_Create(t *testing.T) {
 
 	treemod, err := trees.Create(mockBus, trees.WithTreeStoreDirectory(":memory:"))
 	assert.NoError(t, err)
-
 	got := treemod.GetBus()
 	assert.Equal(t, got, mockBus)
 
--- a/persistence/trees/module_test.go
+++ b/persistence/trees/module_test.go
@@ -142,6 +142,7 @@ func validatorId(i int) string {
 func createMockBus(t *testing.T, runtimeMgr modules.RuntimeMgr) *mockModules.MockBus {
 	t.Helper()
 	ctrl := gomock.NewController(t)
+
 	mockBus := mockModules.NewMockBus(ctrl)
 	mockModulesRegistry := mockModules.NewMockModulesRegistry(ctrl)
 
@@ -170,6 +171,5 @@ func createMockBus(t *testing.T, runtimeMgr modules.RuntimeMgr) *mockModules.Moc
 	mockBus.EXPECT().
 		PublishEventToBus(gomock.Any()).
 		AnyTimes()
-
 	return mockBus
 }
--- a/persistence/trees/trees.go
+++ b/persistence/trees/trees.go
@@ -260,9 +260,6 @@ func (t *treeStore) updateMerkleTrees(pgtx pgx.Tx, txi indexer.TxIndexer, height
 		}
 	}
 
-	if err := t.Commit(); err != nil {
-		return "", fmt.Errorf("failed to commit: %w", err)
-	}
 	return t.getStateHash(), nil
 }
 
--- a/persistence/trees/trees.go
+++ b/persistence/trees/trees.go
@@ -305,7 +305,7 @@ func (t *treeStore) getStateHash() string {
 ////////////////////////////////
 
 // Savepoint generates a new savepoint for the tree store and saves it internally.
-func (t *TreeStore) Savepoint() error {
+func (t *treeStore) Savepoint() error {
 	w, err := t.save()
 	if err != nil {
 		return err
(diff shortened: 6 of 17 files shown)


shared/modules/persistence_module.go
```diff
@@ -37,6 +37,18 @@ type PersistenceModule interface {
 	HandleDebugMessage(*messaging.DebugMessage) error
 }
 
+// Tx is fed to an AtomicStore by a StoreManager to atomically apply a block to the persistence layer.
+type Tx struct {
```
h5law, line 41: Can we not just use the block protobuf directly? I feel adding this `Tx` struct confuses with `Transaction` struct and `pgx.Tx`

dylanlott: This is gone now, refactored it out cause the new approach didn't need it.

persistence/trees/module.go
```diff
@@ -24,35 +31,52 @@ func (*treeStore) Create(bus modules.Bus, options ...modules.TreeStoreOption) (m
...
+		if ok {
+			mod.TXI = txi
 		}
 	}
 }
 
-func (t *treeStore) setupTrees() error {
-	if t.treeStoreDir == ":memory:" {
+func (t *TreeStore) GetModuleName() string  { return modules.TreeStoreModuleName }
+func (t *TreeStore) Start() error           { return nil }
```
h5law, line 69: Is there a reason this is no longer a submodule when the Start and Stop functions are noop?

persistence/trees/module_test.go
```diff
@@ -0,0 +1,101 @@
...
+func TestTreeStore_Create(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockRuntimeMgr := mockModules.NewMockRuntimeMgr(ctrl)
+	mockBus := createMockBus(t, mockRuntimeMgr)
+
+	genesisStateMock := createMockGenesisState(nil)
+	persistenceMock := preparePersistenceMock(t, mockBus, genesisStateMock)
+
+	mockBus.EXPECT().GetPersistenceModule().Return(persistenceMock).AnyTimes()
+	persistenceMock.EXPECT().GetBus().AnyTimes().Return(mockBus)
```
h5law, line 33:
```suggestion

```

Don't think you need this as you already cover this in the persistence mock creation

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,277 @@
...
+
+		context.RollbackToSavePoint()
+
+		_, err = context.ComputeStateHash()
+		assert.Error(t, err)
+	})
+}
+
+// createMockBus returns a mock bus with stubbed out functions for bus registration
+func createMockBus(t *testing.T, runtimeMgr modules.RuntimeMgr) *mockModules.MockBus {
```
h5law, line 99: NIT: I would consider moving all these helpers to a file named `trees_utils_test.go`

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,277 @@
...
+	runtimeMgr := runtime.NewManager(cfg, genesisState)
+
+	bus, err := runtime.CreateBus(runtimeMgr)
+	assert.NoError(t, err)
+
+	persistenceMod, err := persistence.Create(bus)
+	assert.NoError(t, err)
+
+	return persistenceMod.(modules.PersistenceModule)
+}
```
h5law, line 151:
```suggestion
}

```

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,277 @@
...
+func createAndInsertDefaultTestApp(db *persistence.PostgresContext) (*coreTypes.Actor, error) {
```
h5law, line 152: can we not pass in `t *testing.T` here to avoid the fatal logs?

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,277 @@
...
+		DefaultStake,
+		DefaultChains,
+		DefaultPauseHeight,
+		DefaultUnstakingHeight)
 }
 
-func TestTreeStore_New(t *testing.T) {
-	// TODO: Write test case for the NewStateTrees function
-	t.Skip("TODO: Write test case for NewStateTrees function")
+func newTestApp() (*coreTypes.Actor, error) {
```
h5law, line 183: Same as above - pass in `t *testing.T`

runtime/bus.go
```diff
@@ -71,6 +71,10 @@ func (m *bus) GetPersistenceModule() modules.PersistenceModule {
 	return getModuleFromRegistry[modules.PersistenceModule](m, modules.PersistenceModuleName)
 }
 
+func (m *bus) GetTreeStoreModule() modules.TreeStoreModule {
```
h5law, line 74: Is this the reason you are making TreeStore a full module? If so #855 addresses this and allows Submodules to be called in this way

shared/modules/persistence_module.go
```diff
@@ -47,6 +47,13 @@ type PersistenceModule interface {
 	GetLocalContext() (PersistenceLocalContext, error)
 }
 
+// AtomicStore defines the interface for stores to implement to guarantee atomic commits to the persistence layer
+type AtomicStore interface {
```
h5law, line 51: :fire:

persistence/trees/module.go
dylanlott: I think this is necessary because the submodule stuff isn't merged into main yet and I'm not rebased on top of that branch.

h5law: You could base off of `ibc/initial_stores_olsh..._delete_later` which includes the `ibc/initial_stores` and the submodule PRs if needed am using it as interim until merged in main

dylanlott: I have rebased this on top of #896 which covers the Submodule updates.

persistence/context.go
```diff
@@ -36,18 +36,22 @@ type PostgresContext struct {
...
 	return nil
 }
 
-// TECHDEBT(#327): Guarantee atomicity betweens `prepareBlock`, `insertBlock` and `storeBlock` for save points & rollbacks.
-func (p *PostgresContext) RollbackToSavePoint(bytes []byte) error {
-	p.logger.Info().Bool("TODO", true).Msg("RollbackToSavePoint not fully implemented")
-	return p.tx.Rollback(context.TODO())
+// RollbackToSavepoint triggers a rollback for the current pgx transaction and the underylying submodule stores.
+func (p *PostgresContext) RollbackToSavePoint() error {
+	ctx := context.TODO()
```
h5law, line 49:
```suggestion
	ctx, _ := p.getCtxAndTx
```
Unsure if this will affect things but each `PostgresContext` already has a `context.Context` so we could reuse it?

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,73 @@
...
+	require.NotEmpty(t, hash0)
+
+	err = ts.Savepoint()
+	require.NoError(t, err)
+
+	for _, treeName := range stateTreeNames {
+		err := ts.merkleTrees[treeName].tree.Update([]byte("foo"), []byte("bar"))
+		require.NoError(t, err)
+	}
+	err = ts.Commit()
```
h5law, line 43:
```suggestion
	require.NoError(t, ts.Commit())
```

persistence/trees/trees.go
```diff
@@ -1,7 +1,17 @@
 // package trees maintains a set of sparse merkle trees
-// each backed by the KVStore interface. It offers an atomic
+// each backed by the `KVStore` interface. It offers an atomic
 // commit and rollback mechanism for interacting with
 // that core resource map of merkle trees.
```
h5law, line 4: "that core resource" doesnt make sense to me

persistence/trees/trees.go
```diff
@@ -247,6 +255,77 @@ func (t *TreeStore) getStateHash() string {
...
+// save commits any pending changes to the trees and serializes the treeStore to a WorldState object
+// with new readers and writers.
+func (t *TreeStore) save() (*Worldstate, error) {
+	if err := t.Commit(); err != nil {
+		return nil, err
+	}
+
+	w := &Worldstate{
+		TreeStoreDir: t.treeStoreDir,
+		MerkleTrees:  map[string]*stateTree{},
```
h5law, line 305:
```suggestion
		MerkleTrees:  map[string][]byte,
```
My personal opinion here is that we should store the tree name and root hash and then during the load phase we should do the importing. This would save time as we are more likely to call `save` a lot more than `load` so allocating the SMTs in memory each save seems overkill when it is during load that we need them.

During load we can use `w.TreeStoreDir` and the same pattern we use during `setupTrees()` to get each tree's nodestore

persistence/trees/trees.go
```diff
@@ -247,6 +255,77 @@ func (t *TreeStore) getStateHash() string {
...
+		w.MerkleTrees[treeName] = &stateTree{
+			name:      treeName,
+			tree:      tree,
+			nodeStore: nodeStore,
+		}
+	}
+
+	root, nodeStore := t.GetTree(RootTreeName)
+	tree := smt.ImportSparseMerkleTree(nodeStore, smtTreeHasher, root)
+	w.RootTree = &stateTree{
```
h5law, line 320: See the comment above

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,216 @@
...
+	genesisStateNumApplications = 1
+)
 
-// TECHDEBT(#836): Tests added in https://github.com/pokt-network/pocket/pull/836
 func TestTreeStore_Update(t *testing.T) {
-	// TODO: Write test case for the Update method
-	t.Skip("TODO: Write test case for Update method")
+	pool, resource, dbUrl := test_artifacts.SetupPostgresDocker()
+	t.Cleanup(func() {
+		if err := pool.Purge(resource); err != nil {
```
h5law, line 47:
```suggestion
		require.NoError(t, pool.Purge(resource))
```

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,216 @@
...
+		genesisStateNumValidators,
+		genesisStateNumServicers,
+		genesisStateNumApplications,
+		genesisStateNumServicers,
+	)
+
+	runtimeMgr := runtime.NewManager(cfg, genesisState)
+
+	bus, err := runtime.CreateBus(runtimeMgr)
+	assert.NoError(t, err)
```
h5law, line 96: Is there a reason for `assert` over `require`?

AFAIK `assert` fails but keeps the test running, whereas `require` panics on the error.

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,216 @@
...
+
+	persistenceMod, err := persistence.Create(bus)
+	assert.NoError(t, err)
+
+	return persistenceMod.(modules.PersistenceModule)
 }
+func createAndInsertDefaultTestApp(t *testing.T, db *persistence.PostgresContext) (*coreTypes.Actor, error) {
+	app := newTestApp(t)
```
h5law, line 104:
```suggestion
	t.Helper()
	app := newTestApp(t)
```

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,216 @@
...
+		Chains:          DefaultChains,
+		StakedAmount:    DefaultStake,
+		PausedHeight:    DefaultPauseHeight,
+		UnstakingHeight: DefaultUnstakingHeight,
+		Output:          hex.EncodeToString(outputAddr),
+	}
+}
+
+func NewTestPostgresContext(t testing.TB, height int64, testPersistenceMod modules.PersistenceModule) *persistence.PostgresContext {
+	rwCtx, err := testPersistenceMod.NewRWContext(height)
```
h5law, line 145: `testing.TB` also has the [`Helper()` method](https://pkg.go.dev/testing#TB)

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,216 @@
...
+	// TECHDEBT: This should not be part of `NewTestPostgresContext`. It causes unnecessary resets
+	// if we call `NewTestPostgresContext` more than once in a single test.
+	t.Cleanup(func() {
+		resetStateToGenesis(testPersistenceMod)
+	})
+
+	return postgresCtx
+}
+
+func NewTestPersistenceModule(t *testing.T, databaseUrl string) modules.PersistenceModule {
```
h5law, line 164: Unsure if this is in progress, but can we add a `TECHDEBT` or `TODO` comment on some of these functions to move them into `internal` for global use? I have copies of these functions in IBC pretty sure as there aren't any globally accessable ones.

shared/modules/persistence_module.go
```diff
@@ -1,6 +1,6 @@
 package modules
 
-//go:generate mockgen -destination=./mocks/persistence_module_mock.go github.com/pokt-network/pocket/shared/modules PersistenceModule,PersistenceRWContext,PersistenceReadContext,PersistenceWriteContext,PersistenceLocalContext
+//go:generate mockgen -destination=./mocks/persistence_module_mock.go github.com/pokt-network/pocket/shared/modules PersistenceModule,PersistenceRWContext,PersistenceReadContext,PersistenceWriteContext,PersistenceLocalContext,TreeStoreModule
```
h5law, line 3: any reason we are mocking it within persistence now it is a submodule?

shared/modules/treestore_module.go
```diff
@@ -5,7 +5,7 @@ import (
 	"github.com/pokt-network/pocket/persistence/kvstore"
 )
 
-//go:generate mockgen -destination=./mocks/treestore_module_mock.go github.com/pokt-network/pocket/shared/modules TreeStoreModule
+// //go:generate mockgen -destination=./mocks/treestore_module_mock.go github.com/pokt-network/pocket/shared/modules TreeStoreModule
```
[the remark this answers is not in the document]
dylanlott: Yes, this should be deleted, I commented it out to test that it fixed the tests (duplicate mock generation was breaking compilation) so I will push a commit that removes this line.

h5law (changes requested):
First off, amazing work 🚀 I love it <3

To clarify: this PR does **not** contain the rollbacks? Only the savepoints?

Left a few comments my only major one is around the `save` function, I think we should do the importing during the `load` method as this will be ran less frequently 🤞🏼

persistence/trees/trees_test.go
dylanlott: This is "in progress" in that it's being discussed, but what you're describing is similar to what we've identified in the [TestUtils](https://www.notion.so/pocketnetwork/Testutils-9cba9010e18447248e9daa8a3b87e3f2#f1d7f404918541bc86ccde07980e3f09) document. I'll add a TECHDEBT comment here, though. This is definitely a function that would benefit from being in a clearly visible & central place for testing.

shared/modules/persistence_module.go
dylanlott: I believe I have some tests that aren't in PRs yet that will use mocked TreeStores.

h5law: I meant moreso about the location of the mock `persistence_module_mock.go` vs `treestore_module_mock.go`

persistence/context.go
dylanlott: I initially avoided the getCtxAndTx method because of the fact we haven't addressed context handling yet, but it shouldn't effect anything to use the getCtxAndTx method here. I will add a TECHDEBT comment about the context handling, too.

shared/modules/persistence_module.go
dylanlott: Oh, I see what you meant now. Updated to put it in treestore_module instead.

persistence/trees/trees.go
dylanlott: This approach comes at the expense of complicating `Rollback` logic with the load functionality rather than the simple reassignment that `TreeStore` uses right now. That could have longer term consequences where maybe a node gets itself into a position where it can't rollback but it can't move forward, either. Having less complicated and stronger guarantees at the `Rollback` point is safer, IMO.

I suggest addressing this in #897 because saving a `Worldstate` is slightly different than making a `Savepoint`, and I think that's important to handle correctly. I'll add an `OPTIMIZE` comment though, because I think that approach is worth exploring.

dylanlott:
> First off, amazing work 🚀 I love it <3
> 
> To clarify: this PR does **not** contain the rollbacks? Only the savepoints?

This PR implements the rollback functionality for block application in `utility/unit_of_work/module.go`.

> Left a few comments my only major one is around the `save` function, I think we should do the importing during the `load` method as this will be ran less frequently 🤞🏼

I'm going to politely push back on this for now, with the caveat that I added an OPTIMIZE comment, but I do think this is a premature optimization with unclear benefits when weighed against the added `Rollback` function complexity. This should likely be examined in an ADR that more clearly discusses the trade-offs.

Olshansk:
@dylanlott Please add me as a reviewer if/when you need additional feedback on this.

persistence/trees/trees.go
dylanlott: Reworded it to make it clearer 👍

persistence/trees/trees.go
```diff
@@ -251,6 +272,66 @@ func (t *treeStore) getStateHash() string {
...
+	if err != nil {
+		return err
+	}
+	t.Prev = w
+	return nil
+}
+
+// Rollback intentionally can't return an error because at this point we're out of tricks
+// to recover from problems.
+func (t *treeStore) Rollback() {
```
h5law, line 291: I think I got confused between rollback and load save point we previously spoke about

h5law:
> I'm going to politely push back on this for now, with the caveat that I added an OPTIMIZE comment, but I do think this is a premature optimization with unclear benefits when weighed against the added `Rollback` function complexity. This should likely be examined in an ADR that more clearly discusses the trade-offs.

I see your point, I missed the Rollback function in Trees, I got confused between it and the load world state function you had previously discussed with me offline. I think you are right that at a later date we can look into where the imports should go. I will rereview now and approve these as appart from that everything else was addressed 👍🏼

h5law (approved):
Left a reply around the ADR idea at a later date. I think for me it makes more logical sense for the imports to be where the rollback occurs but you are right this can and should be discussed and addressed later it is not necessary this moment in time.

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+	hash1 := ts.getStateHash()
+	require.NotEmpty(t, hash1)
+	require.NotEqual(t, hash0, hash1)
+
+	require.NoError(t, ts.Savepoint())
+	require.NotEmpty(t, ts.Prev.MerkleTrees)
+	require.NotEmpty(t, ts.Prev.RootTree)
+	require.Equal(t, hash1, hex.EncodeToString(ts.Prev.RootTree.tree.Root()))
+
+	hash2 := ts.getStateHash()
```
Olshansk, line 53: Feel free to modify if I'm wrong on this

```suggestion
	// verify that creating a Savepoint does not commit anything to disk
	hash2 := ts.getStateHash()
```

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+		require.NoError(t, err)
+	}
+
+	require.NoError(t, ts.Commit())
+
+	hash1 := ts.getStateHash()
+	require.NotEmpty(t, hash1)
+	require.NotEqual(t, hash0, hash1)
+
+	require.NoError(t, ts.Savepoint())
```
Olshansk, line 48: Can you add a small comment above these 4 lines on what you're validating?

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	hash2 := ts.getStateHash()
+	require.Equal(t, hash2, hash1)
+
+	for _, treeName := range stateTreeNames {
+		require.NotEmpty(t, ts.merkleTrees[treeName])
+		require.NotEmpty(t, ts.Prev.MerkleTrees[treeName])
+	}
+
+	for _, treeName := range stateTreeNames {
```
Olshansk, line 61:
```suggestion
	// insert additional test data into all of the trees
	for _, treeName := range stateTreeNames {
```

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+		require.NoError(t, ts.merkleTrees[treeName].tree.Update([]byte("fiz"), []byte("buz")))
+	}
+
+	ts.Rollback()
```
Olshansk, line 65:
```suggestion
	// rollback the changes made to the trees above BEFORE anything was committed
	ts.Rollback()
```

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	require.NoError(t, ts.Savepoint())
+	require.NotEmpty(t, ts.Prev.MerkleTrees)
+	require.NotEmpty(t, ts.Prev.RootTree)
+	require.Equal(t, hash1, hex.EncodeToString(ts.Prev.RootTree.tree.Root()))
+
+	hash2 := ts.getStateHash()
+	require.Equal(t, hash2, hash1)
+
+	for _, treeName := range stateTreeNames {
```
Olshansk, line 56: Can you add a comment on what this forloop is validating?

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+		require.NotEmpty(t, ts.Prev.MerkleTrees[treeName])
+	}
+
+	for _, treeName := range stateTreeNames {
+		require.NoError(t, ts.merkleTrees[treeName].tree.Update([]byte("fiz"), []byte("buz")))
+	}
+
+	ts.Rollback()
+
+	hash3 := ts.getStateHash()
```
Olshansk, line 67:
```suggestion
	// validate that the state hash is unchanged after new data was inserted but rolled back before commitment
	hash3 := ts.getStateHash()
```

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/golang/mock/gomock"
+	mock_types "github.com/pokt-network/pocket/persistence/types/mocks"
+	mockModules "github.com/pokt-network/pocket/shared/modules/mocks"
+	"github.com/rs/zerolog"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestTreeStore_AtomicUpdates(t *testing.T) {
```
Olshansk, line 15:
```suggestion
func TestTreeStore_AtomicUpdatesWithSuccessfulRollback(t *testing.T) {
```

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for _, treeName := range stateTreeNames {
+		require.NoError(t, ts.merkleTrees[treeName].tree.Update([]byte("fiz"), []byte("buz")))
+	}
+
+	ts.Rollback()
+
+	hash3 := ts.getStateHash()
+	require.Equal(t, hash3, hash2)
+}
```
Olshansk, line 69: Can we add a test for a failed rollback somehow and have it catch the fatal log?

persistence/trees/trees.go
```diff
@@ -1,7 +1,17 @@
...
+// `Commit` must be called after any `Update` calls to persist changes.
+// `Savepoint` is first called to create a new anchor in time.
+// `Update` is called, which will fetch and apply the changes to their
+// respective trees. Finally, `Commit` is called, which persists the
+// currently applied mutations to disk. If `Rollback` is called at any
+// point before committing, it rolls the TreeStore state back to the
+// earlier savepoint. This means that the caller is responsible for
+// correctly managing atomic updates of the TreeStore. In most contexts
+// this is from the perspective of the `utility/unit_of_work` package.
+
```
Olshansk, line 14:
```suggestion
// its core resource - a set of merkle trees.

// - `Update` is called, which will fetch and apply the contextual changes to the respective trees. 
// - `Savepoint` is first called to create a new anchor in time that can be rolled back to
// - `Commit` must be called after any `Update` calls to persist changes applied to disk.
// - If `Rollback` is called at any point before committing, it rolls the TreeStore state back to the
//    earlier savepoint. This means that the caller is responsible for correctly managing atomic updates
//     of the TreeStore.
// In most contexts, this is from the perspective of the `utility/unit_of_work` package.

```

persistence/trees/trees.go
```diff
@@ -88,6 +98,17 @@ type treeStore struct {
 	treeStoreDir string
 	rootTree     *stateTree
 	merkleTrees  map[string]*stateTree
+
+	// Worldstate holds a previous view of the Worldstate
+	// for savepoints and rollbacks purposes.
+	Prev *Worldstate
```
Olshansk, line 104: Can you update the name to something more self-explanatory like `PrevState` or `PrevWorlState` or `PrevCommittedState`?

persistence/trees/trees.go
```diff
@@ -88,6 +98,17 @@ type treeStore struct {
...
+}
+
+// Worldstate holds a ser/deserializable view of the entire tree state.
```
Olshansk, line 107: OPTIONAL NIT
```suggestion
// Worldstate holds a (de)serializable view of the entire tree state.
```

persistence/trees/trees.go
```diff
@@ -251,6 +272,66 @@ func (t *treeStore) getStateHash() string {
...
+func (t *treeStore) Savepoint() error {
+	w, err := t.save()
+	if err != nil {
+		return err
+	}
+	t.Prev = w
+	return nil
+}
+
+// Rollback intentionally can't return an error because at this point we're out of tricks
```
Olshansk, line 289:
```suggestion
// Rollback rolls back to the last saved world state maintained by the treeStore.
// It intentionally can't return an error because at this point we're out of tricks
```

persistence/trees/trees.go
```diff
@@ -251,6 +272,66 @@ func (t *treeStore) getStateHash() string {
...
+// to recover from problems.
+func (t *treeStore) Rollback() {
+	if t.Prev != nil {
+		t.merkleTrees = t.Prev.MerkleTrees
```
Olshansk, line 293: Do we need to do any sort of release or is it all magic?

persistence/trees/trees.go
```diff
@@ -251,6 +272,66 @@ func (t *treeStore) getStateHash() string {
...
+	if err := t.Commit(); err != nil {
+		return nil, err
+	}
+
+	w := &Worldstate{
+		TreeStoreDir: t.treeStoreDir,
+		MerkleTrees:  map[string]*stateTree{},
+	}
+
+	for treeName := range t.merkleTrees {
```
Olshansk, line 314: Can you add some inline comments about how this creates a whole new copy of the trees?

persistence/trees/trees.go
```diff
@@ -251,6 +272,66 @@ func (t *treeStore) getStateHash() string {
...
+		root, nodeStore := t.GetTree(treeName)
+		tree := smt.ImportSparseMerkleTree(nodeStore, smtTreeHasher, root)
+		w.MerkleTrees[treeName] = &stateTree{
```
Olshansk, line 317: Given that the directory where they are stored is the same (with the same name), how do we avoid overrides?

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,225 @@
...
+)
+
+var (
+	DefaultChains = []string{"0001"}
+
+	DefaultDeltaBig   = big.NewInt(100)
+	DefaultAccountBig = big.NewInt(1000000)
+	DefaultStakeBig   = big.NewInt(1000000000000000)
+
+	DefaultAccountAmount = utils.BigIntToString(DefaultAccountBig)
```
Olshansk, line 29: 1. Cant find where `DefaultAccountAmount` is used in this file
2. If we do need it, and we need it to be customized for this local test, make it lowercase and update the test
3. If we can use the default, why not just import from `persistence/test/setup_test.go`?

DItto to other variables too.

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,225 @@
...
+
+	t.Run("should update actor tree and commit", func(t *testing.T) {
+		pmod := newTestPersistenceModule(t, dbUrl)
+		context := NewTestPostgresContext(t, 0, pmod)
+
+		require.NoError(t, context.NewSavePoint())
+
+		hash1, err := context.ComputeStateHash()
+		require.NoError(t, err)
+		require.NotEmpty(t, hash1)
```
Olshansk, line 57: Can you validate against a constant to make the tests more annoying but also more robust?

Open to pushback

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,225 @@
...
-// TECHDEBT(#836): Tests added in https://github.com/pokt-network/pocket/pull/836
 func TestTreeStore_Update(t *testing.T) {
-	// TODO: Write test case for the Update method
-	t.Skip("TODO: Write test case for Update method")
+	pool, resource, dbUrl := test_artifacts.SetupPostgresDocker()
+	t.Cleanup(func() {
+		require.NoError(t, pool.Purge(resource))
+	})
+
+	t.Run("should update actor tree and commit", func(t *testing.T) {
```
Olshansk, line 49:
```suggestion
	t.Run("should update actor trees, commit and modify the state hash", func(t *testing.T) {
```

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,225 @@
...
+		require.NotEqual(t, hash1, hash2)
+	})
+}
+
+func newTestPersistenceModule(t *testing.T, databaseUrl string) modules.PersistenceModule {
+	t.Helper()
+	teardownDeterministicKeygen := keygen.GetInstance().SetSeed(42)
+	defer teardownDeterministicKeygen()
+
+	cfg := &configs.Config{
```
Olshansk, line 76: can you move this into a helper given how much its reused?

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,225 @@
...
+		PausedHeight:    DefaultPauseHeight,
+		UnstakingHeight: DefaultUnstakingHeight,
+		Output:          hex.EncodeToString(outputAddr),
+	}
 }
 
-func TestTreeStore_New(t *testing.T) {
-	// TODO: Write test case for the NewStateTrees function
-	t.Skip("TODO: Write test case for NewStateTrees function")
+func NewTestPostgresContext(t testing.TB, height int64, testPersistenceMod modules.PersistenceModule) *persistence.PostgresContext {
```
Olshansk, line 150: Does this need to be publically exposed?

persistence/trees/trees_test.go
```diff
@@ -1,19 +1,225 @@
...
+		outputBz,
+		false,
+		DefaultStakeStatus,
+		DefaultStake,
+		DefaultChains,
+		DefaultPauseHeight,
+		DefaultUnstakingHeight)
+}
+
+func newTestApp(t *testing.T) *coreTypes.Actor {
```
Olshansk, line 132: Can you leave a few TECHDEBT comments linking to the ticket you and @bryanchriswhite created relating to consolidating testing helpers?

Olshansk (changes requested):

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	mockTxIndexer := mock_types.NewMockTxIndexer(ctrl)
+	mockBus := mockModules.NewMockBus(ctrl)
+	mockPersistenceMod := mockModules.NewMockPersistenceModule(ctrl)
+
+	mockBus.EXPECT().GetPersistenceModule().AnyTimes().Return(mockPersistenceMod)
+	mockPersistenceMod.EXPECT().GetTxIndexer().AnyTimes().Return(mockTxIndexer)
+
+	ts := &treeStore{
+		logger:       &zerolog.Logger{},
```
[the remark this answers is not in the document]
dylanlott: I don't think it is necessary, I only did it because I saw it done in a test somewhere else that just wanted a quick logger implementation for testing. What should our general recommendation be for loggers during testing?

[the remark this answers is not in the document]
dylanlott: I made it use the bus instead 👍

utility/unit_of_work/module.go
```diff
@@ -58,51 +58,55 @@ func (uow *baseUtilityUnitOfWork) ApplyBlock() error {
...
 	}
 
+	// processProposalBlockTransactions indexes the transactions into the TxIndexer.
+	// if it fails, it triggers a rollback to undo the changes that processProposalBlockTransactions
+	// could have caused.
 	log.Debug().Msg("processing transactions from proposal block")
 	txMempool := uow.GetBus().GetUtilityModule().GetMempool()
 	if err := uow.processProposalBlockTransactions(txMempool); err != nil {
-		return err
+		rollErr := uow.revertLastSavePoint()
```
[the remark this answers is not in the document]
dylanlott: changed to `revertToLastSavepoint`

utility/unit_of_work/module.go
```diff
@@ -58,51 +58,55 @@ func (uow *baseUtilityUnitOfWork) ApplyBlock() error {
...
 	log.Debug().Msg("calling beginBlock")
 	if err := uow.beginBlock(); err != nil {
 		return err
 	}
 
+	// processProposalBlockTransactions indexes the transactions into the TxIndexer.
+	// if it fails, it triggers a rollback to undo the changes that processProposalBlockTransactions
+	// could have caused.
 	log.Debug().Msg("processing transactions from proposal block")
 	txMempool := uow.GetBus().GetUtilityModule().GetMempool()
```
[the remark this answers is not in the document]
dylanlott: `tl;dr` - I made `processProposalBlockTransactions` reference the bus to get access to the `Mempool` and now takes no argument.

Longer answer: Most of the discussion with @bryanchriswhite on this topic concluded with the idea that code should typically call the bus when it needs to access a module rather than take references to modules. Specifically for the function above, taking the `Mempool` from arguments as an interface makes it easy to test, but mocking a module on the bus is theoretically just as easy, yields the same flexibility, and still lets us abstract complexity behind interfaces.

shared/modules/treestore_module.go
```diff
@@ -21,13 +21,16 @@ type TreeStoreModule interface {
...
+
+	// Update returns the computed state hash for a given height.
 	// * Height is passed through to the Update function and is used to query the TxIndexer for transactions
 	// to update into the merkle tree set
 	// * Passing a higher height will cause a change but repeatedly calling the same or a lower height will
 	// not incur a change.
 	// * By nature of it taking a pgx transaction at runtime, Update inherits the pgx transaction's read view of the
 	// database.
+	// * Commit must be called after Update to persist any changes it made to disk.
 	Update(pgtx pgx.Tx, height uint64) (string, error)
```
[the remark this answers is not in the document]
dylanlott: I have considered renaming this, yes. Might I propose `GetAndUpdate` or `FetchAndUpdate`.

shared/modules/persistence_module.go
```diff
@@ -72,8 +79,8 @@ type PersistenceRWContext interface {
 // PersistenceWriteContext has no use-case independent of `PersistenceRWContext`, but is a useful abstraction
 type PersistenceWriteContext interface {
 	// Context Operations
-	NewSavePoint([]byte) error
-	RollbackToSavePoint([]byte) error
+	NewSavePoint() error
```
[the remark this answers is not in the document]
dylanlott: I like `SetSavepoint` the most, personally. That better captures the reality that only the most recent savepoint is kept. Will update to match 👍

persistence/context.go
```diff
@@ -36,18 +36,22 @@ type PostgresContext struct {
...
 
-// TECHDEBT(#327): Guarantee atomicity betweens `prepareBlock`, `insertBlock` and `storeBlock` for save points & rollbacks.
-func (p *PostgresContext) RollbackToSavePoint(bytes []byte) error {
-	p.logger.Info().Bool("TODO", true).Msg("RollbackToSavePoint not fully implemented")
-	return p.tx.Rollback(context.TODO())
+// RollbackToSavepoint triggers a rollback for the current pgx transaction and the underylying submodule stores.
+func (p *PostgresContext) RollbackToSavePoint() error {
+	ctx, _ := p.getCtxAndTx()
+	err := p.tx.Rollback(ctx)
+	p.stateTrees.Rollback()
```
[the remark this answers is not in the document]
dylanlott: > Should we pass in ctx 

Yes, probably when we thread context through the module. 

> and get back an err to append?

It was an intentional design choice to make `AtomicStore#Rollback` return nothing, so I would argue no, it shouldn't. It was an attempt by me at [defining errors out of existence by masking](https://thedeployguy.com/2019-06-16-a-philosophy-of-software-design-summary-part-2-cp10-end/).

persistence/trees/trees.go
dylanlott: Not as far as I know - @h5law can speak more to the details of the `smt` side, and correct me if I'm wrong about anything, but as far as I understand, the pointer will get dropped if it's not used, and in the case of a rollback being triggered, the reassignment kicks in which doesn't see the uncommitted tree changes because they haven't been written to disk, and it uses the same underlying connection from the `nodeStore`. Pretty slick, all things considered.

persistence/trees/trees.go
dylanlott: I'm not sure what you mean by overrides. BadgerDB will only allow one connection at a time to a file, and will error if it detects a second attempted connection, is that what you mean?

persistence/context.go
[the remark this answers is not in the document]
dylanlott: After messing with it and testing it locally, the key might actually be to make the `AtomicStore` interface return an error for easier testing, but mask the error from the `persistenceRWContext` layer up instead, allowing it the power to handle and possibly recover from rollback errors. Thinking about this more 👍

persistence/trees/atomic_test.go
dylanlott: Catching the log felt gross so I updated the interface to return an error and catch it in the case where no previous state is assigned.

persistence/context.go
```diff
@@ -47,9 +48,9 @@ func (p *PostgresContext) SetSavePoint() error {
 // RollbackToSavepoint triggers a rollback for the current pgx transaction and the underylying submodule stores.
 func (p *PostgresContext) RollbackToSavePoint() error {
 	ctx, _ := p.getCtxAndTx()
-	err := p.tx.Rollback(ctx)
-	p.stateTrees.Rollback()
-	return err
+	pgErr := p.tx.Rollback(ctx)
+	treesErr := p.stateTrees.Rollback()
+	return multierr.Combine(pgErr, treesErr)
```
h5law, line 53:
```suggestion
	return errors.Join(pgErr, treesErr)
```
Now we are in go 1.20 we can use this which works with `errors.Is()` https://pkg.go.dev/errors@master#Join

persistence/trees/trees_test.go
```diff
@@ -65,6 +65,15 @@ func TestTreeStore_Update(t *testing.T) {
 		require.NotEmpty(t, hash2)
 		require.NotEqual(t, hash1, hash2)
 	})
+
+	t.Run("should fail to rollback when no treestore savepoint is set", func(t *testing.T) {
+		pmod := newTestPersistenceModule(t, dbUrl)
+		context := newTestPostgresContext(t, 0, pmod)
+
+		err := context.RollbackToSavePoint()
+		require.Error(t, err)
```
h5law, line 74:
```suggestion
		require.ErrorIs(t, err, errFailedRollback)
```
Does this work?

persistence/context.go
dylanlott: oh, right! Good call, will update to match.

Olshansk: If you haven't already done this, I think a separate PR just updating this across the board would be good. If you already have, ignore.

utility/unit_of_work/module.go
```diff
@@ -1,12 +1,10 @@
 package unit_of_work
 
 import (
-	"fmt"
-
 	coreTypes "github.com/pokt-network/pocket/shared/core/types"
-	"github.com/pokt-network/pocket/shared/mempool"
 	"github.com/pokt-network/pocket/shared/modules"
 	"github.com/pokt-network/pocket/shared/modules/base_modules"
+	"go.uber.org/multierr"
```
Olshansk, line 7: @dylanlott  Let's change all of this in a small followup commit

persistence/trees/trees_test.go
```diff
@@ -1,24 +1,203 @@
...
+	}); err != nil {
+		log.Fatalf("Error clearing state: %v\n", err)
+	}
 }
 
 // TODO_AFTER(#861): Implement this test with the test suite available in #861
 func TestTreeStore_GetTreeHashes(t *testing.T) {
 	t.Skip("TODO: Write test case for GetTreeHashes method") // context: https://github.com/pokt-network/pocket/pull/915#discussion_r1267313664
-}
+} //
```
Olshansk, line 203:
```suggestion
}
```

persistence/trees/trees.go
```diff
@@ -88,6 +99,17 @@ type treeStore struct {
...
 	rootTree     *stateTree
 	merkleTrees  map[string]*stateTree
+
+	// PrevState holds a previous view of the Worldstate.
+	// The tree store rolls back to this view if errors are encountered during block application.
+	PrevState *Worldstate
+}
+
+// Worldstate holds a (de)serializable view of the entire tree state.
+type Worldstate struct {
```
Olshansk, line 109: Does this need to be publically exposed? 

Since we'll have a CLI for it, might be worth making it local and adding a TODO (with the cli ticket) to add an interface for it?

Olshansk (changes requested):
I didn't have time to do a full re-look at this, but for some reason "Show changes since you're last review" didn't work. Left a few comments and will take a full look tomorrow.

utility/unit_of_work/module.go
dylanlott: Covered in #939

utility/unit_of_work/module.go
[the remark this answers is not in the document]
Olshansk: > st, but mocking a module on the bus is theoretically just as easy, yields the same flexibility, and still lets us abstract complexity behind interfaces.

Yup, that was the design goal. Mock this on the "dependency injection bus" and pass that around everywhere.

shared/modules/treestore_module.go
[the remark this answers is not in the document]
Olshansk: I like your idea and was going to suggest `QueryAndUpdate` given that it takes `pgtx` as an argument. However, maybe `CollectAndUpdate` to avoid being implementation specific but capture the fact that it "collects" all the changes. THoughts?

persistence/context.go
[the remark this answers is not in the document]
Olshansk: Bump on this one.

persistence/trees/trees.go
Olshansk: Can you just leave a comment next to the code explaining this?

persistence/trees/trees.go
Olshansk: Yes, that's what I meant. I guess it's a non-issue?

persistence/trees/trees.go
```diff
@@ -88,6 +99,18 @@ type treeStore struct {
...
+	// The tree store rolls back to this view if errors are encountered during block application.
+	PrevState *worldstate
+}
+
+// worldstate holds a (de)serializable view of the entire tree state.
+// TECHDEBT(#566) - Hook this up to node CLI subcommands
+type worldstate struct {
+	TreeStoreDir string
+	RootTree     *stateTree
+	MerkleTrees  map[string]*stateTree
```
Olshansk, line 113: Apologize if I already brought this up in a previous PR but why do all of these need to be exposed outside this package? If we can change it, let's do it. If it's not, and I'm re-asking the same question, might mean it's an opportunity to add a comment with reasoning.


```suggestion
	prevState *worldstate
}

// worldstate holds a (de)serializable view of the entire tree state.
// TECHDEBT(#566) - Hook this up to node CLI subcommands
type worldstate struct {
	treeStoreDir string
	rootTree     *stateTree
	merkleTrees  map[string]*stateTree
```

persistence/trees/trees.go
```diff
@@ -88,6 +99,18 @@ type treeStore struct {
...
 	merkleTrees  map[string]*stateTree
+
+	// PrevState holds a previous view of the Worldstate.
+	// The tree store rolls back to this view if errors are encountered during block application.
+	PrevState *worldstate
+}
+
+// worldstate holds a (de)serializable view of the entire tree state.
+// TECHDEBT(#566) - Hook this up to node CLI subcommands
+type worldstate struct {
```
Olshansk, line 110: s/worldstate/worldState

persistence/trees/trees.go
```diff
@@ -261,6 +281,68 @@ func (t *treeStore) getStateHash() string {
 	return hexHash
 }
 
+////////////////////////////////
+// AtomicStore Implementation //
+////////////////////////////////
+
+// Savepoint generates a new savepoint for the tree store and saves it internally.
```
Olshansk, line 288:
```suggestion
// Savepoint generates a new savepoint (i.e. a worldState) for the tree store and saves it internally.
```

persistence/trees/trees.go
```diff
@@ -261,6 +281,68 @@ func (t *treeStore) getStateHash() string {
...
+func (t *treeStore) Savepoint() error {
+	w, err := t.save()
+	if err != nil {
+		return err
+	}
+	t.PrevState = w
+	return nil
+}
+
+// Rollback rolls back to the last saved world state maintained by the treeStore.
```
Olshansk, line 298:
```suggestion
// Rollback rolls back to the last saved worldState maintained by the treeStore.
```

persistence/trees/trees.go
```diff
@@ -261,6 +281,68 @@ func (t *treeStore) getStateHash() string {
...
+// Rollback intentionally can't return an error because at this point we're out of tricks
```
Olshansk, line 299: This comment needs to be updated

utility/unit_of_work/uow_leader_test.go
```diff
@@ -0,0 +1,103 @@
...
+			name: "should revert a failed block proposal",
+			args: args{},
+			fields: fields{
+				leaderUOW: func(t *testing.T) *leaderUtilityUnitOfWork {
+					ctrl := gomock.NewController(t)
+
+					mockUtilityMod := newDefaultMockUtilityModule(t, ctrl)
+					mockrwcontext := newDefaultMockRWContext(t, ctrl)
+
+					mockrwcontext.EXPECT().ComputeStateHash().Return("", fmt.Errorf("rollback error"))
```
Olshansk, line 45: NIT: I don't fully understand the choice whitespacing & grouping of functions here, but not a big deal.

I would've personally gone with:
```go
mockbus := mockModules.NewMockBus(ctrl)

mockUtilityMod := newDefaultMockUtilityModule(t, ctrl)
mockbus.EXPECT().GetUtilityModule().Return(mockUtilityMod).AnyTimes()
	
mockrwcontext := newDefaultMockRWContext(t, ctrl)
mockrwcontext.EXPECT().ComputeStateHash().Return("", fmt.Errorf("rollback error"))

luow := NewLeaderUOW(0, mockrwcontext, mockrwcontext)
luow.SetBus(mockbus)
```

utility/unit_of_work/uow_leader_test.go
```diff
@@ -0,0 +1,103 @@
...
+					mockbus := mockModules.NewMockBus(ctrl)
+					mockbus.EXPECT().GetUtilityModule().Return(mockUtilityMod).AnyTimes()
+
+					luow := NewLeaderUOW(0, mockrwcontext, mockrwcontext)
+					luow.SetBus(mockbus)
+
+					return luow
+				},
+			},
+			wantErr: true,
```
Olshansk, line 55: Can you add a test where `wantErr: false`?

persistence/trees/trees.go
```diff
@@ -261,6 +281,68 @@ func (t *treeStore) getStateHash() string {
...
+		w.MerkleTrees[treeName] = &stateTree{
+			name:      treeName,
+			tree:      tree,
+			nodeStore: nodeStore,
+		}
+	}
+
+	root, nodeStore := t.GetTree(RootTreeName)
+	tree := smt.ImportSparseMerkleTree(nodeStore, smtTreeHasher, root)
+	w.RootTree = &stateTree{
```
Olshansk, line 337: s/RootTree/rootTree?

persistence/trees/trees_test.go
```diff
@@ -1,21 +1,201 @@
...
-// TECHDEBT(#836): Tests added in https://github.com/pokt-network/pocket/pull/836
 func TestTreeStore_Update(t *testing.T) {
-	// TODO: Write test case for the Update method
-	t.Skip("TODO: Write test case for Update method")
+	pool, resource, dbUrl := test_artifacts.SetupPostgresDocker()
+	t.Cleanup(func() {
+		require.NoError(t, pool.Purge(resource))
+	})
+
+	t.Run("should update actor trees, commit, and modify the state hash", func(t *testing.T) {
```
Olshansk, line 48:
```suggestion
	t.Run("should update application trees, commit, and modify the state hash", func(t *testing.T) {
```

persistence/trees/trees_test.go
```diff
@@ -1,21 +1,201 @@
...
+		hash2, err := context.ComputeStateHash()
+		require.NoError(t, err)
+		require.NotEmpty(t, hash2)
+		require.NotEqual(t, hash1, hash2)
+	})
+
+	t.Run("should fail to rollback when no treestore savepoint is set", func(t *testing.T) {
+		pmod := newTestPersistenceModule(t, dbUrl)
+		context := newTestPostgresContext(t, 0, pmod)
+
```
Olshansk, line 73: Can you update this test to:

1. Insert the test app
2. not create a savepoint (as exists)
3. Verify hash is unchanged after the failed rollback?

Olshansk (changes requested):

persistence/trees/atomic_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// set a new savepoint
+	require.NoError(t, ts.Savepoint())
+	require.NotEmpty(t, ts.PrevState.MerkleTrees)
+	require.NotEmpty(t, ts.PrevState.RootTree)
+	// assert that savepoint creation doesn't mutate state hash
+	require.Equal(t, hash1, hex.EncodeToString(ts.PrevState.RootTree.tree.Root()))
+
+	// verify that creating a savepoint does not change state hash
+	hash2 := ts.getStateHash()
```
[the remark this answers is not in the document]
h5law: Personally I think this makes it harder to grok 😅

persistence/trees/trees.go
dylanlott: Unless I misunderstand the problem, I believe it's a non-issue. Each `nodeStore` handles the badgerDB reference, which can't be overridden because badgerDB will panic the caller that attempts a second connection.

persistence/trees/atomic_test.go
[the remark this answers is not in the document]
dylanlott: Yeah I agree with Harry here, the `s` thing is more difficult for me to reason about the progression of these hashes and how they should align.

persistence/trees/trees_test.go
dylanlott: Addressed in #944 to unblock this PR for now.

persistence/trees/trees_test.go
Olshansk: @dylanlott Thoughts on this?

persistence/trees/trees_test.go
```diff
@@ -1,111 +1,205 @@
...
+	}
+	if err := m.HandleDebugMessage(&messaging.DebugMessage{
+		Action:  messaging.DebugMessageAction_DEBUG_PERSISTENCE_RESET_TO_GENESIS,
+		Message: nil,
+	}); err != nil {
+		log.Fatalf("Error clearing state: %v\n", err)
+	}
+}
+
+// TODO_AFTER(#861): Implement this test with the test suite available in #861
```
Olshansk, line 202: `test suite available in #861` is this correct?

We want to use the test suite in this PR to implement it?

persistence/trees/trees_test.go
```diff
@@ -1,111 +1,205 @@
...
-func TestTreeStore_Prove(t *testing.T) {
-	nodeStore := kvstore.NewMemKVStore()
-	tree := smt.NewSparseMerkleTree(nodeStore, smtTreeHasher)
-	testTree := &stateTree{
-		name:      "test",
-		tree:      tree,
-		nodeStore: nodeStore,
-	}
+// TECHDEBT(#796): Test helpers should be consolidated in a single place
+func newTestApp(t *testing.T) *core_types.Actor {
```
Olshansk, line 149: OPTIONAL: Per @h5law's comment, we could move the helpers into a separate `utils_test.go` but don't feel strongly about it.

Olshansk (changes requested):
A few last minute nits and then it should be g2g.

@dylanlott Can you please `merge with main` and resolve conflicts to avoid a force push? It allows the reviewer to review `changes since last review`. Ty!

persistence/trees/main_test.go
```diff
@@ -0,0 +1,12 @@
+//go:build test
+
+package trees
```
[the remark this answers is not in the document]
dylanlott: I was following the prescription of [TestUtils](https://www.notion.so/pocketnetwork/Testutils-9cba9010e18447248e9daa8a3b87e3f2?pvs=4#2119f5318e804d689fb29b560ff6f2af) for testing unexported members like previously discussed.

persistence/trees/trees_test.go
dylanlott: I believe that's what @h5law meant out of this, yes.

persistence/trees/main_test.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"crypto/sha256"
+	"hash"
+)
+
+type TreeStore = treeStore
+
+var SMTTreeHasher hash.Hash = sha256.New()
```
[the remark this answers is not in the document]
dylanlott: It can't go in one, it will complain that it's not used if it's declared in the init and causes a compile error.

persistence/trees/trees_test.go
dylanlott: I'd rather create an issue with a `TECHDEBT` comment. There's no easy answer to the imports until we've solved some larger problems.

Olshansk (approved):

PR #862, opened 2023-06-28
Update vault_test.go

okdas:
The latest tag was removed for some reason: https://github.com/docker-library/official-images/commit/32c2455241a13e31b9c46cb109294d2bc6c46ba0

Let's pin `1.13.3`.


h5law (approved):
Assuming this is something we will have to be updating manually from now on?

Otherwise LFG good spot! <3 Seems tests are working so lets :shipit:

okdas:
> Assuming this is something we will have to be updating manually from now on?

Most likely, but I'd treat hashicorp as a database. Meaning, the API that we use should only rarely change. We'd upgrade if this will be a problem, and I, personally, don't see that particular functionality being used often.

PR #876, opened 2023-07-04
[IBC] Add length check in RemovePrefix function

innocent-saeed36:
## Problem:
The `RemovePrefix` function in `prefix.go` currently performs a slice operation on the `path` argument without first checking if the `path` is long enough to accommodate the slice operation. If the length of `path` is less than the combined length of `prefix` and `delimiter`, this will result in a runtime panic due to out-of-range slice indices.

## Solution:
This PR adds a check to ensure the length of `path` is sufficient before the slice operation is performed. If the length of `path` is less than the combined length of `prefix` and `delimiter`, the function returns an empty string.

## Changes:
Added length check to `RemovePrefix` function in `prefix.go`.


Olshansk:
I do not believe this is a necessary check

PR #877, opened 2023-07-04
[IBC] Add nil check on proof for membership and non-membership proof creation

innocent-saeed36:
## Problem:
In `proofs_ics23.go`, the `createMembershipProof` and `createNonMembershipProof` functions call `tree.Prove(key)` and then pass the `proof` to the `convertSMPToExistenceProof` or `convertSMPToExclusionProof` functions without checking if `proof` is `nil`. This could lead to a panic due to a nil pointer dereference if `tree.Prove(key)` returns a `nil` proof.

## Solution:
This PR adds a check to verify if `proof` is `nil` before passing it to the `convertSMPToExistenceProof` or `convertSMPToExclusionProof` functions. If `proof` is `nil`, an error is returned.

## Changes:
Added nil check on `proof` in `createMembershipProof` and `createNonMembershipProof` functions in `proofs_ics23.go`.

### DIFF
--- a/ibc/store/proofs_ics23.go
+++ b/ibc/store/proofs_ics23.go
@@ -63,7 +63,7 @@ func VerifyNonMembership(root ics23.CommitmentRoot, proof *ics23.CommitmentProof
 // in the SMT provided
 func createMembershipProof(tree *smt.SMT, key, value []byte) (*ics23.CommitmentProof, error) {
 	proof, err := tree.Prove(key)
-	if err != nil {
+	if err != nil || proof == nil {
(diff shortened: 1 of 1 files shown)


Olshansk (approved):
@innocent-saeed36 Could you please provide the exact `ChatGPT` prompt you used for this?

@h5law This looks good to me but want a 👍  from you as well

h5law (approved):
Currently the SMT will only return `nil` for the proof when the `err` is non-nil. However, if this logic ever changes its good to have this extra check.

LGTM let's merge it in!

PR #801, opened 2023-06-02
[P2P] [Tooling] Peer discovery `peer connections` subcommand

bryanchriswhite:
## @Reviewer
This PR may be more digestible / reviewable on a commit-by-commit basis. Commits are organized logically and any given line is only modified in a single commit, with few exceptions*.

*(In the interest of preserving the git-time-continuum :police_officer::rotating_light:, this applies in batches of commits between comments or reviews *by humans*, only once "in review")

---

## Description

Adds the following subcommands:
```
Print open peer connections

Usage:
  p1 peer connections [flags]

Flags:
  -h, --help   help for connections

Global Flags:
  -a, --all                     operations apply to both staked & unstaked router peerstores (default)
      --config string           Path to config
      --data_dir string         Path to store pocket related data (keybase etc.) (default "/home/bwhite/.pocket")
  -l, --local                   operations apply to the local (CLI binary's) P2P module instead of being broadcast
      --non_interactive         if true skips the interactive prompts wherever possible (useful for scripting & automation)
      --remote_cli_url string   takes a remote endpoint in the form of <protocol>://<host>:<port> (uses RPC Port) (default "http://localhost:50832")
  -s, --staked                  operations only apply to staked router peerstore (i.e. raintree)
  -u, --unstaked                operations only apply to unstaked router peerstore (i.e. gossipsub)
      --verbose                 Show verbose output
```

## Issue

Related:
- #730

Dependencies:
- #804 
- #891

## Type of change

Please mark the relevant option(s):

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Added `peer connections` subcommand
- Added `P2PModule#GetConnections()` interface method & implementation

## Testing

- [x] `make develop_test`; if any code changes were made
- [ ] `make test_e2e` on [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any code changes were made
- [ ] `e2e-devnet-test` passes tests on [DevNet](https://pocketnetwork.notion.site/How-to-DevNet-ff1598f27efe44c09f34e2aa0051f0dd); if any code was changed
- [ ] [Docker Compose LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md); if any major functionality was changed or introduced
[...]

Issue #730, opened 2023-05-04
[P2P] Peer discovery debug CLI sub-commands


bryanchriswhite:
## Objective

Extend the debug CLI to expose sub-commands for driving peer discovery in the P2P module (analogous to how it currently supports driving consensus). This will facilitate both manual and automated testing efforts.

## Origin Document

Based on a discussion between @dylanlott and myself about end-to-end test planning for peer discovery. It has been made clear in discussions past (probably in documentation somewhere as well) that we want to prioritize supporting robust manual testing/investigation of these dynamic systems (before they get a chance to run away from us).

## Goals

- Expose controls to peer discovery to developers
- Support step-by-step progression through peer discovery
- Support debugging & reproducing peer discovery scenarios (incl. edge cases)

## Deliverable

- [ ] Extend the debug CLI to expose the sub-commands subcommands outlined below (peer CLI UX)
- [ ]  (Bonus): draft a design for reproducible peer discovery failure reports or something
  - Could be generic (e.g. trace based)

### CLI UX - `peer` & `p2p` subcommands

The following commands should send/broadcast debug messages into the network, resulting in nodes to print to their respective logs, similar to the debug subcommand's behavior.

```
peer discovery status # background router only
peer discovery start  # background router only
peer discovery stop   # background router only
peer list [--all | --staked | --unstaked]
peer forget (--all | <forgetter peer ID>)
peer connections
peer interrogate <peer ID>
peer disconnect <peer ID>
```

`p2pCmd` persisten flags: 
- `--trace`

_NOTE: `p2p` commands MAY interact with a remote/standalone node rather than utilize the CLI's own P2P module_

```
p2p send [--trace] <peer ID> <text message>
p2p broadcast [--trace] <text message>
```

## Non-goals / Non-deliverables

- Refactoring/simplifying the `Router` interface
- Additional, unrelated libp2p integration work

## General issue deliverables

- [ ] Update the appropriate CHANGELOG(s)
- [ ] Update any relevant local/global README(s)
- [ ] Update relevant source code tree explanations
- [ ] Add or update any relevant or supporting [mermaid](https://mermaid-js.github.io/mermaid/) diagrams

## Testing Methodology

- [ ] **New CLI subcommands behave as expected**
- [ ] **All tests**: `make test_all`
- [ ] **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)
[...]

Issue #811, opened 2023-06-08
[P2P] Move `Peerstore` interface to a module-shared pkg


bryanchriswhite:
## Objective

Support refactoring of the `unstakedPeerstoreProvider` interface to become part of the `P2PModule` interface to improve readability and maintainability while preventing import cycles.

## Origin Document

Observation made while working on #732 

## Goals

- Avoid import cycle on `Peerstore` which would be created by simply integrating or embedding `unstakedPeerstoreProvider` into `P2PModule`.

## Deliverable

- [ ] Move the `Peerstore` interface from p2p/types/peerstore.go to a shared/core/types (or similar / more appropriate)

## Non-goals / Non-deliverables

- ...

## General issue deliverables

- [ ] Update the appropriate CHANGELOG(s)
- [ ] Update any relevant local/global README(s)
- [ ] Update relevant source code tree explanations
- [ ] Add or update any relevant or supporting [mermaid](https://mermaid-js.github.io/mermaid/) diagrams

## Testing Methodology

- [ ] **All tests**: `make test_all`
- [ ] **LocalNet**: verify a `LocalNet` is still functioning correctly by following the instructions at [docs/development/README.md](https://github.com/pokt-network/pocket/tree/main/docs/development)
- [ ] **k8s LocalNet**: verify a `k8s LocalNet` is still functioning correctly by following the instructions [here](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md)

---

**Creator**: @bryanchriswhite

Issue #810, opened 2023-06-08
[Core] Improve dependency injection for submodules


bryanchriswhite:
## Objective

Improve the dependency injection experience for submodules (e.g. P2P's `PeerstoreProvider`, telemetry's `TimeSeriesAgent`, etc.). I think this may generally apply to submodules which need to embed the `base_modules.InterruptableModule`.

## Origin Document

Observation made while working on #732. In order to complete the peerstore provider refactor (#804).

When retrieving modules from the bus (via the module registry), they are asserted to the correct interface type by their respective `Bus#Get_X_Module()` method. In contrast, retrieving submodules from the registry must be done directly (at the time of writing) which requires additional type assertions and boilerplate in each place any submodule is retrieved from the bus.

## Goals

- Support simplification of submodule interfaces (i.e. don't embed `Module` unless appropriate).
- Support dependency injection of submodules with a developer experience on par with that of modules.

```mermaid
classDiagram 

class IntegratableModule {
    <<interface>>
    +GetBus() Bus
    +SetBus(bus Bus)
}

class ModuleFactoryWithOptions {
    <<interface>>
    Create(bus Bus, options ...ModuleOption) (Module, error)
}


class Module {
    <<interface>>
}
class InjectableModule {
    <<interface>>
    GetModuleName() string
}
class InterruptableModule {
    <<interface>>
    Start() error
    Stop() error
}

Module --|> InjectableModule
Module --|> IntegratableModule
Module --|> InterruptableModule
Module --|> ModuleFactoryWithOptions

class Submodule {
    <<interface>>
}

Submodule --|> InjectableModule
Submodule --|> IntegratableModule

class exampleSubmodule
class exampleSubmoduleFactory {
    <<interface>>
    Create(...) (exampleSubmodule, error)
}

exampleSubmodule --|> Submodule
exampleSubmodule --|> exampleSubmoduleFactory

exampleModule --|> Module
```

### Concrete Example

Below, `rpcPeerstoreProvider` MUST implement `Module` so that it can traverse the `ModuleRegistry` dependency injection system that we currently have, as it's used outside of the P2P module in the CLI. This results in it embedding the noop implementations of `InterruptableModule` and being additionally over-constrained by the `InitializableModule#Create()` interface method:

```mermaid
classDiagram 

class IntegratableModule {
    <<interface>>
    +GetBus() Bus
    +SetBus(bus Bus)
}

class PeerstoreProvider {
    <<interface>>
[...]

PR #891, opened 2023-07-07
[CLI] Consistent config/flag parsing & common helpers

bryanchriswhite:
## @Reviewer
This PR may be more digestible / reviewable on a commit-by-commit basis. Commits are organized logically and any given line is only modified in a single commit, with few exceptions*.

*(In the interest of preserving the git-time-continuum :police_officer::rotating_light:, this applies in batches of commits between comments or reviews *by humans*, only once "in review")

---

## Description

I discovered some inconsistencies in how/whether we parse the config in different subcommands. 

There was also some code that turns out to be common between some existing commands and the ones I'm adding for #730.

## Issue

Related:

- #730 

Dependents:

- #892
- #801 

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- simplified debug message broadcasting by using the p2p modules broadcast method
- factored out common CLI helper functions
  - `FetchPeerstore()`
  - `GetBusFromCmd()`
- ensured config and flag parsing is consistent
- ensured a consistent identity across subsequent usages of the debug CLI

## Testing

## Required Checklist

### If Applicable Checklist


runtime/manager.go
```diff
@@ -104,6 +104,7 @@ func WithRandomPK() func(*Manager) {
 	return WithPK(privateKey.String())
 }
 
+// TECHDEBT(#750): separate conseneus and P2P keys.
```
Olshansk, line 107:
```suggestion
// TECHDEBT(#750): separate consensus and P2P (identity vs communication) keys.
```

app/client/cli/helpers/setup.go
```diff
@@ -4,26 +4,37 @@ import (
...
 // P2PDependenciesPreRunE initializes peerstore & current height providers, and a
 // p2p module which consumes them. Everything is registered to the bus.
 func P2PDependenciesPreRunE(cmd *cobra.Command, _ []string) error {
+
 	// TECHDEBT: this is to keep backwards compatibility with localnet
 	flags.ConfigPath = runtime.GetEnv("CONFIG_PATH", "build/config/config.validator1.json")
+	configs.ParseConfig(flags.ConfigPath)
+
+	// set final `remote_cli_url` value; order of precedence: flag > env var > config > default
+	flags.RemoteCLIURL = viper.GetString("remote_cli_url")
```
Olshansk, line 32: Wow, literally have the same change in #902

app/client/cli/helpers/setup.go
```diff
@@ -4,26 +4,37 @@ import (
...
 	"github.com/pokt-network/pocket/logger"
 	"github.com/pokt-network/pocket/p2p"
 	rpcCHP "github.com/pokt-network/pocket/p2p/providers/current_height_provider/rpc"
 	rpcPSP "github.com/pokt-network/pocket/p2p/providers/peerstore_provider/rpc"
 	"github.com/pokt-network/pocket/runtime"
+	"github.com/pokt-network/pocket/runtime/configs"
 	"github.com/pokt-network/pocket/shared/modules"
 )
 
+// TODO_THIS_COMMIT: add godoc comment explaining what this **is** and **is not**
```
Olshansk, line 19:
```suggestion
// TODO_IN_THIS_COMMIT: add godoc comment explaining what this **is** and **is not**
```

OPTIONAL: Update `build/linters/blockers.go` so we capture other regexes as well

app/client/cli/debug.go
```diff
@@ -210,51 +190,16 @@ func sendDebugMessage(cmd *cobra.Command, debugMsg *messaging.DebugMessage) {
 	}
 
 	// if the message needs to be broadcast, it'll be handled by the business logic of the message handler
+	//
+	// DISCUSS_THIS_COMMIT: The statement above is false. Using `#Send()` will only
+	// be unicast with no opportunity for further propagation.
 	validatorAddress = pstore.GetPeerList()[0].GetAddress()
```
Olshansk, line 196: Unrelated. Thoughts on renaming `validatorAddress` to `randomStakedActorAddress` or `firstStakedActorAddress`?

app/client/cli/debug.go
```diff
@@ -210,51 +190,16 @@ func sendDebugMessage(cmd *cobra.Command, debugMsg *messaging.DebugMessage) {
 	}
 
 	// if the message needs to be broadcast, it'll be handled by the business logic of the message handler
+	//
+	// DISCUSS_THIS_COMMIT: The statement above is false. Using `#Send()` will only
```
Olshansk, line 194: Things have definitely changed/evolved so I'll provide some context & pointers, but would like to get your thoughts/ideas on what the right thing to do is.

For example, take this target to send a tx:
```Makefile
.PHONY: send_local_tx
send_local_tx: ## A hardcoded send tx to make LocalNet debugging easier
	go run -tags=debug app/client/*.go Account Send --non_interactive 00104055c00bed7c983a48aac7dc6335d7c607a7 00204737d2a165ebe4be3a7d5b0af905b0ea91d8 1000
```

- We sent it to one address
- That one address adds it to the mempool
- It needs to be gossiped through the network

In `utility/utility_message_handler.go`, we handle a `TxGossip` like so where `HandleTransaction` just adds it to the local mempool.

```go
func (u *utilityModule) HandleUtilityMessage(message *anypb.Any) error {
	switch message.MessageName() {
	case messaging.TxGossipMessageContentType:
		msg, err := codec.GetCodec().FromAny(message)
		// ...
		if txGossipMsg, ok := msg.(*typesUtil.TxGossipMessage); !ok {
			return fmt.Errorf("failed to cast message to UtilityMessage")
		} else if err := u.HandleTransaction(txGossipMsg.Tx); err != nil {
			return err
		}
		// ...
	default:
		return coreTypes.ErrUnknownMessageType(message.MessageName())
	}
	// ...
}
```

**Our goal is**: Gossip `TxGossip` throughout the network
**Question**: WHat do we change to make it happen? `Account Send`? `s/Send/Broadcast`? How we handle `case messaging.TxGossipMessageContentType`? Wdyt?

app/client/cli/helpers/common.go
```diff
@@ -10,5 +19,44 @@ var (
 	genesisPath = runtime.GetEnv("GENESIS_PATH", "build/config/genesis.json")
 
 	// P2PMod is initialized in order to broadcast a message to the local network
+	// TECHDEBT: prefer to retrieve P2P module from the bus instead.
 	P2PMod modules.P2PModule
 )
+
+// fetchPeerstore retrieves the providers from the CLI context and uses them to retrieve the address book for the current height
```
Olshansk, line 26: Comment is `fetchPeerstore` but function is `FetchPeerstore`. What are we aiming for?

app/client/cli/helpers/common.go
```diff
@@ -10,5 +19,44 @@ var (
...
+func FetchPeerstore(cmd *cobra.Command) (types.Peerstore, error) {
+	bus, err := GetBusFromCmd(cmd)
+	if err != nil {
+		return nil, err
+	}
+	// TECHDEBT(#810, #811): use `bus.GetPeerstoreProvider()` after peerstore provider
```
Olshansk, line 32: #810 is done, so are we only waiting on #811?

app/client/cli/helpers/common.go
```diff
@@ -10,5 +19,44 @@ var (
...
+	}
+	currentHeightProvider := bus.GetCurrentHeightProvider()
+	height := currentHeightProvider.CurrentHeight()
+	pstore, err := pstoreProvider.(peerstore_provider.PeerstoreProvider).GetStakedPeerstoreAtHeight(height)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving peerstore at height %d", height)
+	}
+	// Inform the client's main P2P that a the blockchain is at a new height so it can, if needed, update its view of the validator set
+	err = sendConsensusNewHeightEventToP2PModule(height, bus)
+	if err != nil {
```
Olshansk, line 46:
```suggestion
	if err = sendConsensusNewHeightEventToP2PModule(height, bus); err != nil {
```

app/client/cli/helpers/common.go
```diff
@@ -10,5 +19,44 @@ var (
...
+		return nil, errors.New("sending consensus new height event")
+	}
+	return pstore, nil
+}
+
+// sendConsensusNewHeightEventToP2PModule mimicks the consensus module sending a ConsensusNewHeightEvent to the p2p module
+// This is necessary because the debug client is not a validator and has no consensus module but it has to update the peerstore
+// depending on the changes in the validator set.
+// TODO(#613): Make the debug client mimic a full node.
+func sendConsensusNewHeightEventToP2PModule(height uint64, bus modules.Bus) error {
```
Olshansk, line 56: Let's either add a TECHDEBT comment or validate in this PR if it's necessary.

I prefer the former since ei might require some time to test

Olshansk (changes requested):

app/client/cli/helpers/setup.go
bryanchriswhite: I liked your solution better than what I had done. :sweat_smile: I was trying to consolidate usages of `ParseConfig()` accross the CLI but later discovered that we're not constructing or referencing the config in the same way everywhere.

I think there's probably an issue that needs to be opened about:

1. Clearer separation between CLI and API code
    - In my view, `configs.ParseConfig()` is API code, it feels wrong to me to have all the viper calls that we currently do in there; viper shouldn't be part of our API dependency graph.
2. Parsing the config consistently in all CLI commands
  - Here, we're using the runtime manager to parse files at a dynamic path but [elsewhere](https://github.com/pokt-network/pocket/blob/main/app/client/cli/cmd.go#L50) we're assigning (and [referencing](https://github.com/pokt-network/pocket/blob/main/app/client/cli/utils.go#L314)) a global config var instead, relying on the unconventional behavior in 1.
3. Handling flags consistently in all commands
  I feel like we should have one place where we do `flags.RemoteCLIURL = ...` (and probably all the rest of the viper calls too) rather than duplicating it in multiple pre-run functions. I think we also need to ensure we're consistently overriding flags with viper correctly.

Olshansk: #907 907

app/client/cli/helpers/setup.go
h5law: Added a comment here to capture the meaning of the var

app/client/cli/debug.go
```diff
@@ -198,13 +194,16 @@ func sendDebugMessage(cmd *cobra.Command, debugMsg *messaging.DebugMessage) {
 	//
 	// DISCUSS_THIS_COMMIT: The statement above is false. Using `#Send()` will only
```
Olshansk, line 195:
```suggestion
	// TODO(#936): The statement above is false. Using `#Send()` will only
```

Olshansk (approved):
@h5law One tiny lingering comment (going to merge it in myself).

When you squash & merge this to main, make sure to copy-paste your attribution below the message

```
Co-authored-by: harry <53987565+h5law@users.noreply.github.com>
```

PR #892, opened 2023-07-07
[P2P] [Tooling] Peer discovery `peer list` subcommand

bryanchriswhite:
## Description

Introduce the `peer` and `peer list` subcommands to the CLI.

reviewpad:summary

## Issue

- #730 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- ...

## Testing

## Required Checklist

### If Applicable Checklist


p2p/debug/list.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if err := PrintPeerListTable(peers); err != nil {
+		return fmt.Errorf("printing peer list: %w", err)
+	}
+	return nil
+}
+
+func getPeerstoreProvider() (peerstore_provider.PeerstoreProvider, error) {
+	return nil, nil
+}
```
bryanchriswhite, line 106:
```suggestion
```

p2p/utils/logging.go
```diff
@@ -41,6 +54,63 @@ func LogIncomingMsg(logger *modules.Logger, hostname string, peer types.Peer) {
 	logMessage(logger, msg, hostname, peer)
 }
 
+// Print table prints a table to stdout. Header row is defined by `header`. Row printing
+// behavior is defined by `consumeRows`. Header length SHOULD match row length.
```
Olshansk, line 58: I think something like this is a really good opportunity to just import another library: https://github.com/jedib0t/go-pretty?

p2p/debug/list.go
```diff
@@ -0,0 +1,133 @@
...
+	// is retrievable as a proper submodule.
+	pstoreProviderModule, err := bus.GetModulesRegistry().
+		GetModule(peerstore_provider.PeerstoreProviderSubmoduleName)
+	if err != nil {
+		return fmt.Errorf("getting peerstore provider: %w", err)
+	}
+	pstoreProvider, ok := pstoreProviderModule.(peerstore_provider.PeerstoreProvider)
+	if !ok {
+		return fmt.Errorf("unknown peerstore provider type: %T", pstoreProviderModule)
+	}
```
Olshansk, line 31: OPTIONAL nit: These lines are copy-pasted in a lot of places. I think there's an opportunity to create a shared helper

p2p/debug/list.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/pokt-network/pocket/p2p/providers/peerstore_provider"
+	"github.com/pokt-network/pocket/p2p/types"
+	"github.com/pokt-network/pocket/p2p/utils"
+	"github.com/pokt-network/pocket/shared/modules"
+)
+
+var peerListTableHeader = []string{"Peer ID", "Pokt Address", "ServiceURL"}
+
+func PrintPeerList(bus modules.Bus, routerType RouterType) error {
```
Olshansk, line 15: Please add godoc comments to functions exposed/used outside of the pacakge

p2p/debug/list.go
```diff
@@ -0,0 +1,133 @@
+package debug
```
Olshansk, line 1: debug tag?

app/client/cli/peer/list.go
```diff
@@ -0,0 +1,106 @@
...
+		Message: &anypb.Any{
+			Value: []byte(routerType),
+		},
+	}
+	debugMsgAny, err := anypb.New(debugMsg)
+	if err != nil {
+		return fmt.Errorf("creating anypb from debug message: %w", err)
+	}
+
+	if localFlag {
```
Olshansk, line 68: See my comments in #801. They're similar to this as well

Olshansk (changes requested):
@bryanchriswhite I reviewed #801 first, so some of the comments there are also relevant so PTAL.

app/client/cli/flags/config.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+var Cfg *configs.Config
+
+func ParseConfigAndFlags(_ *cobra.Command, _ []string) error {
+	// by this time, the config path should be set
+	Cfg = configs.ParseConfig(ConfigPath)
+
+	// set final `remote_cli_url` value; order of precedence: flag > env var > config > default
+	RemoteCLIURL = viper.GetString("remote_cli_url")
```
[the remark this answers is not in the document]
bryanchriswhite: I intended to revert this change. This is the direction I started when investigating https://github.com/pokt-network/pocket/pull/891#discussion_r1263340347, I thought I had reverted it but apparently failed.

To answer your question, the reason for this is because viper's only integration with flags is to support setting a viper key based on the flag, but not the other way around. I.e: Viper won't update the flag value. **This only applies to bound flags**

This means that we have to do one of the following consistently for bound flags:

- call `viper.GetString("<flag key>")` (or a helper containing it) anywhere we need the value
- update the flag consistently and reference the flag value anywhere we need it

I opted for the latter option as I felt it was more conventional and easier to read and maintain.

app/client/cli/peer/list.go
h5law: @Olshansk can you clarify which comment you are referring to? And what you think should change here?

p2p/debug/list.go
h5law: adding `//go:build debug` causes the CLI to break, my guess is its not build with the debug tag? Will look into this further

app/client/cli/peer/list.go
Olshansk: Doing a `ctrl f` for `go build` is a good starting point here.

Screenshot 2023-07-25 at 3 47 51 PM

p2p/debug/list.go
Olshansk: Doing a `ctrl f` for `go build` is a good starting point here.

Screenshot 2023-07-25 at 3 47 51 PM

app/client/cli/peer/list.go
h5law: @Olshansk the reason I am tackling this first is that this is the base of #801 as such I feel it would be difficult to address #801 when the changes may need to be downwards propagated. Will go through #801 alongside from now on

Olshansk: Thank you. This one is on me

p2p/debug/list.go
h5law: So the reason these cannot be put under debug is as follows. The cli calls the `PrintPeerList` function. So the `peer` subdir in the cli would also need to be with the debug tag. This calls the helper functions so that would need to be in a debug tag and thus all of the CLI would also need it.

I think it better to change the `debug` dir to something like `printing` or `introspect` perhaps.

WDYT @Olshansk

app/client/cli/peer/list.go
h5law: Have gone through #801 and I think I have already addressed most of the relavent comments in this PR as is

p2p/debug/list.go
```diff
@@ -0,0 +1,130 @@
...
+	// TECHDEBT(#811): use `bus.GetPeerstoreProvider()` after peerstore provider
+	// is retrievable as a proper submodule.
+	pstoreProvider, err := peerstore_provider.GetPeerstoreProvider(bus)
+	if err != nil {
+		return err
+	}
+
+	switch routerType {
+	case StakedRouterType:
+		// TODO_IN_THIS_COMMIT: what about unstaked peers actors?
```
h5law, line 33: @Olshansk Am I missing something here but what does `what about unstaked peers actors` mean?

When using the staked router type we are only looking for staked actors? If we want both we would use `AllRouterTypes`

p2p/debug/list.go
```diff
@@ -0,0 +1,130 @@
...
+		pstore, err := pstoreProvider.GetUnstakedPeerstore()
+		if err != nil {
+			return fmt.Errorf("getting unstaked peerstore: %v", err)
+		}
+
+		peers = pstore.GetPeerList()
+	case AllRouterTypes:
+		routerPlurality = "s"
+
+		// TODO_IN_THIS_COMMIT: what about unstaked peers actors?
```
h5law, line 51: Again here I am confused, are we not already getting both staked and unstaked?

p2p/background/router.go
```diff
@@ -335,33 +330,36 @@ func (rtr *backgroundRouter) setupSubscription() (err error) {
 }
 
 // TECHDEBT(#859): integrate with `p2pModule#bootstrap()`.
```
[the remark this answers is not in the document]
bryanchriswhite: `#bootstrap()` is currently being called in one of the router's setup functions and should be called as part of the P2P module's bootstrapping process.

p2p/debug/list.go
bryanchriswhite: This TODO goes along with the commented code line below. It's referring to the current actor (this local P2P module instance ). My thinking at the time was something along the lines of: if the current actor is unstaked, it may not make sense to ask the peerstore provider for the staked actor set. Perhaps it should indeed work as written and it would be the responsibility of the P2P module to ensure the appropriate peerstore provider implemented is registered and then that of the node to ensure a compatible persistence module implementation is also registered (which may not yet exist), for example in the case of light client actors.

h5law: @bryanchriswhite I appreciate your response during your time away so no rush on this reply. But AFAIK if we are checking whether the node is staked we need to retrieve the staked peerstore anyway? the `p2pModule.isStakedActor()` function is as follows:

```go
func (m *p2pModule) getStakedPeerstore() (typesP2P.Peerstore, error) {
	pstoreProvider, err := peerstore_provider.GetPeerstoreProvider(m.GetBus())
	if err != nil {
		return nil, err
	}

	return pstoreProvider.GetStakedPeerstoreAtHeight(
		m.GetBus().GetCurrentHeightProvider().CurrentHeight(),
	)
}

// isStakedActor returns whether the current node is a staked actor at the current height.
// Return an error if a peerstore can't be provided.
func (m *p2pModule) isStakedActor() (bool, error) {
	pstore, err := m.getStakedPeerstore()
	if err != nil {
		return false, fmt.Errorf("getting staked peerstore: %w", err)
	}

	// Ensure self address is present in current height's staked actor set.
	if self := pstore.GetPeer(m.address); self != nil {
		return true, nil
	}
	return false, nil
}
```

So if we check the addr from `bus.GetP2PModule().GetAddress()` against this staked list and its not there we return the unstaked peerstore? I feel like this defeats the purpose of us passing in the router type in the first place if we are going to ignore it? Or am I misunderstanding?

h5law:
To give an update on the changes I made:

- Merged #891 and then merged main into this branch
- Updated errors to be clear they were errors
- Refactored CLI to follow similar pattern to rest (creator functions)
- Consolidated peerstore provider retrieval into a single functon prior to #811 merger
- Fixed the fix of the flaky P2P test

There is 2 bugs that are being encountered in this branch currently:

1. on k8s localnet the nodes are unable to bootstrap
  - this is unique to k8s localnet and the feat/peer-discovery-list branch
  - this appears to be due to the peerstore using the wrong ips potentially?
2. the p1 peers list command does not work
  - this is only on localnet but presumed also on k8s however issue issue 1 blocks this from being tested
  - error message below
```json
{"level":"debug","time":"2023-07-27T10:29:40+01:00","message":"Creating P2P module"}
{"level":"debug","module":"p2p","time":"2023-07-27T10:29:40+01:00","message":"setupPeerstoreProvider"}
{"level":"debug","module":"p2p","time":"2023-07-27T10:29:40+01:00","message":"loaded peerstore provider..."}
{"level":"fatal","error":"parsing multiaddr from config: resolving peer IP for hostname: validator1: resolving peer IP for hostname: validator1, lookup validator1: no such host","time":"2023-07-27T10:29:40+01:00","message":"Failed to create p2p module"}
```
This is the error found in `p2p/utils/url_conversion.go` `Libp2pMultiaddrFromServiceURL` line 67
This is called when `./bin/p1 peer list` calls 
```go
func sendToStakedPeers(cmd *cobra.Command, debugMsgAny *anypb.Any) error {
    bus, err := helpers.GetBusFromCmd(cmd)
    if err != nil {
        return err
    }

    pstore, err := helpers.FetchPeerstore(cmd)
    if err != nil {
        logger.Global.Fatal().Err(err).Msg("unable to retrieve the pstore")
    }

    if pstore.Size() == 0 {
        logger.Global.Fatal().Msg("no validators found")
    }

    for _, peer := range pstore.GetPeerList() {
        if err := bus.GetP2PModule().Send(peer.GetAddress(), debugMsgAny); err != nil {
            logger.Global.Error().Err(err).Msg("failed to send debug message")
        }
    }
    return nil
}
```
Specifically in the Send call (see app/client/cli/peer/list.go)

Olshansk:
@h5law Appreciate for all the work, improvements, hours spent debugging as well as a concise update summary. I think this is in a good place to pause things for now and have @bryanchriswhite pick it up when he's back.

PR #916, opened 2023-07-18
[WIP][IBC] Implement ICS-02 Client Semantics

h5law:
**_Note: This PR is still a WIP and will be rebased in the future to clean up the commits. Remaining for this PR: Documentation, GetHostClientState(), Tests?, ???_**

## Description

reviewpad:summary

## Issue

Fixes #894 

## Type of change

Please mark the relevant option(s):

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Adds `ClientManager` submodule to manage the client implementations
- Adds the interface definitions and WASM implementations (in part minus the WASM bit) implementaions
- Adds the Pocket `ConsensusState` and `ClientState` protobufs for use in Pocket WASM clients

## Testing

- [ ] `make develop_test`; if any code changes were made
- [ ] `make test_e2e` on [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any code changes were made
- [ ] `e2e-devnet-test` passes tests on [DevNet](https://pocketnetwork.notion.site/How-to-DevNet-ff1598f27efe44c09f34e2aa0051f0dd); if any code was changed
- [ ] [Docker Compose LocalNet](https://github.com/pokt-network/pocket/blob/main/docs/development/README.md); if any major functionality was changed or introduced
- [ ] [k8s LocalNet](https://github.com/pokt-network/pocket/blob/main/build/localnet/README.md); if any infrastructure or configuration changes were made

## Required Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added, or updated, [`godoc` format comments](https://go.dev/blog/godoc) on touched members (see: [tip.golang.org/doc/comment](https://tip.golang.org/doc/comment))
- [ ] I have tested my changes using the available tooling
- [ ] I have updated the corresponding CHANGELOG

### If Applicable Checklist

- [ ] I have updated the corresponding README(s); local and/or global
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added, or updated, [mermaid.js](https://mermaid-js.github.io) diagrams in the corresponding README(s)
- [ ] I have added, or updated, documentation and [mermaid.js](https://mermaid-js.github.io) diagrams in `shared/docs/*` if I updated `shared/*`README(s)

PR #922, opened 2023-07-20
[IBC] Clone `cosmos/ics23` protobuf definitions into IBC repo

h5law:
This PR copies in the `cosmos/ics23` protobuf definition file into the IBC types directory.

Doing so allows the `protoc` compile to compile the protobuf definitions correctly such that they implement the `proto.Message` interface, as when compiled using the `gogoproto` fork used by cosmos they do not. This is to allow for the `ProofSpec` to be included in the light client `PocketClientState` message in #916. In this usecase the `ics23.SmtSpec` cannot be used directly due to the issues mention above.

### Summary generated by Reviewpad on 24 Jul 23 22:27 UTC
This pull request includes the following changes:

- A newly added file containing the implementation of types and functions related to proofs in the ibc package. It includes conversion functions and a definition of a SmtSpec variable.
- Changes in the "proofs.proto" file, including the addition of new enum types, message types, and options. These changes enhance the functionality related to proof generation and verification.
- Changes in the `Makefile`, including the detection of the operating system, addition of OS-specific commands, a new target for downloading proto definitions, modification of imported package path, and a comment header indicating the origin of the `proofs.proto` file.

Please review the entire diff for further details.

## Issue

Fixes N/A

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Copy the `cosmos/ics23` proto definition into the IBC types directory

## Testing

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -318,7 +318,9 @@ protogen_local: go_protoc-go-inject-tag ## Generate go structures for all of the
 	$(PROTOC_SHARED) -I=./p2p/types/proto --go_out=./p2p/types ./p2p/types/proto/*.proto
 
 	# IBC
-	make copy_ics23_proto
+	@if test ! -e "./ibc/types/proto/proofs.proto"; then \
+		make download_ics23_proto; \
+	fi
 	$(PROTOC_SHARED) -I=./ibc/types/proto --go_out=./ibc/types ./ibc/types/proto/*.proto
 
 	# echo "View generated proto files by running: make protogen_show"
@@ -326,15 +328,15 @@ protogen_local: go_protoc-go-inject-tag ## Generate go structures for all of the
 # CONSIDERATION: Some proto files contain unused gRPC services so we may need to add the following
 #                if/when we decide to include it: `grpc--go-grpc_opt=paths=source_relative --go-grpc_out=./output/path`
 
-.PHONY: copy_ics23_proto
-copy_ics23_proto:
-	echo "Downloading cosmos/ics23 proto definitions..."
-	curl -s -o ./ibc/types/proto/proofs.proto https://raw.githubusercontent.com/cosmos/ics23/master/proto/cosmos/ics23/v1/proofs.proto && \
+.PHONY: download_ics23_proto
+download_ics23_proto:
+	echo "Downloading cosmos/ics23 proto definitions..."; \
+	curl -s -o ./ibc/types/proto/proofs.proto https://raw.githubusercontent.com/cosmos/ics23/master/proto/cosmos/ics23/v1/proofs.proto; \
 	$(SEDI) \
-        -e '/^package/{N;d;}' \
-        -e 's@github.com/.*"@github.com/pokt-network/pocket/ibc/types"@g' \
+		-e '/^package/{N;d;}' \
+		-e 's@github.com/.*"@github.com/pokt-network/pocket/ibc/types"@g' \
 		./ibc/types/proto/proofs.proto && \
-	awk 'BEGIN { print "// ===== !! THIS IS CLONED FROM cosmos/ics23 !! =====\n" } { print }' ./ibc/types/proto/proofs.proto > tmpfile && mv tmpfile ./ibc/types/proto/proofs.proto
+	awk 'BEGIN { print "// ===== !! THIS IS CLONED FROM cosmos/ics23 !! =====\n" } { print }' ./ibc/types/proto/proofs.proto > tmpfile && mv tmpfile ./ibc/types/proto/proofs.proto; \
 
 .PHONY: protogen_docker_m1
 ## TECHDEBT: Test, validate & update.
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,15 @@ CWD ?= CURRENT_WORKING_DIRECTIONRY_NOT_SUPPLIED
 #  `VERBOSE_TEST="" make test_persistence` is an easy way to run the same tests without verbose output
 VERBOSE_TEST ?= -v
 
+# Detect OS using the $(shell uname -s) command
+ifeq ($(shell uname -s),Darwin)
+    # Add macOS-specific commands here
+    SEDI = sed -i ''
+else ifeq ($(shell uname -s),Linux)
+    # Add Linux-specific commands here
+    SEDI = sed -i
+endif
+
 .SILENT:
 
 .PHONY: list ## List all make targets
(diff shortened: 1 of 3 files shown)


ibc/types/proto/proofs.proto
```diff
@@ -0,0 +1,222 @@
+syntax = "proto3";
+
+// This file is a clone from the github.com/cosmos/ics23 repo, it has been
```
Olshansk, line 3: 1. I suggest even creating a small bash file that does a `wget` to the files you need (e.g. `https://raw.githubusercontent.com/cosmos/ics23/master/proto/cosmos/ics23/v1/proofs.proto`) so we can update it by running one command if it ever changes.

2. Make sure to highlight this comment by starting with a `// !! IMPORTANT !! This file is a clone`

3. What problems are you encountering or questions do you have? From my point of view, embedding the `.proto` definition instead of the compiled `.pb.go` file is the correct way to go about it.

Olshansk (changes requested):

ibc/types/proto/proofs.proto
h5law: So if you look in #916 I need to embed the `ProofSpec` into the `PocketClientState` message. This is only possible when 1. we import the proto file and 2. the type is a message.

As the cosmos gogoproto compiler doesnt produce types that satisy the `proto.Message` interface we cannot just use the `ics23.SmtSpec` type they define we must recompile using normal protoc (so it produces `proto.Message`) types and then create the type ourselves. This way we can use the proto file, import it and then use the type. Adding the conversion allows us to still use the `cosmos/ics23` library for the methods they define but gives us the ability to use the message aswell

Makefile
```diff
@@ -13,6 +13,15 @@ CWD ?= CURRENT_WORKING_DIRECTIONRY_NOT_SUPPLIED
 #  `VERBOSE_TEST="" make test_persistence` is an easy way to run the same tests without verbose output
 VERBOSE_TEST ?= -v
 
+# Detect OS using the $(shell uname -s) command
+ifeq ($(shell uname -s),Darwin)
```
Olshansk, line 17: Should I install https://daoyuan.li/a-normal-sed-on-mac/?

Makefile
```diff
@@ -309,13 +318,24 @@ protogen_local: go_protoc-go-inject-tag ## Generate go structures for all of the
 	$(PROTOC_SHARED) -I=./p2p/types/proto --go_out=./p2p/types ./p2p/types/proto/*.proto
 
 	# IBC
+	make copy_ics23_proto
```
Olshansk, line 321: Can you add a check to see if the file is available locally or not? This turns `protogen_local` into a target that makes a network request EVERY time whereas we should only do it when we need it.

Makefile
```diff
@@ -309,13 +318,24 @@ protogen_local: go_protoc-go-inject-tag ## Generate go structures for all of the
...
 
 	# echo "View generated proto files by running: make protogen_show"
 
 # CONSIDERATION: Some proto files contain unused gRPC services so we may need to add the following
 #                if/when we decide to include it: `grpc--go-grpc_opt=paths=source_relative --go-grpc_out=./output/path`
 
+.PHONY: copy_ics23_proto
+copy_ics23_proto:
+	echo "Downloading cosmos/ics23 proto definitions..."
+	curl -s -o ./ibc/types/proto/proofs.proto https://raw.githubusercontent.com/cosmos/ics23/master/proto/cosmos/ics23/v1/proofs.proto && \
```
Olshansk, line 332: I know both curl and wget get the job done, but I've seen `wget` used more frequently in these types of contexts

Makefile
```diff
@@ -309,13 +318,24 @@ protogen_local: go_protoc-go-inject-tag ## Generate go structures for all of the
...
 	# IBC
+	make copy_ics23_proto
 	$(PROTOC_SHARED) -I=./ibc/types/proto --go_out=./ibc/types ./ibc/types/proto/*.proto
 
 	# echo "View generated proto files by running: make protogen_show"
 
 # CONSIDERATION: Some proto files contain unused gRPC services so we may need to add the following
 #                if/when we decide to include it: `grpc--go-grpc_opt=paths=source_relative --go-grpc_out=./output/path`
 
+.PHONY: copy_ics23_proto
```
Olshansk, line 329: s/copy/download

Makefile
```diff
@@ -309,13 +318,24 @@ protogen_local: go_protoc-go-inject-tag ## Generate go structures for all of the
...
+copy_ics23_proto:
+	echo "Downloading cosmos/ics23 proto definitions..."
+	curl -s -o ./ibc/types/proto/proofs.proto https://raw.githubusercontent.com/cosmos/ics23/master/proto/cosmos/ics23/v1/proofs.proto && \
+	$(SEDI) \
+        -e '/^package/{N;d;}' \
+        -e 's@github.com/.*"@github.com/pokt-network/pocket/ibc/types"@g' \
+		./ibc/types/proto/proofs.proto && \
+	awk 'BEGIN { print "// ===== !! THIS IS CLONED FROM cosmos/ics23 !! =====\n" } { print }' ./ibc/types/proto/proofs.proto > tmpfile && mv tmpfile ./ibc/types/proto/proofs.proto
```
Olshansk, line 337: Nice

ibc/types/proofs.go
```diff
@@ -0,0 +1,195 @@
+package types
+
+import ics23 "github.com/cosmos/ics23/go"
+
+var SmtSpec = &ProofSpec{
```
Olshansk, line 5: Add a comment with a link to where this is defined in the ics23 repo as well

Olshansk (changes requested):

Makefile
h5law: BSD sed is basically GNU sed with posix mode so no just makes format different

Makefile
h5law: wget is not installed out of the box on most linux distros as such I think it better to use curl which is part of the coreutils.

Olshansk: Noted regarded `coreutils`. I have found `wget` (verified by chatgpt) to be the more common default, but not a block by any means. 
Screenshot 2023-07-25 at 1 49 27 PM

Olshansk (approved):

PR #924, opened 2023-07-21
[Consensus] Add MinBlockTime to delay mempool reaping

### LINKED ISSUE #705: [Consensus] Block time - implement pacemaker configs to set min block time in auto-mode
## Objective

Implement a block time configuration feature in the Pacemaker Protocol, which allows setting a minimum block time in auto-mode for improved scalability and lower resource consumption while maintaining consistent Commit times.

## Origin Document

?

## Goals

- Develop a configuration setting to define a minimum block time in the Pacemaker Protocol's auto-mode.
- Optimize the Pacemaker Protocol to work efficiently with the new block time configuration, ensuring a balance between responsiveness, network speed, and consistent Commit times.

## Deliverable

- [ ] Implement a new configuration setting in the Pacemaker Protocol that allows (users?) to set a minimum block time in auto-mode
- [ ] Update the Pacemaker Protocol, specifically the QuorumCertificate processing and StepTimeouts mechanism, to work efficiently with the new block time configuration while maintaining consistent Commit times and upholding the SLA with the network's Applications.

- Overhauling the entire Pacemaker Protocol or consensus algorithm.
- Developing additional features unrelated to the block time configuration.

## General issue deliverables

## Testing Methodology

---

red-0ne:
## Description

This PR implements delaying mempool reaping by `MinBlockTime`, resetting the timer at every new height

All the logic lives in `consensus/pacemaker/module.go`, and the actual delay is happening `consensus/hotstuff_leader.go`

reviewpad:summary

## Issue

Fixes #705 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Major breaking change
- [ ] Documentation
- [ ] Other 

## List of changes

- Change #1
- Change #2
- ...

## Testing

## Required Checklist

### If Applicable Checklist

### DIFF
--- a/consensus/e2e_tests/pacemaker_test.go
+++ b/consensus/e2e_tests/pacemaker_test.go
@@ -172,6 +172,56 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
 	advanceTime(t, clockMock, paceMakerTimeout+10*time.Millisecond)
 }
 
+func TestPacemakerMinBlockTime(t *testing.T) {
+	// Test preparation
+	clockMock := clock.NewMock()
+	timeReminder(t, clockMock, time.Second)
+
+	// UnitTestNet configs
+	paceMakerTimeoutMsec := uint64(10000)
+	consensusMessageTimeout := time.Duration(paceMakerTimeoutMsec / 5) // Must be smaller than pacemaker timeout because we expect a deterministic number of consensus messages.
+	paceMakerMinBlockTimeMsec := uint64(5000)                          // Make sure it is larger than the consensus message timeout
+	runtimeMgrs := GenerateNodeRuntimeMgrs(t, numValidators, clockMock)
+	for _, runtimeConfig := range runtimeMgrs {
+		consCfg := runtimeConfig.GetConfig().Consensus.PacemakerConfig
+		consCfg.TimeoutMsec = paceMakerTimeoutMsec
+		consCfg.MinBlockTimeMsec = paceMakerMinBlockTimeMsec
+	}
+	buses := GenerateBuses(t, runtimeMgrs)
+
+	// Create & start test pocket nodes
+	eventsChannel := make(modules.EventsChannel, 100)
+	pocketNodes := CreateTestConsensusPocketNodes(t, buses, eventsChannel)
+	err := StartAllTestPocketNodes(t, pocketNodes)
+	require.NoError(t, err)
+
+	// Debug message to start consensus by triggering next view
+	for _, pocketNode := range pocketNodes {
+		TriggerNextView(t, pocketNode)
+	}
+
+	consMod := pocketNodes[1].GetBus().GetConsensusModule()
+
+	newRoundMessages, err := waitForProposalMsgs(t, clockMock, eventsChannel, pocketNodes, 1, uint8(consensus.NewRound), 0, 0, numValidators*numValidators, consensusMessageTimeout, true)
+	require.NoError(t, err)
+	whenBroadcast := clockMock.Now()
+	broadcastMessages(t, newRoundMessages, pocketNodes)
+	finishedBroadcast := uint64(clockMock.Now().Sub(whenBroadcast).Milliseconds())
+	beforePrepareTimeout := time.Duration((paceMakerMinBlockTimeMsec - finishedBroadcast) * uint64(time.Millisecond))
+
+	step := typesCons.HotstuffStep(consMod.CurrentStep())
+	require.Equal(t, consensus.NewRound, step)
+
+	advanceTime(t, clockMock, clock.Duration(beforePrepareTimeout))
+	step = typesCons.HotstuffStep(consMod.CurrentStep())
+	// Should still be blocking proposal step
+	require.Equal(t, consensus.NewRound, step)
+
+	//advanceTime(t, clockMock, 8000*time.Millisecond)
+	//step = typesCons.HotstuffStep(consMod.CurrentStep())
+	//require.Equal(t, consensus.Prepare, step)
+}
+
 // TODO: Implement these tests and use them as a starting point for new ones. Consider using ChatGPT to help you out :)
 
 func TestPacemakerDifferentHeightsCatchup(t *testing.T) {
--- a/consensus/hotstuff_leader.go
+++ b/consensus/hotstuff_leader.go
@@ -35,8 +35,6 @@ func (handler *HotstuffLeaderMessageHandler) HandleNewRoundMessage(m *consensusM
 		return
 	}
 
-	// DISCUSS: Do we need to pause for `MinBlockFreqMSec` here to let more transactions or should we stick with optimistic responsiveness?
-
 	if err := m.didReceiveEnoughMessageForStep(NewRound); err != nil {
 		m.logger.Info().Fields(hotstuffMsgToLoggingFields(msg)).Msgf("⏳ Waiting ⏳for more messages; %s", err.Error())
 		return
@@ -64,6 +62,11 @@ func (handler *HotstuffLeaderMessageHandler) HandleNewRoundMessage(m *consensusM
 	// TODO: Add test to make sure same block is not applied twice if round is interrupted after being 'Applied'.
 	// TODO: Add more unit tests for these checks...
 	if m.shouldPrepareNewBlock(highPrepareQC) {
+		doPrepare := m.paceMaker.ProcessDelayedBlockPrepare()
+		if !doPrepare {
+			m.logger.Info().Msg("skip prepare new block")
+			return
+		}
 		block, err := m.prepareBlock(highPrepareQC)
 		if err != nil {
 			m.logger.Error().Err(err).Msg(typesCons.ErrPrepareBlock.Error())
--- a/consensus/pacemaker/module.go
+++ b/consensus/pacemaker/module.go
@@ -3,6 +3,7 @@ package pacemaker
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	consensusTelemetry "github.com/pokt-network/pocket/consensus/telemetry"
@@ -38,6 +39,7 @@ type Pacemaker interface {
 	PacemakerDebug
 
 	ShouldHandleMessage(message *typesCons.HotstuffMessage) (bool, error)
+	ProcessDelayedBlockPrepare() bool
 
 	RestartTimer()
 	NewHeight()
@@ -48,9 +50,10 @@ type pacemaker struct {
 	base_modules.IntegrableModule
 	base_modules.InterruptableModule
 
-	pacemakerCfg    *configs.PacemakerConfig
-	roundTimeout    time.Duration
-	roundCancelFunc context.CancelFunc
+	pacemakerCfg         *configs.PacemakerConfig
+	roundTimeout         time.Duration
+	roundCancelFunc      context.CancelFunc
+	latestPrepareRequest latestPrepareRequest
 
 	// Only used for development and debugging.
 	debug pacemakerDebug
@@ -58,6 +61,15 @@ type pacemaker struct {
 	logger *modules.Logger
 	// REFACTOR: logPrefix should be removed in exchange for setting a namespace directly with the logger
 	logPrefix string
+}
+
+// Structure to handle delaying block preparation (reaping the block mempool)
+type latestPrepareRequest struct {
+	m              sync.Mutex
+	ch             chan bool
+	cancelFunc     context.CancelFunc
+	blockProposed  bool
+	deadlinePassed bool
 }
 
 func CreatePacemaker(bus modules.Bus, options ...modules.ModuleOption) (modules.Module, error) {
@@ -85,6 +97,13 @@ func (*pacemaker) Create(bus modules.Bus, options ...modules.ModuleOption) (modu
 		debugTimeBetweenStepsMsec: m.pacemakerCfg.GetDebugTimeBetweenStepsMsec(),
 		quorumCertificate:         nil,
 	}
+	m.latestPrepareRequest = latestPrepareRequest{
+		m:              sync.Mutex{},
+		ch:             nil,
+		cancelFunc:     nil,
+		blockProposed:  false,
+		deadlinePassed: false,
+	}
 
 	return m, nil
 }
@@ -92,6 +111,7 @@ func (*pacemaker) Create(bus modules.Bus, options ...modules.ModuleOption) (modu
 func (m *pacemaker) Start() error {
 	m.logger = logger.Global.CreateLoggerForModule(m.GetModuleName())
 	m.RestartTimer()
+	m.RestartBlockProposalTimer()
 	return nil
 }
 
@@ -171,6 +191,7 @@ func (m *pacemaker) ShouldHandleMessage(msg *typesCons.HotstuffMessage) (bool, e
 
 func (m *pacemaker) RestartTimer() {
 	// NOTE: Not deferring a cancel call because this function is asynchronous.
+	// DISCUSS: Should we have a lock to manipulate m.roundCancelFunc?
 	if m.roundCancelFunc != nil {
 		m.roundCancelFunc()
 	}
@@ -187,6 +208,45 @@ func (m *pacemaker) RestartTimer() {
 			}
 		case <-clock.After(m.roundTimeout + timeoutBuffer):
 			return
+		}
+	}()
+}
+
+func (m *pacemaker) RestartBlockProposalTimer() {
+	m.latestPrepareRequest.m.Lock()
+	defer m.latestPrepareRequest.m.Unlock()
+
+	if m.latestPrepareRequest.ch != nil {
+		m.latestPrepareRequest.ch <- false
+	}
+
+	if m.latestPrepareRequest.cancelFunc != nil {
+		m.latestPrepareRequest.cancelFunc()
+	}
+
+	m.latestPrepareRequest.blockProposed = false
+	m.latestPrepareRequest.deadlinePassed = false
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	m.latestPrepareRequest.cancelFunc = cancel
+	clock := m.GetBus().GetRuntimeMgr().GetClock()
+	minBlockTime := time.Duration(m.pacemakerCfg.MinBlockTimeMsec * uint64(time.Millisecond))
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			return
+		case <-clock.After(minBlockTime):
+			m.latestPrepareRequest.m.Lock()
+			defer m.latestPrepareRequest.m.Unlock()
+
+			if m.latestPrepareRequest.ch != nil {
+				m.latestPrepareRequest.blockProposed = true
+				m.latestPrepareRequest.ch <- true
+				m.latestPrepareRequest.ch = nil
+			}
+
+			m.latestPrepareRequest.deadlinePassed = true
 		}
 	}()
 }
@@ -225,6 +285,7 @@ func (m *pacemaker) InterruptRound(reason string) {
 
 func (m *pacemaker) NewHeight() {
 	defer m.RestartTimer()
+	defer m.RestartBlockProposalTimer()
 
 	consensusMod := m.GetBus().GetConsensusModule()
 	consensusMod.ResetRound(true)
@@ -241,6 +302,29 @@ func (m *pacemaker) NewHeight() {
 		CounterIncrement(
 			consensusTelemetry.CONSENSUS_BLOCKCHAIN_HEIGHT_COUNTER_NAME,
 		)
+}
+
+func (m *pacemaker) ProcessDelayedBlockPrepare() bool {
+	m.latestPrepareRequest.m.Lock()
+
+	if m.latestPrepareRequest.blockProposed {
+		return false
+	}
+
+	if m.latestPrepareRequest.deadlinePassed {
+		return true
+	}
+
+	// there is already a block preparer candidate, we cancel it and start a new one
+	// DISCUSS: This is needed if we want to get the latest QC for the block proposal.
+	if m.latestPrepareRequest.ch != nil {
+		m.latestPrepareRequest.ch <- false
+	}
+
+	m.latestPrepareRequest.ch = make(chan bool)
+	m.latestPrepareRequest.m.Unlock()
+
+	return <-m.latestPrepareRequest.ch
 }
 
 func (m *pacemaker) startNextView(qc *typesCons.QuorumCertificate, forceNextView bool) {
--- a/consensus/doc/CHANGELOG.md
+++ b/consensus/doc/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Implement minimum block production pace by delaying block preparation
+
 ## [0.0.0.54] - 2023-06-13
 
 - Fix tests
--- a/runtime/configs/config.go
+++ b/runtime/configs/config.go
@@ -122,6 +122,7 @@ func NewDefaultConfig(options ...func(*Config)) *Config {
 				TimeoutMsec:               defaults.DefaultPacemakerTimeoutMsec,
 				Manual:                    defaults.DefaultPacemakerManual,
 				DebugTimeBetweenStepsMsec: defaults.DefaultPacemakerDebugTimeBetweenStepsMsec,
+				MinBlockTimeMsec:          defaults.DefaultPacemakerMinBlockTimeMsec,
 			},
 		},
 		Utility: &UtilityConfig{
--- a/runtime/configs/proto/consensus_config.proto
+++ b/runtime/configs/proto/consensus_config.proto
@@ -15,4 +15,5 @@ message PacemakerConfig {
   uint64 timeout_msec = 1;
   bool manual = 2;
   uint64 debug_time_between_steps_msec = 3;
+  uint64 min_block_time_msec = 4;
 }
--- a/runtime/defaults/defaults.go
+++ b/runtime/defaults/defaults.go
@@ -43,6 +43,7 @@ var (
 	DefaultPacemakerTimeoutMsec               = uint64(10000)
 	DefaultPacemakerManual                    = true
 	DefaultPacemakerDebugTimeBetweenStepsMsec = uint64(1000)
+	DefaultPacemakerMinBlockTimeMsec          = uint64(5000)
 	// utility
 	DefaultUtilityMaxMempoolTransactionBytes = uint64(1024 ^ 3) // 1GB V0 defaults
 	DefaultUtilityMaxMempoolTransactions     = uint32(9000)
--- a/runtime/docs/CHANGELOG.md
+++ b/runtime/docs/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add a new MinBlockTimeMsec config field to consensus
+
 ## [0.0.0.44] - 2023-06-26
 
 - Add a new ServiceConfig field to servicer config
--- a/runtime/manager_test.go
+++ b/runtime/manager_test.go
@@ -1775,6 +1775,7 @@ func TestNewManagerFromReaders(t *testing.T) {
 							TimeoutMsec:               10000,
 							Manual:                    true,
 							DebugTimeBetweenStepsMsec: 1000,
+							MinBlockTimeMsec:          5000,
 						},
 						ServerModeEnabled: true,
 					},


consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,56 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+		TriggerNextView(t, pocketNode)
+	}
+
+	consMod := pocketNodes[1].GetBus().GetConsensusModule()
+
+	newRoundMessages, err := waitForProposalMsgs(t, clockMock, eventsChannel, pocketNodes, 1, uint8(consensus.NewRound), 0, 0, numValidators*numValidators, consensusMessageTimeout, true)
+	require.NoError(t, err)
+	whenBroadcast := clockMock.Now()
+	broadcastMessages(t, newRoundMessages, pocketNodes)
+	finishedBroadcast := uint64(clockMock.Now().Sub(whenBroadcast).Milliseconds())
```
h5law, line 209: does this not == 0? you dont advance time between whenBroadcast and finishedBroacast

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,56 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+	beforePrepareTimeout := time.Duration((paceMakerMinBlockTimeMsec - finishedBroadcast) * uint64(time.Millisecond))
+
+	step := typesCons.HotstuffStep(consMod.CurrentStep())
+	require.Equal(t, consensus.NewRound, step)
+
+	advanceTime(t, clockMock, clock.Duration(beforePrepareTimeout))
```
h5law, line 215: so this advances by min block time

consensus/e2e_tests/pacemaker_test.go
red-0ne: I need to check that again, but leader should wait for other's messages before preparing the block. I understand that broadcasting will stop at when the leader got all its messages, but not necessarily start processing them.

consensus/e2e_tests/pacemaker_test.go
red-0ne: Yet it should not pas to the Prepare step

red-0ne: `advanceTime(t, clockMock, clock.Duration(0))` makes consensus step into Prepare step, while removing that line does not. I suspect how the mock handles with go routines.

Olshansk:
@red-0ne Please make sure to update the issues & PR meta fields in the future. Updated it in this one for reference.

Screenshot 2023-07-24 at 11 06 09 AM

consensus/pacemaker/module.go
```diff
@@ -60,6 +63,16 @@ type pacemaker struct {
 	logPrefix string
 }
 
+// Structure to handle delaying block preparation (reaping the block mempool)
```
Olshansk, line 66:
```suggestion
// Structure to handle delaying block preparation (reaping the block mempool)
// by adding a delay before the next prepare request
```

consensus/pacemaker/module.go
```diff
@@ -60,6 +63,16 @@ type pacemaker struct {
...
+type latestPrepareRequest struct {
+	m              sync.Mutex
```
Olshansk, line 68: Can you add comments to each of the types. The names alone don't make it clear to me how/why it sused.

consensus/pacemaker/module.go
```diff
@@ -60,6 +63,16 @@ type pacemaker struct {
 	logPrefix string
 }
 
+// Structure to handle delaying block preparation (reaping the block mempool)
+type latestPrepareRequest struct {
```
Olshansk, line 67: s/latestPrepareRequest/prepareStepDelayer`?

consensus/pacemaker/module.go
```diff
@@ -171,6 +192,7 @@ func (m *pacemaker) ShouldHandleMessage(msg *typesCons.HotstuffMessage) (bool, e
 
 func (m *pacemaker) RestartTimer() {
 	// NOTE: Not deferring a cancel call because this function is asynchronous.
+	// DISCUSS: Should we have a lock to manipulate m.roundCancelFunc?
```
Olshansk, line 195: If we do, we need to add a test case where it's an issue.

1. Have you hit issues with it?
2. Why do you think we need it?
3. If not, a a TECHDEBT and tend it it separately? `// TECHDEBT: Add a mutex for this cancel function and include a test with the implmentation`

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,87 @@ func (m *pacemaker) NewHeight() {
...
 }
 
+// This is called each time there is a NewRound message received by the leader from replicas
+// With the introduction of MinBlockTimeMsec delay, multiple concurrent calls may happen
+// It makes sure that:
+//   - Block proposal is made by only one of the possible `HotstuffLeaderMessageHandler.HandleNewRoundMessage()` concurrent (because delayed) calls
+//   - If the timer expires, the first call to this method will trigger the block proposal
+//   - If a late message is received after a block is proposed by another call, the late message is discarded
+//   - Reads and affectations to pacemaker.latestPrepareRequest state are protected by a mutex
+func (m *pacemaker) ProcessDelayedBlockPrepare(currentHeight uint64) chan bool {
```
Olshansk, line 275: Does this need to be publically exposed?


If so, make sure it's part of the pacemaker interface under the `shared/` module?

Olshansk, line 275: s/ProcessDelayedBlockPreparation/delayBlockPrepation ?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,87 @@ func (m *pacemaker) NewHeight() {
...
+	defer m.latestPrepareRequest.m.Unlock()
+
+	// Prepare channel for the the current request
+	ch := make(chan bool)
+
+	// First time to build a block for current height, cancel previous timer if any
+	if m.latestPrepareRequest.delayedHeight < currentHeight {
+		// Discard the request building the old height
+		if m.latestPrepareRequest.ch != nil {
+			m.latestPrepareRequest.ch <- false
```
Olshansk, line 286: Should we close the channel?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,87 @@ func (m *pacemaker) NewHeight() {
...
+		return ch
+	}
+
+	if m.latestPrepareRequest.blockProposed {
+		go func() { ch <- false }()
+
+		return ch
+	}
+
+	if m.latestPrepareRequest.ch != nil {
```
Olshansk, line 333:
```suggestion
	// If there is a channel signaling to propose a block or not, make sure it is not proposed
	if m.latestPrepareRequest.ch != nil {
```

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,87 @@ func (m *pacemaker) NewHeight() {
...
+//   - Reads and affectations to pacemaker.latestPrepareRequest state are protected by a mutex
+func (m *pacemaker) ProcessDelayedBlockPrepare(currentHeight uint64) chan bool {
+	m.latestPrepareRequest.m.Lock()
+	defer m.latestPrepareRequest.m.Unlock()
+
+	// Prepare channel for the the current request
+	ch := make(chan bool)
+
+	// First time to build a block for current height, cancel previous timer if any
+	if m.latestPrepareRequest.delayedHeight < currentHeight {
```
Olshansk, line 283: I don't fully understand the need for `delayedHeight` yet. Can you elaborate (and update the comments)?

There's no way around it?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,87 @@ func (m *pacemaker) NewHeight() {
...
+		if m.latestPrepareRequest.ch != nil {
+			m.latestPrepareRequest.ch <- false
+		}
+		// Discard the timer for the old height
+		if m.latestPrepareRequest.cancelFunc != nil {
+			m.latestPrepareRequest.cancelFunc()
+		}
+
+		m.latestPrepareRequest.ch = ch
+		m.latestPrepareRequest.blockProposed = false
```
Olshansk, line 294: We should emphaize that all of these are mutually exclusive scenarios/triggers for block proposal. Is that correct?

Olshansk (changes requested):
I did not review `consensus/e2e_tests/pacemaker_test.go` yet.

Just minor comments / nits. 

Interesting approach but I do want to look at it with fresh eyes tomorrow. Just want to see if we can make it a bit simpler.

Other than that, it's a fucking awesome start!

consensus/pacemaker/module.go
red-0ne: Yes, I was unsure if the channel would emit it last value once closed. If that's the case, we can safely close it.

consensus/e2e_tests/pacemaker_test.go
red-0ne: See new impl.

consensus/e2e_tests/utils_test.go
```diff
@@ -840,6 +840,8 @@ func advanceTime(t *testing.T, clck *clock.Mock, duration time.Duration) {
 	clck.Add(duration)
 	t.Logf("[⌚ CLOCK ⏩] advanced by %v", duration)
 	logTime(t, clck)
+	// Give goroutines a chance to run
+	clck.Add(0)
```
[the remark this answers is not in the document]
red-0ne: This clock mock implementation seems to advance time synchronously not giving a chance to go routines to trigger. This seems to do the trick though, I can get that back to paceMaker tests but it allowed to have exact time precision.

consensus/pacemaker/module.go
red-0ne: Yes, it's basically called from the consensus module. Will make sure `shared/` is aligned

consensus/pacemaker/module.go
red-0ne: Was resolving an imaginary problem. The idea was that if a timer of an "old" proposal is active and another block proposal wants to setup its own timer; we should discard the old one.

This implies that the same node could to prepare two blocks consecutively without the network deciding on the first one. This should not happen because
* Of that timer (even though it's enforced at the leader level only --discuss this)
* If that node got to propose a block, then there's very low probability that it will be a leader for the next block (highly depends on leader election module)
* The protocol doesn't allow you to propose block h+1 if h is not decided on yet

Yet I'm not sure this could never happen :sweat_smile:

red-0ne:
>  Just want to see if we can make it a bit simpler.

We can. True that this implementation is intricate. It's coupling timer registration with the client/listener logic which should not. I will split this into a timer registration function and a listening one. It would be much simpler to reason about.

consensus/pacemaker/module.go
red-0ne: Updating the interface made some errors to raise at `consensus/module_consensus_pacemaker.go#L12`
```go
var _ modules.ConsensusPacemaker = &consensusModule{}
```

Which makes me question the assignment :thinking:

red-0ne:
@Olshansk the logic should be more readable and ready to review now. Which makes me think about the nature of the delay.

We are currently enforcing it at the leader level. Which means that if the leader wants to prepare the block earlier, there's nothing enforcing it not to do so.

But if we place the delay at the new round message emission point. All (honest) replicas will delay their new round messages so that the leader could not reach the consensus and build the block earlier. This would mean that each block proposal time is enforced by the whole network instead of by its leader only. We could floor the delay imposed by replicas (eg. -10%) and let the leader that window to fine grain the timer.

The problem I see with it, is that it makes reaching consensus always a hurry task, an adjustable offset could do the trick and even be dynamic. Also choreographing the replicas to distribute their deadlines over MinBlockTime period may also be worth digging.

This should not be that much of a change :smile: Since at the new round construction level, any node knows if it's the leader of that round or not:
* If it is a leader
  * Registers the timer at that exact moment (the earliest a node knows there's a new round and that it's the leader)
  * Sends the new round to itself (acts like a replica)
  * If it reaches new round quorum it prepares the block by applying the delay logic according to its local timer
* If it's a replica
  * Registers the timer at that exact moment (the earliest that replica commits for the new round --should be same as leader)
  * Delays broadcasting new round message to the leader by MinBockTime - AdjustableOffset ± (Randomness or distribution function)

--
This way, for a 15minutes block time the whole network could eg. enforce that a block is proposed no earlier than 14minutes, and leader makes sure it lands at the 15minutes mark

This may be overprotocoling too :rofl:

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+	require.Equal(t, consensus.NewRound, step)
+
+	// Advance time just enough to trigger minBlockTime
+	advanceTime(t, clockMock, pivotTime)
+	step = typesCons.HotstuffStep(leaderConsensusModule.CurrentStep())
+
+	// Time advanced by minBlockTime
+	require.Equal(t, uint64(clockMock.Now().UnixMilli()), paceMakerMinBlockTimeMsec)
+	// Leader is at proposal step
+	require.Equal(t, consensus.Prepare, step)
```
h5law, line 264: 🔥

consensus/e2e_tests/utils_test.go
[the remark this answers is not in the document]
h5law: Is there any documentation explaining this in their godocs you could link to?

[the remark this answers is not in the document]
red-0ne: Looking into the clock modules code, it seems that `clck.Add(duration)` executes as following:
`clck.Add->time.Sleep(1 * time.Millisecond)`. This may be easily tested, otherwise it may be happening somewhere here [`benbjohnson/clock/clock.go#L101C7-L101C7`](https://github.com/benbjohnson/clock/blob/master/clock.go#L101C7-L101C7)
```
// Continue to execute timers until there are no more before the new time.
for {
	if !m.runNextTimer(t) {
		break
	}
}
```

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
 		)
 }
 
+// This is called each time the system wants a delayed signal to build a block
```
Olshansk, line 268: Let's follow proper godoc comments: https://tip.golang.org/doc/comment

Unrelated request: If you are finding that I'm making requests that you didn't read about in https://github.com/pokt-network/pocket/blob/main/docs/development/CODE_REVIEW_GUIDELINES.md?
```suggestion
// RegisterMinBlockTimeDelay should be called when a delay should be introduced into proposing a new block
```

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+func (m *pacemaker) RegisterMinBlockTimeDelay() {
+	// Discard any previous timer
+	// DISCUSS: This should not happen, Identify cases where an active timer has to be discarded, remove cancellation logic if none
+	if m.prepareStepDelayer.cancelFunc != nil {
+		m.prepareStepDelayer.cancelFunc()
+	}
```
Olshansk, line 274:
```suggestion
	// Discard any previous timer if one exists
	if m.prepareStepDelayer.cancelFunc != nil {
		m.logger.Warn().Msg("RegisterMinBlockTimeDelay has an existing timer which should not happen. Releasing for now...")
		m.prepareStepDelayer.cancelFunc()
	}
```

consensus/hotstuff_leader.go
```diff
@@ -64,6 +62,15 @@ func (handler *HotstuffLeaderMessageHandler) HandleNewRoundMessage(m *consensusM
 	// TODO: Add test to make sure same block is not applied twice if round is interrupted after being 'Applied'.
 	// TODO: Add more unit tests for these checks...
 	if m.shouldPrepareNewBlock(highPrepareQC) {
+		// Place this where we want to start the timer
+		// Placed here, timer starts when we receive enough NewRound messages
+		m.paceMaker.RegisterMinBlockTimeDelay()
```
Olshansk, line 67:
```suggestion
		// Leader should prepare a new block. Introducing a delay based on configurations.
		m.paceMaker.RegisterMinBlockTimeDelay()
```

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
 		)
 }
 
+// This is called each time the system wants a delayed signal to build a block
+func (m *pacemaker) RegisterMinBlockTimeDelay() {
```
Olshansk, line 269: Shold this be called `RegisterMin...` or `StartMin..`?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+	}
+
+	m.prepareStepDelayer.blockProposed = false
+	m.prepareStepDelayer.deadlinePassed = false
+
+	minBlockTime := time.Duration(m.pacemakerCfg.MinBlockTimeMsec * uint64(time.Millisecond))
+	ctx, cancel := context.WithCancel(context.TODO())
+	m.prepareStepDelayer.cancelFunc = cancel
+
+	go func() {
```
Olshansk, line 283: Can you add a comment above `go func` explaining what's going on?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+		select {
+		case <-ctx.Done():
```
Olshansk, line 285:
```suggestion
		// only called if the delay timer is explicitly cancelled
		case <-ctx.Done():
```

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+			return
+		case <-m.GetBus().GetRuntimeMgr().GetClock().After(minBlockTime):
```
Olshansk, line 287:
```suggestion
		// called after the minimum block delay is exhausted meaning it is time to propse a new block
		case <-m.GetBus().GetRuntimeMgr().GetClock().After(minBlockTime):
```

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+			m.prepareStepDelayer.m.Lock()
+			defer m.prepareStepDelayer.m.Unlock()
+
+			// After the timeout, if there was any candidate request waiting for a signal, tell it to build the block
```
Olshansk, line 291: What is a `candidate request waiting for a signal`? Please update the comment for future readers.

consensus/pacemaker/module.go
```diff
@@ -60,6 +64,16 @@ type pacemaker struct {
 	logPrefix string
 }
 
+// Structure to handle delaying block preparation (reaping the block mempool)
```
Olshansk, line 67:
```suggestion
// prepareStepDelayer handles delays block preparation (reaping the block mempool)
```

consensus/pacemaker/module.go
```diff
@@ -60,6 +64,16 @@ type pacemaker struct {
...
+// by adding a delay before the next prepare request
```
Olshansk, line 68:
```suggestion
// by adding a delay before the next prepare request to prevent block creation
// from moving at the speed of the network (i.e. responsivness).
// See Tendermint, which is not responsive, for similar configurations; https://docs.tendermint.com/v0.33/tendermint-core/configuration.html
```

consensus/pacemaker/module.go
```diff
@@ -60,6 +64,16 @@ type pacemaker struct {
...
+type prepareStepDelayer struct {
+	m              sync.Mutex         // Mutex locking access to this structure and prevent inconsistent state
+	ch             chan bool          // Whenever there is a block proposal request arriving before the timeout, this channel is used to signal to it to whether build the block or not (if better candidate request with higher QC)
+	cancelFunc     context.CancelFunc // This is used to cancel an ongoing timeout. It should not happen, but future code changes may no longer preserve this guarantee, so this feature maintain its own cancellation logic
+	blockProposed  bool
+	deadlinePassed bool
```
Olshansk, line 74:
```suggestion
	delayExhausted bool // the delay for this step/round has already passed, so should create the block ASAP
```

consensus/pacemaker/module.go
```diff
@@ -60,6 +64,16 @@ type pacemaker struct {
 	logPrefix string
 }
 
+// Structure to handle delaying block preparation (reaping the block mempool)
+// by adding a delay before the next prepare request
+type prepareStepDelayer struct {
+	m              sync.Mutex         // Mutex locking access to this structure and prevent inconsistent state
+	ch             chan bool          // Whenever there is a block proposal request arriving before the timeout, this channel is used to signal to it to whether build the block or not (if better candidate request with higher QC)
+	cancelFunc     context.CancelFunc // This is used to cancel an ongoing timeout. It should not happen, but future code changes may no longer preserve this guarantee, so this feature maintain its own cancellation logic
+	blockProposed  bool
```
Olshansk, line 73:
```suggestion
	blockProposed  bool // a flag to capture whether a block was proposed while the timer was ticking
```

consensus/pacemaker/module.go
```diff
@@ -85,6 +99,13 @@ func (*pacemaker) Create(bus modules.Bus, options ...modules.ModuleOption) (modu
 		debugTimeBetweenStepsMsec: m.pacemakerCfg.GetDebugTimeBetweenStepsMsec(),
 		quorumCertificate:         nil,
 	}
+	m.prepareStepDelayer = prepareStepDelayer{
+		m:              sync.Mutex{},
+		ch:             nil,
+		cancelFunc:     nil,
+		blockProposed:  false,
+		deadlinePassed: false,
```
Olshansk, line 107: Thoughts on adding `minBlockTime := time.Duration(m.pacemakerCfg.MinBlockTimeMsec * uint64(time.Millisecond))` directly to this structure?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+			// From now on, build the block ASAP
+			m.prepareStepDelayer.deadlinePassed = true
+
+			// No need to cancel the context anymore
+			m.prepareStepDelayer.cancelFunc = nil
+		}
+	}()
+}
+
+// This is called when conditions are met by the leader to build a block but still needs to wait for the MinBlockTimeMsec delay before reaping the mempool
```
Olshansk, line 307:
```suggestion
// DelayBlockPreparation is called when conditions are met by the leader to build a block but still needs to wait for the MinBlockTimeMsec delay before reaping the mempool.
```

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+// With MinBlockTimeMsec delay, multiple concurrent calls may happen
+// It makes sure that:
+//   - Block proposal is made by only one of the possible `HotstuffLeaderMessageHandler.HandleNewRoundMessage()` concurrent (because delayed) calls
+//   - If the timer expires, the first call to this method will trigger the block proposal
+//   - If a late message is received AFTER the a block is marked as proposed by another call, the late message is discarded
+//   - Reads and affectations to pacemaker.DelayBlockPreparation state are protected by a mutex
```
Olshansk, line 313: affectations?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+func (m *pacemaker) DelayBlockPreparation() bool {
+	m.prepareStepDelayer.m.Lock()
+
```
Olshansk, line 316: Why not just execute `defer m.prepareStepDelayer.m.Unlock()` here?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+	if m.prepareStepDelayer.blockProposed {
+		m.prepareStepDelayer.m.Unlock()
+		return false
+	}
+
+	// If there already is a channel signaling block proposal, make sure it gives up
```
Olshansk, line 322:
```suggestion
	// If there already is a channel signaling block proposal, make sure it permits block proposal
```

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+	if m.prepareStepDelayer.ch != nil {
+		m.prepareStepDelayer.ch <- false
+		close(m.prepareStepDelayer.ch)
+	}
+
+	// Deadline has passed, no need to have a channel, propose a block now
+	if m.prepareStepDelayer.deadlinePassed {
+		m.prepareStepDelayer.blockProposed = true
```
Olshansk, line 330: Should this be `shoulProposeBlock` instead of `blockProposed` or am I misunderstanding?

consensus/pacemaker/module.go
```diff
@@ -243,6 +265,82 @@ func (m *pacemaker) NewHeight() {
...
+		return true
+	}
+
+	// We still need to wait, create a channel and discard the old candidate if any
+	ch := make(chan bool)
+	m.prepareStepDelayer.ch = ch
+	// We cannot defer the unlock here because the channel read is blocking
+	m.prepareStepDelayer.m.Unlock()
+
+	return <-ch
```
Olshansk, line 341: Can you update the function comments that this is a synchronous blocking function?

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+func TestPacemaker_MinBlockTime_BlockPrepAsapAfterTrigger(t *testing.T) {
+	t.Skip()
+}
+
+// TODO: Block preparation is always discarded if a new one with better QC is received within minBlockTime.
+func TestPacemaker_MinBlockTime_AllowOnlyLatestBlockPrep(t *testing.T) {
+	t.Skip()
+}
+
+// TODO: Mempool reaped is the one present at minBlockTime or later.
```
Olshansk, line 277: WOuldn't this happen by default? Need to either explain this in more detail or delete

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+func TestPacemaker_MinBlockTime_DelayReapMempool(t *testing.T) {
+	t.Skip()
+}
+
+// TODO: Successive blocks timings are at least minBlockTime apart.
+func TestPacemaker_MinBlockTime_BehaviorAcrossMultipleBlocks(t *testing.T) {
```
Olshansk, line 283: This looks important. Can you implement it in this PR?

```suggestion
func TestPacemaker_MinBlockTime_ConsecutiveBlocksAtLeastMinBlockTimeApart(t *testing.T) {
```

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
 	advanceTime(t, clockMock, paceMakerTimeout+10*time.Millisecond)
 }
 
+func TestPacemaker_MinBlockTime(t *testing.T) {
```
Olshansk, line 175:
```suggestion
func TestPacemaker_MinBlockTime_???(t *testing.T) {
```

Happy Path? Sad Path? Need a suffix

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+	// Test preparation
+	clockMock := clock.NewMock()
+	timeReminder(t, clockMock, time.Second)
+
+	// UnitTestNet configs
+	paceMakerTimeoutMsec := uint64(300000)
+	consensusMessageTimeout := time.Duration(paceMakerTimeoutMsec / 5) // Must be smaller than pacemaker timeout because we expect a deterministic number of consensus messages.
+	paceMakerMinBlockTimeMsec := uint64(5000)                          // Make sure it is larger than the consensus message timeout
```
Olshansk, line 183: Regarding `// Make sure it is larger than the consensus message timeout`. 

Can you just do math to guarantee it?

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+
+	// Debug message to start consensus by triggering next view
+	// Get leader out of replica set
+	for id, pocketNode := range pocketNodes {
+		TriggerNextView(t, pocketNode)
+		if id == leaderId {
+			leader[id] = pocketNode
+		} else {
+			replicas[id] = pocketNode
+		}
```
Olshansk, line 213: Can we not just use `pocketNodes[leaderId]` below and avoid create `replias` and `leader` altogether?

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+		numReplicas, // We want new round messages from replicas only
+		consensusMessageTimeout, false,
+	)
+	require.NoError(t, err)
+
+	// Broadcast new round messages to leader to build a block
+	broadcastMessages(t, newRoundMessages, leader)
+
+	var step typesCons.HotstuffStep
+	var pivotTime = 1 * time.Millisecond // Min time it takes to switch from NewRound to Prepare step
```
Olshansk, line 232: Why is it called `pivotTime`?
```suggestion
	pivotTime := 1 * time.Millisecond // Min time it takes to switch from NewRound to Prepare step
```

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+
+	// Give go routines time to trigger
+	advanceTime(t, clockMock, 0)
+
+	// We get consensus module from leader to get its POV
+	leaderConsensusModule := leader[leaderId].GetBus().GetConsensusModule()
```
Olshansk, line 238:
```suggestion
	leaderConsensusModule := pocketNodes[leaderId].GetBus().GetConsensusModule()
```

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+		nodeLeader := pocketNode.GetBus().GetConsensusModule().GetLeaderForView(1, 0, uint8(consensus.NewRound))
+		require.Equal(t, typesCons.NodeId(nodeLeader), leaderId)
+	}
+
+	// Timer is blocking the proposal step
+	step = typesCons.HotstuffStep(leaderConsensusModule.CurrentStep())
+	require.Equal(t, consensus.NewRound, step)
+
+	// Advance time right before minBlockTime triggers
+	advanceTime(t, clockMock, time.Duration(paceMakerMinBlockTimeMsec*uint64(time.Millisecond))-pivotTime)
```
Olshansk, line 251: Can you assign `time.Duration(paceMakerMinBlockTimeMsec*uint64(time.Millisecond))-pivotTime` to a variable above this with a self explantory name?

consensus/e2e_tests/pacemaker_test.go
```diff
@@ -172,6 +172,118 @@ func forcePacemakerTimeout(t *testing.T, clockMock *clock.Mock, paceMakerTimeout
...
+
+	// Should still be blocking proposal step
+	step = typesCons.HotstuffStep(leaderConsensusModule.CurrentStep())
+	require.Equal(t, consensus.NewRound, step)
+
+	// Advance time just enough to trigger minBlockTime
```
Olshansk, line 257: Good test and great comments 🙌 !

Olshansk (changes requested):
@red-0ne 

1. Can you create a new GitHub Issue for your [comment](https://github.com/pokt-network/pocket/pull/924#issuecomment-1652121297). The idea has legs and it's important to discuss but outside the scope of this ticket.

2. You have great diagrams in the github description, but it'll get lost and forgotten. Can you add it to a README either in this PR or another one?

red-0ne:
> You have great diagrams in the github description, but it'll get lost and forgotten. Can you add it to a README either in this PR or another one?

I will create pace maker documentation along the requested changes.

Olshansk:
> I will create pace maker documentation along the requested changes.

Sounds great! 

- Please `re-request` the review when I should take another look
- For comments that are tended to and don't need an action, go ahead and resolve them. If more conversation is needed, leave them open and we will keep the discussion going.

_Note to self: Add the above to the code review guidelines._

consensus/pacemaker/module.go
red-0ne: At the end of this function
```go
func (m *pacemaker) DelayBlockPreparation() bool {
  ...
  m.prepareStepDelayer.ch = ch
  // We cannot defer the unlock here because the channel read is blocking
  m.prepareStepDelayer.m.Unlock()

  return <-ch
}
```
 we have `return <-ch` which will be blocking the call and preventing the `defer ...m.Unlock()` from being called  IIUC. The desired behavior is to `Unlock` right after the last assignment made to `m.prepeareStepDelayer`

consensus/pacemaker/module.go
red-0ne: `ProcessDelayedBlockPrepare` should only be used by `Consensus` module (specifically `hotstuff_leader`, and maybe `hotstuff_replica`). Since the PaceMaker is acting like a submodule to `Consensus`, there is nothing to expose through the `Bus`.

red-0ne:
> I will create pace maker documentation along the requested changes.

PaceMaker documentation added to [PR#915](https://github.com/pokt-network/pocket/pull/951/commits/30d1115922db81fb43f8eff66adee041a82d8e0d)

red-0ne (commented):
@Olshansk , I resolved most of the change requests. Only remains these two discussions:

* Exposing ` ProcessDelayedBlockPrepare` on `/shared` which I don't think we should do.
* Deferring `mutex.Unlock`

Anything else to re-check before finalizing it?

consensus/pacemaker/module.go
```diff
@@ -39,7 +39,7 @@ type Pacemaker interface {
 	PacemakerDebug
 
 	ShouldHandleMessage(message *typesCons.HotstuffMessage) (bool, error)
-	RegisterMinBlockTimeDelay()
+	StartMinBlockTimeDelay()
 	DelayBlockPreparation() bool
```
Olshansk, line 43:
```suggestion
	// WARNING: DelayBlockPreparation is a synchronous blocking call that acquires a mutex and prevents block propagation until its complete.
	DelayBlockPreparation() bool
```

consensus/pacemaker/module.go
```diff
@@ -39,7 +39,7 @@ type Pacemaker interface {
 	PacemakerDebug
 
 	ShouldHandleMessage(message *typesCons.HotstuffMessage) (bool, error)
-	RegisterMinBlockTimeDelay()
+	StartMinBlockTimeDelay()
```
Olshansk, line 42:
```suggestion
	// StartMinBlockTimeDelay configures `prepareStepDelayer` in preparation for delaying block proposal
	StartMinBlockTimeDelay()
```

consensus/pacemaker/module.go
Olshansk: > we have return <-ch which will be blocking the call and preventing the defer ...m.Unlock() from being called IIUC.
> The desired behavior is to Unlock right after the last assignment made to m.prepeareStepDelayer

These are super valuable comments. Can you add them to the code?

Olshansk (changes requested):

Olshansk:
@red-0ne Just a couple more lingering edits/suggestions. 

Also, I wanted to confirm that you tested this with Localnet and everything worked fine end-to-end? 👍 
?

Olshansk:
@red-0ne 

> The delay is effectively applied when running in non-manual mode.

That makes sense. This is intended for `non-manual mode` only. Wdyt of adding an assertion for this?

> Even if I had issues with block building (which seems to be coming from somewhere else).

Consider trying to revert to the `main` branch and seeing if you see the same issues in that context. If that's the case, and it's an issue on main, I'll jump in to debug immediately. If not, let me know if/when I should dive deeper into the code in this PR.

red-0ne:
@Olshansk 
> That makes sense. This is intended for `non-manual mode` only. Wdyt of adding an assertion for this?

The only way I see to test this is through real clock (and wait for consensus on new round to happen). Any ideas?
 
> Consider trying to revert to the `main` branch and seeing if you see the same issues in that context. If that's the case, and it's an issue on main, I'll jump in to debug immediately. If not, let me know if/when I should dive deeper into the code in this PR.

Created an issue for that #965, it is not blocking this work.

Olshansk:
> You mean enable the feature only when manual is false ?

Yes.

If `manual == true`  => `skip code path for minBlockTime`

Olshansk (changes requested):
This PR has 106 lines changed relative to main. Please merge with main and resolve conflicts and re-request a review then.

PR #874, opened 2023-07-04
[E2E Test] Utilities for State Sync Test

Olshansk:
### Summary generated by Reviewpad on 03 Aug 23 20:11 UTC
This pull request introduces several changes across multiple files. 

Here is a summary of the changes:

1. `go.mod`:
   - The `golang.org/x/text` dependency is now listed explicitly instead of being an indirect dependency.

2. `scenario_test.go`:
   - Added a new feature called "State Sync Namespace" that includes various commands and waits for specific amounts of time.
   - Code improvements and TODO comments have been added.

3. `FAQ.md`:
   - Updated an issue related to starting LocalNet with SELinux on an operating system. Replaced the command `make compose_and_watch` with `make lightweight_localnet` to avoid permission denied errors.

4. `.gitignore`:
   - Removed the entry "main" from the list of ignored files.
   - Removed the entries "rpc/server.gen.go" and "rpc/client.gen.go" from the list of ignored files.
   - Added the entry "**/gomock_reflect_*/" to ignore mock temporary files.

5. `e2e/README.md`:
   - Added a new section on `Keywords`.
   - Modified scenario descriptions and code examples to replace instances of "Validator" with "Node".
   - Included a flowchart depicting the E2E scenarios with updated terminology.

6. Consensus module files:
   - Added a new logging statement in the `HandleDebugMessage` function.
   - Simplified the handling of the `DEBUG_CONSENSUS_RESET_TO_GENESIS` action.

7. `.tiltignore`:
   - Removed the entry "main" from the list of ignored files.
   - Removed the entries "rpc/server.gen.go" and "rpc/client.gen.go" from the list of ignored files.

8. `CHANGELOG.md`:
   - Updated build commands and added a new section on `Keywords`.

9. `persistence/docs/CHANGELOG.md`:
   - Several changes related to deprecation, addition, and fixing of functions and issues.

10. `validator.feature`:
    - Renamed file from "valdator.feature" to "validator.feature".
    - Updated scenario titles and step descriptions to use more descriptive terminology.
    - Replaced references to "validator" with "node".

11. `tilt_helpers.go`:
    - Added a new file containing functions related to syncing network configuration and checking package installation.

12. `debug.go`:
    - Added new debug commands and subcommands.
    - Updated existing functions and added new functions for debug actions.
    
13. `account.feature`:
    - Added a new file containing scenarios for testing node account functionalities.

14. `README.md` files:
    - Updated sections, titles, dependencies, and instructions in various README.md files.

[...]

PR #954, opened 2023-07-31
hack: 😴 sleep enough for cli debug subcommands to broadcast

0xBigBoss:
When running commands like `p1 dui TriggerNextView`, seems the program exits faster than the p2p module can broadcast. This PR introduces a magic number to give the p2p module _enough_ time to broadcast the debug messages when run as a subcommand.

This is working for my e2e testing for now.

### DIFF
--- a/app/client/cli/debug.go
+++ b/app/client/cli/debug.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"os"
+	"time"
 
 	"github.com/manifoldco/promptui"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


app/client/cli/debug.go
```diff
@@ -50,7 +51,10 @@ func newDebugUISubCommands() []*cobra.Command {
 			Use:               promptItem,
 			PersistentPreRunE: helpers.P2PDependenciesPreRunE,
 			Run: func(cmd *cobra.Command, _ []string) {
+				// TECHDEBT: this is a magic number, but an alternative would be to have the p2p module wait until connections are open and to flush the message correctly
```
Olshansk, line 54:
```suggestion
				// TECHDEBT(#874): this is a magic number, but an alternative would be to have the p2p module wait until connections are open and to flush the message correctly
```

Olshansk (approved):
@0xBigBoss I have a similar hack for this in #874, so would appreciate if you review that as well.

Feel free to merge this in to unblock `main` in the meantime.

Screenshot 2023-07-31 at 9 46 50 AM

PR #969, opened 2023-08-05
[utility] fix failed to create servicer

0xBigBoss:
Adds servicer pk to validator1 docker compose deployment

Fixes `failed to create servicer` in the utility module for validator1 deployment.

There are probably more places this needs to be fixed, but this unblocked me after rebasing off main.

### DIFF
--- a/build/config/config.validator1.json
+++ b/build/config/config.validator1.json
@@ -56,13 +56,14 @@
   },
   "servicer": {
     "enabled": true,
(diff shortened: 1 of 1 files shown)


Olshansk:
@0xBigBoss I put up a similar patch last week, so please see https://github.com/pokt-network/pocket/pull/968.

Also fixes the corresponding test.

